PR #23, opened 2015-08-19
ignite-1127 Query with result size more then one page doesn't increase Query executions count metric

agura:


svladykin:
Looks good, but I think for all the additive values we need to use LongAdder8 instead of Atomics.

PR #27, opened 2015-08-22
IGNITE-884 (Cache store usability issues)

vkulichenko:


dspavlov:
https://issues.apache.org/jira/browse/IGNITE-884 is closed as Won't fix. Probably we can close this PR for now? AFAIK, PR can be reopened later.

PR #30, opened 2015-08-24
ignite-1261 App context should be transient

agura:


agura:
Changes was made without this PR in ignite-1.4 branch.

PR #34, opened 2015-08-25
Ignite 1277 main

iveselovskiy:
1) enabled data cache backups for IGFS to provide data resilience.
2) added tests to check the failovers with backups.


iveselovskiy:
Merged

PR #35, opened 2015-08-26
ignite-1226:Need to add method that returns names of all available caches

kcheng-mvp:
Fix bug: https://issues.apache.org/jira/browse/IGNITE-1226


agoncharuk:
Ken, my comments are below:
 * I do not think that we need to ever return system cache names to a user. System caches may be a subject to change without notice, and a reference to a system cache can be never obtained by a user. This method should always return only user caches.
 * I think the return type may be weaken to Collection<String>
 * You do not need to create an instance of VisorCacheMetrics to determine is this is a user cache. The correct if statement should be `if (ca.context().started() && ca.context().userCache())`

agoncharuk:
Ken,

I actually missed one important use-case during the review. If this method is invoked on a client node that has never called `getOrCreateCache()`, there will be no cache contexts available. We need to use `registeredCaches` map inside the cache processor to craft a collection of available cache names.
Below is the method for cache processor that will return cache names regardless of the availability of cache contexts:

    public Collection<String> publicCacheNames() {
        return F.viewReadOnly(
            F.view(registeredCaches.values(),
                new IgnitePredicate<DynamicCacheDescriptor>() {
                    @Override public boolean apply(DynamicCacheDescriptor desc) {
                        return desc.cacheType().userCache();
                    }
                }),
            new C1<DynamicCacheDescriptor, String>() {
                @Override public String apply(DynamicCacheDescriptor desc) {
                    return desc.cacheConfiguration().getName();
                }
            });
    }

kcheng-mvp:
Thank you very much!
I will go through the code and do the necessary changes.

Thanks,
kcheng

kcheng-mvp:
Just now I write a unit test for this case : Call this method on the node
without any calling "getOrCreateCache()", it works as expected.

public void testAllCachesNamesInCluster(){
    Ignite g0 = grid(0);
    Ignite g1 = grid(1);
    Ignite g2 = grid(2);


    g0.getOrCreateCache("Cache1");
    g1.getOrCreateCache("Cache2");
    g1.getOrCreateCache("Cache3");

    Collection<String> caches = g0.cacheNames();
    assertTrue(caches.contains("Cache1"));
    assertTrue(caches.contains("Cache2"));
    assertTrue(caches.contains("Cache3"));
    assertEquals(3, g0.cacheNames().size());

    caches = g1.cacheNames();
    assertTrue(caches.contains("Cache1"));
    assertTrue(caches.contains("Cache2"));
    assertTrue(caches.contains("Cache3"));
    assertEquals(3, g1.cacheNames().size());

    caches = g2.cacheNames();
    assertTrue(caches.contains("Cache1"));
    assertTrue(caches.contains("Cache2"));
    assertTrue(caches.contains("Cache3"));
    assertEquals(3, g1.cacheNames().size());
}


Did I miss something?


By the way I did some other test:

I found there a method "public Collection<IgniteCacheProxy<?, ?>>
publicCaches()" in

org.apache.ignite.internal.processors.cache.GridCacheProcessor


on the contrary, if I implements cacheNames() as this way


GridCacheProcessor cacheProcessor = ctx.cache();
for(IgniteCacheProxy cache:cacheProcessor.publicCaches()){
    cache.getName();
    cacheNames.add(cache.getName());
}
return cacheNames;


The last case will fail.


That's to say, if call method

public Collection<IgniteCacheProxy<?, ?>> publicCaches() {
    Collection<IgniteCacheProxy<?, ?>> res = new
ArrayList<>(jCacheProxies.size());

    for (Map.Entry<String, IgniteCacheProxy<?, ?>> entry :
jCacheProxies.entrySet()) {
        if (entry.getValue().context().userCache())
            res.add(entry.getValue());
    }

    return res;
}

on a node which never call "getOrCreateCache()", it would not return anything.


I think the same issue exists in the method "public <K, V>
IgniteCacheProxy<K, V> jcache(@Nullable String name) " as well.


The correct behaviour should be always return the existing caches, it
does not matter there are caches created on the current node,right?

it's a bug right? If so I am going to log a ticket and fix it together
with the current one.

please confirm.


Thanks,

kcheng


Thanks,
kcheng

modules/core/src/test/java/org/apache/ignite/internal/ClusterMetricsSelfTest.java
```diff
@@ -95,6 +95,42 @@ public void testTaskExecution() {
...
+        Ignite g1 = grid(1);
+        Ignite g2 = grid(2);
+        Ignite g3 = grid(3);
+
+
+        g0.getOrCreateCache("Cache1");
+        g1.getOrCreateCache("Cache2");
+        g1.getOrCreateCache("Cache3");
+
+        Thread.sleep(20);
```
kcheng-mvp: When I run this unit test code at my desktop, I found this case would fail sporadic, from the log I can see the some of the node does not contains the cache, as the cache does not publish the the node. so here I let it sleep 20ms. with this line the test can pass always.

agoncharuk:
Ken, please add the following test: start 3 server nodes and a client node, start a cache on a server node and then call cacheNames() on the client node. It should fail in the current implementation as cacheContext is not created on the client node unless the cache is accessed.

kcheng-mvp:
Hi agoncharuk,
yes, just as you said, if the node start with client mode 
```
 cfg.setClientMode(true);
```
Then the case will fail, I have implemented the method as you advised.
please help to reivew

dmagda:
Ken, thanks for the contribution!

Reviewed, did minor changes & merge to 1.5 and master branch.

kcheng-mvp:
merged

PR #37, opened 2015-08-26
Portables: optimize cache related implementation and make it backward compatible

dmagda:
At the Ignite's layer now it's possible to register additional classes outside of GridPortableContext.


dmagda:
Integrated into the apache master

PR #129, opened 2015-10-04
IGNITE-708: Remove refreshPartitions on ResendTimeoutObject.

murador:
IGNITE-708:  From what I could see, the refresh partition in case of timeout is redundant, since each node has a refresh partition which runs at regular intervals in the method body() of GridCachePartitionExchangeManager retrieved from the background process ExchangeWorker. Thus eliminating the refreshPartition in ResendTimeoutObject class should not have any impact on the operation, but an increase in performance when we avoid sending redundant message.


dspavlov:
Closing this PR as the ticket is duplicate https://issues.apache.org/jira/browse/IGNITE-708

Feel free to reopen if it is needed

PR #130, opened 2015-10-05
[Ignite-429] Implement IgniteStormStreamer to stream data from Apache Storm

chandresh-pancholi:
Made streamer with java1.7 compatible and added javadoc


chandresh-pancholi:
[Ignite-429] Implement IgniteStormStreamer to stream data from Apache Storm #130

PR #152, opened 2015-10-13
IGNITE-1635 Added reproduciable unit-test for the bug

ashutakGG:
https://issues.apache.org/jira/browse/IGNITE-1635

Added reproduciable unit-test for the bug.

Also added test for https://issues.apache.org/jira/browse/IGNITE-1616, but test do not reproduce the issue.


ashutakGG:
Merged.

PR #153, opened 2015-10-14
IGNITE-1622 - Fixed cache.clear() with near cache

vkulichenko:


agoncharuk:
Val,

There is a javadoc entry for the missing readers variable in clearLocally(K) in IgniteInternalCache. Is it just javadoc or a really missed argument for the function? 

Otherwise looks good.

PR #158, opened 2015-10-15
IGNITE-1682 .Net: Remove RemoteListen from Events API

ptupitsyn:


yp19910928:
Now.net does not have RemoteListen, what is used to achieve this function?

PR #171, opened 2015-10-22
IGNITE-1759 .Net: Guid read/write is not platform-safe.

ptupitsyn:


ptupitsyn:
Close because of old target branch

PR #174, opened 2015-10-26
ignite-1792 ContinuousProcessor should use ALWAYS_TRUE predicate instead of null

agura:


dmagda:
Changes look good. Merged to the master.

PR #180, opened 2015-10-28
ignite-1717 Portable marshaller fixed for cases when serialization proxy should be serialized by optimized marshaller

agura:


agura:
Merged.

PR #191, opened 2015-11-01
ignite-1817 Deprecate RandomEvictionPolicy and IgniteCache.randomEntry method

agura:


agura:
Merged into ignite-1.5 branch.

PR #252, opened 2015-11-20
ignite-1544 Make sure objects number stored in a cache can be bigger than Integer.MAX_VALUE

agura:


agura:
Merged into ignite-1.5.

PR #309, opened 2015-12-09
IGNITE-1993 remove non-standard SQL clause so table for JDBC discovery…

westbury:
Note that the tests do not run locally.  They fail because portable/test1 and portable/test2 are not available (see IGNITE-1462).  However I note that the instructions at https://cwiki.apache.org/confluence/display/IGNITE/How+to+Contribute instruct to use -DskipTests so I did that.

This pull request supercedes the patch as per Denis Magda's comment.


westbury:
This commit has now been updated to protect correctly against the race condition.

dspavlov:
https://issues.apache.org/jira/browse/IGNITE-1993 is closed and merged, closing the PR

PR #324, opened 2015-12-11
 IGNITE-1979 Support case insensitive nonquoted cache names in SQL

VladimirErshov:
IGNITE-1979 Support case insensitive nonquoted cache names in SQL

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
+++ b/modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
@@ -1787,6 +1787,8 @@ public class CacheConfiguration<K, V> extends MutableConfiguration<K, V> {
     }
 
     /**
+     * TODO
+     *
      * @return Schema name for current cache according to SQL ANSI-99. Could not be {@code null}.
      */
     public String getSqlSchema() {
@@ -1803,6 +1805,7 @@ public class CacheConfiguration<K, V> extends MutableConfiguration<K, V> {
      */
     public void setSqlSchema(String sqlSchema) {
         A.ensure((sqlSchema == null), "Schema could not be null");
+        A.ensure((sqlSchema.isEmpty()), "Schema could not be empty");
 
         this.sqlSchema = sqlSchema;
     }
--- a/modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
+++ b/modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
@@ -1775,7 +1775,7 @@ public class CacheConfiguration<K, V> extends MutableConfiguration<K, V> {
     }
 
     /**
-     * Gets timeout in milliseconds after which long query warning will be printed.
+     * Sets timeout in milliseconds after which long query warning will be printed.
      *
      * @param longQryWarnTimeout Timeout in milliseconds.
      * @return {@code this} for chaining.
@@ -1787,11 +1787,12 @@ public class CacheConfiguration<K, V> extends MutableConfiguration<K, V> {
     }
 
     /**
-     * TODO
+     * Gets custom name of the sql schema. If custom sql schema is not set then {@code null} will be returned and
+     * quoted case sensitive name will be used as sql schema.
      *
-     * @return Schema name for current cache according to SQL ANSI-99. Could not be {@code null}.
+     * @return Schema name for current cache according to SQL ANSI-99. Could be {@code null}.
      */
-    public String getSqlSchema() {
+    @Nullable public String getSqlSchema() {
         return sqlSchema;
     }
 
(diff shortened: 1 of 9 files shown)


modules/visor-console/src/main/scala/org/apache/ignite/visor/commands/cache/VisorCacheCommand.scala
```diff
@@ -17,23 +17,21 @@
 
 package org.apache.ignite.visor.commands.cache
 
+import java.lang.{Boolean => JavaBoolean}
+import java.util.{Collection => JavaCollection, Collections, UUID}
+
 import org.apache.ignite._
 import org.apache.ignite.cluster.ClusterNode
 import org.apache.ignite.internal.util.typedef._
+import org.apache.ignite.internal.visor.cache._
```
avinogradovgg: Please discuss import rearrangement with scala team.

modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
```diff
@@ -365,15 +365,18 @@
     private IgnitePredicate<ClusterNode> nodeFilter;
 
     /** */
-    private boolean sqlEscapeAll;
+    private String sqlSchema;
```
svladykin: Did really make sense to reorder the fields?

modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
```diff
@@ -1783,6 +1787,27 @@ public long getLongQueryWarningTimeout() {
     }
 
     /**
+     * @return Schema name for current cache according to SQL ANSI-99. Could not be {@code null}.
```
svladykin: All the public APIs must have a description (other than params and returns).

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -642,7 +648,7 @@ private void removeTable(TableDescriptor tbl) throws IgniteCheckedException {
         if (log.isDebugEnabled())
             log.debug("Removing query index table: " + tbl.fullTableName());
 
-        Connection c = connectionForThread(tbl.schema());
```
svladykin: Why did we drop `TableDescriptor.schema()` method?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -1516,6 +1526,9 @@ private void createSqlFunctions(String schema, Class<?>[] clss) throws IgniteChe
 
     /** {@inheritDoc} */
     @Override public void registerCache(CacheConfiguration<?,?> ccfg) throws IgniteCheckedException {
+        space2schema.put(ccfg.getName(), ccfg.getSqlSchema() == null ? '\"' + ccfg.getName() + '\"' : ccfg.getSqlSchema());
+        schema2space.put(ccfg.getSqlSchema() == null ? '\"' + ccfg.getName() + '\"' : ccfg.getSqlSchema(), ccfg.getName());
```
svladykin: The code above looks cryptic to me. Lets calculate schema name once, assign it to a variable and put everywhere we need to.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -1530,6 +1543,9 @@ private void createSqlFunctions(String schema, Class<?>[] clss) throws IgniteChe
 
     /** {@inheritDoc} */
     @Override public void unregisterCache(CacheConfiguration<?, ?> ccfg) {
+        schema2space.remove(ccfg.getSqlSchema() == null ? '\"' + ccfg.getName() + '\"' : ccfg.getSqlSchema());
```
svladykin: We will do `String schema = schema(ccfg.getName());` on line 1549 why do not use schema name from there instead of calculating it. It is a third time I see schema name is calculated in the code. DRY!!!

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/GridSqlQuerySplitter.java
```diff
@@ -135,9 +135,10 @@ else if (expr instanceof GridSqlColumn)
      * @param stmt Prepared statement.
      * @param params Parameters.
      * @param collocated Collocated query.
+     * @param igniteH2Indexing Indexing implementation.
      * @return Two step query.
      */
-    public static GridCacheTwoStepQuery split(JdbcPreparedStatement stmt, Object[] params, boolean collocated) {
+    public static GridCacheTwoStepQuery split(JdbcPreparedStatement stmt, Object[] params, boolean collocated, IgniteH2Indexing igniteH2Indexing) {
```
svladykin: I don't really like this passing indexing instance. Probably we just have to collect schemas here and convert them to spaces afterwards.

modules/visor-console/src/main/scala/org/apache/ignite/visor/commands/cache/VisorCacheCommand.scala
```diff
@@ -17,23 +17,21 @@
...
 import org.apache.ignite.visor.visor._
-
 import org.jetbrains.annotations._
 
-import java.lang.{Boolean => JavaBoolean}
-import java.util.{Collection => JavaCollection, Collections, UUID}
-
-import org.apache.ignite.internal.visor.cache._
-import org.apache.ignite.internal.visor.util.VisorTaskUtils._
-
```
svladykin: Are you sure it is safe to drop all the imports above? Scala can change behavior in presence or absence of some imports.

modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
VladimirErshov: It is better to gather all three sql* fields together, I suppose.

modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
VladimirErshov: Agree. Will fix.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
VladimirErshov: It's the consequences of doing IgniteH2Indexing#schema method non-static. This approach allows us not to create any extra fields in TableDescriptor class: like String schemaName for ex.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
VladimirErshov: It's a good point, but actually I saw one more over here, which is more interesting - schema called after deletion from cache, and that is a problem. Going to fix both.
And about DRY (exclamation point, exclamation point, exclamation point) - here was a Rule of Three applied. Nonetheless, I'm going to think about possible simplification here. Thanks.

PR #335, opened 2015-12-15
Ignite 2016: Kafka Connect Integration

shroman:
Kafka Connect is a new standard method "for scalably and reliably streaming data between Apache Kafka and other systems" from the latest Kafka 0.9.0.


modules/kafka/pom.xml
```diff
@@ -34,6 +35,12 @@
     <version>1.5.1.final-SNAPSHOT</version>
     <url>http://ignite.apache.org</url>
 
+    <properties>
+        <kafka.version>0.9.0.0</kafka.version>
```
dmagda: We store all the versions numbers in {ignite}/parent/pom.xml.

Please define new or override existing versions there.

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSinkTask.java
```diff
@@ -0,0 +1,161 @@
...
+     * Initializes grid client from configPath.
+     *
+     * @param props Task properties.
+     */
+    @Override public void start(Map<String, String> props) {
+        igniteConfigFile = props.get(IgniteSinkConstants.CACHE_CFG_PATH);
+        cacheName = props.get(IgniteSinkConstants.CACHE_NAME);
+
+        if (props.containsKey(IgniteSinkConstants.CACHE_ALLOW_OVERWRITE))
+            StreamerContext.getStreamer().allowOverwrite(
```
dmagda: Can there be a case when these settings (CACHE_ALLOW_OVERWRITE, CACHE_PER_NODE_DATA_SIZE,...)  are different among tasks?

If yes, the we can't share a single data streamer instance among different tasks.
If such a case is impossible then it doesn't make sense to "override" the setting of the streamer each time a new task is started.

shroman: With the current implementation, we are writing from the specified topics to the specified cache, and the settings are same for all tasks.
# One can implement it differently if there's a strong demand from users though.

Right, it will "override" the settings of the streamer each time a new task is started, but the overhead is minimal and happens only once. What do you propose?

dmagda: I would just ignore setting the settings one more time if a streamer for a particular cache already exists.

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSinkTask.java
```diff
@@ -0,0 +1,161 @@
...
+     */
+    public static class StreamerContext {
+        /** Constructor. */
+        private StreamerContext() {
+        }
+
+        /** Instance holder. */
+        private static class Holder {
+            private static final Ignite IGNITE = Ignition.start(igniteConfigFile);
+            private static final IgniteDataStreamer STREAMER = IGNITE.dataStreamer(cacheName);
```
dmagda: This will prevent us from streaming to different ignite caches in a scope of a single Java process.

We should implement it in a way that it would be possible to have several IgniteSinks that stream data to different caches in a single Java process. Is this possible?

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSinkTask.java
shroman: Ok, I will check if static variable "cacheName" is not null, settings will be ignored.

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSinkTask.java
shroman: It will complicate settings. In addition, 
1. a user may still want to have 2 tasks running on one cache and n tasks on another. Since you cannot call _Ignition.start(...)_ on the same cache, you will need a more elaborate singleton implementation.
2. Different caches can have different data types, and Kafka worker cannot have multiple converters specified.
So if you wish to work with different cache (and different data types for keys and values), you have to start another process.
That is the way I would like to proceed for now unless there are special requirements from users (and even then there is 2. I described above).

dmagda: Ok, got you. Thanks for the explanation.

PR #383, opened 2015-12-26
IGNITE-944 HTTP REST log command doesn't work

samaitra:


modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
```diff
@@ -0,0 +1,149 @@
...
+        if (req0.to() != -1) {
+            to = req0.to();
+        }
+        String path = DEFAULT_LOG_PATH;
+        Path filePath = getPath(req0, path);
+        if (from > to) {
+            return new GridFinishedFuture<>(new GridRestResponse(GridRestResponse.STATUS_FAILED,
+                    "Log file start from cannot be greater than to"));
+        }
+        switch (req.command()) {
```
dmagda: No need to you 'switch' operator there.
Just go ahead with basic "if (req.command == LOG)"

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
```diff
@@ -0,0 +1,149 @@
...
+        return SUPPORTED_COMMANDS;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public IgniteInternalFuture<GridRestResponse> handleAsync(GridRestRequest req) {
+        assert req != null;
+        assert SUPPORTED_COMMANDS.contains(req.command());
```
dmagda: Since only one cmd is supported by this handler there is no need to create SUPPORTED_COMMANDS list.

More optimal is to check that req.command == LOG

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
```diff
@@ -0,0 +1,149 @@
...
+        if (log.isDebugEnabled())
+            log.debug("Handling log REST request: " + req);
+        GridRestLogRequest req0 = (GridRestLogRequest) req;
+        int from = DEFAULT_FROM;
```
dmagda: Let's implement the following use cases:
1) if both 'to' and 'from' are set in a request then use them as is;
2) if neither 'to' nor 'from' is set then return the first line from the log;
3) if 'to' is set and 'from' is not or vice verse then just an error code because according to the doc [1] if either parameter is set the other is mandatory

[1] https://apacheignite.readme.io/docs/rest-api#log

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
```diff
@@ -0,0 +1,149 @@
...
+            filePath = Paths.get(path);
+        } else {
+            filePath = Paths.get(ctx.config().getIgniteHome() + "/" + path);
+
+        }
+        return filePath;
+    }
+
+    private List getContent(int from, int to, Path filePath) throws IgniteCheckedException {
+        List content = new ArrayList<>();
```
dmagda: Can't we use StringBuilder instead by just putting lines into it and returning a final String at the end.

This will be more memory optimal approach.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
```diff
@@ -0,0 +1,149 @@
...
+        try (BufferedReader reader = Files.newBufferedReader(filePath, Charset.defaultCharset())) {
+            String line = null;
+            int start = from;
+            int lineCounter = 0;
+            while ((line = reader.readLine()) != null
+                    && start < to) {
+                if (lineCounter >= from) {
```
dmagda: Don't see a reason why we need both 'lineCounter' and 'start'.

Isn't it enough just to increment 'from' and stop the loop once 'from' == 'to'?

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
```diff
@@ -0,0 +1,149 @@
...
+                    content.add(line);
+                    start++;
+                }
+                lineCounter++;
+            }
+        } catch (IOException e) {
+            throw new IgniteCheckedException("Unable to open file " + filePath.getFileName());
```
dmagda: The reason of IOException can be of other kind. I would thrown IgniteCheckedException(e) instead.

modules/core/src/test/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandlerTest.java
```diff
@@ -0,0 +1,60 @@
...
+        GridLogCommandHandler gridLogCommandHandler = new GridLogCommandHandler(super.newContext());
+        Collection<GridRestCommand> commandCollection = gridLogCommandHandler.supportedCommands();
+        assertFalse(commandCollection.contains(GridRestCommand.VERSION));
+    }
+
+    public void testHandleAsync() throws Exception {
+        GridTestKernalContext ctx = super.newContext();
+        GridLogCommandHandler gridLogCommandHandler = new GridLogCommandHandler(ctx);
+        GridRestLogRequest gridLogRestRequest = new GridRestLogRequest();
+        gridLogRestRequest.to(5);
```
dmagda: Please add additional test for the following cases:
- 'to' and 'from' are not set;
- 'path' is not set;
- 'from' is bigger or equal to 'to'

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
samaitra: @dmagda Yes, StringBuilder can be used but in that case a long list of log lines will be returned as single json object.  The list of String creates a json array with each log line as a separate entry in the array. It is looks much cleanly formatted in json view.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
samaitra: @dmagda makes sense :)

samaitra:
Hi,

I have updated the PR with review feedback.

Regards
Saikat

dmagda:
Hi Saikat, thanks!

I'm on holidays and will be able to review your changes at the end of the
next week.


Denis

samaitra:
Thank you Denis. Let me know if any further changes are required.

Regards
Saikat

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
```diff
@@ -0,0 +1,158 @@
...
+            if (req0.from() != -1) {
+                from = req0.from();
+            }
+            int to = DEFAULT_TO;
+            if (req0.to() != -1) {
+                to = req0.to();
+            }
+            String path = DEFAULT_LOG_PATH;
+            Path filePath = getPath(req0, path);
+            if (from >= to) {
```
dmagda: Check this condition before and only after that calculate the path - Path filePath = getPath(req0, path);

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
```diff
@@ -0,0 +1,158 @@
...
+        if (req.command() == LOG) {
+            if (log.isDebugEnabled())
+                log.debug("Handling log REST request: " + req);
+            GridRestLogRequest req0 = (GridRestLogRequest) req;
+            try {
+                validateRange(req0);
+            } catch (IgniteCheckedException e){
+                return new GridFinishedFuture<>(new GridRestResponse(GridRestResponse.STATUS_FAILED, e.getMessage()));
+            }
+            int from = DEFAULT_FROM;
```
dmagda: Saikat,

Lets follow the Rest API docs [1] and implement support of 'from' and 'to' parameters the following way:
- if 'from' is set but 'to' is not then return an error;
- if 'to' is set but 'from' is not then return an error as well;
- if neither 'from' nor 'to' is set then return the first log line.
- if both parameters are set then just check that 'from' is less than 'to' and use them as is

[1] https://apacheignite.readme.io/docs/rest-api#section-request-parameters

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
```diff
@@ -0,0 +1,158 @@
...
+            if (req0.to() != -1) {
+                to = req0.to();
+            }
+            String path = DEFAULT_LOG_PATH;
+            Path filePath = getPath(req0, path);
+            if (from >= to) {
+                return new GridFinishedFuture<>(new GridRestResponse(GridRestResponse.STATUS_FAILED,
+                        "Log file start from cannot be greater than to"));
+            }
+            if(req0.command() == LOG) {
```
dmagda: No need to check this twice. I've already done this at the beginning of the method.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
dmagda: I see your idea. However according to the current REST API is should be returned as a single string.

[1] https://apacheignite.readme.io/docs/rest-api#section-response-example-

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
samaitra: updated the check

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
samaitra: Denis,
*The validateRange method checks for the first 2 point.
*The DEFAULT_FROM=0 and DEFAULT_TO=1 return first log line in absence of 'from' and 'to' 
* if both 'from' and 'to' is set then we are checking that 'from' is less than 'to'.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
samaitra: makes sense. removed the extra check.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
samaitra: updated to use StringBuilder.

PR #403, opened 2016-01-14
gg-8700

ilantukh:


dspavlov:
Is this PR still actual? It is not linked to any JIRA ticket, so I suggest closing it for now.

PR #420, opened 2016-01-24
Ignite 944 validation to check for non existing lines

samaitra:


samaitra:
@dmagda 

Hey Denis,

Can you please review this change for validation of non existing lines.

Regards
Saikat

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/log/GridLogCommandHandler.java
```diff
@@ -161,6 +161,10 @@ private String readLog(int from, int to, File logFile) throws IgniteCheckedExcep
 
                 start++;
             }
+
+            if (start < from || start > (to + 1)){
```
dmagda: I think that you can simply do the following check 

if (content.size() == 0)
thrown new IgniteCheckedException(...)

samaitra: yes, makes sense. Thanks !!!

samaitra:
@dmagda 

Thank you Denis. I have updated the files as per review comments.

Regards
Saikat

samaitra:
@dmagda 

Can you please review and share feedback.

Regards
Saikat

PR #451, opened 2016-02-03
IGNITE-2539 test and commented a way to reproduce

VladimirErshov:


dspavlov:
IGNITE-2539 | Resolved
-- | --

PR #502, opened 2016-02-19
fix IGNITE-2650

ruskim:
- Added checkClient to to IgniteSpiConsistencyChecked annotation with default value true.
  If it's false, than spi consistency will not be checked for client nodes.
- Added IgniteSpiConsistencyChecked to NoopSwapSpaceSpi and FileSwapSpaceSpi. That guaranties, that all nodes in grid will have the same swap space provider. Client nodes ignore this check, because clients don't have any data.
- Cache with swap creation will fail, if nodes in grid have NoopSwapSpaceSpi


dspavlov:
https://issues.apache.org/jira/browse/IGNITE-2650 was merged, closing the PR

PR #549, opened 2016-03-11
IGNITE-2404

ryanzz:
For TcpDiscoverySpi and TcpCommunicationSpi, if port range value is 0, then implementation will try bind only to the port provided by setLocalPort(int) method and fail if binding to this port did not succeed.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
+++ b/modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
@@ -958,7 +958,7 @@ public class TcpCommunicationSpi extends IgniteSpiAdapter
      */
     @IgniteSpiConfiguration(optional = true)
     public void setLocalPortRange(int locPortRange) {
-        this.locPortRange = locPortRange;
+        this.locPortRange = locPortRange == 0 ? 1 : locPortRange;
     }
 
     /** {@inheritDoc} */
--- a/modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
+++ b/modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
@@ -958,7 +958,7 @@ public class TcpCommunicationSpi extends IgniteSpiAdapter
      */
     @IgniteSpiConfiguration(optional = true)
     public void setLocalPortRange(int locPortRange) {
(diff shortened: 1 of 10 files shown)


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -958,7 +958,7 @@ public void setLocalPort(int locPort) {
      */
     @IgniteSpiConfiguration(optional = true)
     public void setLocalPortRange(int locPortRange) {
-        this.locPortRange = locPortRange;
+        this.locPortRange = locPortRange == 0 ? 1 : locPortRange;
```
dmagda: If the uses sets '0' as a range then getLocalPortRange() must return '0' as well. Please remove the modification at this part of the code and refer to my suggestions below.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -4861,27 +4861,25 @@ private void checkConnection() {
                     else
                         srvrSock = new ServerSocket(port, 0, spi.locHost);
 
-                    break;
+                    if (log.isInfoEnabled())
```
dmagda: Let's do the following here

Create 'lastPort' variable
int lastPort = spi.locPortRange  == 0 ? spi.locPort : spi.locPort + spi.locPortRange - 1;

for (port = spi.locPort; port <= lasPort; port++)
....

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoverySpi.java
```diff
@@ -686,7 +688,7 @@ public TcpDiscoverySpi setLocalPort(int locPort) {
      */
     @IgniteSpiConfiguration(optional = true)
     public TcpDiscoverySpi setLocalPortRange(int locPortRange) {
-        this.locPortRange = locPortRange;
+        this.locPortRange = locPortRange == 0 ? 1 : locPortRange;
```
dmagda: The same as above. Please don't set locPortRange to 1 here because getLocalPortRange() method must return '0' if the latest is set.

Modify loops in TcpCommunicationSpi the same way as I proposed in ServerImpl above.
Look for usages of TcpCommunicationSpi.locPortRange and you will find a couple of places where locPortRange is used.

PR #560, opened 2016-03-17
Ignite-2730: Ignite Events Source Streaming to Kafka.

shroman:


modules/kafka/README.txt
```diff
@@ -127,3 +127,80 @@ k1,v1
...
+Connector can be found in 'optional/ignite-kafka.' It and its dependencies have to be on the classpath of a Kafka running instance,
+as described in the following subsection.
+
+### Setting up and Running
+
+1. Put the following jar files on Kafka's classpath
+- ignite-kafka-connect-x.x.x-SNAPSHOT.jar
+- ignite-core-x.x.x-SNAPSHOT.jar
+- cache-api-1.0.0.jar
+- ignite-spring-1.5.0-SNAPSHOT.jar
```
dmagda: I would avoid referring to specific versions of libraries there because otherwise this doc has to be updated each time a new version of a library is released.

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/CacheEventFilter.java
```diff
@@ -0,0 +1,36 @@
...
+package org.apache.ignite.stream.kafka.connect;
+
+import java.io.Serializable;
+import org.apache.ignite.events.CacheEvent;
+
+/**
+ * Filter for {@link IgniteSourceTask}.
+ * Enables filtering out cache events that do not match user-specified match conditions.
+ */
+public interface CacheEventFilter extends Serializable {
```
dmagda: Probably you can reuse org.apache.ignite.lang.IgnitePredicate there that was created for purposes like that one.

shroman: You are right, and that's what I was actually doing ;) but forgot to remove CacheEventFilter...

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSourceConnector.java
```diff
@@ -0,0 +1,81 @@
...
+
+        for (int i = 0; i < maxTasks; i++)
+            taskConfigs.add(taskProps);
+
+        return taskConfigs;
+    }
+
+    /** {@inheritDoc} */
+    @Override public void stop() {
+        // No-op.
```
dmagda: Is it expected? Don't we want to give a chance to stop running active tasks?

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSourceTask.java
```diff
@@ -0,0 +1,309 @@
...
+                }
+                catch (Exception e) {
+                    log.error("Failed to instantiate the provided filter! " +
+                        "User-enabled filtering is ignored!", e);
+                }
+            }
+        }
+
+        IgniteBiPredicate<UUID, CacheEvent> locLsnr = new IgniteBiPredicate<UUID, CacheEvent>() {
+            @Override public boolean apply(UUID id, CacheEvent evt) {
```
dmagda: Please don't use anonymous classes for both remote and local listener because in that case their enclosing class (IgniteSourceTask) will be serialized and sent over the wire along with them. Just move implementation of the listeners to "private static class".

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSourceTask.java
```diff
@@ -0,0 +1,309 @@
...
+                }
+
+                return true;
+            }
+        };
+
+        IgnitePredicate<CacheEvent> rmtLsnr = new IgnitePredicate<CacheEvent>() {
+            @Override public boolean apply(CacheEvent evt) {
+
+                Affinity affinity = IgniteGrid.getIgnite().affinity(cacheName);
```
dmagda: Remote filter will be executed on server nodes that can be started as standalone processes using ignite.sh or similar method. Looking at the implementation of IgniteGrid.getIgnite() I should say that you will trigger starting of one more Ignite instance on a remote side. 

To get an instance of Ignite on a remote side please use @IgniteInstanceResource annotation
private static class RemoteFilter ... {
@IgniteInstanceResource
Ignite ignite;
....
}

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSourceConnector.java
shroman: Tasks are stopped in IgniteSourceTask. I have looked through other reference implementations, and all of them do nothing.

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSourceTask.java
shroman: Thank you! I haven't thought about it.

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSourceTask.java
```diff
@@ -0,0 +1,326 @@
...
+ */
+public class IgniteSourceTask extends SourceTask {
+    /** Logger. */
+    private static final Logger log = LoggerFactory.getLogger(IgniteSourceTask.class);
+
+    /** Event buffer size. */
+    private static int evtBufSize = 100000;
+
+    /** Event buffer. */
+    private static BlockingQueue<CacheEvent> evtBuf = new LinkedBlockingQueue<>(evtBufSize);
```
dmagda: Are you sure you want to share this buffer among different IgniteSourceTasks?
All of them will use a single instance which will be re-created each time a new IgniteSourceTask is started.

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSourceTask.java
```diff
@@ -0,0 +1,326 @@
...
+    private static UUID rmtLsnrId;
+
+    /** Local listener. */
+    private static TaskLocalListener locLsnr = new TaskLocalListener();
+
+    /** Remote filter. */
+    private static TaskRemoteFilter rmtLsnr = new TaskRemoteFilter();
+
+    /** User-defined filter. */
+    private static IgnitePredicate<CacheEvent> filter;
```
dmagda: Filter is unique for every tasks but you marked it "static" meaning that all tasks, running in a single JVM, will use a single instance of the filter.

Please review usage of "static" keyword in this class keeping in mind that if a field is static it will be shared among all IgniteSourceTasks running in a single VM.

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSourceTask.java
shroman: Thanks for pointing to it. It was used within TaskLocalListener which is "private static", but I'll take away "static" from TaskLocalListener then too.

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSourceTask.java
shroman: In the current implementation IgnitePredicate filter is the same for all tasks, and also initialized after _if (cacheName != null)_ check, so I will keep it static. I removed _static_ from listeners since we won't share the queue.

shroman: I think we want listeners (both local and remote) _static_ to receive a notification once -- no matter how many tasks we have running. _TaskLocalListener_ is static and uses the buffer which has to be static then. Same for the user-defined filter.
Do you have an idea to make it better?

Btw, making everything _static_ helps to avoid duplicate notifications in standalone Kafka Connect mode. In a distributed mode (when multiple Sources are started) the user will have to handle it. I will mention it in the docs.

dmagda: In fact the number of times both remote and local listeners will be called depends on the number of times you call remoteListen(locLsnt, remoteFilter).

Every time you call remoteListen() you register new one remote filter on target nodes. Each localListener corresponds to a particular remote filter. If remote filter returns "true" then the notification will be delivered to its local listener.

In your case despite of the fact that the local listener is static it will be called many times since it corresponds to many remote listeners that are started upon specific caches. In any case you can leave the local listener static if you want to share the queue among all notifications coming from different caches.

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSourceTask.java
```diff
@@ -0,0 +1,328 @@
...
+
+            throw new ConnectException(e);
+        }
+        finally {
+            stopped = false;
+        }
+    }
+
+    /** {@inheritDoc} */
+    @Override public List<SourceRecord> poll() throws InterruptedException {
```
dmagda: Will the method be called by Kafka from several threads? Is the number of Threads equal to number of IgniteSourceTask? If so then it's better to introduce a queue (non static) per task to avoid possible threads contention.

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSourceTask.java
shroman: I see.
We have _if (cacheName != null)_ check at the top of _start(...)_ not to create configurations again and again for tasks because they are the same. In this case, in my understanding, we have one remote filter and one listener only, no matter how many tasks we have.

shroman: Btw, the current implementation treats one cache -> 1...n Kafka topics

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSourceTask.java
shroman: Yes, I think so. "Kafka Connect provides each task with a dedicated thread" http://docs.confluent.io/2.0.0/connect/devguide.html#task-example-source-task

modules/kafka/src/main/java/org/apache/ignite/stream/kafka/connect/IgniteSourceTask.java
dmagda: I see. So according to the current implementation you can start only one remote filter and local listener in a single VM meaning that you have to start several JVM processes if you want to get events from several caches. Am I correct? If yes then now it's clear why you use "static" throughout the code and ok let's have this as a first implementation. Please merge you changes then and update docs if needed.

shroman: Yes, that's correct. I am also not very happy with _static_, but let's see how it goes and if we have feedback from users who want to treat many caches from the same VM, we enhance the current solution.
Even then, we will need a kind of singleton for every Ignite instance (because calling _start()_ more than once gives an exception -- btw, is there any good solution for this except using singletons?) and only one remote listener for each to avoid duplicates. Is my understanding correct?

PR #588, opened 2016-03-30
IGNITE-2668

vldpyatkov:
MBean creation fails if cache name contains ':' character
Write attribute, which contains cache name, in quote


vldpyatkov:
It had been merging through cherry-pick .

PR #597, opened 2016-04-01
Ignite-2822 Continuous query local listener can be notified with empty list of events

dmagda:


dspavlov:
https://issues.apache.org/jira/browse/IGNITE-2822 is closed, closing the PR

PR #617, opened 2016-04-07
Ignite 1481  It is possible to configure local cache with affinity function

kcheng-mvp:
bug fix for https://issues.apache.org/jira/browse/IGNITE-1481


kcheng-mvp:
close as it's merged

PR #621, opened 2016-04-07
Ignite 2645: Assertion error in ATOMIC cachce for invokeAll and cache store

ashutakGG:
https://issues.apache.org/jira/browse/IGNITE-2645


dspavlov:
IGNITE-2645 | Closed
-- | --

PR #656, opened 2016-04-19
Ignite-2854 Deadlock detection for pessimistic transactions

agura:


agura:
Merged into master branch.

PR #661, opened 2016-04-21
'key' is a reserved keyword in mysql and needs to be escaped

krosenvold:


akuznetsov-gridgain:
I do not like this fix.
You are using MySql specific back tick escaping syntax that will failed on all other RDBMs.
I think proper fix is to rename key to something else, for example: akey, lead, item....
I agree that it will be hard to find a good synonym for "key". :)

akuznetsov-gridgain:
After some more thinking of this fix, I came to decision that we do not need to change anything at all.
Thouse DFLT_ constants just a some defaults and it is impossible to have defaults that will fit all RDBMs. User can easily fix issue with "key" for MySQL by specifying appropriate SQL statement in CacheJdbcBlobStore via setCreateTableQuery() method.
Let's close this PR as "Won't fix"? Thoughts?

krosenvold:
In its current state any mysql user is going to have the same troubles as I did and it's not entirely trivial to identify and fix this problem. It's just a few hours worth of trouble I won't ever get back, which is also why  I created this PR. 

I agree with your analysis/sentiments mostly.  The thing is, I am really unsure if this class is really a toy class not intended for any serious use or not. Personally I took the stance that this class might have been useful to get something up and running quickly, but it's probably not a good idea in a production environment. But of course, if you're a mysql user it's definitely not going to be quick, you're actually just going to be wasting your time. So I do not think this PR is a good idea; the best real solution is probably to rename the field.

krosenvold:
A reasonable compromise might be to make it simpler to change the field names.

akuznetsov-gridgain:
Ok, lets rename to akey.
Could you update your PR?

akuznetsov-gridgain:
I merged into ignite-1.6 and master.
ignite-1.6 will be released by the end of May
Thanks.

PR #663, opened 2016-04-22
ignite-3041

ascherbakoff:
Implement config variations test for IgniteServices


ascherbakoff:
Moved to gridgain repo.

PR #666, opened 2016-04-22
IGNITE-2893 Atomic structures use entry processors instead of transactions

edouard-chevalier:


dspavlov:
https://issues.apache.org/jira/browse/IGNITE-2893 is resolved, closing PR

PR #676, opened 2016-04-28
ignite-2982

ascherbakoff:
Various fails on batch operations to cache.


ascherbakoff:
Moved development to gridgain repo.

PR #686, opened 2016-05-04
Correct cluster misses aggregation

krnlMustard:
When merging metrics from a cluster, node hits instead of node misses were being aggregated into merged misses.  Thus for clusters the CacheMetrics were reporting the same value for both hits and misses.


krnlMustard:
Realized this is already fixed on master.  Checked wrong branch when I forked.

PR #691, opened 2016-05-06
IGNITE-3023

jcfiorenzano:
Rename keyBackups to Backups for performance advices printed out of a node, due to the name of the property in the configuration file is Backups instead of keyBackups


dspavlov:
https://issues.apache.org/jira/browse/IGNITE-3023 is closed

PR #696, opened 2016-05-06
Ignite services with different class loaders (Ignite 3056)

ashutakGG:
https://issues.apache.org/jira/browse/IGNITE-3056


ashutakGG:
Created new branch and PR based on ignite-1.6 branch instead of master. See https://github.com/apache/ignite/pull/700

PR #713, opened 2016-05-12
ignite-2382 JDBC drivers improvements

agura:


agura:
Merged into ignite-1.6 and master branches.

PR #740, opened 2016-05-21
IGNITE-2929 IgniteContext should not have type parameters

f7753:
Currently implementation of `SparkContext` has type parameters `[K, V]` which means that all the RDDs that are created by a particular instance of context have to be of the same type.
Type parameters on`IgniteContext` is removed.`fromCache` method is parameterized instead.


f7753:
@agoncharuk  Would you please help me to review this PR?

PR #745, opened 2016-05-23
IGNITE-2888

ilantukh:


dspavlov:
@ilantukh  could you please close the PR if you don't want to apply the patch. https://issues.apache.org/jira/browse/IGNITE-2329 is unassigned

PR #761, opened 2016-05-27
IGNITE-3207 Rename IgniteConfiguration.gridName

f7753:
Deprecate IgniteConfiguration.gridName
Add IgniteConfiguration.localInstanceName
Rename related parameters in Ignition class (and other places, if present)
Update Javadoc: clearly state that this name only works locally and has no effect on topology.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
+++ b/modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
@@ -213,7 +213,11 @@ public class IgniteConfiguration {
     public static final Long DFLT_FAILURE_DETECTION_TIMEOUT = new Long(10_000);
 
     /** Optional grid name. */
+    @Deprecated
     private String gridName;
+
+    /** Optional local instance name. */
+    private String localInstanceName;
 
     /** User attributes. */
     private Map<String, ?> userAttrs;
@@ -503,7 +507,7 @@ public class IgniteConfiguration {
         failureDetectionTimeout = cfg.getFailureDetectionTimeout();
         ggHome = cfg.getIgniteHome();
         ggWork = cfg.getWorkDirectory();
-        gridName = cfg.getGridName();
+        localInstanceName = cfg.getLocalInstanceName();
         igfsCfg = cfg.getFileSystemConfiguration();
         igfsPoolSize = cfg.getIgfsThreadPoolSize();
         hadoopCfg = cfg.getHadoopConfiguration();
@@ -562,9 +566,22 @@ public class IgniteConfiguration {
      * @return Optional grid name. Can be {@code null}, which is default grid name, if
      *      non-default grid name was not provided.
      */
+    @Deprecated
     public String getGridName() {
         return gridName;
     }
+
+    /**
+     * Gets optional local instance name. Returns {@code null} if non-default local instance
+     * name was not provided.
+     *
+     * @return Optional local instance name. Can be {@code null}, which is default local
+     * instance name, if non-default local instance name was not provided.
+     */
+    public String getLocalInstanceName() {
+        return localInstanceName;
+    }
+
 
     /**
      * Whether or not this node should be a daemon node.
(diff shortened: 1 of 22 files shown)


modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -503,7 +507,7 @@ public IgniteConfiguration(IgniteConfiguration cfg) {
         failureDetectionTimeout = cfg.getFailureDetectionTimeout();
         ggHome = cfg.getIgniteHome();
         ggWork = cfg.getWorkDirectory();
-        gridName = cfg.getGridName();
```
dmagda: This line mustn't be removed for compatibility reasons.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -562,11 +566,24 @@ public IgniteConfiguration(IgniteConfiguration cfg) {
      * @return Optional grid name. Can be {@code null}, which is default grid name, if
      *      non-default grid name was not provided.
```
dmagda: Add the following line to the Java doc right after @return

"@deprecated Use {@link #setInstanceName} instead.

Refer to IgniteConfiguration.setMarshallerCachePoolSize as an example.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -613,12 +630,19 @@ public IgniteConfiguration setDaemon(boolean daemon) {
      *      grid name.
      * @return {@code this} for chaining.
      */
+    @Deprecated
     public IgniteConfiguration setGridName(String gridName) {
         this.gridName = gridName;
 
         return this;
     }
```
dmagda: Add missing Java doc.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -562,11 +566,24 @@ public IgniteConfiguration(IgniteConfiguration cfg) {
...
 
     /**
+     * Gets optional local instance name. Returns {@code null} if non-default local instance
+     * name was not provided.
+     *
+     * @return Optional local instance name. Can be {@code null}, which is default local
+     * instance name, if non-default local instance name was not provided.
+     */
+    public String getLocalInstanceName() {
```
dmagda: rename here and everywhere from "localInstanceName" to "instanceName" and add more details to Java doc saying that this name has no effect on cluster topology and must be set in cases when several nodes are started in a single JVM process or you just need to set some name to all the nodes.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -613,12 +630,19 @@ public IgniteConfiguration setDaemon(boolean daemon) {
      *      grid name.
      * @return {@code this} for chaining.
      */
+    @Deprecated
```
dmagda: The same as above: add @deprecated comment block to Java doc.

modules/core/src/test/java/org/apache/ignite/testframework/junits/GridAbstractTest.java
```diff
@@ -1240,13 +1240,23 @@ protected IgniteConfiguration getConfiguration(String gridName) throws Exception
     /**
      * @return Generated unique test grid name.
      */
+    @Deprecated
     public String getTestGridName() {
```
dmagda: Remove this method at all please.

f7753:
Thanks, I'll create another PR.

f7753:
If my understanding is not biased, the test code  below:
`/** {@inheritDoc} */
        IgniteConfiguration c = super.getConfiguration(gridName);`

should be changed to `/** {@inheritDoc} */
    @Override protected IgniteConfiguration getConfiguration(String instanceName) throws Exception {
        IgniteConfiguration c = super.getConfiguration(instanceName);`

Any ideas?

CC @dmagda

dmagda:
Yes, you can change it in such a way because you just rename method's parameter "gridName" to "instanceName".

PR #762, opened 2016-05-27
IGNITE-2616

vldpyatkov:
IGNITE-2616
NonHeap memory usage metrics don't work as expected.


modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsSnapshot.java
```diff
@@ -380,6 +383,8 @@ public ClusterMetricsSnapshot(ClusterGroup p) {
 
             nonHeapCommitted += m.getNonHeapMemoryCommitted();
 
+            addCacheOffHeapAllocatedSize((TcpDiscoveryNode)node);
+
             nonHeapUsed += m.getNonHeapMemoryUsed();
```
dmagda: I would put this line into the new method "addCacheOffHeapAllocatedSize" and rename the latest to "nonHeapMemoryUsage".

modules/core/src/test/java/org/apache/ignite/internal/processors/closure/ClusterMetricsOnCacheSelfTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+            CacheConfiguration<Integer, Object> onHeapTieredCfg = cacheConfiguration(gridName);
+            onHeapTieredCfg.setName(ON_HEAP_TIERED_NAME);
+            onHeapTieredCfg.setStatisticsEnabled(true);
+            onHeapTieredCfg.setMemoryMode(CacheMemoryMode.ONHEAP_TIERED);
+            onHeapTieredCfg.setOffHeapMaxMemory(MAX_VALS_AMOUNT * VAL_SIZE);
+
+            FifoEvictionPolicy plc = new FifoEvictionPolicy();
+            plc.setMaxMemorySize(MAX_VALS_AMOUNT * VAL_SIZE);
+            plc.setMaxSize(VAL_SIZE);
```
dmagda: set plc.setMaxSize to -1 so that the eviction considers max memory size only.

modules/core/src/test/java/org/apache/ignite/internal/processors/closure/ClusterMetricsOnCacheSelfTest.java
```diff
@@ -0,0 +1,172 @@
...
+                for (int i = start; i < start + MAX_VALS_AMOUNT; i++)
+                    cache.put(i, val);
+
+                latch.countDown();
+
+                return null;
+            }
+        }, THREAD_COUNT, "test");
+
+        latch.await(timeout, TimeUnit.SECONDS);
```
dmagda: There is no need to way while the cache is propagated. GridTestUtils.runMultiThreaded will block execution of your Thread until all the Threads created by this method are finished. See the source code of GridTestUtils.runMultiThreaded for more details.

modules/core/src/test/java/org/apache/ignite/internal/processors/closure/ClusterMetricsOnCacheSelfTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+        assertEquals(0, prevTieredOffHeapSize);
+        assertEquals(0, prevValuesOffHeapSize);
+
+        long prevClusterNonHeapMemoryUsed = grid(0).cluster().metrics().getNonHeapMemoryUsed();
+
+        fillCache(tiered, getTestTimeout());
+
+        assertTrue(tiered.metrics().getOffHeapAllocatedSize() > (MAX_VALS_AMOUNT - 5) * VAL_SIZE + prevTieredOffHeapSize);
+        assertEquals(prevValuesOffHeapSize, values.metrics().getOffHeapAllocatedSize());
```
dmagda: prevValuesOffHeapSize is equal to 0 here correct? Then you can use 0 directly instead.

modules/core/src/test/java/org/apache/ignite/internal/processors/closure/ClusterMetricsOnCacheSelfTest.java
```diff
@@ -0,0 +1,172 @@
...
+        // prevClusterNonHeapMemoryUsed = 69_277_368
+        // (MAX_VALS_AMOUNT - 5) * VAL_SIZE = 49_807_360
+        // tiered.metrics().getOffHeapAllocatedSize() = 51_012_531
+
+        assertTrue((MAX_VALS_AMOUNT - 5) * VAL_SIZE + prevClusterNonHeapMemoryUsed < grid(0).cluster().metrics().getNonHeapMemoryUsed());
```
dmagda: prevClusterNonHeapMemoryUsed should be 0 here. You can delete it from the calculation for the simplicity of understanding your code.

modules/core/src/test/java/org/apache/ignite/internal/processors/closure/ClusterMetricsOnCacheSelfTest.java
```diff
@@ -0,0 +1,172 @@
...
+            cacheInit = true;
+
+            return cfg.setCacheConfiguration(offHeapValuesCfg, onHeapTieredCfg);
+        }
+
+        return cfg;
+    }
+
+    public void testAllocatedMemory() throws Exception {
+        final IgniteCache tiered = grid(0).getOrCreateCache(ON_HEAP_TIERED_NAME);
```
dmagda: rename to "onHeapCache"

modules/core/src/test/java/org/apache/ignite/internal/processors/closure/ClusterMetricsOnCacheSelfTest.java
```diff
@@ -0,0 +1,172 @@
...
+        final IgniteCache values = grid(0).getOrCreateCache(OFF_HEAP_VALUE_NAME);
```
dmagda: rename to "offHeapCache"

modules/core/src/test/java/org/apache/ignite/internal/processors/closure/ClusterMetricsOnCacheSelfTest.java
```diff
@@ -0,0 +1,172 @@
...
+ */
+public class ClusterMetricsOnCacheSelfTest extends GridCacheAbstractSelfTest {
+    /** */
+    private static final int VAL_SIZE = 512 * 1024; // bytes
+
+    /** */
+    private static final int MAX_VALS_AMOUNT = 100;
+
+    /** */
+    private static final int NODES_CNT = 2;
```
dmagda: Test it using a single node. Otherwise you should consider that metrics are updated once in TcpDiscoverySpi.getHeartbeatFrequency

modules/core/src/test/java/org/apache/ignite/internal/processors/closure/ClusterMetricsOnCacheSelfTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+    /** Init caches only once */
+    private boolean cacheInit = false;
```
dmagda: You don't need this even if you start many nodes in your test. The cache will be defined in the configurations of all the nodes but in any case will be created only once. Please remove this.

modules/core/src/test/java/org/apache/ignite/internal/processors/closure/ClusterMetricsOnCacheSelfTest.java
```diff
@@ -0,0 +1,172 @@
...
+import org.apache.ignite.configuration.CacheConfiguration;
+import org.apache.ignite.configuration.IgniteConfiguration;
+import org.apache.ignite.internal.processors.cache.GridCacheAbstractSelfTest;
+import org.apache.ignite.internal.util.IgniteUtils;
+import org.apache.ignite.testframework.GridTestUtils;
+
+/**
+ * Tests cluster metrics affected by cache state.
+ */
+public class ClusterMetricsOnCacheSelfTest extends GridCacheAbstractSelfTest {
```
dmagda: You don't need to create a special test class. Just add additional test into ClusterNodeMetricsSelfTest

modules/core/src/test/java/org/apache/ignite/internal/processors/closure/ClusterMetricsOnCacheSelfTest.java
vldpyatkov: Correctly set 0 otherwise assert works.

PR #768, opened 2016-05-31
IGNITE-3151

vldpyatkov:
Using IgniteCountDownLatch sometimes drives to dead lock.


modules/core/src/main/java/org/apache/ignite/internal/processors/datastructures/GridCacheCountDownLatchImpl.java
```diff
@@ -249,20 +252,16 @@ private void initializeLatch() throws IgniteCheckedException {
                 internalLatch = CU.outTx(
                     retryTopologySafe(new Callable<CountDownLatch>() {
                         @Override public CountDownLatch call() throws Exception {
-                            try (IgniteInternalTx tx = CU.txStartInternal(ctx, latchView, PESSIMISTIC, REPEATABLE_READ)) {
```
dmagda: You must leave this code as-is because a latch with the same name can be being initialized on several nodes at the same time. This distributed transaction protects from this case.

modules/core/src/main/java/org/apache/ignite/internal/processors/datastructures/GridCacheCountDownLatchImpl.java
```diff
@@ -236,6 +236,9 @@ public GridCacheCountDownLatchImpl(String name,
     @Override public void onUpdate(int cnt) {
         assert cnt >= 0;
 
+        if (initGuard.get())
```
dmagda: Add comments to this code to make it clear for other developers why this code is there.

It can sound this way - "if the latch is being initialized on this node as well we have to wait for initialization completion to avoid possible race conditions"

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/datastructures/partitioned/IgniteCountDownLatchDeadlockTest.java
```diff
@@ -0,0 +1,210 @@
...
+import org.apache.ignite.internal.processors.cache.datastructures.IgniteAtomicsAbstractTest;
+import org.apache.ignite.lang.IgniteRunnable;
+import org.apache.ignite.resources.IgniteInstanceResource;
+
+import java.util.Collection;
+
+/**
+ * The test will fall if deadlock detected.
+ */
+public class IgniteCountDownLatchDeadlockTest extends IgniteAtomicsAbstractTest {
```
dmagda: I would move your test into IgniteCountDownLatchAbstractSelfTest. Probably in that test class you will see how tests for latches can be done easily.

modules/core/src/main/java/org/apache/ignite/internal/processors/datastructures/GridCacheCountDownLatchImpl.java
dmagda: Vlad, lets use the following implementation for "onUpdate" method

------
        CountDownLatch latch0 = internalLatch;

        if (initGuard.get() && latch0 == null) {
            U.awaitQuiet(initLatch);

            latch0 = internalLatch;
        }

        assert latch0 != null;

        while (latch0.getCount() > cnt)
            latch0.countDown();
------

This implementation:
- avoids unnecessary calls to  U.awaitQuiet(initLatch); when the latch is already initialized
- avoids frequent reads from volatile variable

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/datastructures/IgniteCountDownLatchAbstractSelfTest.java
```diff
@@ -334,6 +339,58 @@ public void testLatchMultinode1() throws Exception {
...
+
+        int numOfSrvs = srvsGrp.nodes().size();
+
+        ignite.destroyCache("testCache");
+        IgniteCache<Object, Object> cache = ignite.createCache("testCache");
+
+        for(ClusterNode node: srvsGrp.nodes())
+            cache.put(String.valueOf(node.id()), 0);
+
+        for(int i=0; i<500; i++) {
```
dmagda: There must be a space between an arithmetic or comparison operator and an operand
for(int i=0; i<500; i++) -> for(int i = 0; i < 500; i++)

Check the coding guidelines and setup IntelliJ IDEA plugin from ignite/idea/ignite_codeStyle.xml

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/datastructures/IgniteCountDownLatchAbstractSelfTest.java
```diff
@@ -334,6 +339,58 @@ public void testLatchMultinode1() throws Exception {
...
+            assertTrue(latch2.await(10000));
+        }
+    }
+
+    /**
+     * @param client Ignite client.
+     * @param numOfSrvs Number of server nodes.
+     * @return Ignite latch.
+     */
+    private IgniteCountDownLatch createLatch1(Ignite client, int numOfSrvs) {
```
dmagda: Why do you need to create two different methods when the only difference between them is a latch name?

You can either create method like "createLatch1(Ignite client, int numOfSrvs, latchName)" or just inline creation of a latch into your test method. I prefer the second approach because presently this method body implementation is a one-liner.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/datastructures/IgniteCountDownLatchAbstractSelfTest.java
```diff
@@ -391,4 +448,100 @@ public void testLatchMultinode2() throws Exception {
     @Override public void readExternal(ObjectInput in) throws IOException, ClassNotFoundException {
         // No-op.
     }
+
+    /**
+     * Ignite job
+     */
+    public class IgniteRunnableJob implements IgniteRunnable {
+
```
dmagda: No space here.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/datastructures/IgniteCountDownLatchAbstractSelfTest.java
```diff
@@ -391,4 +448,100 @@ public void testLatchMultinode2() throws Exception {
...
+        /**
+         * Ignite.
+         */
+        @IgniteInstanceResource
+        Ignite igniteInstance;
+
+        /**
+         * Number of iteration.
```
dmagda: /**
   * Number of iteration.
  */

must be changed to 

  /** Number of iterations. */

Apply this for all other places.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/datastructures/IgniteCountDownLatchAbstractSelfTest.java
```diff
@@ -391,4 +448,100 @@ public void testLatchMultinode2() throws Exception {
...
+                if (node.isLocal())
+                    id = String.valueOf(node.id());
+            }
+            return id;
+        }
+
+        /**
+         * @return Ignite.
+         */
+        public Ignite igniteInstance() {
```
dmagda: Why do we need this method here?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/datastructures/IgniteCountDownLatchAbstractSelfTest.java
```diff
@@ -391,4 +448,100 @@ public void testLatchMultinode2() throws Exception {
...
+        public IgniteRunnableJob(IgniteCountDownLatch latch1, IgniteCountDownLatch latch2, int iteration) {
+            this.iteration = iteration;
+            this.latch1 = latch1;
+            this.latch2 = latch2;
+        }
+
+        /**
+         * @return Ignite latch.
+         */
+        IgniteCountDownLatch createLatch1() {
```
dmagda: Why do we need this method and why is it called so? It doesn't create anything.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/datastructures/IgniteCountDownLatchAbstractSelfTest.java
```diff
@@ -391,4 +448,100 @@ public void testLatchMultinode2() throws Exception {
...
+            return latch1;
+        }
+
+        /**
+         * @return Ignite latch.
+         */
+        IgniteCountDownLatch createLatch2() {
+            return latch2;
```
dmagda: Why do we need this method and why is it called so? It doesn't create anything.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/datastructures/IgniteCountDownLatchAbstractSelfTest.java
```diff
@@ -391,4 +448,100 @@ public void testLatchMultinode2() throws Exception {
...
+        }
+
+        /** {@inheritDoc} */
+        @Override
```
dmagda: Override must be on the same line with function name

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/datastructures/IgniteCountDownLatchAbstractSelfTest.java
```diff
@@ -391,4 +448,100 @@ public void testLatchMultinode2() throws Exception {
...
+        public void run() {
+
```
dmagda: No space

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/datastructures/IgniteCountDownLatchAbstractSelfTest.java
```diff
@@ -391,4 +448,100 @@ public void testLatchMultinode2() throws Exception {
...
+            for(ClusterNode node:igniteInstance.cluster().forServers().nodes()) {
+                Integer val = (Integer) cache.get(String.valueOf(node.id()));
+                assertEquals(val, (Integer)iteration);
+            }
+
+            latch1.countDown();
+
+            assertTrue(latch1.await(10000));
+
+            cache.put(getUID(), (iteration + 1));
```
dmagda: I don't get why you need this cache and why you store iterations count per node. 

It's guaranteed that a broadcast closure will be executed on all the nodes from a specified cluster group. Please remove this cache and iteration keeping logic unless you have a significant reason for that.

PR #773, opened 2016-06-02
IGNITE-3226

vldpyatkov:
Load test: iteration over cache partitions using scan queries and performing transactions

### DIFF
--- a/modules/yardstick/config/benchmark-capitalization.properties
+++ b/modules/yardstick/config/benchmark-capitalization.properties
@@ -67,7 +67,7 @@ w=60
 d=300
 
 # Threads count.
-t=64
+t=8
 
 # Sync mode.
 sm=PRIMARY_SYNC
--- a/modules/yardstick/config/benchmark-capitalization.properties
+++ b/modules/yardstick/config/benchmark-capitalization.properties
@@ -1,0 +1,84 @@
+
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+now0=`date +'%H%M%S'`
+
+# JVM options.
+JVM_OPTS=${JVM_OPTS}" -DIGNITE_QUIET=false"
+
+# Uncomment to enable concurrent garbage collection (GC) if you encounter long GC pauses.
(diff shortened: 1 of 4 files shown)


modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/load/IgniteCapitalizationBenchmark.java
```diff
@@ -0,0 +1,277 @@
...
+
+        return clientBuilder.build();
+    }
+
+
+
+    /** {@inheritDoc} */
+    @Override public boolean test(Map<Object, Object> ctx) throws Exception {
+
+        Map<UUID, List<Integer>> partitionsMap = personCachePartitions();
```
dmagda: This can be done once at setUp state. We suppose that the topology won't change during the test.

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/load/IgniteCapitalizationBenchmark.java
```diff
@@ -0,0 +1,277 @@
...
+ * Ignite capitalization benchmark.
+ */
+public class IgniteCapitalizationBenchmark extends IgniteAbstractBenchmark {
+
+    static final String PERSON_CACHE = "person";
+    static final String DEPOSIT_CACHE = "deposit";
+    static SqlFieldsQuery findDepositQuery;
+
+    static {
+        findDepositQuery = new SqlFieldsQuery("SELECT _key FROM \"" + DEPOSIT_CACHE + "\".Deposit WHERE affKey=?").setLocal(true);
```
dmagda: It's better to recreate this object all the time in the test like it's done in practice. There is no need to send it over the wire in the serialized form.

modules/yardstick/config/ignite-capitalization-config.xml
```diff
@@ -0,0 +1,128 @@
...
+            </bean>
+        </property>
+    </bean>
+
+    <bean name="person" class="org.apache.ignite.configuration.CacheConfiguration">
+        <property name="name" value="person"/>
+
+        <property name="cacheMode" value="PARTITIONED"/>
+
+        <property name="atomicityMode" value="ATOMIC"/>
```
dmagda: Cache must be:
- transactional;
- off heap tired;
- number of backups = 1;

modules/yardstick/config/ignite-capitalization-config.xml
```diff
@@ -0,0 +1,128 @@
...
+            </list>
+        </property>
+    </bean>
+
+    <bean name="deposit" class="org.apache.ignite.configuration.CacheConfiguration">
+        <property name="name" value="deposit"/>
+
+        <property name="cacheMode" value="PARTITIONED"/>
+
+        <property name="atomicityMode" value="ATOMIC"/>
```
dmagda: Cache must be:
- transactional;
- off heap tired;
- number of backups = 1;

modules/yardstick/config/benchmark-capitalization.properties
```diff
@@ -0,0 +1,84 @@
...
+b=1
+
+# Warmup.
+w=60
+
+# Duration.
+d=300
+
+# Threads count.
+t=64
```
dmagda: Let's set number of threads equal to 8 by default because the operation (capitalization) that is being tested is quite expensive in terms of CPU usage.

PR #776, opened 2016-06-03
IGNITE-2616

vldpyatkov:
NonHeap memory usage metrics don't work as expected.


modules/core/src/test/java/org/apache/ignite/internal/ClusterNodeMetricsSelfTest.java
```diff
@@ -75,7 +96,84 @@
...
+        plc.setMaxMemorySize(MAX_VALS_AMOUNT * VAL_SIZE);
+        plc.setMaxSize(0);
+
+        onHeapTieredCfg.setEvictionPolicy(plc);
+
+        return cfg.setCacheConfiguration(offHeapValuesCfg, onHeapTieredCfg);
+    }
+
+    /**
+     *
```
dmagda: Missing " @throws Exception If failed."

Follow our documentation guidelines.

modules/core/src/test/java/org/apache/ignite/internal/ClusterNodeMetricsSelfTest.java
```diff
@@ -75,7 +96,84 @@
...
+        prevTieredOffHeapSize = onHeapCache.metrics().getOffHeapAllocatedSize();
+
+        fillCache(offHeapCache, getTestTimeout());
+
+        assertTrue(offHeapCache.metrics().getOffHeapAllocatedSize() > (MAX_VALS_AMOUNT - 5) * VAL_SIZE);
+        assertEquals(prevTieredOffHeapSize, onHeapCache.metrics().getOffHeapAllocatedSize());
+        assertTrue((MAX_VALS_AMOUNT - 5) * VAL_SIZE + prevClusterNonHeapMemoryUsed < ignite.cluster().metrics().getNonHeapMemoryUsed());
+    }
+
+    /** Fill cache with values. */
```
dmagda: Documentation doesn't align with our guidelines. See how the documentation looks like in the tests.

modules/core/src/test/java/org/apache/ignite/internal/ClusterNodeMetricsSelfTest.java
```diff
@@ -75,7 +96,84 @@
...
+
+        onHeapTieredCfg.setEvictionPolicy(plc);
+
+        return cfg.setCacheConfiguration(offHeapValuesCfg, onHeapTieredCfg);
+    }
+
+    /**
+     *
+     */
+    public void testAllocatedMemory() throws Exception {
```
dmagda: Non descriptive test name. Rename to "testNonHeapClusterMetrics"

And this test has to be moved into "ClusterMetricsSelfTest" because it checks **cluster** metrics.
Here you must have the same but for node metrics - ignite.cluster().node(i).getNonHeapMemoryMaximum().

modules/core/src/test/java/org/apache/ignite/internal/ClusterNodeMetricsSelfTest.java
```diff
@@ -75,7 +96,84 @@
         cfg.setCacheConfiguration();
         cfg.setMetricsUpdateFrequency(0);
 
-        return cfg;
+        CacheConfiguration<Integer, Object> offHeapValuesCfg = defaultCacheConfiguration();
```
dmagda: These caches are not needed for every test from the class. Create it dymically in your test's method body.

modules/core/src/test/java/org/apache/ignite/internal/ClusterNodeMetricsSelfTest.java
```diff
@@ -52,6 +60,19 @@
     /** Number of messages. */
     private static final int MSG_CNT = 3;
 
+    /** */
+    private static final int VAL_SIZE = 512 * 1024; // bytes
+
+    /** */
+    private static final int MAX_VALS_AMOUNT = 100;
+    public static final int DEFAUTL_HEARTBEAT = 2_000;
```
dmagda: Comment and space is missed.

modules/core/src/test/java/org/apache/ignite/internal/ClusterNodeMetricsSelfTest.java
```diff
@@ -267,7 +365,7 @@ public boolean apply() {
 
         assert metrics0.getHeapMemoryUsed() > 0;
         assert metrics0.getHeapMemoryTotal() > 0;
-        assert metrics0.getNonHeapMemoryMaximum() > 0;
+//        assert metrics0.getNonHeapMemoryMaximum() > 0;
```
dmagda: Why have you commented out this? I've checked through the code and see that you have to fix the issue for node metrics as well. Presently the fix is done only for cluster metrics. Fix node metrics as well and uncomment this line. Moreover simply extend this test method by adding your code.

modules/core/src/test/java/org/apache/ignite/internal/ClusterNodeMetricsSelfTest.java
```diff
@@ -75,7 +96,84 @@
...
+
+                for (int i = start; i < start + MAX_VALS_AMOUNT; i++)
+                    cache.put(i, val);
+
+
+                return null;
+            }
+        }, THREAD_COUNT, "test");
+
+        IgniteUtils.sleep(2*DEFAUTL_HEARTBEAT);
```
dmagda: Don't do this artificial delays. It slowdowns the whole testing process.

modules/core/src/test/java/org/apache/ignite/internal/ClusterNodeMetricsSelfTest.java
```diff
@@ -75,7 +96,84 @@
...
+        assertTrue((MAX_VALS_AMOUNT - 5) * VAL_SIZE + prevClusterNonHeapMemoryUsed < ignite.cluster().metrics().getNonHeapMemoryUsed());
+    }
+
+    /** Fill cache with values. */
+    private static void fillCache(final IgniteCache<Integer, Object> cache, long timeout) throws Exception{
+        final int THREAD_COUNT = 4;
+        final byte[] val = new byte[VAL_SIZE];
+        final AtomicInteger keyStart = new AtomicInteger(0);
+
+        GridTestUtils.runMultiThreaded(new Callable<Void>() {
```
dmagda: I don't see an significant reason why we need to initialize caches this way. Usually in tests caches are initialized with "cache.put" or "cache.putAll" because the data set is small. Follow the same approach.

modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsSnapshot.java
```diff
@@ -380,7 +383,7 @@ public ClusterMetricsSnapshot(ClusterGroup p) {
 
             nonHeapCommitted += m.getNonHeapMemoryCommitted();
 
-            nonHeapUsed += m.getNonHeapMemoryUsed();
+            nonHeapMemoryUsage(node);
```
dmagda: In fact when you fix this issue for node metrics (node.metrics().getNonHeapMemoryUsed();) you're no longer need to have this method here at all because node.metrics().getNonHeapMemoryUsed(); will return a valid value.

PR #799, opened 2016-06-14
When multiple addresses resolve to same name, TcpDiscoverySharedFsIpFinder fails on unregister

krosenvold:
On my mac, I will have an address that reads <hostname>/ip:port as well as ip:port. These will resolve
to the same filename and subsequently give a stacktrace upon unregister.

Fixed with testcase


dmagda:
Thanks, the changes are merged!

dspavlov:
Closing merged PR

PR #801, opened 2016-06-14
ignite-3209 Waiting for affinity topology in case of failover for affinity call

agura:


agura:
Merged into master branch.

PR #802, opened 2016-06-14
ignite-2680

ascherbakoff:
Terminating running SQL queries.


modules/core/src/main/java/org/apache/ignite/internal/IgniteSchedulerImpl.java
```diff
@@ -109,6 +112,20 @@ public IgniteSchedulerImpl(GridKernalContext ctx) {
         }
     }
```
svladykin: Where is {@inheritDoc}?

PR #812, opened 2016-06-17
IGNITE-1018

AndreyVel:
Node local map is not available when LifecycleEventType.BEFORE_NODE_START is processed in lifecycle bean


dspavlov:
https://issues.apache.org/jira/browse/IGNITE-1018 is closed, closing the PR

PR #831, opened 2016-06-27
ignite-3372: IgniteDataStreamer: pre-loading starvation if multiple streamers preload the same cache

dmagda:


dspavlov:
https://issues.apache.org/jira/browse/IGNITE-3372 is closed

PR #838, opened 2016-06-28
IGNITE-3389 metadata result set throws NPE when closed - fixed

aboudnik:


dspavlov:
https://issues.apache.org/jira/browse/IGNITE-3389 is closed, closing the PR

PR #839, opened 2016-06-29
IGNITE-3391 add param tableName for queryEntity

f7753:
Currently the SQL table name is always the same as class name. In some cases user may want to override this and use different name in queries.
`tableName` property is added to `QueryEntity`.


dspavlov:
https://issues.apache.org/jira/browse/IGNITE-3391

PR #852, opened 2016-07-05
ignite-3426: Compute Engine: job ID is generated in a non unique way

dmagda:


dspavlov:
https://issues.apache.org/jira/browse/IGNITE-3426 is closed

PR #864, opened 2016-07-08
Ignite-3055

vldpyatkov:


modules/core/src/main/java/org/apache/ignite/internal/processors/datastreamer/DataStreamerImpl.java
```diff
@@ -453,6 +458,22 @@ private void leaveBusy() {
     }
 
     /** {@inheritDoc} */
+    @Override public void timeout(long timeout) {
+        if (timeout < -1)
+            throw new IllegalArgumentException();
```
dmagda: Add description to "IllegalArgumentException" saying that this is a wrong timeout value.

modules/core/src/main/java/org/apache/ignite/internal/processors/datastreamer/DataStreamerImpl.java
```diff
@@ -453,6 +458,22 @@ private void leaveBusy() {
...
+
+        if (timeout < 1)
```
dmagda: If timeout is '0' then throw IAE as well.

PR #871, opened 2016-07-17
IGNITE-2527 - fixed

mwadhwa:
Issue:
Calling {{SchedulerFuture.nextExecutionTime()}} (and probably other methods) on a completed future leads to this exception:
{noformat}
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 0
    at org.apache.ignite.internal.processors.schedule.ScheduleFutureImpl.nextExecutionTime(ScheduleFutureImpl.java:455)
{noformat}
In particular, this happens if {{nextExecutionTime()}} right after calling {{IgniteScheduler.scheduleLocal()}} with an invalid pattern (see code example below). The exception is lost in this case and is never printed out or thrown.
{code}
SchedulerFuture schedulerFuture = ignite.scheduler().scheduleLocal(() -> {}, "{55} 53 3/5 \* \* *");
long nextExecutionTime = schedulerFuture.nextExecutionTime();
{code}
Root Cause:
schedulerFuture.nextExecutionTime() is trying to caste Empty Array element (first element) to Long. Thats why It is throwing ArrayOutOfBoundException
schedulerFuture.nextExecutionTime() internally invokes nextExecutionTimes() which in case If scheduler is finished will returns Empty Long Array.
As schedulerFuture.nextExecutionTime() returns Long, It has to caste Array Elements returned from nextExecutionTimes() to Long which is empty in this case. Hence throwing ArrayOutOfBoundException
Resolution:
keeping schedulerFuture.nextExecutionTime() default termination value as -1.


dspavlov:
https://issues.apache.org/jira/browse/IGNITE-2527 is closed, closing the PR

PR #876, opened 2016-07-20
IGNITE-3516

kromulan:
H2 Indexing unregister cache does not call TableDescriptor's GridH2Table close and optionally GridLuceneIndex close, which causes OFFHEAP_TIERED cache to leak memory.


dspavlov:
https://issues.apache.org/jira/browse/IGNITE-3516 is closed, closing the PR

PR #886, opened 2016-07-23
Ignite 2294 SQL DML

alexpaschenko:


modules/core/src/main/java/org/apache/ignite/cache/query/QueryCursor.java
```diff
@@ -17,6 +17,7 @@
 
 package org.apache.ignite.cache.query;
 
+import java.sql.ResultSet;
```
svladykin: what is this?

modules/core/src/main/java/org/apache/ignite/cache/query/SqlFieldsQuery.java
```diff
@@ -42,7 +42,7 @@
  *
  * @see IgniteCache#query(Query)
  */
-public final class SqlFieldsQuery extends Query<List<?>> {
```
svladykin: Why this changed?

modules/core/src/main/java/org/apache/ignite/cache/query/QueryCursor.java
alexpaschenko: Unused, removed

modules/core/src/main/java/org/apache/ignite/cache/query/SqlFieldsQuery.java
alexpaschenko: Reverted back.

svladykin:
Main notices for now:

- Move everything you've added to `IgniteH2Indexing` to separate component like `DmlExecutor`.
- `SqlSplitter` should not be touched at all.

svladykin:
It also makes sense to wait until all other feature branches will be merged to master.

alexpaschenko:
@svladykin,

Regarding some of your comments:

1. Use assertEquals(a, b) instead of assert a.eaquals(b) in all tests
 **There are no keyword based assertions in any new tests that this branch introduces. Please specify where exactly you found those.**
2. JdbcPreparedStatement.executeBatch must throw unsupported exception
**Done, removed tests as well.**
3. JdbcPreparedStatement.ensureArgsSize: if we know exact number of parameters from getNativeStatement(), then we have to check parameter index here.
**We don't know exact number of params here - PreparedStatement interface does not give that number and H2 library is not among dependencies of ignite-core, therefore we can't cast to H2's JdbcPreparedStatement either.**
4. JdbcQueryTask constuctor: why locQry became Boolean?
**Typo, fixed.**
5. JdbcStatement.updateCounterFromQueryResult and updateCnt always use long internally, convert to int if public API needs int.
**Done.**
6. Avoid using UUID.randomUUID()
**Could you please elaborate on this? JdbcStatement has been using randomUUID for a long time, nothing new added here. And besides that, the only uses of randomUUID I see in this diff are in tests.**
7. executeBatch have to throw SQLFeatureNotSupportedException
**Done, removed test as well.**
8. QueryCursorImpl.isResSet rename to isQry **Done.**
9. QueryCursorImpl.forUpdateResult: itemCnt must be long, fieldsMeta must not be empty **Done.**
10. IgniteSQLException.toJdbcException: always wrap getCause() to another SQLException **Done.**

The rest is WIP.

svladykin:
Also lets change `UUID.randomUUID()` to `IgniteUuid.randomUuid()` in V2 jdbc task.

dspavlov:
 
- or ./scripts/apply_pull_request.sh from Apache Ignite source.
 
Feel free to reopen this PR if it is anyway actual.

PR #889, opened 2016-07-25
ignite-3563 Support distributedJoins flag in JDBC driver

agura:


agura:
Merged into master branch.

PR #891, opened 2016-07-25
ignite-3573 Support of Distributed joins queries in REST API

agura:


agura:
Merged into master branch.

PR #925, opened 2016-08-05
IGNITE-3406 - Interceptor and continuous query can get incorrect old …

dkarachentsev:
…value during rebalancing
- Fix interceptor beforePup() on rebalancing.


dspavlov:

PR #927, opened 2016-08-07
IGNITE-3407 HTTP REST: query commands without pageSize failed with NPE

samaitra:


dspavlov:
https://issues.apache.org/jira/browse/IGNITE-3407 is closed, closing the PR

PR #928, opened 2016-08-08
IGNITE-3633 GridCacheUtils - validate hash code presence in binary objs.

alexpaschenko:


modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryObjectEx.java
```diff
@@ -38,4 +38,11 @@
      * @throws BinaryObjectException If failed.
      */
     @Nullable public BinaryType rawType() throws BinaryObjectException;
+
+    /**
+     * Check if flag set.
```
dmagda: There should be a carriage return symbol between method's description and the first @param specified.

modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryClassDescriptor.java
```diff
@@ -826,7 +826,14 @@ private boolean preWrite(BinaryWriterExImpl writer, Object obj) {
      * @param obj Object.
      */
     private void postWrite(BinaryWriterExImpl writer, Object obj) {
-        writer.postWrite(userType, registered, obj instanceof CacheObjectImpl ? 0 : obj.hashCode());
+        if (obj instanceof CacheObjectImpl)
+            writer.postWrite(userType, registered, 0, false);
+        else {
+            boolean flagSet = obj instanceof BinaryObjectEx &&
```
dmagda: I would always pass `true` as a value for `isHashCode` set parameter  if an object being written is neither `CacheObjectImpl` nor `BinaryObjectEx` because otherwise the code that validates key's hash code should fail if some custom key is used and if this key is not created with the builder. You need to add a test to check this.

In short my suggestion is to add the following code:

`if (obj instanceof BinaryObjectEx) {
boolean flagSet = ((BinaryObjectEx)obj).isFlagSet(BinaryUtils.FLAG_EMPTY_HASH_CODE);

writer.postWrite(userType, registered, obj.hashCode(), !flagSet);
}
else {
// the rest of the cases. 
writer.postWrite(userType, registered, obj.hashCode(),  true);
}
`

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheUtilsSelfTest.java
```diff
@@ -104,7 +121,8 @@ public boolean equals(String obj) {
 
     /**
      */
-    public void testCacheKeyValidation() {
+    @SuppressWarnings("ResultOfMethodCallIgnored")
+    public void testCacheKeyValidation() throws IgniteCheckedException {
```
dmagda: Let's add the test and support custom keys that are essentially binary objects. Refer to one of my review note above.

PR #934, opened 2016-08-09
IGNITE-3138 - IgniteDataStreamer: failures are not shown on the strea…

vldpyatkov:
…ming side


modules/core/src/main/java/org/apache/ignite/internal/processors/datastreamer/DataStreamerImpl.java
```diff
@@ -837,20 +845,13 @@ private void doFlush() throws IgniteCheckedException {
...
         for (IgniteInternalFuture<?> f : activeFuts) {
             if (!f.isDone()) {
                 if (activeFuts0 == null)
-                    activeFuts0 = new ArrayList<>((int)(activeFuts.size() * 1.2));
+                    activeFuts0 = new ArrayList<>(activeFuts.size() / 2);
 
                 activeFuts0.add(f);
             }
-            else {
-                f.get();
```
dmagda: You mustn't delete this call (f.get()) because otherwise the user code, that called flush() or close() methods, can miss an Exception that happened during a pre-loading of a batch of data.

Actually current logic is supposed to store the flushing as soon as at least one exception occurs. We can leave it as is for now.

modules/core/src/test/java/org/apache/ignite/internal/processors/datastreamer/DataStreamerFailuresTest.java
```diff
@@ -0,0 +1,176 @@
+package org.apache.ignite.internal.processors.datastreamer;
```
dmagda: Apache Ignite copyright is missing.

modules/core/src/test/java/org/apache/ignite/internal/processors/datastreamer/DataStreamerFailuresTest.java
```diff
@@ -0,0 +1,176 @@
...
+            IgniteFuture[] futures = loadData(client);
+
+            checkFeatures(futures);
+        }
+        finally {
+            stopAllGrids();
+        }
+    }
+
+    private void checkFeatures(IgniteFuture[] futures) {
```
dmagda: Documentation is missing. Also 'checkFeatures' naming doesn't make much sense to me. Do you mean something like 'testFailed'?

modules/core/src/test/java/org/apache/ignite/internal/processors/datastreamer/DataStreamerFailuresTest.java
```diff
@@ -0,0 +1,176 @@
...
+        finally {
+            stopAllGrids();
+        }
+    }
+
+    /**
+     * Test fail on receiver, when streamer invokes from client node.
+     * @throws Exception If fail.
+     */
+    public void testFromClient() throws Exception {
```
dmagda: In addition as I mentioned earlier in my previous review notes let's add the following test case - Try to add one more test when a client is loading the data and all the servers are stopped in the middle.

dspavlov:
IGNITE-3138 | Resolved
-- | --

PR #936, opened 2016-08-09
Ignite 1.6.5

devozerov:


dspavlov:
PR seems to be for testing only, closing the PR

PR #952, opened 2016-08-11
IGNITE-3678: Test for partition distribution

pperalta:
This submission includes a test that will assert the even and safe distribution of partitions when server nodes are forcibly killed.

Additionally, the following are included:

(1) A new `Assertion` interface. This looks similar on the surface to `IgnitePredicate` - however there are two subtle differences:
- it returns `void`, not `boolean`
- it throws `AssertionError`
Since this does not return `boolean`, the caller does not have to check for the result of the assertion; if the assertion fails an exception is thrown which contains the details of the failed assertion.

(2) `GridRollingRestartAbstractTest`: a new abstract test class that is extended for tests that require:
- multiple JVMs
- the first JVM to be a client (i.e. not own any partitions)
- the server JVMs to be killed forcibly (not gracefully) while the test executes

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/GridJavaProcess.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/GridJavaProcess.java
@@ -191,7 +191,10 @@ public final class GridJavaProcess {
 
         killProc.waitFor();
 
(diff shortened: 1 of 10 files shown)


pperalta:
Merged as 632ff9c0098e1d62c106cc6a2c3b3c49ab1005d7

PR #998, opened 2016-08-27
fix: Unexpected exception during cache update  java.lang.ArrayIndexOu…

wjw465150:
Unexpected exception during cache update java.lang.ArrayIndexOutOfBound

java array use Integer range,this field is:  
/*\* Stripes. */ 
    @GridToStringExclude 
    private final LruStripe[] lrus; 

Integer overflow!


wjw465150:
addIdx and addIdx overflow!

PR #1001, opened 2016-08-29
IGNITE-3601 Read-only optimistic transaction shouldn't throw an excep…

vldpyatkov:
…tion if entry version was not changed.

Read only serializable transactions are commiting without check.


dspavlov:
IGNITE-3601 | Resolved
-- | --

PR #1010, opened 2016-08-30
IGNITE-3808 Reproducer for missing EntryProcessr invocation

pperalta:
This test reproduces [IGNITE-3808.](https://issues.apache.org/jira/browse/IGNITE-3808) It is intermittent; every 3 or 4 invocations the test will fail with a message like this:

```
junit.framework.AssertionFailedError: Unexpected clear bit: 180424
```


AMashenkov:
Sorry, test is ok. But I can't to reproduce test failure on x86. Do you use different arch to run test?

dspavlov:
IGNITE-3808 | Resolved
-- | --

PR #1017, opened 2016-09-01
IGNITE-3809: Fix for ArrayIndexOutOfBoundsException in GridUnsafeLru.

shroman:


wjw465150:
the same with:https://github.com/apache/ignite/pull/998

modules/core/src/main/java/org/apache/ignite/internal/util/offheap/unsafe/GridUnsafeLru.java
```diff
@@ -156,7 +156,7 @@ long entry(short order, long qAddr) {
      * @throws GridOffHeapOutOfMemoryException If failed.
      */
     long offer(int part, long addr, int hash) throws GridOffHeapOutOfMemoryException {
-        return lrus[addIdx.getAndIncrement() % cnt].offer(part, addr, hash);
+        return lrus[U.safeAbs(addIdx.getAndIncrement()) % cnt].offer(part, addr, hash);
```
wjw465150: there will add U.safeAbs!
long prePoll() {
 -        int idx = rmvIdx.getAndIncrement();		 +        int idx = U.safeAbs(rmvIdx.getAndIncrement());
  		  
          // Must try to poll from each LRU.		          // Must try to poll from each LRU.
          for (int i = 0; i < lrus.length; i++) {		          for (int i = 0; i < lrus.length; i++) {
 @@ -625,4 +625,4 @@ private void entry(long qAddr, long addr) {
              return S.toString(LruStripe.class, this);		              return S.toString(LruStripe.class, this);
          }		          }
      }		      }

shroman:
Thanks for pointing to rmvIdx!
Yes, we have faced the same problem. Is IGNITE-3809 is a duplicate of the same issue? I cannot find it on JIRA.

PR #1059, opened 2016-09-14
IGNITE-3897 Test transactional behaviour of caches during rolling restart

leveyj:
This submission includes 2 new tests for the IgniteCacheFailoverTestSuite that validate the results of an aggregation function running against a cluster undergoing rolling restart.

Also, the following supporting Assertions are included:
(1) EventualAssertion - based on the Assertion interface added in IGNITE-3678, this assertion has a timeout and retests the assertion with exponential back-off until successful, or failure.
(2) CacheNodeSafeAssertion - an assertion to make the rolling restart of nodes deterministic, in that it ensures that the primary and backup partitions are distributed such that no data is lost.


yzhdanov:
Hi John!

Thank you for your contribution! I have just had a chance to reivew it.

First question. Why did you file this issue? Was there a discussion on dev list? Can you porvide a link if any?

I fixed some naming/formatting issues in GridCacheRebalancingAggregationTest and pushed everything to ignite-3897-pr1059. Please review and consider going furhter.

As far as GridCacheRebalancingAggregationTest - it obviously has a bug. If on some random iteration update thread throws exception, test does not fail. This should fail the test for sure.

As a general note I would say that when you dealing with restart/failover tests one thread doing updates is not enough. You should send massive and variable load to your tests topology from multiple threads prefferrably having contended keys between them.

I would aslo encourage you to take a look at our current failover threads. It seems we already have tests of the kind:
IgniteCacheCrossCacheTxFailoverTest
GridCacheTxNodeFailureSelfTest
etc

Please let me know if you have any questions.

leveyj:
Hi Yakov,
Thank you for the feedback - I will look at ignite-3897-pr1059 and update this PR with the changes, and also look at the failure of GridCacheRebalancingAggregationTest to fail on Exception - thank you for spotting that.

Just a little background - these restart tests began with https://issues.apache.org/jira/browse/IGNITE-3789 - these tests `kill -9 <pid>` a node in the grid, rather than sending a Callable to shut down nodes gracefully. We find that this type of test can uncover unusual/unintended behaviours. 

 IgniteCacheCrossCacheTxFailoverTest and GridCacheTxNodeFailureSelfTest don't override `GridAbstractTest.isMultiJvm()` to start external processes for testing, and GridAbstractTest doesn't currently have the ability to kill nodes - this ability was added in GridRollingRestartAbstractTest#stopGrid(int idx)

If these new tests we're adding are duplicating the existing failover tests, perhaps I could modify the existing tests so they start nodes as separate processes and randomly stop nodes with `kill -9` instead of a graceful `stop` callable? Or perhaps I can keep these as a distinct type of failover test, and I will try to reuse the existing tests - simply run them in the context of a grid undergoing "RollingRestart" (which is achieved by subclassing GridRollingRestartAbstractTest).

Let me know what you think would be the best way to get test coverage in this area (whether by reusing GridRollingRestartAbstractTest, or modifying the existing failover tests to allow them to be run while a rolling restart is in progress). 
Thanks!
John

leveyj:
Hi @yzhdanov , I've merged your review feedback into the branch. I added code to ensure the update thread logs any exceptions and also the test checks that the thread exited without encountering an exception. 

If the updateThread exited, either `while (restartThread.isAlive());` or an exception occurred, so the test that `assertEquals(getMaxRestarts(), restartThread.getRestartTotal());` was trying to ensure that updateThread exited due to restartThread/test finishing, rather than an exception - but I've made it test explicitly that the updateThread hasn't encountered an exception - it's much cleaner/clearer this way, thanks for questioning it. 

Cheers
John

PR #1122, opened 2016-09-27
IGNITE-3957: clear swap space when cache is destroyed

amartianov:


dspavlov:
IGNITE-3957 | Resolved
-- | --

PR #1170, opened 2016-10-18
IGNITE-3448: Wrong count returned by count distinct and similar queries.

AMashenkov:
Fixed. 
Aggregates with DISTINCT work correctly with non collocated grouping.

### DIFF
--- a/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/GridSqlQuerySplitter.java
+++ b/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/GridSqlQuerySplitter.java
@@ -302,8 +302,10 @@ public class GridSqlQuerySplitter {
 
         boolean hasDistinctAggregate = false;
 
-        for (int i = 0, len = mapExps.size(); i < len; i++)
-            hasDistinctAggregate |= hasAggregates(mapExps.get(i), true);
+        if (!collocatedGroupBy) {
+            for (int i = 0, len = mapExps.size(); i < len; i++)
+                hasDistinctAggregate |= hasAggregates(mapExps.get(i), true);
+        }
 
         boolean aggregateFound = hasDistinctAggregate;
 
--- a/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/GridSqlQuerySplitter.java
+++ b/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/GridSqlQuerySplitter.java
@@ -39,6 +39,8 @@ import static org.apache.ignite.internal.processors.query.h2.opt.GridH2Collocati
 import static org.apache.ignite.internal.processors.query.h2.sql.GridSqlFunctionType.AVG;
 import static org.apache.ignite.internal.processors.query.h2.sql.GridSqlFunctionType.CAST;
 import static org.apache.ignite.internal.processors.query.h2.sql.GridSqlFunctionType.COUNT;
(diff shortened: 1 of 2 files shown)


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/GridSqlQuerySplitter.java
```diff
@@ -702,12 +718,15 @@ private static boolean splitSelectExpression(List<GridSqlElement> mapSelect, Lis
      * @param el Expression.
      * @return {@code true} If expression contains aggregates.
      */
-    private static boolean hasAggregates(GridSqlElement el) {
-        if (el instanceof GridSqlAggregateFunction)
+    private static boolean hasAggregates(GridSqlElement el, boolean distinct) {
```
svladykin: Where is javadoc param description for `distinct`? Also I don't like the naming of this new parameter: even when looking at the code I can't clearly understand what this parameter means. It would be even better to move the whole expression after `el instanceof GridSqlAggregateFunction &&` to a separate method with clear name and javadoc.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/GridSqlQuerySplitter.java
```diff
@@ -653,14 +669,14 @@ private static boolean splitSelectExpression(List<GridSqlElement> mapSelect, Lis
...
-        if (!collocated && hasAggregates(el)) {
+        if (!collocated && hasAggregates(el, false)) {
             aggregateFound = true;
 
             if (alias == null)
                 alias = alias(isHaving ? HAVING_COLUMN : columnName(idx), el);
 
             // We can update original alias here as well since it will be dropped from mapSelect.
-            splitAggregates(alias, 0, mapSelect, idx, true);
+            splitAggregates(alias, 0, mapSelect, idx, hasDistinctAggregate /* && !collocated*/, true);
```
svladykin: Why we have this commented part in the code?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/GridSqlQuerySplitter.java
```diff
@@ -773,53 +796,81 @@ private static void splitAggregate(
...
+                    rdcAgg = aggregate(agg.distinct(), agg.type()).addChild(column(mapAggAlias.alias()));
+                }
 
                 break;
 
             case COUNT_ALL: // CAST(SUM( COUNT(*) ) AS BIGINT)
-            case COUNT: // CAST(SUM( COUNT(x) ) AS BIGINT)
-                mapAgg = aggregate(agg.distinct(), agg.type()).resultType(GridSqlType.BIGINT);
+            case COUNT: // CAST(SUM( COUNT(x) ) AS BIGINT) or CAST(COUNT(DISTINCT x) AS BIGINT)
+
```
svladykin: No need for empty line

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/GridSqlQuerySplitter.java
```diff
@@ -773,53 +796,81 @@ private static void splitAggregate(
...
+                if (hasDistinctAggregate) /* and has no collocated group by */ {
+                    assert agg.type == COUNT;
```
svladykin: Why do we use field access here instead of a method call?

PR #1175, opened 2016-10-20
Ignite 514

abeliak:
Add sorting procedure if client call putAll method with non sorted map. Work with only binary or only objects keys.


dspavlov:
IGNITE-514 | Closed
-- | --

PR #1194, opened 2016-10-28
Ignite sql builder

andreevym:
I want to make useful methods when I have object with unnecessary parameters (if age == null). I want to make fast and easy ignite sql query.

Add test for SqlBuilderSelfTest. refactoring - in test need to have Only test, not one of something unknown. assert need to have in testMethod, not in other class method. move SqlSelfTest in sql package for easy to find and use

example
SqlBuilder.create().more("age",10).build(PersonTest.class)


andreevym:
God please, you see this "pull request" and please merge in "master" branch.
Amen.
amen

PR #1347, opened 2016-12-14
Ignite-4429: Deadlock in IgniteAtomicSequence when used inside transaction

ascherbakoff:
Fixed.


dspavlov:
IGNITE-4429 | Resolved
closing PR

PR #1398, opened 2017-01-04
IGNITE-4519 Updating ignite-aws java sdk pom version

chandresh-pancholi:
I have tested this with AWS S3 account. All the tests were passing.

### DIFF
--- a/modules/cloud/pom.xml
+++ b/modules/cloud/pom.xml
@@ -33,7 +33,7 @@
     <url>http://ignite.apache.org</url>
 
     <properties>
-        <jcloud.version>1.9.0</jcloud.version>
+        <jcloud.version>2.0.0</jcloud.version>
     </properties>
 
     <dependencies>
@@ -52,13 +52,13 @@
         <dependency>
             <groupId>org.apache.jclouds.labs</groupId>
             <artifactId>google-compute-engine</artifactId>
-            <version>${jcloud.version}</version>
+            <version>1.9.3</version>
         </dependency>
 
         <dependency>
             <groupId>org.apache.jclouds.labs</groupId>
             <artifactId>docker</artifactId>
-            <version>${jcloud.version}</version>
+            <version>1.9.3</version>
         </dependency>
 
         <dependency>
--- a/modules/gce/pom.xml
+++ b/modules/gce/pom.xml
@@ -44,13 +44,13 @@
         <dependency>
             <groupId>com.google.api-client</groupId>
             <artifactId>google-api-client</artifactId>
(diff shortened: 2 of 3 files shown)


modules/cloud/pom.xml
```diff
@@ -33,7 +33,7 @@
     <url>http://ignite.apache.org</url>
 
     <properties>
-        <jcloud.version>1.9.0</jcloud.version>
+        <jcloud.version>2.0.0</jcloud.version>
```
zshamrock: What is the reason of declaring the property here, and not to use it in the dependencies above? And why they are different? I think this is confusing at least, if no explanatory comments are added describing why it is that.

modules/cloud/pom.xml
```diff
@@ -52,13 +52,13 @@
         <dependency>
             <groupId>org.apache.jclouds.labs</groupId>
             <artifactId>google-compute-engine</artifactId>
-            <version>${jcloud.version}</version>
+            <version>1.9.3</version>
```
zshamrock: Why this version is different from the defined `jcloud.version` property above?

modules/cloud/pom.xml
```diff
@@ -52,13 +52,13 @@
...
         </dependency>
 
         <dependency>
             <groupId>org.apache.jclouds.labs</groupId>
             <artifactId>docker</artifactId>
-            <version>${jcloud.version}</version>
+            <version>1.9.3</version>
```
zshamrock: Why this version is different from the defined `jcloud.version` property above?

modules/cloud/pom.xml
chandresh-pancholi: Earlier <jcloud.version> was same for every dependencies. But with latest version upgrade allcompute dependency is 2.0.0 and others are 1.9.3.
That was the reason of change.

modules/cloud/pom.xml
chandresh-pancholi: Please check above comment.

modules/cloud/pom.xml
zshamrock: Just wondering, why not that add another property, like `<jcloud.compute.version>1.9.3</jcloud.compute.version>`, and use it instead of duplicated `1.9.3` literals below? And it will also be auto-documented, at least it will be clear what is for.

Also if `jcloud.version` property is not used, I guess, it then could be removed from the `properties` all together.

What do you think?

modules/cloud/pom.xml
chandresh-pancholi: #1402 Pushed the changes

PR #1404, opened 2017-01-08
IGNITE-4530 Add support to provide AWSCredentialsProvider to TcpDisco…

zshamrock:
…veryS3IpFinder in addition to just AWSCredentials

Addressing https://issues.apache.org/jira/browse/IGNITE-4530.

Just to review by the maintainers whether it meets all the project criteria.

### DIFF
--- a/modules/aws/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/s3/TcpDiscoveryS3IpFinder.java
+++ b/modules/aws/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/s3/TcpDiscoveryS3IpFinder.java
@@ -21,7 +21,6 @@ import com.amazonaws.AmazonClientException;
 import com.amazonaws.ClientConfiguration;
 import com.amazonaws.auth.AWSCredentials;
 import com.amazonaws.auth.AWSCredentialsProvider;
-import com.amazonaws.auth.AWSStaticCredentialsProvider;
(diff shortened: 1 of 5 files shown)


modules/aws/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/s3/TcpDiscoveryS3IpFinder.java
```diff
@@ -310,15 +312,27 @@ public void setClientConfiguration(ClientConfiguration cfg) {
...
+     * Sets AWS credentials. Either use {@link #setAwsCredentials(AWSCredentialsProvider)} or this one.
      * <p>
      * For details refer to Amazon S3 API reference.
      *
      * @param cred AWS credentials.
      */
     @IgniteSpiConfiguration(optional = false)
     public void setAwsCredentials(AWSCredentials cred) {
-        this.cred = cred;
+        this.credProvider = new AWSStaticCredentialsProvider(cred);
```
dmagda: I would pass AWSCredentials directly to AmazonS3Client object letting Amazon lib to decide what kind of provider to use. It's clear that Amazon presently uses AWSStaticCredentialsProvider by default but the logic might change in the future and Ignite needs to be as robust as possible.

dspavlov:
https://issues.apache.org/jira/browse/IGNITE-4530 is closed, closing the PR

PR #1408, opened 2017-01-09
IGNITE-4374

daradurvs:
Check of "Basic JVM Configuration" added.
Link to the tuning article added.
https://apacheignite.readme.io/docs/jvm-and-system-tuning


modules/core/src/main/java/org/apache/ignite/internal/suggestions/JvmPerformanceSuggestions.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    @NotNull static List<String> getRecommendedOptions() {
+        List<String> options = new LinkedList<>();
+        // option '-server' isn't in input arguments
+        if (!U.jvmName().toLowerCase().contains("server"))
+            options.add(SERVER);
+
+        List<String> args = U.jvmArgs();
+
+        if (!args.stream().anyMatch(o -> o.startsWith(XMX)))
```
yzhdanov: 1. Project needs to be buildable under jdk7
2. -Xmx has the same effect as -mx, same for -Xms and -ms

modules/core/src/main/java/org/apache/ignite/internal/suggestions/JvmOptions.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.suggestions;
+
+/**
+ * Java Virtual Machine options.
+ */
+class JvmOptions {
```
yzhdanov: I don't like the idea of separate class when it is used from single place only. Can we inline this to ConfigurationSuggestions class?

modules/core/src/main/java/org/apache/ignite/internal/suggestions/JvmPerformanceSuggestions.java
```diff
@@ -0,0 +1,67 @@
...
+        if (!args.stream().anyMatch(o -> o.startsWith(MAX_DIRECT_MEMORY_SIZE)))
+            options.add(MAX_DIRECT_MEMORY_SIZE + "=size[g|G|m|M|k|K]");
+
+        if (!args.contains(USE_TLAB))
+            options.add(USE_TLAB);
+
+        if (!args.contains(DISABLE_EXPLICIT_GC))
+            options.add(DISABLE_EXPLICIT_GC);
+
+        if (!args.contains(USE_COMPRESSED_OOPS))
```
yzhdanov: compressed pointers will be used if heap size is below 32g

modules/core/src/main/java/org/apache/ignite/internal/suggestions/ConfigurationSuggestions.java
```diff
@@ -31,10 +32,12 @@
 /**
  * Grid performance suggestions.
  */
-public class GridPerformanceSuggestions {
+public class ConfigurationSuggestions {
     /** */
     private static final boolean disabled = Boolean.getBoolean(IGNITE_PERFORMANCE_SUGGESTIONS_DISABLED);
 
+    private static final String SUGGESTIONS_LINK = "https://apacheignite.readme.io/docs/jvm-and-system-tuning";
```
yzhdanov: I think we should suggest user to configure gc logging and doing heap dump on outofmemory as described at this page.

Also, can we check if OS is CentOS based and check as much OS properties as we can?

daradurvs: > Also, can we check if OS is CentOS based and check as much OS properties as we can?
Yes, I can do it, but I think will better do it within the separate task where it will be described.

> ..configure gc logging...
Check on existence "-Xloggc"?

daradurvs:
Commited into master-branch.

PR #1419, opened 2017-01-10
IGNITE-4531 Use correct property name in BinaryProperty.

alexpaschenko:


dspavlov:
IGNITE-4531 | Resolved
-- | --

PR #1436, opened 2017-01-18
IGNITE-4501: Improvement of connection in a cluster of new node

SharplEr:
-- Add "region id" implementation: RegionNodeComparator which trying to use the node's attribute "CLUSTER_REGION_ID" to ordering in the ring, if any.
-- Modified TcpDiscoveryNodesRing#nextNode for using nodeComparator for find next node.
-- Add tests for RegionNodeComparator and any future nodeComparator implementation.
-- Copy tests from TcpDiscoverySelfTest to RegionTcpDiscoverySelfTest but with using CLUSTER_REGION_ID
-- Add test RegionTcpDiscoveryStressTest for testing correct ring building.


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/RegionNodeComparator.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Comparator;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * New behavior for node ordering. Sort firstly by region ID, and secondly by topology version.
+ * Node without region ID (it is equal null) is less the node with the region ID.
+ */
+public class RegionNodeComparator implements Comparator<TcpDiscoveryNode> {
+    /**
+     * {@inheritDoc}
```
yzhdanov: In Ignite's code inherit doc is usually single-line comment

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/RegionNodeComparator.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    @Override
```
yzhdanov: @Override should go on the same line in this case

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/RegionNodeComparator.java
```diff
@@ -0,0 +1,46 @@
...
+    public int compare(@NotNull TcpDiscoveryNode firstNode, @NotNull TcpDiscoveryNode secondNode) {
+        Long firstRegionId = firstNode.getClusterRegionId();
+        Long secondRegionId = secondNode.getClusterRegionId();
+
+        if (firstRegionId == null && secondRegionId == null)
```
yzhdanov: There are plenty of formatting issues - please refer to https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNode.java
```diff
@@ -174,8 +177,7 @@ public TcpDiscoveryNode(UUID id,
         int discPort,
         DiscoveryMetricsProvider metricsProvider,
         IgniteProductVersion ver,
-        Serializable consistentId)
```
yzhdanov: Not critical but I would move ") {" to a new line

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNode.java
```diff
@@ -631,6 +649,15 @@ public boolean isCacheClient() {
         clientRouterNodeId = U.readUuid(in);
     }
 
+    /**
+     * Return cluster region id.
+     *
+     * @return Cluster region id or null if not set.
+     */
+    public @Nullable Long getClusterRegionId() {
```
yzhdanov: 1. First annotation, then modifiers
2. Why getClusterRegionId() is java.lang.Long but not primitive? Would not it be fine if it was 0 for cases if it is not set?

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNodesRing.java
```diff
@@ -29,15 +30,7 @@
...
 
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.NavigableSet;
-import java.util.TreeSet;
-import java.util.UUID;
```
yzhdanov: we use per-class import only

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNode.java
```diff
@@ -251,6 +253,22 @@ public void lastSuccessfulAddress(InetSocketAddress lastSuccessfulAddr) {
      */
     public void setAttributes(Map<String, Object> attrs) {
         this.attrs = U.sealMap(attrs);
+        Object id = this.attrs.get("CLUSTER_REGION_ID");
```
yzhdanov: CLUSTER_REGION_ID can only be set via attributes which is not convenient.

User should have an opportunity to set it:
1. via IgniteConfiguration.setClusterRegionId(long)
2. via system property - IgniteSystemProperties.IGNITE_CLUSTER_REGION_ID - which is very flexible and allows me to pass the ID even to ignite.sh
3. I would remove public attribute name.

PR #1473, opened 2017-01-27
IGNITE-3196

daradurvs:
Ideally, we shouldn't check sign at serialization, because the used approach:
BigInteger intVal = val.unscaledValue();
byte[] vals = intVal.toByteArray();
#toByteArray() - already including at least one sign bit, which is (ceil((this.bitLength() + 1)/8)). (This representation is compatible with the (byte[]) constructor.)
Therefore, at deserialization we just read  byte[] vals and scale, also we use default constructor which will define a sign from byte[] vals.

But for .NET & C++ compatibility support, we always serialize only positive numbers, also we write a sign in the first byte, because it is always positive.

We can't use the scale for save a sign, because it can be the negative in Java.


daradurvs:
Commited into ignite-2.0

PR #1499, opened 2017-02-06
ignite-4646

gvvinblade:


modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridChunkedMessageUtils.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.ignite.plugin.extensions.communication.MessageWriter;
+import sun.nio.ch.DirectBuffer;
+
+
+import java.nio.ByteBuffer;
+
+/**
+ * Helper methods to deal with chunked messages
+ */
+public abstract class GridChunkedMessageUtils {
```
gvvinblade: A better name is needed
Probably wrong place

PR #1507, opened 2017-02-08
IGNITE-4663: skip class loading when deserialize == false

tledkov-gridgain:


voipp:
why do u have so many commits in one pull request??

tledkov-gridgain:
Thanks for your comment. I've fixed the base branch

PR #1560, opened 2017-02-21
IGNITE-4372 Ids quoting fix

alexpaschenko:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/DmlAstUtils.java
```diff
@@ -109,6 +109,8 @@ public static GridSqlQuery selectForInsertOrMerge(GridSqlColumn[] cols, List<Gri
                     args[i].addChild(row[i]);
             }
 
+            sel.getSQL();
```
svladykin: Looks like this must be removed.

PR #1572, opened 2017-02-23
automatically detect IGNITE_HOME in service.sh

lammic:
it would be handy to have service.sh detect IGNITE_HOME, without relying on other files to be sourced before calling it


ribamar-santarosa:
No plans to review this PR?

PR #1581, opened 2017-02-28
IGNITE-1094 Ignite.createCache(CacheConfiguration) hangs if some exception occurs during cache initialization

voipp:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/IgniteDynamicCacheStartSelfTest.java
```diff
@@ -1336,4 +1339,26 @@ public void testStartStopSameCacheMultinode() throws Exception {
...
+    /**
+     *
+     */
+    public void testBrokenStoreFactory() {
+        final CacheConfiguration<Integer, String> cacheConfiguration = new CacheConfiguration<>();
+
+        cacheConfiguration.setName("brokenCache");
+        cacheConfiguration.setCacheStoreFactory(new BrokenStoreFactory());
+
+        Assert.assertNull(grid(0).createCache(cacheConfiguration));
```
yzhdanov: Ignite.createCache() should never return null. It should return cache instance or throw exception. In your case this should be an exception thrown from factory. And cache should be stopped on all nodes.

How many nodes are in your test? 1? I would prefer you test such critical changes with more than 1 node.

voipp: ok

PR #1603, opened 2017-03-08
IGNITE-4211

daradurvs:


modules/spring/src/main/java/org/apache/ignite/cache/spring/SpringCache.java
```diff
@@ -74,6 +75,25 @@
...
+    @SuppressWarnings("unchecked")
+    @Override public <T> T get(Object key, Callable<T> valueLoader) {
+        Object val = cache.get(key);
+
+        if (val == null) {
+            try {
+                val = valueLoader.call();
+            }
+            catch (Exception ignored) {
+                return null;
```
voipp: may be throw some sensible exception?

daradurvs: valueLoader.call() - throws an java.lang.Exception if unable to compute a result.

PR #1642, opened 2017-03-20
Ignite-4828

endian675:
Improve the distribution of keys within partitions


endian675:
some kind of Git screw-up

PR #1651, opened 2017-03-21
ignite-4211

daradurvs:
PR after removing ignite-2.0


daradurvs:
Partially merged to master.

PR #1652, opened 2017-03-21
IGNITE-3682 GridFunc: move all inner anonymous classes to separate top-level classes

daradurvs:
PR after removing ignite-2.0


daradurvs:
Merged to the maser branch.

PR #1676, opened 2017-03-27
IGNITE-4501: Improvement of connection in a cluster of new node

SharplEr:
Reopen that PR: https://github.com/apache/ignite/pull/1436


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNodesRing.java
```diff
@@ -260,12 +279,8 @@ public long maxInternalOrder() {
         rwLock.readLock().lock();
 
         try {
-            if (maxInternalOrder == 0) {
```
yzhdanov: Please explain this change

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNodesRing.java
```diff
@@ -393,7 +421,7 @@ public void clear() {
                 nodesMap.put(locNode.id(), locNode);
 
             nodeOrder = 0;
-            maxInternalOrder = 0;
+            maxInternalOrder = locNode.internalOrder();
```
yzhdanov: seems incorrect to me. clear() should reset everything

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -545,6 +545,12 @@
     public static final String IGNITE_MEMORY_PER_BYTE_COPY_THRESHOLD = "IGNITE_MEMORY_PER_BYTE_COPY_THRESHOLD";
 
     /**
+     * Cluster region ID used for sorting nodes in ring for discovery.
+     * <p>
+     * Defaults to {@code} Long.MIN_VALUE,
```
SharplEr: Okay.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -474,6 +474,9 @@
     /** */
     private boolean lateAffAssignment = DFLT_LATE_AFF_ASSIGNMENT;
 
+    /** */
+    private long clusterRegionId = Long.MIN_VALUE;
```
SharplEr: I'm thinking it's more predictable behavior: all nodes with default configuration less or equal than nodes with region id.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -576,6 +579,7 @@ public IgniteConfiguration(IgniteConfiguration cfg) {
         utilityCachePoolSize = cfg.getUtilityCacheThreadPoolSize();
         waitForSegOnStart = cfg.isWaitForSegmentOnStart();
         warmupClos = cfg.getWarmupClosure();
+        clusterRegionId = cfg.getClusterRegionId();
```
SharplEr: Okay.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNode.java
```diff
@@ -519,7 +529,7 @@ public void onClientDisconnected(UUID newId) {
      */
     public TcpDiscoveryNode clientReconnectNode(Map<String, Object> nodeAttrs) {
         TcpDiscoveryNode node = new TcpDiscoveryNode(id, addrs, hostNames, discPort, metricsProvider, ver,
-            null);
+            null, clusterRegionId);
```
SharplEr: I guess we don't need to watch that.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNodesRing.java
SharplEr: Right now nodes.last() don't given nodes with maximum of internal order, because I changed order. So I update maxInternalOrder more often in other place. Also nodes.last()==null branch is unreachable.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNodesRing.java
SharplEr: .clear() put locNode in nodes. So in old version of .maxInternalOrder() we called nodes.last() and got this the new maxInternalOrder. I just do it immediately here. I think it's better than create two TreeSet with a same values but in difference ordering.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNodesRing.java
```diff
@@ -483,21 +509,23 @@ public void clear() {
...
-
-            while (iter.hasNext()) {
-                TcpDiscoveryNode node = iter.next();
-
-                if (locNode.equals(node))
-                    break;
+            Iterator<TcpDiscoveryNode> filtered = serverNodes(excluded, locNode).iterator();
+            if (filtered.hasNext()) {
+                return filtered.next();
+            }
```
SharplEr: My mistake.

PR #1728, opened 2017-04-04
IGNITE-4022

vldpyatkov:
IgniteServices soesn't throw an exception if there are no server nodes


modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
```diff
@@ -999,6 +999,8 @@ private void reassign(GridServiceDeployment dep, AffinityTopologyVersion topVer)
                             }
                         }
                     }
+                    else
+                        log.warning("The service " + cfg.getName() + " was not have assigned nodes.");
```
dmekhanikov: The message is not grammatically correct. `"The service " + cfg.getName() + " wasn't assigned to any nodes."` sounds better. Also `U.warn()` method should be used here.

PR #1749, opened 2017-04-06
IGNITE-4524: Index direct lookup for SQL MIN() and MAX()

skalashnikov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/database/tree/BPlusTree.java
```diff
@@ -245,7 +245,12 @@
             g.backId(0L); // Usually we'll go left down and don't need it.
 
             int cnt = io.getCount(pageAddr);
-            int idx = findInsertionPoint(io, pageAddr, 0, cnt, g.row, g.shift);
+
+            int idx;
+            if (g.findLast)
+                idx = io.isLeaf()? cnt - 1: -cnt - 1; //(-cnt - 1) mimics not_found result of findInsertionPoint
```
svladykin: Looks wrong. The adjustment `-cnt - 1` is done on line 266 ` idx = fix(idx);`

svladykin: Please also add comments about `cnt == 0` case.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/database/tree/BPlusTree.java
```diff
@@ -2212,13 +2278,19 @@ private long doAskNeighbor(BPlusIO<L> io, long pageAddr, boolean back) {
...
+
         /**
          * @param row Row.
+         * @param findLast find last row.
          */
-        Get(L row) {
-            assert row != null;
+        Get(L row, boolean findLast) {
+            if (!findLast)
+                assert row != null;
```
svladykin: `assert findLast ^ row != null`

PR #1764, opened 2017-04-10
ML: distributed algebra

ybabak:
Apache ticket: https://issues.apache.org/jira/browse/IGNITE-4572


modules/math/src/main/java/org/apache/ignite/math/exceptions/SingularMatrixException.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.apache.ignite.IgniteException;
+
+/**
+ * Exception to be thrown when a non-singular matrix is expected.
+ */
+public class SingularMatrixException extends IgniteException {
+
+    public SingularMatrixException(){
+        super("Singular matrix expected.");
```
vladisav: This should be "Regular (or non-singular) matrix expected."

zaleslaw:
This is an initial commit for ML, all code currently is in codebase and was totally refactored for many times

PR #1798, opened 2017-04-14
IGNITE-4182: fixed

akaigoro:
detailed diagnostics from H2 driver retained as wrapper exception's message.


rfqu:
"why didn't we see the correct error message down the stack?" - because the wrapper exception is itself wrapped, and so is done several times. If user have long enough space to see the full stack trace, he has a chance to see original error message down the stack, but in most cases, the system trims stack trace to reasonable size and deletes the bottom part with the original error message.

svladykin:
The following test always shows the original message "This is my fail!" in the stack trace to me. Even for depth 1000. 

```java
    public void testRecursiveException() {
        doTestRecursive(1000);
    }

    private void doTestRecursive(int x) {
        if (x <= 0)
            throw new IllegalStateException("This is my fail!");

        try {
            doTestRecursive(x - 1);
        }
        catch (Exception e) {
            throw new IllegalStateException("Not my fail!", e);
        }
    }
```

dspavlov:
IGNITE-4182 | Resolved
Closing PR, feel free to reopen if needed

PR #1819, opened 2017-04-18
Ignite 2.0

devozerov:


modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
```diff
@@ -122,17 +122,14 @@
     /** Default lock timeout. */
     public static final long DFLT_LOCK_TIMEOUT = 0;
 
-    /** Initial default cache size. */
-    public static final int DFLT_START_SIZE = 1500000;
```
ankurjha7: While this has been removed. The same configuration is still visible in the documentation . Also what has been provided as the alternative to this ?

dspavlov:
Closing old release PR

PR #1829, opened 2017-04-19
IGNITE-5018 Javadocs in Ignite ML module.

ybabak:
- verified with diffs overview and execution of code style inspections.


zaleslaw:
This is an obsolete PR

PR #1858, opened 2017-04-21
IGNITE-4523

ascherbakoff:
Allow distributed SQL query execution over explicit set of partitions.


modules/indexing/pom.xml
```diff
@@ -31,7 +31,7 @@
     </parent>
 
     <artifactId>ignite-indexing</artifactId>
-    <version>2.0.0-SNAPSHOT</version>
+    <version>2.1.0-SNAPSHOT</version>
```
svladykin: Looks like we have some extra commits from master here.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheProxy.java
```diff
@@ -780,6 +780,13 @@ private ClusterGroup projection(boolean loc) {
 
                 final SqlQuery p = (SqlQuery)qry;
 
+                if (ctx.isReplicated() && p.getPartitions() != null)
```
svladykin: We should not look at `ctx.isReplicated()` here anymore because we are allowed to run any type of query from any cache (partitioned or replicated). This check probably must be somewhere else.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheProxy.java
```diff
@@ -794,6 +801,13 @@ private ClusterGroup projection(boolean loc) {
 
                 SqlFieldsQuery p = (SqlFieldsQuery)qry;
 
+                if (ctx.isReplicated() && p.getPartitions() != null)
```
svladykin: The same as above.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridReduceQueryExecutor.java
```diff
@@ -1063,7 +1135,13 @@ private GridThreadLocalTable fakeTable(Connection c, int idx) {
             List<ClusterNode> owners = cctx.topology().owners(p);
 
             if (F.isEmpty(owners)) {
-                if (!F.isEmpty(dataNodes(cctx.name(), NONE)))
+                // Handle special case: no mapping is configured for a partition.
+                if (F.isEmpty(cctx.affinity().assignment(NONE).get(p))) {
+                    partLocs[p] = Collections.emptySet(); // Mark unmapped partition.
```
svladykin: Lets add a constant `UNMAPPED_PARTS` for `Collections.emptySet()` and use it everywhere.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridReduceQueryExecutor.java
```diff
@@ -1088,6 +1166,9 @@ private GridThreadLocalTable fakeTable(Connection c, int idx) {
                 for (int p = 0, parts =  extraCctx.affinity().partitions(); p < parts; p++) {
                     List<ClusterNode> owners = extraCctx.topology().owners(p);
 
+                    if (partLocs[p] == Collections.<ClusterNode>emptySet())
```
svladykin: Use a constant instead, see above.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridReduceQueryExecutor.java
```diff
@@ -1119,6 +1200,9 @@ private GridThreadLocalTable fakeTable(Connection c, int idx) {
                     return null; // Retry.
 
                 for (Set<ClusterNode> partLoc : partLocs) {
+                    if (partLoc == Collections.<ClusterNode>emptySet())
```
svladykin: Use a constant instead, see above.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridReduceQueryExecutor.java
```diff
@@ -1134,6 +1218,10 @@ private GridThreadLocalTable fakeTable(Connection c, int idx) {
         for (int p = 0; p < partLocs.length; p++) {
             Set<ClusterNode> pl = partLocs[p];
 
+            // Skip unmapped partitions.
+            if (pl == Collections.<ClusterNode>emptySet())
```
svladykin: Use a constant instead, see above.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheProxy.java
ascherbakoff: Added eager check for isReplicatedOnly mode and another when all caches are prepared and determined to be replicated only. In such case setting explicit partitions will throw an exception.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridReduceQueryExecutor.java
ascherbakoff: Done for all cases.

PR #1865, opened 2017-04-24
IGNITE-3487: hidden _key and _val columns

skalashnikov:


modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
```diff
@@ -875,4 +912,53 @@ public static void validateCacheConfiguration(CacheConfiguration ccfg) throws Ig
...
+        }
+
+        /** {@inheritDoc} */
+        @Override public Object value(Object key, Object val) throws IgniteCheckedException {
+            return isKey ? key : val;
+        }
+
+        /** {@inheritDoc} */
+        @Override public void setValue(Object key, Object val, Object propVal) throws IgniteCheckedException {
+            //No-op
```
svladykin: Are you sure it must be no-op? Don't we use this method in DML?

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
```diff
@@ -61,6 +63,9 @@
  */
 public class QueryUtils {
     /** */
+    public static final String _KEY = "_key";
+
+    /** */
     public static final String _VAL = "_val";
```
svladykin: It is preferable to use constants from `IgniteH2Indexing` here.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -3733,6 +3790,103 @@ public void onDrop() {
...
+        @Override public SearchRow prepareProxyIndexRow(SearchRow row) {
+            if (row == null)
+                return null;
+
+            Value[] data = new Value[row.getColumnCount()];
+            for (int idx = 0; idx < data.length; idx++)
+                data[idx] = row.getValue(idx);
+
+            if (keyAliasColumnId > 0) {
+                if (row.getValue(keyAliasColumnId) == null && row.getValue(KEY_COL) != null)
```
svladykin: The following code is the same for key and value, please move it into a separate method.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
skalashnikov: We skip such properties using `desc.isKeyValueOrVersionColumn(i)`.

PR #1885, opened 2017-04-28
IGNITE-4575: SQL support for enum

skalashnikov:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/opt/GridH2QueryContext.java
```diff
@@ -88,16 +90,21 @@
     /** */
     private GridH2CollocationModel qryCollocationMdl;
 
+    /** */
+    private GridKernalContext ctx;
```
svladykin: I'd prefer to have indexing instance here instead.

dspavlov:
IGNITE-4575 | Closed
-- | --

PR #1918, opened 2017-05-06
IGNITE-5119 Append extra message to exception when unable to determine affinity

WilliamDo:
Add detail on cache name to exception message when cache is not available and unable to determine affinity.


dspavlov:
IGNITE-5119 | Closed
-- | --

PR #1935, opened 2017-05-12
IGNITE-4324 Check node list is non-empty before query execution

WilliamDo:


dspavlov:
IGNITE-4324 | Closed
-- | --

PR #1936, opened 2017-05-15
Fixed segmented indices snapshots.

AMashenkov:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/opt/GridH2Table.java
```diff
@@ -247,20 +269,20 @@ private boolean snapshotInLock() {
     /**
      * @param qctx Query context.
      */
-    public void snapshotIndexes(GridH2QueryContext qctx) {
+    public void snapshotIndexes(GridH2QueryContext qctx, int segment) {
```
svladykin: I do not like that we do not pass `segment` to `doSnapshotIndexes `.

PR #1940, opened 2017-05-15
Ignite 2.0 Tests suite check

dspavlov:
For run TC test only, don't merge


dspavlov:
Created by mistake, base branch should be not master

dspavlov:
Created by mistake, base branch should be not master

PR #1942, opened 2017-05-15
IGNITE-5225: Fixed bug in TcpCommunicationSpi.

AMashenkov:


AMashenkov:
Fixed:
1. Thread pool isn't shutdown in case of Exception in some cases.
2. Unresolved ClusterNode addresses cause NPE while node address being checked for reachability.

PR #1967, opened 2017-05-17
IGNITE-5227: StackOverflowError in GridCacheMapEntry#checkOwnerChanged()

mcherkasov:
StackOverflowError occurs due recursion in methods:
GridCacheMapEntry.checkOwnerChanged -> GridDistributedCacheEntry.checkThreadChain -> GridDistributedCacheEntry.recheck -> GridCacheMapEntry.checkOwnerChanged

I broke this cycle by passing CacheLockCandidates of current checking thread into GridCacheMapEntry.checkOwnerChanged method, which allow us to miss GridDistributedCacheEntry.checkThreadChain invocation for owner from the current checking thread. 

The check for the whole candidate chain is moved to GridDistributedCacheEntry.checkThreadChain method, as you can see there  already was cycle for this, I've just removed break and added checks in case we don't  acquired  the lock for new candidate in chain.


dspavlov:

PR #1976, opened 2017-05-20
IGNITE-5248: Detect a 32-bit JVM using too large init page size

vinx13:


dspavlov:
System.getProperty("sun.arch.data.model") may return null on non hotspot VM. I suggest to flip equals as 
"32".equals(System.getProperty("sun.arch.data.model"))

agoncharuk:
I also think you should catch the IllegalArgumentException and make a suggestion to a user to check the JVM, because it may be running on a different VM (sun.arch.data.model may return null, as Dmitriy suggested)

PR #1989, opened 2017-05-22
IGNITE-4724: SQL: corrected result type for AVG()

skalashnikov:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/GridSqlQuerySplitter.java
```diff
@@ -1868,7 +1869,13 @@ private void splitAggregate(
 
                     GridSqlElement sumDownRdc = aggregate(false, SUM).addChild(column(cntMapAggAlias));
 
-                    rdcAgg = op(GridSqlOperationType.DIVIDE, sumUpRdc, sumDownRdc);
+                    if (!isFractionalType(agg.resultType().type())) {
+                        sumUpRdc =  function(CAST).resultType(GridSqlType.BIGINT).addChild(sumUpRdc);
+                        sumDownRdc = function(CAST).resultType(GridSqlType.BIGINT).addChild(sumDownRdc);
```
svladykin: I don't think we need to cast a divisor.

skalashnikov: @svladykin 
The divisor is <SUM()> and H2 promotes its argument type. So <long> becomes <decimal>. Thus, without cast on divisor we can't get integer division anymore.

svladykin: But it is always `SUM(COUNT(expr))` which is always BIGINT already, no?

Another thing worth fixing here division by 0 in case of empty result set.

skalashnikov: No, H2 makes this SUM return `DECIMAL` instead of `BIGINT`. You can check the `DataType.getAddProofType()` and its usages. 

As for the division by 0, I added the test `testEmptyCacheAggregates` for this case. Not sure how, but it works OK for empty resultset.

svladykin: Ok, I see.

Yes, luckily AVG from empty data set returns `null` and `null / 0` is `null` which is correct.

PR #2026, opened 2017-05-29
IGNITE-5299 Don't set IgniteLock as broken if lock is failoverSafe

ezhuravl:


dspavlov:
Closing because
IGNITE-5299 | Resolved

PR #2053, opened 2017-05-31
IGNITE-5364 Remove contention on DataStructure creation or removing

mcherkasov:
All DSs are stored in one Map which itself is stored in utilityCache, this makes high contention on DS creation or removing, it requires lock on the key and manipulation with the Map under the lock. So all threads in cluster should wait for this lock to create or remove DS.

I don't see any reason to store all DS in one map, we already have utilityCache and can save DSs directly in utilityCache, to distinguish DS with other objects in utilityCache I use composite key, the first part of which is DATA_STRUCTURES_KEY, second one is DS's name, also DS type can be added, this will allow us to create different DS with the same name.


modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteReentrantLockBenchmark.java
```diff
@@ -0,0 +1,53 @@
...
+            if(val == null)
+                val = 0;
+            else
+                val++;
+
+            cache.put(key, val);
+        }
+        finally {
+            lock.unlock();
+            lock.removed();
```
yzhdanov: Why "removed()"? Did you mean "close()"?

mcherkasov: yes, right, I confused it with close(). Thank you!

PR #2074, opened 2017-06-03
IGNITE 5368 Cassandra store should support private/public fields serialization for POJO persistence strategy

irudyak:


irudyak:
Reviewed and merged to master.

PR #2089, opened 2017-06-05
IGNITE-5355 Implement class for generate Jira based release report

achetaev:


achetaev:
Implement class for generate Apache Ignite Jira based release reports

PR #2104, opened 2017-06-07
IGNITE-5170 .NET: Compute peer deployment example

ptupitsyn:


modules/platforms/dotnet/examples/Apache.Ignite.Examples/Compute/PeerLoadingExample.cs
```diff
@@ -0,0 +1,93 @@
...
+namespace Apache.Ignite.Examples.Compute
+{
+    using System;
+    using Apache.Ignite.Core;
+    using Apache.Ignite.Core.Compute;
+
+    /// <summary>
+    /// Example demonstrating closure execution with peer assembly loading.
+    /// Should be run with standalone Apache Ignite.NET node.
+    /// "-assembly" command line argument is not required on remote nodes: assemblies are loaded automatically when needed.
```
dmagda: I would rephrase the doc this way:

Before executing the example start one or more standalone Apache Ignite.NET nodes not passing  "-assembly" command line argument. That argument is not required on remote nodes because the assemblies are loaded automatically whenever is needed.

modules/platforms/dotnet/examples/Apache.Ignite.Examples/Compute/PeerLoadingExample.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// 
+    /// Start one or more standalone nodes with Apache.Ignite.exe (see below) and keep them running,
+    /// modify and re-run this example, and observe how new versions of this assembly are deployed and executed.
```
dmagda: Let's clarify saying "... keep them running. Modify HelloAction closure and re-run the example to see that the modified version of the closure will be executed because of automatic deployment of the updated assembly"

PR #2167, opened 2017-06-20
IGNITE-5533 Proper DROP of dynamic tables.

alexpaschenko:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/opt/GridH2Table.java
```diff
@@ -452,6 +460,45 @@ private void ensureNotDestroyed() {
...
+                    // This call implicitly removes both idx and its proxy, if any, from idxs.
+                    database.removeSchemaObject(ses, idx);
+
+                    // We have to call destroy here if we are who has removed this index from the table.
+                    if (idx instanceof GridH2IndexBase)
+                        ((GridH2IndexBase)idx).destroy();
+                }
+            }
+
+            if (SysProperties.CHECK) {
```
svladykin: Why do we rely on some H2 flags in our code?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/opt/GridH2Table.java
```diff
@@ -452,6 +460,45 @@ private void ensureNotDestroyed() {
...
+                for (SchemaObject obj : database.getAllSchemaObjects(DbObject.INDEX)) {
+                    Index idx = (Index) obj;
+                    if (idx.getTable() == this)
+                        DbException.throwInternalError("index not dropped: " + idx.getName());
+                }
+            }
+
+            database.removeMeta(ses, getId());
+            invalidate();
```
svladykin: Why do we remove the table itself here?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/opt/GridH2Table.java
alexpaschenko: I think it makes sense here as we aim to remove table and its indexes in terms of H2 – if we don't, we won't be able to, say, create an index with a pre-existent name.
H2 itself does the same when it drops the table.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/opt/GridH2Table.java
alexpaschenko: Because for dynamic DDL we want to remove the table completely in order to be able to reuse names, for example.

PR #2228, opened 2017-07-03
IGNITE-5153 CPP: Introduced "varint" encoding in C++

isapego:


dspavlov:

PR #2243, opened 2017-07-05
IGNITE-5607: HttpSessionBindingListener is not supported for clustere…

yucigou:
Ignite WebSessionV2 uses genuineSes as the original HttpSession.

Therefore, when setting an attribute or setting the maxInactiveInterval, Ignite should tell the original HttpSession about it.

Otherwise, when the web container (such as Tomcat) thinks that a session expires, or is invalidated, or a session attribute gets removed, etc., session attributes' HttpSessionBindingListener's valueUnbound callback function will not get fired.

So once the original HttpSession gets updated with the session attributes and the maxInactiveInterval, the web container will transitively trigger the session attributes' HttpSessionBindingListener's valueUnbound callback function when a session expires, etc.

(By the way, tested with our app, and our issue is fixed:
http://apache-ignite-developers.2346864.n4.nabble.com/It-seems-WebSession-s-removeAttribute-does-not-support-HttpSessionBindingListener-td19184.html)

### DIFF
--- a/modules/web/src/main/java/org/apache/ignite/cache/websession/WebSessionV2.java
+++ b/modules/web/src/main/java/org/apache/ignite/cache/websession/WebSessionV2.java
@@ -168,6 +168,8 @@ class WebSessionV2 implements HttpSession {
         maxInactiveInterval = interval;
 
         maxInactiveIntervalChanged = true;
+        
+        genuineSes.setMaxInactiveInterval(interval);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


yucigou:
The mechanism is similar to invalidate() of the WebSessionV2 class. The {@link #invalidate()} action will be delegated to the genuine session. Similarly, actions setAttribute(), removeAttribute(), and setMaxInactiveInterval() should be delegated to the genuine session. This way, the web container can do to the session whatever it promises to do, such as calling the HttpSessionBindingListener's valueUnbound callback function, etc.

yucigou:
Register expired and removed event listeners with the WebSession cache or the binary cache, in order to invoke them when a session expires or a session attribute gets removed, etc.

PR #2290, opened 2017-07-13
IGNITE-5382: SQL: Added per-schema connection lookup

skalashnikov:


dspavlov:

PR #2298, opened 2017-07-13
IGNITE-5750 Change HMSM format to hh:mm:ss.mmm

YevIgn:


modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
```diff
@@ -889,4 +889,4 @@ public static double parseDouble(@Nullable String s, double dflt) {
             return dflt;
         }
     }
-}
```
YevIgn: Done. Thanks for pointing out.

c0s:
Looks good! Let me commit this.

PR #2305, opened 2017-07-14
IGNITE-5078: refix according to latets changes

dspavlov:
https://issues.apache.org/jira/browse/IGNITE-5078


dspavlov:
Not actual, re-fix in newer code base will be simpler that resurrecting this PR

PR #2319, opened 2017-07-18
IGNITE-5775 JobsProcessor: Remove job from activeJobs before invoke o…

ezhuravl:
…nCollision

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/IgniteComputeJobOneThreadTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/IgniteComputeJobOneThreadTest.java
@@ -25,7 +25,7 @@ import org.apache.ignite.spi.collision.fifoqueue.FifoQueueCollisionSpi;
 import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
 
 /**
- *
+ * Test of absence of gaps between jobs in compute
  */
 public class IgniteComputeJobOneThreadTest extends GridCommonAbstractTest {
     @Override protected IgniteConfiguration getConfiguration(String name) throws Exception {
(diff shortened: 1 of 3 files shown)


modules/core/src/test/java/org/apache/ignite/internal/IgniteComputeJobOneThreadTest.java
```diff
@@ -0,0 +1,74 @@
...
+                @Override
+                public void run() {
+
+                }
+            });
+        }
+
+        fut.get();
+    }
+}
```
yzhdanov: missing empty line in the end

modules/core/src/test/java/org/apache/ignite/internal/IgniteComputeJobOneThreadTest.java
```diff
@@ -0,0 +1,74 @@
...
+     * @throws Exception If failed.
+     */
+    public void testNoTimeout() throws Exception {
+        Ignite ignite = ignite(0);
+
+        IgniteFuture fut = null;
+
+        for (int i = 0; i < 10000; i++) {
+            fut =  ignite.compute().runAsync(new IgniteRunnable() {
+                @Override
```
yzhdanov: @Override should be on the same line

modules/core/src/test/java/org/apache/ignite/internal/IgniteComputeJobOneThreadTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.ignite.lang.IgniteFuture;
+import org.apache.ignite.lang.IgniteRunnable;
+import org.apache.ignite.spi.collision.fifoqueue.FifoQueueCollisionSpi;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+
+/**
+ *
+ */
+public class IgniteComputeJobOneThreadTest extends GridCommonAbstractTest {
+    @Override protected IgniteConfiguration getConfiguration(String name) throws Exception {
```
yzhdanov: missing comment

modules/core/src/test/java/org/apache/ignite/internal/IgniteComputeJobOneThreadTest.java
yzhdanov: Can we change this test so it asserts some conditions and add it to the suite?

PR #2366, opened 2017-07-31
IGNITE-5882 Duplicated dependency in pom.xml of core module

daradurvs:


daradurvs:
Merged to the master branch.

PR #2377, opened 2017-08-02
ignite-gg-12573 Check cluster state from daemon node, using compute

dgovorukhin:


dspavlov:

PR #2382, opened 2017-08-03
IGNITE-5910 Method stopGrid(name) doesn't work in multiJvm mode

daradurvs:


modules/core/src/test/java/org/apache/ignite/testframework/junits/multijvm/IgniteProcessProxy.java
```diff
@@ -295,10 +300,32 @@ public static IgniteProcessProxy ignite(String igniteInstanceName) {
...
+                    }
+
+                    return true;
+                }
+            }, EventType.EVT_NODE_LEFT);
+
+            proxy.remoteCompute().runAsync(new StopGridTask(igniteInstanceName, cancel));
+
+            try {
+                assert rmtNodeStoppedLatch.await(NODE_LEFT_TIMEOUT, TimeUnit.MILLISECONDS) : "Remote node has not stopped [id=" + rmNodeId + ']';
```
SomeFire: Too long string.

modules/core/src/test/java/org/apache/ignite/testframework/junits/multijvm/IgniteProcessProxy.java
```diff
@@ -295,10 +300,32 @@ public static IgniteProcessProxy ignite(String igniteInstanceName) {
      * @param cancel If {@code true} then all jobs currently will be cancelled.
      */
     public static void stop(String igniteInstanceName, boolean cancel) {
-        IgniteProcessProxy proxy = gridProxies.get(igniteInstanceName);
+        final IgniteProcessProxy proxy = gridProxies.get(igniteInstanceName);
```
SomeFire: No need final in Java8.

modules/core/src/test/java/org/apache/ignite/testframework/junits/multijvm/IgniteProcessProxy.java
```diff
@@ -295,10 +300,33 @@ public static IgniteProcessProxy ignite(String igniteInstanceName) {
...
+                    }
+
+                    return true;
+                }
+            }, EventType.EVT_NODE_LEFT);
+
+            proxy.remoteCompute().runAsync(new StopGridTask(igniteInstanceName, cancel));
+
+            try {
+                assert rmtNodeStoppedLatch.await(NODE_LEFT_TIMEOUT, TimeUnit.MILLISECONDS) : "Remote node has not " +
```
dspavlov: please fix here, assertion has side effect. If assertions disabled, nothign will be awaited.

daradurvs: Are you sure since this is the common construction which is used in the whole testing framework?
If assertions are disabled then the huge part of the testing framework will not be able to work properly.

dspavlov: Yes, I'm sure. We consider assertions from JUnit as best practice

daradurvs: Are you talking about JUnit 3 assertions from GridAbstactTest? They are not available here since this is not testing class.
I only able to use this construction:
`if (rmtNodeStoppedLatch.await(NODE_LEFT_TIMEOUT, TimeUnit.MILLISECONDS))
                    throw new AssertException("Remote node has not stopped [id=" + rmNodeId + ']');`

PR #2410, opened 2017-08-07
IGNITE-5732 Provide API to test compatibility with old releases

daradurvs:
Special module contains testing framework which provides methods for testing backward compatibility.
It allows to start Ignite-cluster in multiversion mode. It means that nodes with different build version is allowed to join topology.


modules/compatibility/src/test/java/org/apache/ignite/compatibility/persistence/DummyPersistenceTest.java
```diff
@@ -0,0 +1,128 @@
...
+/** */
+public class DummyPersistenceTest extends IgniteCompatibilityAbstractTest {
+    /** */
+    private static final String TEST_CACHE_NAME = DummyPersistenceTest.class.getSimpleName();
+
+    /** {@inheritDoc} */
+    @Override protected void beforeTest() throws Exception {
+        super.beforeTest();
+
+        System.out.println("*****" + System.getProperty("java.class.path"));
```
voipp: forgot to remove test sys out ?

daradurvs: No, it's for debugging on the TeamCity.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/CompatibilityTestIgniteNodeRunner.java
```diff
@@ -0,0 +1,172 @@
...
+        cfg.setNodeId(id);
+
+        final Ignite ignite = Ignition.start(cfg);
+
+        // If 'id' equals 'syncId' then the starting node is the first node and
+        // there was no need to check the join to topology.
+        if (!id.equals(syncId)) {
+            GridTestUtils.waitForCondition(new GridAbsPredicate() {
+                @Override public boolean apply() {
+                    boolean found = ignite.cluster().node(syncId) != null;
```
voipp: adding awaitPartitionMapExchange() lets you wait for node join cluster and that topology version is updated on all nodes.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/CompatibilityTestIgniteNodeRunner.java
```diff
@@ -0,0 +1,172 @@
...
+
+                    if (found)
+                        X.println(IgniteCompatibilityAbstractTest.SYNCHRONIZATION_LOG_MESSAGE_JOINED + id);
```
voipp: may be print messages to some logger, so user can define its own way to print messages.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/CompatibilityTestIgniteNodeRunner.java
daradurvs: There is an another goal: synchronization of testing framework between two separate JVM.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/IgniteCompatibilityAbstractTest.java
```diff
@@ -0,0 +1,331 @@
...
+            final UUID syncId = ((IgniteProcessProxy)rmJvmInstance).getId();
+            final UUID nodeId = cfg.getNodeId();
+
+            ignite = super.startGrid(igniteInstanceName, cfg, ctx);
+
+            assert ignite.configuration().getNodeId() == nodeId : "Started node has unexpected node id.";
+
+            assert GridTestUtils.waitForCondition(new GridAbsPredicate() {
+                @Override public boolean apply() {
+                    return ignite.cluster().node(syncId) != null;
```
voipp: may be add awaitPartitionMapExchange()

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/CompatibilityTestIgniteNodeRunner.java
daradurvs: No, I can't do it. I only can grab a process output of separate JVM.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/IgniteCompatibilityAbstractTest.java
daradurvs: There is a reverse synchronization in case of starting a the first node in local JVM when there are nodes in another JVM.

daradurvs: By the same token, exchange doesn't start before cluster activation.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/CompatibilityTestIgniteNodeRunner.java
daradurvs: By the same token, exchange doesn't start before cluster activation.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/util/MavenUtils.java
```diff
@@ -0,0 +1,208 @@
...
+     *
+     * @param artifact Artifact identifier, must match pattern [groupId:artifactId:version].
+     * @throws Exception In case of an error.
+     */
+    private static void downloadArtifact(String artifact) throws Exception {
+        X.println("Downloading artifact... Identifier: " + artifact);
+
+        exec("mvn dependency:get -Dartifact=" + artifact);
+
+        X.println("Download is finished");
```
voipp: why println not logging

modules/compatibility/src/test/java/org/apache/ignite/compatibility/spi/MultiVersionClusterTest.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /** */
+    public void testJoinMultiVersionTopologyLocalFirst2() throws Exception {
+        try {
+            IgniteEx ignite = startGrid(0);
+
+            assertEquals(1, topologyVersion(ignite));
+
+            startGrid(1, "2.3.0-SNAPSHOT", new PostConfigurationClosure());
```
nizhikov: `2.3.0-SNAPSHOT` definitely should be a constant

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/CompatibilityTestIgniteNodeRunner.java
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * Stores {@link IgniteInClosure} to file as xml.
+     *
+     * @param clos IgniteInClosure.
+     * @return A name of file where the closure was stored.
+     * @throws IOException In case of an error.
+     * @see #readClosureFromFileAndDelete(String)
+     */
+    @Nullable public static String storeToFile(@Nullable IgniteInClosure clos) throws IOException {
```
nizhikov: Why doesn't IgniteInClosure have generic parameter?

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/CompatibilityTestIgniteNodeRunner.java
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * Stores {@link IgniteInClosure} to file as xml.
+     *
+     * @param clos IgniteInClosure.
+     * @param fileName A name of file where the closure was stored.
+     * @throws IOException In case of an error.
+     * @see #readClosureFromFileAndDelete(String)
+     */
+    public static void storeToFile(@NotNull IgniteInClosure clos, @NotNull String fileName) throws IOException {
```
nizhikov: Why doesn't IgniteInClosure have generic parameter?

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/IgniteCompatibilityAbstractTest.java
```diff
@@ -0,0 +1,331 @@
...
+ */
+public abstract class IgniteCompatibilityAbstractTest extends GridCommonAbstractTest {
+    /** Using for synchronization of nodes startup in case of starting remote nodes first. */
+    public static final String SYNCHRONIZATION_LOG_MESSAGE_JOINED = "Remote node has joined, id: ";
+
+    /** Using for synchronization of nodes startup in case of starting remote nodes first. */
+    public static final String SYNCHRONIZATION_LOG_MESSAGE_PREPARED = "Remote node has prepared, id: ";
+
+    /** Waiting seconds of the join of a node to topology. */
+    protected static final int NODE_JOIN_TIMEOUT = 30;
```
nizhikov: Why timeout specified in seconds?
Can we rename constant to `NODE_JOIN_TIMEOUT_SEC` or set a timeout in milliseconds?

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/IgniteCompatibilityAbstractTest.java
```diff
@@ -0,0 +1,331 @@
...
+    /**
+     * Checks if the default DB working directory is empty.
+     *
+     * @return {@code true} if the default DB working directory is empty or doesn't exist, otherwise {@code false}.
+     * @throws IgniteCheckedException In case of an error.
+     * @see #getDefaultDbWorkPath()
+     * @see #deleteDefaultDBWorkDirectory()
+     */
+    @SuppressWarnings("ConstantConditions")
+    protected boolean defaultDBWorkDirectoryIsEmpty() throws IgniteCheckedException {
```
nizhikov: I think method name should be `isDefaultDBWorkDirectoryEmpty`

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/IgniteCompatibilityAbstractTest.java
```diff
@@ -0,0 +1,331 @@
...
+     * Uses an ignite-core artifact in the Maven local repository, if it isn't exists there, it will be downloaded and
+     * stored via Maven.
+     *
+     * @param idx Index of the grid to start.
+     * @param ver Ignite version.
+     * @param cfgClos IgniteInClosure for post-configuration.
+     * @return Started grid.
+     * @throws Exception If failed.
+     */
+    protected IgniteEx startGrid(int idx, String ver, IgniteInClosure<IgniteConfiguration> cfgClos) throws Exception {
```
nizhikov: Can we use some sort of enum for `ver` variable?

modules/compatibility/src/test/resources/META-INF/services/org.apache.ignite.plugin.PluginProvider
```diff
@@ -0,0 +1 @@
+org.apache.ignite.compatibility.testframework.plugins.TestCompatibilityPluginProvider
```
nizhikov: Is this provider would be available for regular test run?
If yes, we definitely should exclude it from regular test classpath.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/persistence/VarintPersistenceTest.java
```diff
@@ -0,0 +1,195 @@
...
+/** */
+public class VarintPersistenceTest extends IgniteCompatibilityAbstractTest {
+    /** */
+    private static final String TEST_CACHE_NAME = VarintPersistenceTest.class.getSimpleName();
+
+    /** {@inheritDoc} */
+    @Override protected void beforeTest() throws Exception {
+        super.beforeTest();
+
+        if (!defaultDBWorkDirectoryIsEmpty())
```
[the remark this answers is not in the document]
daradurvs: This test class will be removed.
But, I agree we should provide PersistenceAbstractClasss, thanks!

modules/compatibility/src/test/java/org/apache/ignite/compatibility/spi/MultiVersionClusterTest.java
```diff
@@ -0,0 +1,151 @@
...
+ */
+public class MultiVersionClusterTest extends IgniteCompatibilityAbstractTest {
+    /** */
+    public void testJoinMultiVersionTopologyLocalFirst() throws Exception {
+        try {
+            IgniteEx ignite = startGrid(0);
+
+            assertEquals(1, topologyVersion(ignite));
+
+            startGrid("testMultiVersion", "2.3.0-SNAPSHOT", new PostConfigurationClosure());
```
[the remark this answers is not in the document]
daradurvs: This test will be removed.
There is framework-self-test with a package which installed in the Maven local repositoty.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/spi/MultiVersionClusterTest.java
daradurvs: This is dummy test and will be removed.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/CompatibilityTestIgniteNodeRunner.java
daradurvs: Do you think this is required?
I think no because given instance will be serialized to a file, without any type information.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/CompatibilityTestIgniteNodeRunner.java
daradurvs: Do you think this is required?
I think no because given instance will be serialized to a file, without any type information.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/IgniteCompatibilityAbstractTest.java
daradurvs: Agree, will be fixed.

modules/compatibility/src/test/resources/META-INF/services/org.apache.ignite.plugin.PluginProvider
daradurvs: No, this plugin will be used only in tests in the compatibility-module and won't be used in any other module like ignite-core.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/IgniteCompatibilityAbstractTest.java
daradurvs: I thought about it.
Not sure is it required.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/IgniteCompatibilityAbstractTest.java
daradurvs: Agree. Will be renamed.
It will be used with definition TimeUnit.SECONDS in countDownLatch .

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/junits/CompatibilityTestIgniteNodeRunner.java
nizhikov: Is `storeToFile` common method? Or it will be used only in compatibility module with specific closure types?

If first - we should use  
`public static void <T> storeToFile(@NotNull IgniteInClosure<T> clos, @NotNull String fileName) throws IOException {`
If second:
`public static void storeToFile(@NotNull IgniteInClosure<Ignite> clos, @NotNull String fileName) throws IOException {`

daradurvs: It will be used only in the compatibility module with specific closure types:
IgniteInClosure<IgniteConfiguration>
IgniteInClosure<Ignite>

daradurvs:
Merged to the master branch by the [commit](https://github.com/apache/ignite/commit/c6571448c5d75af3093436323f9f339602a5c083)

PR #2418, opened 2017-08-09
[IGNITE-6002] Make log message more pretty when instance name is null

sarutak:
When instance name is null, log message containing "null" is shown like as follows.

```
[09-08-2017 14:49:38][INFO ][grid-timeout-worker-#15%null%][IgniteKernal]
Metrics for local node (to disable set 'metricsLogFrequency' to 0)
    ^-- Node [id=de0743ec, name=null, uptime=00:01:00:007]
```


dspavlov:

PR #2425, opened 2017-08-09
IGNITE-5736 Add test of backward-compatibility

daradurvs:


daradurvs:
Split into 2 issues, see #2549 and #2410.

daradurvs:
Split into 2 issues, see #2549 and #2410.

PR #2428, opened 2017-08-10
IGNITE-4991 Do not print out system properties when IGNITE_TO_STRING_…

ezhuravl:
…INCLUDE_SENSITIVE is set to false


c0s:
Looks good!

PR #2457, opened 2017-08-16
IGNITE-6088 Socket#shutdownOutput in ServerImpl leads to UnsupportedO…

ezhuravl:
…perationException on SSLSocket


dspavlov:

PR #2489, opened 2017-08-20
ignite-5714 thread id amended to data structure

voipp:


voipp:
need to run tests for it

PR #2501, opened 2017-08-23
IGNITE-5280: SparseDistributedMatrix refactoring

ybabak:


modules/ml/src/main/java/org/apache/ignite/ml/math/MatrixStorage.java
```diff
@@ -55,6 +55,13 @@
     public int storageMode();
 
     /**
+     * @return Matrix storage mode.
```
oignatenko: looks like a typo, I would expect javadoc to say "access mode" instead of "storage mode" here

modules/ml/src/main/java/org/apache/ignite/ml/math/impls/matrix/DenseLocalOnHeapMatrix.java
```diff
@@ -128,6 +128,6 @@ private DenseLocalOnHeapMatrix(DenseLocalOnHeapMatrix orig, int acsMode) {
 
     /** */
```
oignatenko: {@inheritDoc} seems to be missing here, compare to javadocs in other methods of this class

ybabak:
merged

PR #2502, opened 2017-08-23
Ignite-4756 Print info about partition distribution to log.

vadopolski:
Added tests, printDistribution method, IGNITE_PART_DISTRIBUTION_WARN_THRESHOLD system property.


vadopolski:
Leave a comment

PR #2532, opened 2017-08-28
IGNITE-5839: Unclear exception from BinaryObjectBuilder::build call when builder is reused.

AMashenkov:
Fix BinaryObjectBuilder reusage issue.


AMashenkov:
Already in master

PR #2558, opened 2017-08-31
IGNITE-6217  Add benchmark to compare JDBC drivers and native SQL execution

tledkov-gridgain:


modules/yardstick/src/main/java/org/apache/ignite/yardstick/jdbc/JdbcSqlInsertBenchmark.java
```diff
@@ -0,0 +1,53 @@
...
+
+        PreparedStatement delete = singleDelete.get();
+        delete.setLong(1, insertKey);
+
+        try {
+            insert.executeUpdate();
+            delete.executeUpdate();
+        }
+        catch (Exception ign){
+            // collision occurred, ignoring
```
pavel-kuznetsov: Do we need to log/count collisions?

modules/yardstick/src/main/java/org/apache/ignite/yardstick/jdbc/NativeSqlUpdateRangeBenchmark.java
```diff
@@ -0,0 +1,78 @@
...
+            long id = ThreadLocalRandom.current().nextLong(args.range() - args.sqlRange()) + 1;
+            long maxId = id + args.sqlRange() - 1;
+
+            qry.setArgs(id, maxId);
+
+            expRsSize = args.sqlRange();
+        }
+
+        try (FieldsQueryCursor<List<?>> cursor =
+                 ((IgniteEx)ignite()).context().query().querySqlFieldsNoCache(qry, false)) {
```
pavel-kuznetsov: is it worth to move this code to parent class?
```
final FieldsQueryCursor<List<?>> execQuery(SqlFieldsQuery qry){
    return  ((IgniteEx)ignite()).context().query().querySqlFieldsNoCache(qry, false);
}```

PR #2566, opened 2017-08-31
IGNITE-6234 null check for schemaIds

sergey-chugunov-1985:


kristoffSC:
Hi,
I'm facing this issue to, but in a slightly different scenario. 

However I wonder how `schemaIds` can be null since BinaryMetadata constructor sets this field to `Collections.emptySet();` in a "worst" case.

PR #2583, opened 2017-09-04
IGNITE-6252 Cassandra Cache Store Session does not retry if statement…

sunnychanwork:
… failed

This patch will check to ensure that we retry any prepare statement failure when we submit the async request to Cassandra


sunnychanwork:
This bug actually is not being fixed by that specific PR. That PR only address the fact that the preparedStatement is reinitialized for the next item, but it doesn't address that fact that the current items needs to retry. In fact, the unit test provided in #3088 is incorrect - There are 10 items of data being processed by only assert that there are 9 items in the processed count.

sunnychanwork:
Even with the patch the error still occurs - here is the sequence:

1. block around [231](https://github.com/apache/ignite/blob/6330f0bb74d6b986d040a8017f3dfbde361d5457/modules/cassandra/store/src/main/java/org/apache/ignite/cache/store/cassandra/session/CassandraSessionImpl.java#L231) throws InvalidQueryException. data item is not processed and not added to the list of future
2. prepStatEx record the exception in [251](https://github.com/apache/ignite/blob/6330f0bb74d6b986d040a8017f3dfbde361d5457/modules/cassandra/store/src/main/java/org/apache/ignite/cache/store/cassandra/session/CassandraSessionImpl.java#L251)
3. error set to keep the exception throw (line [276](https://github.com/apache/ignite/blob/6330f0bb74d6b986d040a8017f3dfbde361d5457/modules/cassandra/store/src/main/java/org/apache/ignite/cache/store/cassandra/session/CassandraSessionImpl.java#L276))
4. line [282](https://github.com/apache/ignite/blob/6330f0bb74d6b986d040a8017f3dfbde361d5457/modules/cassandra/store/src/main/java/org/apache/ignite/cache/store/cassandra/session/CassandraSessionImpl.java#L282) resets the prepStatEx to null
5. line [309](https://github.com/apache/ignite/blob/6330f0bb74d6b986d040a8017f3dfbde361d5457/modules/cassandra/store/src/main/java/org/apache/ignite/cache/store/cassandra/session/CassandraSessionImpl.java#L309) didn't check the cached "error", so the error in 231 is ignored and it will not retry and return 310

PR #2608, opened 2017-09-07
IGNITE-6190: Let pass collections as arguments to SQL queries.

shroman:


dspavlov:
IGNITE-6190 | Closed
-- | --

PR #2710, opened 2017-09-21
IGNITE-6334 Throttle writing threads during ongoing checkpoint

glukos:


rpuch:
Just a note for future archeologists: https://issues.apache.org/jira/browse/IGNITE-6334 mentions Token Bucket algorithm, but in reality this is not what was implemented. In the Token Bucket algorithm, non-complying requests are delayed till they become complying (so, in our case, this would mean wait till the memory resources we protect become available). But in the code, a different approach is used: when the protected resources are scarse, slow down writer threads; in conjunction with a sufficient buffer size, this makes the protection reliable.

PR #2714, opened 2017-09-21
IGNITE-898: Fixed error when path to Ignite folder contains whitespaces

alexzaitzev:
Changes in build-classpath.sh fix the problem. setenv.sh was also changed to prevent problems with user libs


vveider:
Checked with spaces in `JAVA_HOME` and `IGNITE_HOME` respectively -- node starts.

bin/ignite.bat
```diff
@@ -119,8 +118,10 @@ if %ERRORLEVEL% neq 0 (
 ::
 :: Process 'restart'.
 ::
-set RANDOM_NUMBER_COMMAND="!JAVA_HOME!\bin\java.exe" -cp %CP% org.apache.ignite.startup.cmdline.CommandLineRandomNumberGenerator
-for /f "usebackq tokens=*" %%i in (`!RANDOM_NUMBER_COMMAND!`) do set RANDOM_NUMBER=%%i
+set RANDOM_NUMBER_COMMAND="!JAVA_HOME!\bin\java.exe" -cp "%CP%" org.apache.ignite.startup.cmdline.CommandLineRandomNumberGenerator
+for /f "usebackq tokens=*" %%i in (`"!RANDOM_NUMBER_COMMAND!"`) do (
+    set RANDOM_NUMBER=%%i
```
vveider: Round brackets wrapping is not related to task

bin/include/build-classpath.sh
```diff
@@ -42,6 +42,9 @@ case "`uname`" in
 esac
 
 includeToClassPath() {
+    SAVEIFS=$IFS
```
vveider: Can it be replaced with following construction?
```
while IFS=$(echo -en "\n\b") read file
do
    ...
done < $1/*
```

bin/include/setenv.sh
```diff
@@ -56,13 +56,18 @@ esac
 #
 IGNITE_LIBS="${IGNITE_HOME}/libs/*"
 
+SAVEIFS=$IFS
```
vveider: Can it be replaced with following construction?
```
while IFS=$(echo -en "\n\b") read file
do
    ...
done < $1/*
```

bin/ignite.bat
alexzaitzev: I added it for better readability. Should I create a new task for this?

bin/include/build-classpath.sh
alexzaitzev: Maybe, but why? It adds one more level of nesting. This is not good.

bin/ignite.bat
vveider: Yes. We must stick to task only, no concomitant changes should be introduced that do not relate the task at least indirectly.

bin/include/build-classpath.sh
vveider: There will be no additional nesting, as I suggest to replace `for` loop with `while` loop, because `while` loop can use temporary IFS override, i.e.:
```
while IFS=$(echo -en "\n\b") read file
do
    if [ -d ${file} ] && [ -d "${file}/target" ]; then
        ...
    fi
done < $1/*
```

alexzaitzev: Ok, it can be done, but your example doesn't work. Maybe we just leave my option?

dspavlov:

PR #2737, opened 2017-09-25
IGNITE-6286: fix org.h2.jdbc.JdbcSQLException

schernolyas:
changes:
1) develop test for check all supported types for parameterized queries
2) add support BigDecimal type as type of paraameter's value


schernolyas:
A lot of changes that not linked with my PR

PR #2759, opened 2017-09-26
ignite-5714-6

voipp:


voipp:
cannot catch exception.Restart tests

voipp:
cannot catch exception. Seems it will never be thrown

voipp:
checked ^ exception woul never be thrown. It is never used

voipp:
waiting for tests

PR #2787, opened 2017-10-02
IGNITE-6542 Reliably close SocketChannel in TcpCommunicationSpi.

alamar:


alamar:
Also fix forceClose() in GridTcpNioCommunicationClient which became wrong when migrated from int to bool.

alamar:
Best viewed in Idea with whitespace changes off

PR #2798, opened 2017-10-03
IGNITE-6554: Atomic cache remove operations are logged in WAL as update records

dspavlov:
Probably fix for atomic cache remove operations


dspavlov:
Replaced by #2800

PR #2801, opened 2017-10-04
removed excluding ML from examples/src

ghost:


dspavlov:
@zaleslaw  could you please validate if we need this change? It is not related to any JIRA, so I suggest closing it

zaleslaw:
Yes, we should close this PR

PR #2805, opened 2017-10-05
IGNITE-6556 setSchema() is case sensitive in the wrong way in Thin JD…

tledkov-gridgain:
…BC driver


alamar:
It seems to be working for my use case!

PR #2811, opened 2017-10-05
IGNITE-6234 Initialize schemaIds as empty set if schemas field is null

kristoffSC:
Initialize schemaIds as an empty set if schemas field is null during BinaryMetada deserialization.  
This behavior is concise with BinaryMetadata constructor. 

This should prevent NPE for some situations when BinarryMetadata object is created from deserialization and we call hasSchema(int i) on it. - https://issues.apache.org/jira/browse/IGNITE-6234


dspavlov:
The ticket is resolved, closing PR. Feel free to reopen if it is needed.

PR #2870, opened 2017-10-17
IGNITE-2766 Deterministically reopen cache after client reconnect.

alamar:


chrisheaththomas:
We have dropped use of Ignite cache until this issue is resolved.

alamar:
@chrisheaththomas can you please state what  problem you are facing? Because I could not reproduce IGNITE-2766.

PR #2885, opened 2017-10-19
GG-12974 Don't use org.apache.spark.Logging in Ignite-spark

apopovgg:


dspavlov:

PR #2900, opened 2017-10-22
ignite-6667

ascherbakoff:
Allow discovery cache instance reuse if only minor topology change has occured.


dspavlov:
The ticket is resolved, closing PR. Feel free to reopen if it is needed.

PR #2922, opened 2017-10-25
IGNITE-6747 Use _HOST instead of FDQN when configuring principal for IgfsSecondaryFileSystem with kerberos enable

Reidddddd:


Reidddddd:
@VladimirErshov , @tledkov-gridgain would you mind giving a review, if there're available bandwidth. Thanks.

Reidddddd:
I wanted to rebase master branch, but it seems mess up.

PR #2945, opened 2017-10-30
IGNITE-6788: Ignite WAL reader fails on Tx marker

dspavlov:
NPE during reading of Tx marker record for persistent store with new style folder naming


dspavlov:
Merged

PR #2953, opened 2017-10-31
IGNITE-6640: Introduction of models import/export

ybabak:


modules/ml/src/main/java/org/apache/ignite/ml/FileExporter.java
```diff
@@ -0,0 +1,88 @@
...
+    /** {@inheritDoc} */
+    @Override public void save(D d, String path) {
+        ObjectOutputStream outStream = null;
+        try {
+            outStream = new ObjectOutputStream(new FileOutputStream(path));
+
+            outStream.writeObject(d);
+        } catch (IOException e) {
+            if (log != null)
+                log.error("Error opening file", e);
```
oignatenko: please add dot at the end of this message, as well as at toher similar messages that lack dots at the end

modules/ml/src/main/java/org/apache/ignite/ml/KMeansModelFormat.java
```diff
@@ -0,0 +1,77 @@
...
+import java.io.Serializable;
+import java.util.Arrays;
+import org.apache.ignite.ml.math.DistanceMeasure;
+import org.apache.ignite.ml.math.Vector;
+
+/**
+ * K-means model representation.
+ *
+ * @see Exportable
+ * @see Exporter
```
oignatenko: it's hard to tell how "@see Export/able" references are relevant here. In order to help reader see the connection consider either extending first sentence to explain that representation is used for export or replacing these see-references with something like "@see org.apache.ignite.ml.clustering.KMeansModel#saveModel"

modules/ml/src/main/java/org/apache/ignite/ml/FileExporter.java
```diff
@@ -0,0 +1,88 @@
...
+     * @param log Logger.
+     */
+    public void setLog(IgniteLogger log) {
+        this.log = log;
+    }
+
+    /** {@inheritDoc} */
+    @Override public void save(D d, String path) {
+        ObjectOutputStream outStream = null;
+        try {
```
oignatenko: consider either try-with-resources here or adding a comment helping reader to understand why it is better to abstain of this. Same to load method

modules/ml/src/test/java/org/apache/ignite/ml/LocalModelsTest.java
```diff
@@ -0,0 +1,80 @@
...
+    @After
+    public void cleanUp() throws IOException {
+        Files.deleteIfExists(Paths.get(mdlFilePath));
+    }
+
+    /**
+     *
+     */
+    @Test
+    public void importExportKMeansModelTest(){
```
oignatenko: consider adding some code that would cover testing of KMeansModel#distanceMeasure - currently IDEA reports this method as unused

modules/ml/src/main/java/org/apache/ignite/ml/clustering/KMeansModel.java
```diff
@@ -18,13 +18,16 @@
...
 import org.apache.ignite.ml.math.DistanceMeasure;
 import org.apache.ignite.ml.math.Vector;
 
 /**
  * This class encapsulates result of clusterization.
  */
-public class KMeansModel implements ClusterizationModel<Vector, Integer> {
+public class KMeansModel implements ClusterizationModel<Vector, Integer>, Exportable<KMeansModelFormat> {
     /** Centers of clusters. */
     private Vector[] centers;
```
oignatenko: consider either making centers and distance final, as suggested by code inspection or adding a comment explaining why it's better to keep these non-final

modules/ml/src/main/java/org/apache/ignite/ml/FileExporter.java
YuriBabak: used  try-with-resources.

modules/ml/src/main/java/org/apache/ignite/ml/FileExporter.java
YuriBabak: fixed

modules/ml/src/main/java/org/apache/ignite/ml/clustering/KMeansModel.java
YuriBabak: added final modifier

modules/ml/src/test/java/org/apache/ignite/ml/LocalModelsTest.java
YuriBabak: this changes(and test too) about import/export, not about kmeans model. But yes we should increase coverage.

modules/ml/src/main/java/org/apache/ignite/ml/KMeansModelFormat.java
YuriBabak: Currently this is intro for model import/export mechanism and I think that this javadoc good enough for now.

PR #2961, opened 2017-11-01
IGNITE-6324 Transactional cache data partially available after crash.

dgovorukhin:


modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/ConsistentIdMapper.java
```diff
@@ -43,6 +44,21 @@ public ConsistentIdMapper(GridDiscoveryManager discoveryManager) {
     }
 
     /**
+     * Map topology version to list of consistent id.
+     *
+     * @param topVer Topology version.
+     * @return List of consistent ids.
+     */
+    public List<Object> mapToConsistentId(AffinityTopologyVersion topVer) {
```
agoncharuk: We should not calculate this list on-demand. Since the list is fixed on topologyVersion, this list should be moved to the DiscoCache

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -389,10 +408,19 @@ public Checkpointer getCheckpointer() {
             final FileLockHolder preLocked = kernalCtx.pdsFolderResolver()
                 .resolveFolders()
                 .getLockedFileLockHolder();
+
             if (preLocked == null)
                 fileLockHolder = new FileLockHolder(storeMgr.workDir().getPath(), kernalCtx, log);
 
             persStoreMetrics.wal(cctx.wal());
+
+            recoveryIo = new RecoveryIoImp(kernalCtx, log);
```
agoncharuk: RecoveryIo looks redundant, we do not use such a pattern. Instead, we probably should add a recovery processor which will add this listener on startup and manager recovery future(s).

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -1612,11 +1738,97 @@ private PageMemoryEx getPageMemoryForCacheGroup(int grpId) throws IgniteCheckedE
     }
 
     /**
+     *
+     */
+    private void restoreUpdates(final RecoveryContext recCtx) throws IgniteCheckedException {
+        final Map<T2<Integer, Integer>, T2<Integer, Long>> partStates = new HashMap<>();
```
agoncharuk: Map of tuples may be too heap-consuming. Check if CachePartitionPartialCountersMap can be used or modified for this use-case.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/recovery/IgniteRecoveryFuture.java
```diff
@@ -0,0 +1,31 @@
...
+
+package org.apache.ignite.internal.processors.cache.persistence.recovery;
+
+import org.apache.ignite.internal.IgniteInternalFuture;
+import org.apache.ignite.internal.pagemem.wal.WALIterator;
+
+/**
+ *
+ */
+public interface IgniteRecoveryFuture<R> extends IgniteInternalFuture<R> {
```
agoncharuk: I think having recovery future dependant on WAL iterator is abstraction leak. We should simply have an init() method on it, and it will construct the required iterator internally.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/recovery/RecoveryFuture.java
```diff
@@ -0,0 +1,503 @@
...
+        this.recoveryIo = recoveryIo;
+
+        recoveryIo.receive(new CI2<String, Message>() {
+            @Override public void apply(final String constId, final Message msg) {
+                initFut.listen(new CI1<IgniteInternalFuture>() {
+                    @Override public void apply(IgniteInternalFuture f) {
+                        try {
+                            f.get();
+
+                            if (msg instanceof TxStateRequest) {
```
agoncharuk: Split different message processing into different methods

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/recovery/RecoveryFuture.java
```diff
@@ -0,0 +1,503 @@
...
+                                    G.stop(true);
+                                }
+                            }).start();
+                        }
+                    }
+                });
+            }
+        });
+
+        recoveryIo.onNodeLeft(new CI1<String>() {
```
agoncharuk: We usually place onNodeLeft callbacks on futures, this is another indication that RecoveryIo should be removed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/recovery/RecoveryIoImp.java
```diff
@@ -0,0 +1,155 @@
...
+    @Override public void onMessage(UUID nodeId, Object msg, byte plc) {
+        IgniteBiInClosure<String, Message> handler = msgHandler;
+
+        if (handler != null && msg instanceof Message) {
+            String constId = idMapper.mapToConsistentId(
+                ctx.discovery().topologyVersionEx(), nodeId
+            ).toString();
+
+            handler.apply(constId, (Message)msg);
+        }
```
agoncharuk: We should not silently drop a message. If a listener is supposed to be always installed before a first message is sent, we should at least print a warning.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/serializer/RecordDataV1Serializer.java
```diff
@@ -315,7 +315,7 @@ public RecordDataV1Serializer(GridCacheSharedContext cctx) {
                 long msb = in.readLong();
                 long lsb = in.readLong();
                 boolean hasPtr = in.readByte() != 0;
-                int idx = hasPtr ? in.readInt() : 0;
+                long idx = hasPtr ? in.readLong() : 0;
```
agoncharuk: Does not this break compatibility?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
```diff
@@ -285,6 +293,79 @@
...
+    }
+
+    /**
+     *
+     */
+    private FileWALPointer findCandidate(Collection<IgniteInternalTx> txs) {
+        FileWALPointer pntCandidate = null;
+
+        for (IgniteInternalTx tx : txs)
+            if (tx instanceof IgniteTxAdapter) {
```
agoncharuk: Do not use instanceof, just add a method to the IgniteInternalTx interface

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
```diff
@@ -285,6 +293,79 @@
...
+                IgniteTxAdapter txAdapter = (IgniteTxAdapter)tx;
+
+                WALPointer ptr = txAdapter.prepPtr;
+
+                if (ptr != null && ptr instanceof FileWALPointer) {
+                    FileWALPointer ptr0 = (FileWALPointer)ptr;
```
agoncharuk: Also, do not cast to FileWALPointer here, add Comparable to wal pointer interface.

PR #2992, opened 2017-11-07
IGNITE-6500 Bug in ignite's CassandraCacheStoreFactory

kotamrajuyashasvi:
POJO fields of java wrapper type are not retaining null values from Cassandra persistent store, while using ignite's CassandraCacheStoreFactory.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
@@ -217,6 +217,9 @@ public class GridCachePartitionExchangeManager<K, V> extends GridCacheSharedMana
     /** For tests only. */
     private volatile AffinityTopologyVersion exchMergeTestWaitVer;
 
+    /** For tests only. */
+    private volatile List mergedEvtsForTest;
+
     /** Distributed latch manager. */
     private ExchangeLatchManager latchMgr;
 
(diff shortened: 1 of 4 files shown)


modules/cassandra/store/src/main/java/org/apache/ignite/cache/store/cassandra/persistence/PojoField.java
```diff
@@ -201,6 +201,8 @@ public Annotation getAnnotation(Class clazz) {
     public void setValueFromRow(Row row, Object obj, Serializer serializer) {
         if (calculatedField())
             return;
+        if (row.isNull(col))
```
dspavlov: Spacing between semantic block is required by code style. So I suggest to insert new empty line before if.

irudyak: This logic is not 100% correct. The best way to do it is to slightly modify logic of **PropertyMappingHelper.getCassandraColumnValue** method for all primitive types like **int**, **long** and etc. Here is the sample for **int/Integer** types (same way for all other types):

Instead of this code for **int/Integer** types:

        if (Integer.class.equals(clazz) || int.class.equals(clazz))
            return row.getInt(col);

It should be something like this:

        if (Integer.class.equals(clazz))
            return row.isNull(col) ? null : row.getInt(col);

        if (int.class.equals(clazz)) {
            if (row.isNull(col)) {
                throw new IllegalArgumentException("Can't cast null value from Cassandra table column '" + col +
                        "' to " + "int value used in domain object model");
            }

            return row.getInt(col);
        }

PR #3010, opened 2017-11-09
IGNITE-6818 - Handle half open connection in communication.

dkarachentsev:


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -658,6 +654,18 @@ private void onFirstMessage(final GridNioSession ses, Message msg) {
                 }
             }
 
+            private void closeStaleConnections(ConnectionKey connKey) {
```
alamar: Method description needed?

dspavlov:
IGNITE-6818 | Resolved
-- | --

PR #3088, opened 2017-11-24
IGNITE-6853: Cassandra cache store does not clean prepared statements…

jasonman107:
IGNITE-6853: Cassandra cache store does not clean prepared statements cache when remove old cassandra session


modules/cassandra/store/pom.xml
```diff
@@ -140,6 +140,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
```
irudyak: Mockito dependency should be specified in the "dependencyManagement" section of parent POM. Such way you don't need to specify <version> tag in all the modules which are going to reuse the dependecy - it will be automatically resolved from parent POM.

jasonman107: added dependencyManagement to the parent POM, such that the child POM doesn't need to specify the version.

However, unless we add it to the parent as a 'dependency', then the child pom still needs to add it as a dependency.

irudyak:
Please resolve the conflicts with master branch.

jasonman107:
Resolved the conflicts.  Please merge if it looks good.  Thanks.

PR #3100, opened 2017-11-28
IGNITE-5960 Ignite Continuous Query might not send update request

sunnychanwork:
During continuous query setup, there are potential updates in flight that would potentially have a situation like this:

T1 updates E1, lsnrs!=null, obtain update counter 1
T2 updates E2, lsnrs==null, obtain update counter 2
T3 updates E3, lsnrs!=null, obtain update counter 3

Notice that as E1 E2 and E3 are different there are no locks between them and they can proceed in parallel. As a result, the sequence of updates being sent to CQManager will be 1,3 with 2 missing and it will wait for update 2 forever which will never come.

To fix this I propose to use a ReadWrite lock to ensure that the Map Entry update will complete before setting up new continuous query.


dspavlov:

PR #3104, opened 2017-11-28
Ignite-6339-hdr-buf

agura:


agura:
Merged to ignite-6339

PR #3117, opened 2017-11-30
IGNITE-6880: KNN Regression and Classification

zaleslaw:
Also added Labeled Dataset and loading from txt file


oignatenko:
Please consider generifying LabeledDataset by using same type parameters as of underlying LabeledVector. Per my reading of code in LabeledDatasetTest this could lead to more type safe code. Such a change would probably better be done per separate ticket, because per my trial attempt to do this it may involve quite a lot of modifications - in particular member "data" would better change from array to list

modules/ml/src/main/java/org/apache/ignite/ml/knn/models/KNNModelFormat.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+package org.apache.ignite.ml.knn.models;
+
+import java.io.Serializable;
+import java.util.Arrays;
+import org.apache.ignite.ml.math.distances.DistanceMeasure;
+import org.apache.ignite.ml.structures.LabeledDataset;
+
+/** */
```
ybabak: Please add javadoc

ybabak: And same for getters

modules/ml/src/main/java/org/apache/ignite/ml/knn/models/KNNModel.java
```diff
@@ -0,0 +1,238 @@
+/*
```
ybabak: no package-info for package org.apache.ignite.ml.knn.models. Please add it

modules/ml/src/main/java/org/apache/ignite/ml/math/distributed/keys/impl/SparseMatrixKey.java
```diff
@@ -17,17 +17,18 @@
 
 package org.apache.ignite.ml.math.distributed.keys.impl;
 
-import java.io.Externalizable;
-import java.io.IOException;
```
ybabak: please check import format

modules/ml/src/main/java/org/apache/ignite/ml/structures/LabeledDataset.java
```diff
@@ -0,0 +1,428 @@
...
+
+    /**
+     * Creates new Labeled Dataset by matrix and vector of labels
+     * @param mtx Given matrix with rows as observations
+     * @param lbs Labels of observations
+     * @param featureNames Column names
+     * @param isDistributed Use distributed data structures to keep data
+     */
+    public LabeledDataset(double[][] mtx, double[] lbs, String[] featureNames, boolean isDistributed) {
+        assert mtx != null;
```
ybabak: please format method

PR #3138, opened 2017-12-04
IGNITE-7090

tonyschak:
IGNITE-7090 added onNodeRelease to the stop method to ensure a node stopping will release permits for future use.


modules/core/src/main/java/org/apache/ignite/internal/processors/datastructures/GridCacheSemaphoreImpl.java
```diff
@@ -588,6 +588,10 @@ private void initializeSemaphore() throws IgniteCheckedException {
 
         // Try to notify any waiting threads.
         sync.releaseShared(0);
+
+        //ensure semaphore is release if last node IGNITE-7090
+        UUID nodeID = ctx.localNodeId();
+        onNodeRemoved(nodeID);
```
vladisav: This requires cache operations therefore might hang; it should be done from separate thread (see invocation of onNode Removed in DataStructuresProcessor, maybe even reconsider moving this call to onKernalStop there, since same thing might be applicable to IgniteLock as well, what do you think?)

PR #3143, opened 2017-12-05
IGNITE-7085 When PDS is enabled and IGNITE_BINARY_SORT_OBJECT_FIELDS …

asfedotov:
…is set and IgniteCache#put is called node hangs


dspavlov:

PR #3148, opened 2017-12-05
Ignite-6846 Add metrics for entry processor invocations

voipp:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheMetricsSnapshot.java
```diff
@@ -981,5 +1123,20 @@ public CacheMetricsSnapshot(CacheMetrics loc, Collection<CacheMetrics> metrics)
         keysToRebalanceLeft = in.readLong();
         rebalancingBytesRate = in.readLong();
         rebalancingKeysRate = in.readLong();
+
+        // 6 long and 5 float values give 68 bytes in total.
+        if (in.available() >= 68) {
```
dspavlov: IMO we should avoid such magic numbers and extract constant, which value is coming from 6 long_size, 5 float instead of writting it in comment.

voipp: @dspavlov creating constant for the only and small usage looks superflows.
Do you find the comment incomplete ?

dspavlov: I mean we can use code to describe meaning of 68, so I think it would be example of good practice for Igniters. 

We all know most efforts in development is coming from code reading, not writting; so clean code will help all community.

voipp: @dspavlov  I agree, the code should be cleaner.
I changed the if clause to `if (in.available() >= 6 * 8/*size in bytes for long type*/ + 5 * 4/*size in bytes for float type*/ )`

PR #3158, opened 2017-12-06
IGNITE-7086 - Backups are not updated when ReadFromBackup=true and Re…

dkarachentsev:
…adThrough happens.


dspavlov:

PR #3236, opened 2017-12-15
IGNITE-5136: GridLogThrottle memory leak

SomeFire:


modules/core/src/main/java/org/apache/ignite/internal/util/GridLogThrottle.java
```diff
@@ -29,28 +30,72 @@
...
+     * @throws InterruptedException If interrupted.
+     */
+    static void onGridStop() throws InterruptedException {
+        synchronized (mux) {
+            if (cleaningTimer == null)
+                return;
+
+            cleaningTimer.cancel();
+
+            cleaningTimer = null;
```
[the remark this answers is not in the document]
SomeFire: Cancel is necessary because it stops the timer thread. Null will clear some space in case if JVM will not stopped with node stop.

modules/core/src/main/java/org/apache/ignite/internal/util/GridLogThrottle.java
```diff
@@ -29,28 +30,72 @@
...
      */
     public static void throttleTimeout(int timeout) {
         throttleTimeout = timeout;
+
+        synchronized (mux) {
+            if (cleaningTimer != null)
+                cleaningTimer.cancel();
+
+            refreshCleaningTimer();
+        }
```
1vanan: Why we should cancel and refresh timer in this place?

modules/core/src/test/java/org/apache/ignite/util/GridLogThrottleTest.java
```diff
@@ -34,82 +38,127 @@ public GridLogThrottleTest() {
...
+        log.reset();
 
-        // OMITTED.
-        LT.error(log, new RuntimeException("Test exception 1."), "Test1");
+        U.sleep(LT.throttleTimeout() * 2);
+        info("Slept for throttle timeout: " + LT.throttleTimeout() * 2);
 
-        // LOGGED.
-        LT.error(log, new RuntimeException("Test exception 2."), "Test");
+        assertEquals(0, LT.errorsSize());
```
1vanan: Why we test logger's timeout twice?

modules/core/src/test/java/org/apache/ignite/util/GridLogThrottleTest.java
```diff
@@ -34,82 +38,127 @@ public GridLogThrottleTest() {
...
+        assertEquals(1, LT.errorsSize());
 
-        info("Slept for throttle timeout: " + LT.throttleTimeout());
+        LT.error(log, new RuntimeException("Test exception 1."), "Test");
+        assertEquals(1, LT.errorsSize());
 
-        //LOGGED.
-        LT.info(log(), "Test info message.");
+        U.sleep(LT.throttleTimeout() * 2);
+        info("Slept for throttle timeout: " + LT.throttleTimeout() * 2);
```
1vanan: Can we change info message?
The info is about that we slept for timeout, but we slept twice as much.

modules/core/src/test/java/org/apache/ignite/util/GridLogThrottleTest.java
```diff
@@ -34,82 +38,127 @@ public GridLogThrottleTest() {
...
         LT.throttleTimeout(1000);
 
-        // LOGGED.
-        LT.error(log, new RuntimeException("Test exception 1."), "Test");
+        String sep = System.getProperty("line.separator");
 
-        // OMITTED.
-        LT.error(log, new RuntimeException("Test exception 1."), "Test");
+        // LOGGED.
+        LT.error(log, new RuntimeException("Test exception 1."), "Test msg");
```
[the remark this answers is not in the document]
SomeFire: I'm not sure that it is good idea. Messages are local and there are many of them. Creating separate variables for them all will produce more redundant lines of code.

modules/core/src/test/java/org/apache/ignite/util/GridLogThrottleTest.java
SomeFire: We check that it works not only once.

modules/core/src/main/java/org/apache/ignite/internal/util/GridLogThrottle.java
```diff
@@ -231,6 +283,16 @@ private static boolean replace(IgniteBiTuple<Class<? extends Throwable>, String>
...
     }
 
+    /**
+     * Method iterates though map and removes outdated entries (compares entry timestamp with current timestamp minus
+     * timeout). It protects the map from causing memory leak.
+     */
+    private static void cleanUpOldEntries() {
+        long curTs = U.currentTimeMillis();
+
+        errors.values().removeIf(aLong -> aLong <= curTs - throttleTimeout);
```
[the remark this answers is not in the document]
SomeFire: Done.

modules/core/src/test/java/org/apache/ignite/util/GridLogThrottleTest.java
SomeFire: Done.

modules/core/src/test/java/org/apache/ignite/util/GridLogThrottleTest.java
```diff
@@ -34,82 +38,97 @@ public GridLogThrottleTest() {
...
+        assertEquals(2, LT.errorsSize());
+        log.reset();
 
-        //OMMITED.
-        LT.info(log(), "Test info message.");
+        LT.throttleTimeout(300);
 
-        //OMMITED.
-        LT.info(log(), "Test info message.");
+        assertTrue("Error list wasnt cleared.",
```
1vanan: Correct this message, please.

PR #3237, opened 2017-12-15
ignite-7195: limit toString of arrays and additional values; add "and…

abeliak:
ignite-7195: limit toString of arrays and additional values; add "and N more".


dspavlov:

PR #3280, opened 2017-12-25
IGNITE-7135 IgniteCluster.startNodes() returns successful ClusterStartNodeResult even though the remote process fails.

akuramshingg:
Fix windows startup scripts.
Add startNodes() Windows support through task scheduling.
Check successful node startup by reading the output log.


dspavlov:

PR #3297, opened 2017-12-27
IGNITE-6783: Create common mechanism for group training.

artemmalykh:


modules/ml/src/main/java/org/apache/ignite/ml/math/Isomorphism.java
```diff
@@ -0,0 +1,42 @@
+/*
```
oignatenko: unused class, please consider deletion

modules/ml/src/main/java/org/apache/ignite/ml/math/util/StreamUtil.java
```diff
@@ -0,0 +1,32 @@
+/*
```
oignatenko: unused class, please consider deletion

modules/ml/src/main/java/org/apache/ignite/ml/trainers/group/GroupTrainer.java
```diff
@@ -0,0 +1,206 @@
...
+     */
+    Ignite ignite;
+
+    /**
+     * Construct an instance of this class.
+     *
+     * @param cache Cache on which training is performed.
+     * @param ignite Ignite instance.
+     */
+    public GroupTrainer(
```
oignatenko: given that class is declared package private, public modifier here looks redundant. Consider either declaring class public or removing public modifier from constructor

modules/ml/src/main/java/org/apache/ignite/ml/trainers/group/Metaoptimizer.java
```diff
@@ -0,0 +1,142 @@
+/*
```
oignatenko: please consider renaming this class with capitalized "O": MetaOptimizer

modules/ml/src/test/java/org/apache/ignite/ml/trainers/group/DistributedWorkersChainTest.java
```diff
@@ -0,0 +1,180 @@
+/*
```
oignatenko: Consider adding this and other tests in this package into an "umbrella" testsuite and adding that new testsuite to IgniteMLTestSuite

modules/ml/src/main/java/org/apache/ignite/ml/math/functions/Functions.java
```diff
@@ -215,15 +215,73 @@
...
+    /**
+     * Transform bi-function of the form (a, b) -> c into a function of form a -> (b -> c).
+     *
+     * @param f Function to be curried.
+     * @param <A> Type of first argument of function to be transformed.
+     * @param <B> Type of second argument of function to be transformed.
+     * @param <C> Type of third argument of function to be transformed.
+     * @return Curried bi-function.
+     */
+    public static <A, B, C> IgniteCurriedBiFunction<A, B, C> curry(IgniteBiFunction<A, B, C> f) {
```
dmitrievanthony: As far as I understand `curry` is an operator that accepts a function with N arguments and M values and produces a function with N-M arguments. So, [currying](https://en.wikipedia.org/wiki/Currying#Definition) in this context is an action of fixing subset of function arguments and the result of the currying should be the function with N-M arguments.
In the current implementation I see that the `curry` method returns the function with the same number of elements as in the accepted function. For me it looks like it should be `IgniteCurryingFunction`, not `Curried`, shouldn't it? 
Or even better, the `curry` method can accept a function and arguments and return actually curried function (just an `IgniteFunction` with one argument in this case) to avoid extra function wrappers.

```
public static <A, B, C> IgniteFunction<B, C> curry(IgniteBiFunction<A, B, C> f, A a) {
  return b -> f.apply(a, b);
}
```

What do you think?

modules/ml/src/main/java/org/apache/ignite/ml/math/functions/IgniteCurriedTriFunction.java
```diff
@@ -0,0 +1,28 @@
...
+package org.apache.ignite.ml.math.functions;
+
+import java.io.Serializable;
+
+/**
+ * Serializable curried tri-function.
+ *
+ * @see IgniteCurriedBiFunction
+ */
+public interface IgniteCurriedTriFunction<A, B, C, D> extends IgniteFunction<A, IgniteCurriedBiFunction<B, C, D>>, Serializable {
```
dmitrievanthony: In according with my comment about `curry` method don't clear why we need this interface.

modules/ml/src/main/java/org/apache/ignite/ml/trainers/Trainer.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.apache.ignite.ml.Model;
+
+/**
+ * Interface to train {@link Model}.
+ *
+ * @param <M> Model type.
+ * @param <T> Type of data to use in training.
+ */
+public interface Trainer<M extends Model, T> {
```
dmitrievanthony: Why this file just has been added? Looks strange for me, as far as I see it's already exists.

modules/ml/src/main/java/org/apache/ignite/ml/trainers/group/BaseLocalProcessorJob.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override public void cancel() {
+
+    }
+
+    /** {@inheritDoc} */
+    @Override public R execute() throws IgniteException {
+        List<ResultAndUpdates<R>> resultsAndUpdates = toProcess().
```
dmitrievanthony: Just a question about style. Do we put dot at the end of a line, not start of the next?

modules/ml/src/main/java/org/apache/ignite/ml/trainers/group/MetaoptimizerDistributedStep.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.ignite.ml.math.functions.IgniteFunction;
+import org.apache.ignite.ml.math.functions.IgniteSupplier;
+import org.apache.ignite.ml.trainers.group.chain.DistributedEntryProcessingStep;
+import org.apache.ignite.ml.trainers.group.chain.EntryAndContext;
+import org.apache.ignite.ml.trainers.group.chain.HasTrainingUUID;
+
+/**
+ * Distributed step
+ *
+ * @param <L>
```
dmitrievanthony: Missing documentation about types.

PR #3310, opened 2017-12-28
IGNITE-7239 - In case of not serializable cache update response, futu…

dkarachentsev:
…re on node requester will never complete


dspavlov:

PR #3317, opened 2017-12-28
IGNITE-7334: MBeans fail to be created for caches with special symbols in names

slukyano:


dspavlov:

PR #3318, opened 2017-12-28
IGNITE-7264: Caches with forward slash "/" in names cause problems for PDS

slukyano:


dspavlov:
The PR related to a resolved ticket, closing

PR #3341, opened 2018-01-09
IGNITE-7277: override equals() & hashCode() methods for JdbcTableMeta…

tledkov-gridgain:
… to prevent tables duplication at the tables metadata result set.


lxiongh:
why delete ignite-7277 branch? @devozerov

PR #3344, opened 2018-01-09
ignite-7365 Switch segment rollover hanging is fixed

agura:


agura:
Merged to master branch.

PR #3353, opened 2018-01-10
IGNITE-7191 JDBC: set socket buffer to 64k by default

apopovgg:


dspavlov:

PR #3354, opened 2018-01-10
IGNITE-7380: Avoid updating PagePartitionCounters in case all counters were not modified

dspavlov:


dspavlov:
To be processed under #3365

PR #3373, opened 2018-01-15
Assert on last fsync position

agura:


agura:
New PR created. See https://issues.apache.org/jira/browse/IGNITE-7412

PR #3430, opened 2018-01-24
ignite-7502 : Disabled BLT for non-persistent caches.

ilantukh:


dspavlov:

PR #3440, opened 2018-01-26
Ignite-7019

x-kreator:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesList.java
```diff
@@ -613,13 +615,20 @@ protected final void put(
...
+            Stripe stripe = getPageForPut(bucket, bag);
 
             final long tailId = stripe.tailId;
             final long tailPage = acquirePage(tailId);
 
             try {
-                long tailAddr = writeLockPage(tailId, tailPage, bucket, lockAttempt++); // Explicit check.
+                long tailAddr = writeLockPage(tailId, tailPage, bucket, lockAttempt++, bag); // Explicit check.
+
+                if (bag != null && bag.isEmpty()) {
```
gvvinblade: the condition should be:

``` 
               if (tailAddr == 0L) {
                    if (bag != null && bag.isEmpty())
                        return;
                    else
                        continue;
                }
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesList.java
```diff
@@ -613,13 +615,20 @@ protected final void put(
...
+                    if (tailAddr != 0L)
+                        writeUnlock(tailId, tailPage, tailAddr, false);
```
gvvinblade: you need to put the check to the finally block and not to call unlock here

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesList.java
```diff
@@ -613,13 +615,20 @@ protected final void put(
         assert bag == null ^ dataAddr == 0L;
 
         for (int lockAttempt = 0; ;) {
-            Stripe stripe = getPageForPut(bucket);
+            Stripe stripe = getPageForPut(bucket, bag);
```
gvvinblade: Here one more check like the one below is needed, because you may create a stripe inside the `getPageForPut` method

```
          if (bag != null && bag.isEmpty())
                return;
```

dspavlov:
 

PR #3468, opened 2018-02-02
IGNITE-7508: Fix contention on system property access in GridKernalContextImpl::isDaemon()

AMashenkov:
Fixed.


amirakhmedov:
I know with this patch you are keeping a consistency with existing logic but why Ignite has "OR" logic? Usually system properties have precedence over configuration files. For me, "OR" over here is kind of misleading.

AMashenkov:
@amirakhmedov ,
agree with you, but changing condition checks order is not related the original issue, so, I've just left order "as is".

Looks like it is legacy-code and I doubt this property is ever useful. I'd think, at first, we should decide either to remove this option at all or fix the order, and then make any changes.
We can't drop property until next major release, so the only way for now is to make smth I suggest in PR.

Feel free to create a ticket for this.

PR #3494, opened 2018-02-08
IGNITE-7438 LSQR solver for Linear Regression

dmitrievanthony:


modules/ml/src/main/java/org/apache/ignite/ml/math/isolve/lsqr/LSQROnHeap.java
```diff
@@ -0,0 +1,102 @@
...
+
+import com.github.fommil.netlib.BLAS;
+import java.util.Arrays;
+import org.apache.ignite.ml.dataset.Dataset;
+import org.apache.ignite.ml.dataset.DatasetBuilder;
+import org.apache.ignite.ml.dataset.PartitionDataBuilder;
+import org.apache.ignite.ml.math.isolve.LinSysPartitionDataOnHeap;
+
+/**
+ * Distributed implementation of LSQR algorithm based on.
```
ybabak: based on what?

PR #3515, opened 2018-02-13
IGNITE-7640: make DiscoveryDataClusterState immutable

SharplEr:
Done. Please review.

**Details:**

1) I removed field `DiscoveryDataClusterState.transitionError` because there is no place in the project where setter was called, so it always was null.

2) Replaced `DiscoveryDataClusterState.transitionRes` with `DiscoveryDataClusterState.transitionFut` to avoid concurrent passing result from `GridClusterStateProcessor#onStateFinishMessage` to `GridClusterStateProcessor#publicApiActiveState` . It also allows me to remove `GridClusterStateProcessor.transitionFuts`. See details in code.

3) Optimized `DiscoveryDataClusterState#setTransitionResult` using because in all cases except one we don't need comparison of request id. See details in code.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
@@ -858,9 +858,6 @@ public class GridDhtPartitionsExchangeFuture extends GridDhtTopologyFutureAdapte
 
         DiscoveryDataClusterState state = cctx.kernalContext().state().clusterState();
 
-        if (state.transitionError() != null)
-            changeGlobalStateE = state.transitionError();
-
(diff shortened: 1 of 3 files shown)


SharplEr:
I'm sorry but I can't do this issue. After strict investigation of tests results, I still see failed tests with my PR. And I truly don't understand a reason.
I found two type of unclear using of `DiscoveryDataClusterState`.

**1. Concurrency using**
`DiscoveryDataClusterState.transitionRes` + `GridClusterStateProcessor.transitionFuts` are working together like a future: `DiscoveryDataClusterState.transitionRes` used for value passing and `GridClusterStateProcessor.transitionFuts` used as a latch.

**2. Serialization**
There is logic based on the fact some fields are `transient` and some other is not. It breaks some invariants which I think I see in code (object become partly initialized).

But some bugs still here, so maybe there is something else.

I think we should make clear using of `DiscoveryDataClusterState` before this issue became resolvable.

Or maybe someone else will be lucky.

PR #3527, opened 2018-02-14
IGNITE-7681: SQL COPY performance fixes and performance test

gg-shq:


gg-shq:
Did a rebase + merge, which produced wrong history of commits, so cherry-picked commits to another branch. The correct pull request is:
https://github.com/apache/ignite/pull/3528

PR #3577, opened 2018-02-27
IGNITE-7796: Adopt kNN classification example to the new datasets

zaleslaw:


zaleslaw:
Was merged by @ybabak

PR #3581, opened 2018-02-28
IGNITE-5357 replicated cache reads load balancing was added

DALukjanenko:


dspavlov:

PR #3607, opened 2018-03-06
IGNITE-7876: Adopt SVM Binary Classification to the Partitioned Dataset

zaleslaw:
* Removed Multiclass-classification related code
* Fixed tests
* Adopted Model and Trainer


modules/ml/src/main/java/org/apache/ignite/ml/svm/SVMLinearBinaryClassificationTrainer.java
```diff
@@ -32,33 +35,50 @@
...
-        final int weightVectorSizeWithIntercept = data.colSize() + 1;
+        assert datasetBuilder != null;
+
+        PartitionDataBuilder<K, V, SVMPartitionContext, LabeledDataset<Double, LabeledVector>> partDataBuilder = new SVMPartitionDataBuilderOnHeap<>(
+            featureExtractor,
+            lbExtractor,
+            cols
+        );
+
+        this.dataset = datasetBuilder.build(
```
dmitrievanthony: Dataset is `AutoCloseable` object, so after `fit` it would be great to close it to clear resources. Try-with-resources is preferable.

modules/ml/src/main/java/org/apache/ignite/ml/svm/SVMLinearBinaryClassificationTrainer.java
```diff
@@ -67,34 +87,32 @@
...
 
     /** */
-    private Vector calculateUpdates(LabeledDataset data, Vector weights) {
-        Vector copiedWeights = weights.copy();
-        Vector deltaWeights = initializeWeightsWithZeros(weights.size());
-
-        final int amountOfObservation = data.rowSize();
+    private Vector calculateUpdates(Vector weights) {
+        return dataset.compute(data -> {
+            Vector copiedWeights = weights.copy();
```
dmitrievanthony: Looks like we can avoid `copy` call because `weights` are passed via serialization-deserialization procedure anyway, so it's not the same object.

modules/ml/src/main/java/org/apache/ignite/ml/svm/SVMLinearBinaryClassificationTrainer.java
```diff
@@ -32,33 +35,50 @@
...
+    private int amountOfLocIterations = 100;
 
     /** Regularization parameter. */
-    private double lambda = 0.2;
+    private double lambda = 0.4;
 
-    /** This flag enables distributed mode for this algorithm. */
-    private boolean isDistributed;
+    /** Dataset. */
+    private Dataset<SVMPartitionContext, LabeledDataset<Double, LabeledVector>> dataset;
```
dmitrievanthony: Would be better to make `dataset` a variable within the `fit` method (not a class field). And close if immediately after fitting as well.

modules/ml/src/main/java/org/apache/ignite/ml/svm/SVMLinearBinaryClassificationTrainer.java
```diff
@@ -32,33 +35,50 @@
...
+            cols
+        );
+
+        this.dataset = datasetBuilder.build(
+            (upstream, upstreamSize) -> new SVMPartitionContext(),
+            partDataBuilder
+        );
+
+
+        final int weightVectorSizeWithIntercept = cols + 1;
```
dmitrievanthony: Do we need it to be final?

modules/ml/src/main/java/org/apache/ignite/ml/svm/SVMLinearBinaryClassificationTrainer.java
```diff
@@ -67,34 +87,32 @@
...
+            final int amountOfObservation = data.rowSize();
 
-        Vector tmpAlphas = initializeWeightsWithZeros(amountOfObservation);
-        Vector deltaAlphas = initializeWeightsWithZeros(amountOfObservation);
+            Vector tmpAlphas = initializeWeightsWithZeros(amountOfObservation);
+            Vector deltaAlphas = initializeWeightsWithZeros(amountOfObservation);
 
-        for (int i = 0; i < this.getAmountOfLocIterations(); i++) {
-            int randomIdx = ThreadLocalRandom.current().nextInt(amountOfObservation);
+            for (int i = 0; i < this.getAmountOfLocIterations(); i++) {
```
dmitrievanthony: It looks like due to  'this.getAmountOfLocIterations()' the whole instance of the trainer will be serialized every time this lambda will be sent anyway. Am I right? The same question about `initializeWeightsWithZeros` method calls.

Bye the way, it might be not a problem if we remove field with dataset. But in this case we need to mark the class as `Serializable` anyway.

modules/ml/src/main/java/org/apache/ignite/ml/svm/SVMPartitionDataBuilderOnHeap.java
```diff
@@ -0,0 +1,86 @@
...
+        this.xExtractor = xExtractor;
+        this.yExtractor = yExtractor;
+        this.cols = cols;
+    }
+
+    /** {@inheritDoc} */
+    @Override public LabeledDataset<Double, LabeledVector> build(Iterator<UpstreamEntry<K, V>> upstreamData, long upstreamDataSize,
+                                                                 C ctx) {
+
+        double[][] x = new double[Math.toIntExact(upstreamDataSize)][cols];
```
dmitrievanthony: It's not required to extend `X` matrix by one additional column here like in another linear methods, did I get it right?

modules/ml/src/main/java/org/apache/ignite/ml/svm/SVMLinearBinaryClassificationTrainer.java
```diff
@@ -67,34 +87,32 @@
...
-    private Vector calculateUpdates(LabeledDataset data, Vector weights) {
-        Vector copiedWeights = weights.copy();
-        Vector deltaWeights = initializeWeightsWithZeros(weights.size());
-
-        final int amountOfObservation = data.rowSize();
+    private Vector calculateUpdates(Vector weights) {
+        return dataset.compute(data -> {
+            Vector copiedWeights = weights.copy();
+            Vector deltaWeights = initializeWeightsWithZeros(weights.size());
+            final int amountOfObservation = data.rowSize();
```
dmitrievanthony: Do we need it to be final?

PR #3610, opened 2018-03-06
IGNITE-7864 Control utility: Add confirm on dangerous operations

macrergate:


dspavlov:
IGNITE-7864 | Closed
-- | --

PR #3617, opened 2018-03-07
IGNITE-6860 Add logging and additional info while BinaryMarshaller fa…

ivandasch:
…ils to serialize object.


dspavlov:

PR #3619, opened 2018-03-10
IGNITE-7654 Add processing for JTS Geometry 

ivandasch:
Now geometry deserializes to WKT.  Also, add Spatial JDBC Example


modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
```diff
@@ -984,6 +985,28 @@ public static boolean isGeometryClass(Class<?> cls) {
         return GEOMETRY_CLASS != null && GEOMETRY_CLASS.isAssignableFrom(cls);
     }
 
+    /**
+     * Checks if the given class is GEOMETRY.
```
tyrasd: is this method description correct?

PR #3636, opened 2018-03-14
IGNITE-7912 Manage unused WAL segments thru control.sh

ivandasch:
from all nodes.


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -886,6 +886,46 @@ private boolean hasIndex(long absIdx) {
         return deleted;
     }
 
+    /** {@inheritDoc} */
+    @Override public File[] canBeTruncated(WALPointer low, WALPointer high){
```
macrergate: This mainly copy-pasted from truncate method. How about make refactoring to avoid copy-paste?

PR #3665, opened 2018-03-21
ignite-8004 set compatibilityMode to true in collectGridNodeData

akalash:


modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/GridClusterStateProcessor.java
```diff
@@ -636,46 +636,50 @@ protected IgniteCheckedException concurrentStateChangeError(boolean activate) {
...
-                        int lastId = joiningNodeState.baselineTopology().id();
+            return;
+        }
 
-                        historyToSend = bltHist.tailFrom(lastId);
-                    }
-                    else
-                        historyToSend = bltHist;
-                }
+        BaselineTopologyHistory historyToSend = null;
```
dspavlov: Abbreviation should be used

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/GridClusterStateProcessor.java
```diff
@@ -636,46 +636,50 @@ protected IgniteCheckedException concurrentStateChangeError(boolean activate) {
...
+        DiscoveryDataClusterState joiningNodeState = null;
+
+        try {
+            if (joiningNodeData.joiningNodeData() != null)
+                joiningNodeState = marsh.unmarshal(
+                    (byte[])joiningNodeData.joiningNodeData(),
+                    U.resolveClassLoader(ctx.config())
+                );
+        }
+        catch (IgniteCheckedException e) {
```
dspavlov: Stacktrace of exception is not logged, eception  is ignored. Is it reasonable to log? Is it reasonable to rename to 'ignored'?

PR #3682, opened 2018-03-22
IGNITE-7902 Enable either swap space or persistence but not both

aealeksandrov:
Enabling both swap and persistence at the same time for a data region does not make sense. An exception should be thrown in this case.

See discussion - http://apache-ignite-developers.2346864.n4.nabble.com/Enabling-swap-space-and-Ignite-Persistence-td27595.html

IgniteCheckedException will be thrown in case if swap and persistence were enabled at the same time.

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheDataRegionConfigurationTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheDataRegionConfigurationTest.java
@@ -16,15 +16,19 @@
  */
 package org.apache.ignite.internal.processors.cache;
 
+import java.util.concurrent.Callable;
 import org.apache.ignite.IgniteCache;
 import org.apache.ignite.IgniteCheckedException;
 import org.apache.ignite.configuration.CacheConfiguration;
+import org.apache.ignite.configuration.DataPageEvictionMode;
 import org.apache.ignite.configuration.DataRegionConfiguration;
 import org.apache.ignite.configuration.DataStorageConfiguration;
 import org.apache.ignite.configuration.IgniteConfiguration;
 import org.apache.ignite.internal.IgniteEx;
 import org.apache.ignite.internal.mem.IgniteOutOfMemoryException;
+import org.apache.ignite.testframework.GridTestUtils;
 import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.jetbrains.annotations.Nullable;
 
 /**
  *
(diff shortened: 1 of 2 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheDataRegionConfigurationTest.java
```diff
@@ -169,4 +173,157 @@ public void testProperlySizedMemoryPolicy() throws Exception {
...
+                startGrid(0);
+                return null;
+            }
+        }, IgniteCheckedException.class, "Failed to start processor: GridProcessorAdapter []");
+
+        // Setting the page eviction threshold greater than 0.999
+        invCfg.setEvictionThreshold(BIG_EVICTION_THRESHOLD);
+        memCfg.setDataRegionConfigurations(invCfg);
+
+        GridTestUtils.assertThrows(log(), new Callable<Object>() {
```
slukyano: I'd rather have one `assertThrows` per test, to have the checks run independently. In it cumbersome in JUnit 3, but I'd usually do it like
```
public void testA() { check("a"); }
public void testB() { check("b"); }
private void check(String param) { /* actual test code goes here */ }
```

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheDataRegionConfigurationTest.java
```diff
@@ -169,4 +173,157 @@ public void testProperlySizedMemoryPolicy() throws Exception {
...
+                startGrid(0);
+                return null;
+            }
+        }, IgniteCheckedException.class, "Failed to start processor: GridProcessorAdapter []");
+    }
+
+    /**
+     * Verifies that {@link IgniteCheckedException} is thrown when page eviction threshold isn't between 0.5 and 0.999.
+     */
+    public void testSetInvalidEviction() {
```
slukyano: Ideally, additional test cases not related to the fix should be committed under a separate JIRA issue (like "Add more test cases to CacheDataRegionConfigurationTest"), but personally I'm OK with piggybacking them here as well.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheDataRegionConfigurationTest.java
aealeksandrov: Ok I will fix.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheDataRegionConfigurationTest.java
aealeksandrov: Got it.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheDataRegionConfigurationTest.java
aealeksandrov: Updated.

slukyano:
Looks good.

PR #3692, opened 2018-03-23
Suspend resume benchmark

voipp:


modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteGetTxExplicitBenchmark.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Ignite benchmark that performs get operation inside transaction.
+ */
+public class IgniteGetTxExplicitBenchmark extends IgniteCacheAbstractBenchmark<Integer, Object> {
+    /** IgniteTransactions. */
+    private IgniteTransactions txs;
+
+    /** Get operation callable. */
+    private Callable<Void> clo;
```
SharplEr: Rename to  `c` or `call` according to [abbreviation rules](https://cwiki.apache.org/confluence/display/IGNITE/Abbreviation+Rules). `clo` is very rarely used in the project.

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
```diff
@@ -0,0 +1,114 @@
...
+    private IgniteTransactions txs;
+
+    /** ExecutorService. */
+    private ExecutorService ex;
+
+    /**
+     * First keyCounter that can be used by transaction in certain thread.
+     * The last one can be calculated based on keys per thread
+     */
+    private ThreadLocal<Integer> firstKey = new ThreadLocal<>();
```
SharplEr: Maybe better use `ThreadLocal.withInitial()`?

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
```diff
@@ -0,0 +1,114 @@
...
+
+    /** Number of keys, allocated to every thread. */
+    private int keysNumbPerThread = 0;
+
+    /** Helps to calculate first key for every thread. */
+    private AtomicInteger keyCntr = new AtomicInteger(0);
```
SharplEr: You also can put `keyCntr` into `Supplier` in `ThreadLocal.withInitial()` as a final field. You can create a static class for it. It helps you to separate a key per thread logic from benchmarking logic, I guess.

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteGetTxExplicitBenchmark.java
voipp: fixed

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
voipp: Or just override initialValue() in ThreadLocal. Tnx

voipp: fixed

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
voipp: @SharplEr  Its just a benchmark, creating new inner class will no doubt over-complicate it

SharplEr: Okay, but you can still put `keyCntr` inside the `ThreadLocal` child you created. You don't use `keyCntr` in `IgniteTxSuspendResumeBenchmark` class.

voipp: ok, lets move it there

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
```diff
@@ -0,0 +1,113 @@
...
+        int key = ThreadLocalRandom.current().nextInt(origin, origin + keysNumbPerThread);
+
+        try (Transaction tx = txs.txStart(args.txConcurrency(), args.txIsolation())) {
+            IgniteCache<Integer, Object> cache = cacheForOperation();
+
+            cache.put(key, new SampleValue(key));
+
+            tx.suspend();
+
+            ex.submit(new Callable<Object>() {
```
SharplEr: Use `Callable<Void>`.

SharplEr: Or lambda:)

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteGetTxExplicitBenchmark.java
```diff
@@ -0,0 +1,80 @@
...
+
+        if (args.preloadAmount() > args.range())
+            throw new IllegalArgumentException("Preloading amount (\"-pa\", \"--preloadAmount\") " +
+                "must by less then the range (\"-r\", \"--range\").");
+
+        loadCachesData();
+
+        txs = ignite().transactions();
+
+        call = new Callable<Void>() {
```
SharplEr: Maybe better to use a lambda expression?

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
```diff
@@ -0,0 +1,113 @@
...
+    private IgniteTransactions txs;
+
+    /** ExecutorService. */
+    private ExecutorService ex;
+
+    /**
+     * First keyCounter that can be used by transaction in certain thread.
+     * The last one can be calculated based on keys per thread
+     */
+    private ThreadLocal<Integer> firstKey = new ThreadLocal<Integer>() {
```
SharplEr: Can be `final`

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
```diff
@@ -0,0 +1,113 @@
...
+        /** Helps to calculate first key for every thread. */
+        private AtomicInteger keyCntr = new AtomicInteger(0);
```
SharplEr: Can be `final`

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.ignite.IgniteTransactions;
+import org.apache.ignite.transactions.Transaction;
+import org.apache.ignite.transactions.TransactionState;
+import org.apache.ignite.yardstick.cache.model.SampleValue;
+import org.yardstickframework.BenchmarkConfiguration;
+
+/**
+ * Ignite benchmark that performs suspend\resume operations.
+ * Concurrent threads must operate different keys, so there will be no waiting on key locks.
+ * So, all keys are separated on different groups, defined by first key and last key(first key plus keys number per thread).
```
SharplEr: The line is too long (125 > 120)

SharplEr: Also add space here `key_(first`

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * First keyCounter that can be used by transaction in certain thread.
+     * The last one can be calculated based on keys per thread
+     */
+    private ThreadLocal<Integer> firstKey = new ThreadLocal<Integer>() {
+        /** Helps to calculate first key for every thread. */
+        private AtomicInteger keyCntr = new AtomicInteger(0);
+
+        @Override protected Integer initialValue() {
+            return keyCntr.getAndAdd(keysNumbPerThread);
```
SharplEr: Okay, but now you have a bit weird link from `firstKey` to `this.keysNumbPerThread`. And I'm not sure about `HB` between `ThreadLocal.initialValue` and `IgniteTxSuspendResumeBenchmark.setUp`. I think it's better to move `firstKey` initialization in `setUp` as all other fields. Then you can use `keysNumbPerThread` inside `initialValue` as a constant without any race.

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
```diff
@@ -0,0 +1,113 @@
...
+
+            ex.submit(new Callable<Object>() {
+                @Override public Object call() {
+                    tx.resume();
+
+                    tx.commit();
+
+                    return null;
+                }
+            }).get();
```
SharplEr: Why you use `ex.submit(...).get()`?

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /** {@inheritDoc} */
+    @Override public void tearDown() throws Exception {
+        super.tearDown();
+
+        ex.shutdown();
+    }
+
+    /** {@inheritDoc} */
+    @Override public boolean test(Map<Object, Object> ctx) throws Exception {
```
SharplEr: `Transaction` creation and `key` putting are parts of your benchmark. So the result will not pure. I think it's possible to move write to javadoc how to calculate a pure result. Like `score_of_this_benchmark - score_of_IgnitePutTxBenchmark`.

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
voipp: why? in any case null would be returned

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteGetTxExplicitBenchmark.java
voipp: how will it improve current code?)

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
voipp: ok, let it be final

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
voipp: fixed

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteGetTxExplicitBenchmark.java
```diff
@@ -0,0 +1,80 @@
...
+            @Override public Void call() throws Exception {
+                IgniteCache<Integer, Object> cache = cacheForOperation();
+
+                int key = nextRandom(args.range());
```
SharplEr: I think it's better to create a local variable `int range = args.range()` and pass it into `Callable`. Because if one day someone removes `final` from `IgniteAbstractBenchmark.args` field then this code will be broken.

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
voipp: check the javadoc for setUp() : "This method is invoked right before the test(Map) method".
You also can check up BenchmarkDriverStartUp and BenchmarkRunner classes to find out how it works

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteGetTxExplicitBenchmark.java
SharplEr: As I know we have stopped support `Java 7`, so you **can** use lambda when it simplifies your code. You can still use anonymous classes, I'm just giving you advise.

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
voipp: because we want to wait for operation finish and chek whether exception happened

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteTxSuspendResumeBenchmark.java
SharplEr: Yes, and `Void` type exists exactly for this situation. You also use `Void` in another class :)

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteGetTxExplicitBenchmark.java
voipp: @SharplEr  In Ignite 2.4 java 8 must be used. In Ignite 2.3 java 7 must be used, so lambdas are prohibited

PR #3715, opened 2018-03-29
IGNITE-7771 Names of Ignite JMX beans should not be quoted unless required

aealeksandrov:
Names of Ignite JMX beans and bean groups are currently quoted if they contain non-alphanumeric characters. This leads to names with spaces, e.g. Thread Pools, appearing as "Thread Pools". Moreover, Thread Pools and "Thread Pools" are recognized by JMX as distinct names, so code accessing that MBean needs to take that into account.

It would be better not to quote the bean and group names unless they contain specific control characters. That way users won't need to quote names in their JMX clients.

Logic of generation of the ObjectName instances was updates. Also new tests were added.

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/GridMBeansTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/GridMBeansTest.java
@@ -17,12 +17,19 @@
 
 package org.apache.ignite.internal;
 
+import java.lang.management.ManagementFactory;
+import org.apache.ignite.Ignite;
 import org.apache.ignite.configuration.ExecutorConfiguration;
 import org.apache.ignite.configuration.IgniteConfiguration;
 import org.apache.ignite.internal.util.IgniteUtils;
 import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
 
 import javax.management.ObjectName;
+
+import static org.apache.ignite.IgniteSystemProperties.IGNITE_MBEAN_APPEND_CLASS_LOADER_ID;
+import static org.apache.ignite.IgniteSystemProperties.IGNITE_MBEAN_APPEND_JVM_ID;
+import static org.apache.ignite.IgniteSystemProperties.getBoolean;
+import static org.apache.ignite.internal.util.IgniteUtils.JMX_DOMAIN;
 
 /**
  * Tests for the standard JMX beans registered by the kernal.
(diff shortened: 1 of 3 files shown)


modules/core/src/test/java/org/apache/ignite/internal/GridMBeansTest.java
```diff
@@ -76,11 +83,83 @@ public void testExecutorBeans() throws Exception {
...
+
+        assertTrue(grid().configuration().getMBeanServer().isRegistered(new ObjectName(EXPECTED_BEAN_NAME)));
+    }
+
+    private String getExpectedBeanName(String domain, String instanceName, String grp, String name) {
+        StringBuffer sb = new StringBuffer();
+
+        sb.append(domain + ":");
+
+        if (getBoolean(IGNITE_MBEAN_APPEND_CLASS_LOADER_ID, true)) {
```
slukyano: Would it be possible to set both IGNITE_MBEAN_APPEND_CLASS_LOADER_ID and IGNITE_MBEAN_APPEND_JVM_ID to false, and then just hardcode the entire name string like "igniteInstanceName=mygrid,group=Thread Pools,name=GridSystemExecutor"? Then I guess it would be possible to check these names in the checkBean() test using them instead of makeMBeanName() call (basically, I'd rather get rid of the makeMBeanName() in this test).

aealeksandrov: I will check.

aealeksandrov: done

modules/core/src/test/java/org/apache/ignite/internal/GridMBeansTest.java
```diff
@@ -52,6 +52,10 @@ public GridMBeansTest() {
      * This implementation registers adds custom executors to the configuration.
      */
     @Override protected IgniteConfiguration getConfiguration(String igniteInstanceName) throws Exception {
+        System.setProperty(IGNITE_MBEAN_APPEND_CLASS_LOADER_ID, "false");
```
slukyano: Could you please set the properties back to the initial values after the test completes?
I'd suggest to override `setUp()` and `tearDown()` methods for that.

modules/core/src/test/java/org/apache/ignite/internal/GridMBeansTest.java
```diff
@@ -41,9 +41,39 @@
...
+    private String jvmIdProp;
+
     /** Create test and auto-start the grid */
     public GridMBeansTest() {
-        super(true);
+        super(false);
+    }
+
+    /**
+     * {@inheritDoc}
```
slukyano: Minor: can be a one-line doc (`/** {@inheritDoc} */`)

slukyano:
Looks good!

modules/core/src/test/java/org/apache/ignite/internal/GridMBeansTest.java
aealeksandrov: +

PR #3735, opened 2018-04-02
IGNITE-8106 Collect suppressed exceptions from causes.

alamar:


modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
```diff
@@ -748,6 +746,29 @@ public static Throwable getCause(Throwable throwable, String[] mtdNames) {
...
     }
 
+    /**
+     * Collects suppressed exceptions from throwable and all it causes.
+     *
+     * @param t Throwable.
+     * @return List of suppressed throwables.
+     */
+    public static List<Throwable> getSuppresseds(Throwable t) {
+        assert t != null;
```
alamar: Remove

PR #3742, opened 2018-04-03
IGNITE-8110

antkr:
Fixed ms to nanos transformation.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/store/GridCacheWriteBehindStore.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/store/GridCacheWriteBehindStore.java
@@ -895,7 +895,7 @@ public class GridCacheWriteBehindStore<K, V> implements CacheStore<K, V>, Lifecy
         protected Thread thread;
 
         /** Cache flushing frequence in nanos. */
-        protected long cacheFlushFreqNanos = cacheFlushFreq * 1000;
+        protected long cacheFlushFreqNanos = cacheFlushFreq * 1000 * 1000;
 
         /** Writer lock. */
         private final Lock flusherWriterLock = new ReentrantLock();
--- a/modules/core/src/test/java/org/apache/ignite/internal/processors/cache/store/GridCacheWriteBehindStoreSelfTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/processors/cache/store/GridCacheWriteBehindStoreSelfTest.java
@@ -120,12 +120,17 @@ public class GridCacheWriteBehindStoreSelfTest extends GridCacheWriteBehindStore
 
         try {
             long start = System.nanoTime();
+
             for ( int i =0; i < threshold/2; i ++ )
                 store.write(new CacheEntryImpl<>(i, "v" + i));
+
             U.sleep(FLUSH_FREQUENCY + 300);
+
             for ( int i = threshold/2; i < threshold; i ++ )
                 store.write(new CacheEntryImpl<>(i, "v" + i));
+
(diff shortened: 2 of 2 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/store/GridCacheWriteBehindStoreSelfTest.java
```diff
@@ -106,6 +106,42 @@ public void testSimpleStoreWithoutCoalescing() throws Exception {
         testSimpleStore(false);
     }
 
+    /**
+     * Checks that write behind cache flush frequency was correctly adjusted to nanos
+     * Expecting putAllCnt to be less or equal than elapsed time divided by flush frequency
```
dspavlov: Please use dot at the end of sentence

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/store/GridCacheWriteBehindStoreSelfTest.java
```diff
@@ -106,6 +106,42 @@ public void testSimpleStoreWithoutCoalescing() throws Exception {
...
+    public void testSimpleStoreFlushFrequencyWithoutCoalescing() throws Exception {
+        initStore(1, false);
+
+        long writeBehindFlushFrequencyNanos = FLUSH_FREQUENCY * 1000 * 1000;
+        int threshold = store.getWriteBehindStoreBatchSize() / 10;
+
+        try {
+            long start = System.nanoTime();
+
+            for (int i =0; i < threshold / 2; i++)
```
dspavlov: seems code reformat (including automatic) will be useful here

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/store/GridCacheWriteBehindStore.java
```diff
@@ -895,7 +895,7 @@ private void wakeUp() {
         protected Thread thread;
 
         /** Cache flushing frequence in nanos. */
-        protected long cacheFlushFreqNanos = cacheFlushFreq * 1000;
+        protected long cacheFlushFreqNanos = cacheFlushFreq * 1000 * 1000;
```
dspavlov: Just wondering, do we have some constant like NANOS_IN_MILLIS in Ignite?

antkr: No such public constants are present in Ignite, we have private NANOS_IN_MICROSECOND in CacheMetricsImpl and NANOS_IN_SECOND in IntervalBasedMeasurement.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/store/GridCacheWriteBehindStoreSelfTest.java
antkr: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/store/GridCacheWriteBehindStoreSelfTest.java
antkr: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/store/GridCacheWriteBehindStore.java
dspavlov: May be in future we should create 1 place for such constants. Thank you for reply

PR #3815, opened 2018-04-13
IGNITE-7024

NSAmelchev:
Implementation of network compression.

- Added new module Compression for compression algorithms implementations.
- Added compression filter to NIO filter chain.
- Added flag to a configuration that enables compression between some nodes. (If flag true at least one of nodes, network traffic between this nodes will be compressed)
- Added compression state flag to sessions interface and to TCP Handshake msg.
- Added JMH benchmark.
- Some changes in next classes to support compression: GridNioServer, TcpCommunicationSPI


modules/core/src/main/java/org/apache/ignite/internal/util/nio/compression/GridNioCompressionHandler.java
```diff
@@ -0,0 +1,246 @@
...
+ */
+
+package org.apache.ignite.internal.util.nio.compression;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.concurrent.locks.ReentrantLock;
+import org.apache.ignite.IgniteLogger;
+import org.apache.ignite.internal.util.nio.GridNioException;
```
SharplEr: Remove unnecessary `import`.

modules/core/src/main/java/org/apache/ignite/internal/util/nio/compression/GridNioCompressionHandler.java
```diff
@@ -0,0 +1,246 @@
...
+import org.apache.ignite.internal.util.nio.GridNioSession;
+import org.jetbrains.annotations.Nullable;
+
+import static org.apache.ignite.internal.util.nio.compression.CompressionEngineResult.BUFFER_OVERFLOW;
+import static org.apache.ignite.internal.util.nio.compression.CompressionEngineResult.OK;
+
+/**
+ * Class that encapsulate the compression state, compress and decompress logic.
+ */
+public final class GridNioCompressionHandler {
```
SharplEr: Decompose into two classes: one thread safe, and one not as a base.

modules/core/src/main/java/org/apache/ignite/internal/util/nio/compression/GridNioCompressionHandler.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/util/nio/compression/GridNioCompressionHandler.java
NSAmelchev: Done.

PR #3828, opened 2018-04-16
IGNITE-8154

antkr:


modules/jms11/src/test/java/org/apache/ignite/stream/jms11/IgniteJmsStreamerTest.java
```diff
@@ -499,6 +505,64 @@ public void testQueueMultipleThreads() throws Exception {
...
+            });
+
+            jmsStreamer.setDestination(dest);
+
+            try {
+                jmsStreamer.start();
+
+                fail("Expecting security exception");
+            } catch (IgniteException ie) {
+                assertTrue("Caught IgniteException, expected: ", ie.getCause().getCause() instanceof SecurityException);
```
slukyano: The message here is printed when the assertion fails, not passes, so it usually reads like "Unexpected exception...".

modules/jms11/src/test/java/org/apache/ignite/stream/jms11/IgniteJmsStreamerTest.java
```diff
@@ -499,6 +505,64 @@ public void testQueueMultipleThreads() throws Exception {
...
+            }
+
+            while(lsnrExceptions.isEmpty())
+                Thread.sleep(200);
+
+            assertTrue(lsnrExceptions.size() > 0);
+
+            try {
+                jmsStreamer.stop();
```
slukyano: Is it OK if the exception is NOT thrown (now the test will pass with either IgniteException or no exception)?

modules/jms11/src/main/java/org/apache/ignite/stream/jms11/JmsStreamer.java
```diff
@@ -150,6 +152,9 @@
     /** Message listeners. */
     private Set<IgniteJmsMessageListener> listeners = Collections.newSetFromMap(new ConcurrentHashMap<IgniteJmsMessageListener, Boolean>());
 
+    /** The Jms {@link ExceptionListener} xception listener to use. */
```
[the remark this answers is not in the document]
antkr: Done.

modules/jms11/src/test/java/org/apache/ignite/stream/jms11/IgniteJmsStreamerTest.java
```diff
@@ -499,6 +505,64 @@ public void testQueueMultipleThreads() throws Exception {
...
     }
 
+    /**
+     * Test for ExceptionListener functionality.
+     *
+     * @throws Exception If fails.
+     */
+    public void testExceptionListener() throws Exception {
+        // restart broker with auth plugin
+        if (broker.isStarted()) {
```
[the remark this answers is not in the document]
antkr: Done, missed this.

modules/jms11/src/test/java/org/apache/ignite/stream/jms11/IgniteJmsStreamerTest.java
antkr: Replaced with assertThrowsWithCause

modules/jms11/src/test/java/org/apache/ignite/stream/jms11/IgniteJmsStreamerTest.java
antkr: Done.

modules/jms11/src/test/java/org/apache/ignite/stream/jms11/IgniteJmsStreamerTest.java
```diff
@@ -537,29 +539,32 @@ public void testExceptionListener() throws Exception {
...
-            } catch (IgniteException ie) {
-                assertTrue("Caught IgniteException, expected: ", ie.getCause().getCause() instanceof SecurityException);
-            }
+                    return null;
+                }
+            }, SecurityException.class);
 
-            while(lsnrExceptions.isEmpty())
-                Thread.sleep(200);
+            latch.await(5, TimeUnit.SECONDS);
```
slukyano: Better to wrap `await()` calls into `assertTrue()` to make sure that the test fails if the `await()` times out.

antkr: Added that.

slukyano:
Looks good to me.

PR #3851, opened 2018-04-17
IGNITE-5994: IgniteInternalCache.invokeAsync().get() can return null

SharplEr:
For https://issues.apache.org/jira/browse/IGNITE-5994


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -2571,10 +2571,10 @@ protected boolean put0(final K key, final V val, final CacheEntryPredicate filte
                 if (resMap != null) {
                     assert resMap.isEmpty() || resMap.size() == 1 : resMap.size();
 
-                    return resMap.isEmpty() ? null : resMap.values().iterator().next();
+                    return resMap.isEmpty() ? new CacheInvokeResult<T>() : resMap.values().iterator().next();
```
NSAmelchev: I think it better to remove generic type.

SharplEr:
merged

PR #3908, opened 2018-04-24
IGNITE-8252 NPE is replaced with an IgniteException with informative message

sergey-chugunov-1985:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/latch/ExchangeLatchManager.java
```diff
@@ -228,14 +231,33 @@ public Latch getOrCreate(String id, AffinityTopologyVersion topVer) {
         }
     }
 
+    /**
+     * Gets alive server nodes from disco cache for provided AffinityTopologyVersion.
+     *
+     * @param topVer Topology version.
```
dspavlov: Return tag is missing in javadoc

PR #3910, opened 2018-04-24
IGNITE-7896 FilePageStore truncate now actually remove redundant

ivandasch:
partition page file.


dspavlov:

PR #3936, opened 2018-04-28
IGNITE-8407  Wrong memory size printing in IgniteCacheDatabaseSnaredManager

snuyanzin:
usages of getInitialSize()/getMaxSize()) should go with U.readableSize(*, false)


dspavlov:
IGNITE-8407 | Closed
-- | --

PR #3941, opened 2018-05-02
IGNITE-8293 BinaryUtils#isCustomJavaSerialization fails when only rea…

Celos:
…dObject is declared in a class


agura:
Don't need anymore.

PR #3945, opened 2018-05-05
IGNITE-8436 Added shutdown for services thread pool

rprabhat:


modules/core/src/test/java/org/apache/ignite/internal/GridStartStopSelfTest.java
```diff
@@ -183,5 +187,29 @@ public void testStoppedState() throws Exception {
...
+        executors.put("indexing", ctx.getIndexingExecutorService());
+        executors.put("management", ctx.getManagementExecutorService());
+        executors.put("peerclass", ctx.getPeerClassLoadingExecutorService());
+        executors.put("query", ctx.getQueryExecutorService());
+        executors.put("rest", ctx.getRestExecutorService());
+        executors.put("schema", ctx.getSchemaExecutorService());
+        executors.put("service", ctx.getServiceExecutorService());
+        executors.put("striped", ctx.getStripedExecutorService());
+        executors.put("system", ctx.getSystemExecutorService());
+        if( ctx.customExecutors() != null ) {
```
Mmuzaf: Refer to our code style guide -- https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines or 
you can install Intellij IDEA Plugin -- https://cwiki.apache.org/confluence/display/IGNITE/Abbreviation+Rules and it will highlight you some reqired changes.

Please, apply minor code style changes:

1) One-liner if statements should be without brakets
2) Source code line should not be longer then 120 characters
3) Line break before `if( ctx.customExecutors() != null )` statement
4) Make assert message more informative: missed keyword "ExecutorService" and dot at the end

Someone can forgot for newly added executor service add new line like `executors.put("service", ctx.getServiceExecutorService());` to this test and also forgot to stop his service. In this case this test will not fail.  
I think we should test this type of cases too.

PR #3973, opened 2018-05-10
IGNITE-6055: SQL: Add String length constraint

nizhikov:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/ddl/DdlStatementsProcessor.java
```diff
@@ -621,6 +622,12 @@ private static QueryEntity toQueryEntity(GridSqlCreateTable createTbl) {
 
             if (col.getType() == Value.DECIMAL)
                 decimalInfo.put(e.getKey(), F.t((int)col.getPrecision(), col.getScale()));
+
+            if (col.getType() == Value.STRING || 
+                col.getType() == Value.STRING_FIXED || 
+                col.getType() == Value.STRING_IGNORECASE) {
+                maxLengthInfo.put(e.getKey(), (int)col.getPrecision());
+            }
```
voipp: may be remove braces from here

PR #3986, opened 2018-05-11
IGNITE-8469: release memory in case initialization multi times

Mmuzaf:


modules/core/src/main/java/org/apache/ignite/internal/mem/unsafe/UnsafeMemoryProvider.java
```diff
@@ -41,6 +41,9 @@
     /** */
     private IgniteLogger log;
 
+    /** */
```
dspavlov: I suggest to add javadoc here describing pupose of this field. We now understand it, but after time new community member may require to find out which init is saved in this flag

modules/core/src/main/java/org/apache/ignite/internal/mem/file/MappedFileMemoryProvider.java
```diff
@@ -65,6 +68,9 @@ public MappedFileMemoryProvider(IgniteLogger log, File allocationPath) {
 
     /** {@inheritDoc} */
     @Override public void initialize(long[] sizes) {
+        if(isInit)
```
dspavlov: space after if seems to be missed: if (isinit)

modules/core/src/main/java/org/apache/ignite/internal/mem/unsafe/UnsafeMemoryProvider.java
```diff
@@ -50,9 +53,14 @@ public UnsafeMemoryProvider(IgniteLogger log) {
 
     /** {@inheritDoc} */
     @Override public void initialize(long[] sizes) {
+        if(isInit)
```
dspavlov: space after if seems to be missed: if (isinit)

modules/core/src/main/java/org/apache/ignite/internal/pagemem/impl/PageMemoryNoStoreImpl.java
```diff
@@ -231,7 +231,8 @@ public PageMemoryNoStoreImpl(
         if (lastIdx != SEG_CNT - 1)
             chunks = Arrays.copyOf(chunks, lastIdx + 1);
 
-        directMemoryProvider.initialize(chunks);
+        if(segments == null)
```
dspavlov: spacing after if

PR #3997, opened 2018-05-15
IGNITE-8449 no need for checkpoint lock in case of stale updates

macrergate:


dspavlov:

PR #4000, opened 2018-05-15
tickTime was tuned in attempt to fix tests instability on TC

sergey-chugunov-1985:


dspavlov:
Is this PR still actual? Could you please link it to a ticket?

PR #4010, opened 2018-05-17
IGNITE-8491

vldpyatkov:
Add JMX flag: Is the node in baseline or not?


zstan:
plz, rename checkBaselineInFromMBena and replace nonusable baselineNodes.

PR #4038, opened 2018-05-21
IGNITE-8526 Create web-agent docker image for k8s deployment

vveider:
* added web-agent separate docker image build
 * refactored and unified docker specifications layout


docker/web-agent/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+WORKDIR /opt/ignite
+
+# Add missing software
+RUN apk --update --no-cache add bash
+
+# Copy main binary archive
+COPY ignite-web-agent* ignite-web-agent
+
+# Update permissions
+RUN chmod -R 777 ${AGENT_HOME}
```
nva: Why you apply such wide permission for all files in agent folder. May be you want `RUN chmod +x ${AGENT_HOME}/ignite-web-agent.sh`

docker/web-agent/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+# Entrypoint
+CMD $AGENT_HOME/ignite-web-agent.sh -d ${DRIVER_FOLDER} -n ${NODE_URI} -s ${SERVER_URI} -t ${TOKENS}
```
nva: Is it really necessary to change $ {DRIVER_FOLDER}?

docker/web-agent/Dockerfile
gromtech: I think it makes sense to follow @nva suggestion if agent doesn't add/copy/move/ files from ${AGENT_HOME}.

docker/web-agent/Dockerfile
gromtech: As we provide a capability to specify this option in run parameters, I think we need to do the same for environment variables. Maybe can we just add default values of env variables? For example:

DRIVER_FOLDER = /opt/ignite/drivers
NODE_URI = http://localhost:8080
SERVER_URI = http://localhost
TOKENS = NO_TOKENS

It should help to avoid some issues when client forgets to init some variables:
http://apache-ignite-users.70518.x6.nabble.com/Agent-not-able-to-connect-tc21714.html#a21735

vveider: Agree

docker/web-agent/Dockerfile
vveider: I will experiment with image built with already `chmod +x ignite-web-agent.sh`

docker/web-agent/README.txt
```diff
@@ -0,0 +1,28 @@
...
+Build image
+===========
+1) Build Apache Ignite Web Console module
+
+        mvn clean install -T 2C \
+                          -Pall-java,all-scala,licenses,web-console \
+                          -pl :ignite-web-console -am \
+                          -DskipTests
+
+3) Copy Apache Ignite Web Agent's binary archive to Apache Ignite Web Console Docker module directory
```
akuznetsov-gridgain: It seems item 2) is missing?

docker/web-agent/README.txt
```diff
@@ -0,0 +1,28 @@
...
+
+        cp -rfv ../../modules/web-console/web-agent/target/ignite-web-agent-*.zip ./
```
akuznetsov-gridgain: Please provide a command that will work Windows too.

docker/web-agent/README.txt
akuznetsov-gridgain: "Copy Apache Ignite Web Agent's binary archive to Apache Ignite Web Console Docker module directory"
It is possible to give exact folder name?

docker/web-agent/README.txt
```diff
@@ -0,0 +1,28 @@
...
+
+4) Unpack and remove Apache Ignite Web Agent's binary archive
+
+        unzip ignite-web-agent-*.zip
```
akuznetsov-gridgain: I think command for remove also should be added.

docker/web-agent/README.txt
vveider: Fixed numeration

docker/web-agent/README.txt
vveider: Disagree.
1. We can't possibly provide exact commands for all kind of shells. Bash is mostly used so this is just a bonus. Human readable instructions should be descriptive enough.

docker/web-agent/README.txt
vveider: Fixed

docker/web-agent/README.txt
vveider: Added

docker/web-agent/README.txt
```diff
@@ -0,0 +1,33 @@
...
+        rm -rf ignite-web-agent-*.zip
+
+4) Build docker image
+
+        docker build . -t apacheignite/web-agent[:<version>]
+
+   Prepared image will be available in local docker registry (can be seen
+   issuing `docker images` command)
+
```
akuznetsov-gridgain: May be add "5) Cleanup after build"  section?
With appropriate "rm" commands?

vveider: Fixed

PR #4059, opened 2018-05-24
IGNITE-8587 GridToStringBuilder use ConcurrentHashMap to avoid global…

macrergate:
… locks on classCache


modules/core/src/main/java/org/apache/ignite/internal/util/tostring/GridToStringBuilder.java
```diff
@@ -1737,14 +1727,7 @@ private static void appendVals(SBLimitedLength buf,
...
 
-        rwLock.readLock().lock();
-
-        try {
-            cd = classCache.get(key);
-        }
-        finally {
-            rwLock.readLock().unlock();
-        }
+        cd = classCache.get(key);
```
ivandasch: I would suggest here putIfAbsent

PR #4074, opened 2018-05-25
IGNITE-8293 BinaryUtils#isCustomJavaSerialization fails when only readObject is declared in a class

agura:


agura:
Don't need anymore.

PR #4101, opened 2018-05-31
IGNITE-8666 Add ability of filtering data during datasets creation

dmitrievanthony:
…asetBuilder.


modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/cache/CacheBasedDatasetBuilder.java
```diff
@@ -52,15 +53,31 @@
...
+    }
+
+    /**
+     * Constructs a new instance of cache based dataset builder that makes {@link CacheBasedDataset}.
+     *
+     * @param ignite Ignite instance.
+     * @param upstreamCache Ignite Cache with {@code upstream} data.
+     * @param pred Predicate that filters {@code upstream} data.
+     */
+    public CacheBasedDatasetBuilder(Ignite ignite, IgniteCache<K, V> upstreamCache, IgniteBiPredicate<K, V> pred) {
```
zaleslaw: pred is strange unreadable name

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/cache/util/ComputeUtils.java
```diff
@@ -253,4 +273,38 @@
         IgniteCache<Integer, C> datasetCache = ignite.cache(datasetCacheName);
         datasetCache.put(part, ctx);
     }
+
+    /**
+     * Computes number of entries selected from the cache ny the query.
```
zaleslaw: ny? fix please

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/cache/CacheBasedDatasetBuilder.java
dmitrievanthony: Renamed `pred` -> `filter`.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/cache/util/ComputeUtils.java
dmitrievanthony: Fixed.

PR #4171, opened 2018-06-09
IGNITE-8751 Failure handler accordingly to segmentation policy should be invoked on node segmentation instead of configured failure handler

agura:
… be invoked on node segmentation instead of configured failure handler


agura:
Merged.

PR #4186, opened 2018-06-14
IGNITE-8737: Improve checkpoint logging information

andrewmed:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -869,7 +869,7 @@ private boolean hasIndex(long absIdx) {
     }
 
     /** {@inheritDoc} */
-    @Override public int truncate(WALPointer low, WALPointer high) {
+    @Override public int truncate(WALPointer low, WALPointer high, Collection<String> names) {
```
sergey-chugunov-1985: @andrewmed 

All parameters in the same method in `FsyncModeFileWriteAheadLogManager` class were declared final.
Please do the same here.

Also please fix javadoc near `lastTruncatedArchiveIdx` field in these two classes: after changing signature it became incorrect.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -900,8 +900,11 @@ private boolean hasIndex(long absIdx) {
                 if (!desc.file.delete())
                     U.warn(log, "Failed to remove obsolete WAL segment (make sure the process has enough rights): " +
                         desc.file.getAbsolutePath());
-                else
+                else {
+                    if (names != null)
+                        names.add(desc.file.getName());
```
sergey-chugunov-1985: Code style: blank line before `deleted++` is missed (the same in another place, won't leave comment there).

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/checkpoint/IgniteMassLoadSandboxTest.java
```diff
@@ -297,6 +301,64 @@ public void testDataStreamerContinuousPutMultithreaded() throws Exception {
...
+
+            cache.put(1, new byte[1024]);
+
+            forceCheckpoint();
+
+            Thread.sleep(200); // needed by GridStringLogger
+
+            final String log = log0.toString();
+
+            System.out.println(log);
```
sergey-chugunov-1985: Looks like a leftover from test development, please remove.

sergey-chugunov-1985:
BTW, I applied your patch to master branch and found out that there is no changes in `CheckpointHistory` class which uses `truncate` method.
Because of this I see compilation error in my code although TC passes just fine.
Could you check please what's going on with this class?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/checkpoint/IgniteMassLoadSandboxTest.java
andrewmed: This if for keeping visual control if smth goes wrong. Do we need to delete it?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
andrewmed: fixed

PR #4196, opened 2018-06-15
IGNITE-8755: If size is too big and left shift can cause arithmetic o…

ivandasch:
…veflow, just allocate bytes for required size. So NegativeArrayException will never throw


modules/core/src/main/java/org/apache/ignite/internal/util/io/GridUnsafeDataOutput.java
```diff
@@ -120,15 +120,18 @@ public void bytes(byte[] bytes, int off) {
     /**
      * @param size Size.
      */
-    private void requestFreeSize(int size) {
+    private void requestFreeSize(int size) throws IOException {
+        if (off + size < size)
```
EdShangGG: Please, add the comment that you check overflow in such way.

dspavlov:
The ticket is resolved, closing PR. Feel free to reopen if needed.

PR #4235, opened 2018-06-21
IGNITE-7993 Striped pool can't be disabled

gromtech:
Restored the capability to disable striped pool

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheIoManager.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheIoManager.java
@@ -277,17 +277,17 @@ public class GridCacheIoManager extends GridCacheSharedManagerAdapter {
                             }
                         };
 
-                        GridKernalContext kernalContext = cctx.kernalContext();
+                        GridKernalContext ctx = cctx.kernalContext();
 
                         if (stripe >= 0) {
-                            if (!kernalContext.isStripedExecutorDisabled())
-                                kernalContext.getStripedExecutorService().execute(stripe, c);
+                            if (!ctx.isStripedExecutorDisabled())
+                                ctx.getStripedExecutorService().execute(stripe, c);
                             else
-                                kernalContext.getSystemExecutorService().execute(c);
+                                ctx.getSystemExecutorService().execute(c);
                         }
                         else {
                             try {
-                                kernalContext.pools().poolForPolicy(plc).execute(c);
+                                ctx.pools().poolForPolicy(plc).execute(c);
                             }
                             catch (IgniteCheckedException e) {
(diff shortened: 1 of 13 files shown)


modules/core/src/test/java/org/apache/ignite/internal/util/DisabledStripedExecutorTest.java
```diff
@@ -0,0 +1,113 @@
...
+        IgniteCache<Integer, Integer> cache = ignite.getOrCreateCache("temp-cache");
+
+        Integer key = 1;
+        Integer val = 1;
+
+        cache.put(key, val);
+
+        assertEquals("Cache should contain saved object", val, cache.get(key));
+    }
+}
```
yzhdanov: Need to add balnk line - file should finish with blank one

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheIoManager.java
```diff
@@ -276,11 +277,17 @@ else if (desc.receivedFromStartVersion() != null)
                             }
                         };
 
-                        if (stripe >= 0)
-                            cctx.kernalContext().getStripedExecutorService().execute(stripe, c);
+                        GridKernalContext kernalContext = cctx.kernalContext();
```
yzhdanov: we usually use ctx for kernal context and cctx for cache context. Please rename kernalContext to ctx

modules/core/src/test/java/org/apache/ignite/internal/util/DisabledStripedExecutorTest.java
```diff
@@ -0,0 +1,113 @@
+/*
```
yzhdanov: You need to add more tests including full api and fault-tolerance tests with disabled striped pool.

modules/core/src/test/java/org/apache/ignite/internal/util/DisabledStripedExecutorTest.java
gromtech: Fixed

modules/core/src/test/java/org/apache/ignite/internal/util/DisabledStripedExecutorTest.java
gromtech: Added DisabledStripedExecutorFailoverTest and DisabledStripedExecutorFullApiTest. Please let me know if we need more tests.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheIoManager.java
gromtech: Fixed

gromtech:
There is going to be implemented a strict thread per partition in the nearest future which will make disabling of striped pool impossible. I am closing this PR. New one PR will be created to update Javadoc.

PR #4249, opened 2018-06-22
IGNITE-8426 LongJVMPauseDetector add logger from context

ezhuravl:


mcherkasov:
Fix looks good for me.

modules/core/src/main/java/org/apache/ignite/internal/LongJVMPauseDetector.java
```diff
@@ -41,36 +40,43 @@
...
-    private static final IgniteLogger LOG = new JavaLogger();
+    private final IgniteLogger LOG;
 
     /** Worker reference. */
-    private static final AtomicReference<Thread> workerRef = new AtomicReference<>();
+    private final AtomicReference<Thread> workerRef = new AtomicReference<>();
 
     /** Precision. */
-    private static final int PRECISION = getInteger(IGNITE_JVM_PAUSE_DETECTOR_PRECISION, 50);
+    private final int PRECISION = getInteger(IGNITE_JVM_PAUSE_DETECTOR_PRECISION, 50);
```
AMashenkov: There is no need to move constants into object.

modules/core/src/main/java/org/apache/ignite/internal/LongJVMPauseDetector.java
```diff
@@ -41,36 +40,43 @@
...
+     */
+    public LongJVMPauseDetector(GridLoggerProxy log) {
+        LOG = log;
+    }
 
     /**
      * Starts worker if not started yet.
      */
-    public static void start() {
+    public void start() {
```
AMashenkov: synchronized (LongJVMPauseDetector.class) looks like a bug

modules/core/src/main/java/org/apache/ignite/internal/LongJVMPauseDetector.java
```diff
@@ -41,36 +40,43 @@
...
+    private final long[] longPausesTimestamps = new long[EVT_CNT];
 
     /** Long pauses durations. */
-    private static final long[] longPausesDurations = new long[EVT_CNT];
+    private final long[] longPausesDurations = new long[EVT_CNT];
+
+    /**
+     * @param log Logger.
+     */
+    public LongJVMPauseDetector(GridLoggerProxy log) {
```
AMashenkov: GridLoggerProxy can be safely replaced with Ignite Logger.

PR #4278, opened 2018-06-28
IGNITE-7782 Thin Client lib: Python

dmelnichuk:
Please review this preliminary version of Ignite binary API client library, written in Python 3.

What is done:
- all API operations implemented,
- code is mostly tested (pytest, ~90% covered),
- setuptools integration, 100% PyPI-ready,
- documentation provided (Sphinx/autodoc),
- Complex object support.

Tested with Python 3.4 and 3.6, Ignite 2.7.


modules/platforms/python/pyignite/cache.py
```diff
@@ -0,0 +1,592 @@
...
+            self._client, self._cache_id, key, value,
+            key_hint=key_hint, value_hint=value_hint
+        )
+        result.value = self._process_binary(result.value)
+        return result
+
+    @status_to_exception(CacheError)
+    def clear(self, keys: Optional[list]=None):
+        """
+        Clears the cache without notifying listeners or cache writers.
```
qvad: Removing values from cache without notifying listeners or cache writers more looks like an security issue than correct behaviour.

qvad: Looks like this is JCache specified behaviour

dmelnichuk: Could you please elaborate on what might the security threat be? This operation is not directly related to authentication or encryption.

modules/platforms/python/pyignite/api/key_value.py
```diff
@@ -0,0 +1,1105 @@
...
+    :param binary: (optional) pass True to keep the value in binary form.
+     False by default,
+    :param query_id: (optional) a value generated by client and returned as-is
+     in response.query_id. When the parameter is omitted, a random value
+     is generated,
+    :return: API result data object. Contains zero status and a value
+     retrieved on success, non-zero status and an error description on failure.
+    """
+
+    class CacheGetQuery(Query):
```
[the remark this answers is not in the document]
dmelnichuk: I must agree with you on performance, although I believe the impact should be negligible.

Another question is: where to define these classes? Module level is no go: they will pollute the namespace, worsen readability, clutter docs, and make debugging more difficult…

Now I think about parameterizing `Query` with opcode rather than subclassing it.

modules/platforms/python/pyignite/api/key_value.py
```diff
@@ -0,0 +1,1105 @@
...
+    ], query_id=query_id)
+
+    _, send_buffer = query_struct.from_python({
+        'hash_code': cache_id(cache),
+        'flag': 1 if binary else 0,
+        'key': key,
+    })
+    connection.send(send_buffer)
+
+    response_struct = Response([
```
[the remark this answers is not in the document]
dmelnichuk: Correctly noticed, thanks.

modules/platforms/python/pyignite/cache.py
```diff
@@ -0,0 +1,592 @@
...
+        """
+        Clears the cache key without notifying listeners or cache writers.
+
+        :param key:  key for the cache entry,
+        :param key_hint: (optional) Ignite data type, for which the given key
+         should be converted,
+        """
+        return cache_remove_key(self._client, self._cache_id, key, key_hint)
+
+    @status_to_exception(CacheError)
```
qvad: remove methods may mislead user if it' familiar with Cache.java:

Here is current mapping: java method -> pyignite method 
remove(Key) -> remove_key()
remove(Key, Value) -> remove_if_equals()
removeAll([Keys]) -> remove()
removeAll() -> remove()

I think following mapping is more clear to end user:
remove(Key) -> remove()
remove(Key, Value) -> remove_if_equals()
removeAll([Keys]) -> remove_all()
removeAll() -> remove_all()

modules/platforms/python/pyignite/api/key_value.py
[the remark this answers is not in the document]
qvad: I've write simple test to check performance impact if inner class used:

```
import time

with_class_counter = 0
no_class_counter = 0


class OtherClass:
    a = 1


def with_class():
    global with_class_counter

    class SomeClass:
        a = 1

    a = SomeClass()
    with_class_counter += a.a


def no_class():
    global no_class_counter

    a = OtherClass()
    no_class_counter += a.a


before = int(round(time.time() * 1000))

for i in range(1, 1000000):
    no_class()

after = int(round(time.time() * 1000))

print("Time for method w/o inner class: %s" % str(after-before))


before = int(round(time.time() * 1000))

for i in range(1, 1000000):
    with_class()

after = int(round(time.time() * 1000))

print("Time for method with inner class: %s" % str(after-before))
```

And output:
Time for method w/o inner class: 259
Time for method with inner class: 9220

I think that parametrising Query is good option.

modules/platforms/python/pyignite/cache.py
dmelnichuk: Don't you think it is rather misleading to name a method `remove_all()`, and make it remove *not* all, but only selected keys?

I understand that such naming can have historical reasons. But, frankly, I do not expect that Java programmers familiar with Cache.java would use *Python* Ignite thin client. I think its primary audience are Python developers who are unaware of Cache.java.

qvad: For me as a Java developer it is ok to see removeAll(Collection) in API :)

The question is in a dialog between java developers and python developers. I think it is better when they use common language. In this case method names and it's behaviour came from Java.
Maybe we can add `remove_keys()` and `remove_key()` like it is made for `contains_*()` methods

Current realisation for remove_* methods looks a bit confusing for me

dmelnichuk: A bit too verbose, but leaves no uncertainties:
- `remove(Key)` → `remove_key()`
- `remove(Key, Value)` → `remove_if_equals()`
- `removeAll([Keys])` → `remove_keys()`
- `removeAll()` → `remove_all()`

Can we proceed with this?

qvad: Yes, looks good

modules/platforms/python/pyignite/connection/__init__.py
```diff
@@ -0,0 +1,329 @@
...
+        else:
+            raise ConnectionError('Connection parameters are not valid.')
+
+        self._connect(host, port)
+
+    def _reconnect(self):
+        """
+        Restore the connection using the next node in `nodes` iterable.
+        """
+        for host, port in self.nodes:
```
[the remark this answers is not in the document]
dmelnichuk: Yes, this is the expected behavior.

PR #4281, opened 2018-06-29
IGNITE-8859 - Open upper Java verison bounds

dkarachentsev:


bin/include/functions.sh
```diff
@@ -27,6 +27,28 @@
...
+
+# Extract only major version of java to `version` variable.
+javaMajorVersion() {
+    javaVersion "$1"
+    version="${version%%.*}"
+
+    if [ ${version} -eq 1 ]; then
+        # Version seems starts from 1, we need second number.
+        javaVersion "$1"
+        backIFS=$IFS
```
vveider: AFAIK, there is no need in saving file separator if IFS is redefined in the same string with the command itself without semicolon

bin/include/functions.sh
```diff
@@ -54,22 +76,20 @@ checkJava() {
...
         echo "$0, ERROR:"
-        echo "The version of JAVA installed in JAVA_HOME=$JAVA_HOME is incorrect."
+        echo "The version of JAVA installed in JAVA_HOME=$JAVA_HOME is incompatible."
         echo "Please point JAVA_HOME variable to installation of JDK 1.8 or JDK 9."
         echo "You can also download latest JDK at http://java.com/download"
         exit 1
-    }
+    elif [ $version -gt 9 ]; then
+        echo "$0, WARNING:"
+        echo "The version of JAVA installed in JAVA_HOME=$JAVA_HOME was not tested with Apache Ignite."
```
vveider: I would add detected JAVA version (the fuller the better) to output to help user to understand on which JAVA currently Apache Ignite is trying to start

bin/include/functions.sh
dkarachentsev: If you're talking about backIFS, yes I need to restore it, because if not to do this, it breaks our scripts.

PR #4288, opened 2018-07-02
IGNITE-8905 incorrect assertion was replaced with if-check

sergey-chugunov-1985:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -3226,11 +3226,11 @@ private void processFullMessage(boolean checkCrd, ClusterNode node, GridDhtParti
                                 if (!F.isEmpty(msg.getErrorsMap())) {
                                     Exception e = msg.getErrorsMap().get(cctx.localNodeId());
 
-                                    assert e != null : msg.getErrorsMap();
+                                    if (e != null && e instanceof IgniteNeedReconnectException) {
```
Mmuzaf: instance of operator always return `false` for `e = null`, so here we don't need check for `e != null`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -1564,9 +1564,17 @@ else if (!grp.isLocal())
...
 
                 if (msg.client() && !exchFut.isDone()) {
-                    if (exchFut.initialVersion().compareTo(readyAffinityVersion()) <= 0) {
+                    AffinityTopologyVersion initVer = exchFut.initialVersion();
+                    AffinityTopologyVersion readyVer = readyAffinityVersion();
+
+                    DiscoveryEvent firstEvt = exchFut.firstEvent();
+
+                    if (initVer.compareTo(readyVer) <= 0
+                        && (firstEvt != null && firstEvt.type() != EVT_NODE_LEFT && firstEvt.eventNode().isClient())) {
```
Mmuzaf: Suppose, we missed here EVT_NODE_FAIL event type.

From my point of view we should add new mehtod like this `ExchangeDiscoveryEvents#serverLeftEvent` but for client nodes and use it here. 
Also, I think `firstEvt` always not `null`, so we don't need to check it here but we must sure about it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -1564,9 +1564,17 @@ else if (!grp.isLocal())
...
                         U.warn(log, "Client node tries to connect but its exchange " +
-                            "info is cleaned up from exchange history." +
+                            "info is cleaned up from exchange history " +
```
Mmuzaf: We have a specific string output format described here -- https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines

From my point, we should also use it here. Square, round brackets not at the right places here and should be placed at the end of warn message.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
sergey-chugunov-1985: @Mmuzaf,

Fixed this.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
sergey-chugunov-1985: fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
sergey-chugunov-1985: Good catch with EVT_NODE_FAIL event, I've addressed it.

Null check for ```firstEvt``` is correct here because in case of ExchangeFuture being preempted from exchange queue it is recreated with default value of this field.

I'm not sure I understand your comment about ```serverLeftEvent``` for client nodes. 
This code replies to **client exchange** that was preempted from the queue and thus cannot be finished so client should reconnect.
Additional checks were added to prevent situations when client nodes may be forced to reconnect when server nodes leave topology.

PR #4296, opened 2018-07-03
IGNITE-8564 Clear affinity mapping cache when client node is disconnected.

alamar:


modules/core/src/test/java/org/apache/ignite/internal/processors/datastreamer/DataStreamerClientReconnectAfterClusterRestartTest.java
```diff
@@ -0,0 +1,129 @@
...
+    private void clusterRestart(boolean withAnotherClient, boolean allowOverwrite) throws Exception {
+        try {
+            startGrid(0);
+
+            clientMode = true;
+
+            Ignite client = startGrid(1);
+
+            if (withAnotherClient) {
+                startGrid(2);
```
sk0x50: I think it would be nice to add a comment.
something like as follows: "restarting the server node in order to emulate client reconnect to a new cluster"

modules/core/src/test/java/org/apache/ignite/internal/processors/datastreamer/DataStreamerClientReconnectAfterClusterRestartTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+            try (IgniteDataStreamer<String, String> streamer = client.dataStreamer("test")) {
+                streamer.allowOverwrite(allowOverwrite);
+
+                streamer.addData("k1", "v1");
+            }
+
+            stopGrid(0);
+
+            U.sleep(2_000);
```
sk0x50: I suggest getting rid of Thread.sleep(). It does not look reliable.
Perhaps, Ignite events will do the required trick:

`
            final CountDownLatch disconnect = new CountDownLatch(1);
            final CountDownLatch reconnect = new CountDownLatch(1);

            Ignite client = startGrid(1);
            client.events().localListen(new IgnitePredicate<Event>() {
                @Override public boolean apply(Event event) {
                    reconnect.countDown();

                    return false;
                }
            }, EventType.EVT_CLIENT_NODE_RECONNECTED);
            client.events().localListen(new IgnitePredicate<Event>() {
                @Override public boolean apply(Event event) {
                    disconnect.countDown();

                    return false;
                }
            }, EventType.EVT_CLIENT_NODE_DISCONNECTED);

    ...

            stopGrid(0);

            disconnect.await();

            startGrid(0);

            reconnect.await();

    ...
`

WDYT?

modules/core/src/test/java/org/apache/ignite/internal/processors/datastreamer/DataStreamerClientReconnectAfterClusterRestartTest.java
```diff
@@ -0,0 +1,129 @@
...
+                    assert ce.getCause() instanceof IgniteClientDisconnectedException;
+
+                    ((IgniteClientDisconnectedException)ce.getCause()).reconnectFuture().get();
+                }
+                catch (IgniteClientDisconnectedException icde) {
+                    icde.reconnectFuture().get();
+                }
+            }
+
+            assert false : "Failed to wayt for reconnect!";
```
sk0x50: typo :)

modules/core/src/test/java/org/apache/ignite/internal/processors/datastreamer/DataStreamerClientReconnectAfterClusterRestartTest.java
alamar: I have implemented that, but test now runs slower :( 8s vs 5s for individual tests

PR #4314, opened 2018-07-05
IGNITE-8619 Remote node could not start in ssh connection

1vanan:


modules/ssh/src/main/java/org/apache/ignite/internal/util/nodestart/StartNodeCallableImpl.java
```diff
@@ -266,7 +267,7 @@ public StartNodeCallableImpl(IgniteRemoteStartSpecification spec, int timeout) {
 
                 info("Starting remote node with SSH command: " + startNodeCmd, spec.logger(), log);
 
-                shell(ses, startNodeCmd);
+                shell(ses, startNodeCmd, "\\[(\\d)\\] (\\d)+");
```
NSAmelchev: Maybe to make reg as a static field?

1vanan: Why?
We use this regrex only once.

PR #4318, opened 2018-07-05
IGNITE-8935 toString() or exclusion for most classes accessible from IgniteConfiguration

alamar:


modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioMetricsListener.java
```diff
@@ -17,10 +17,12 @@
 
 package org.apache.ignite.internal.util.nio;
 
+import java.util.EventListener;
+
 /**
  * Metrics listener for NIO communication.
  */
-public interface GridNioMetricsListener {
+public interface GridNioMetricsListener extends EventListener {
```
slukyano: Why is it in this patch?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -2090,18 +2090,10 @@ private void dumpInfo(StringBuilder sb, UUID dstNodeId) {
         }
 
         if (connectionsPerNode > 1) {
-            connPlc = new ConnectionPolicy() {
-                @Override public int connectionIndex() {
-                    return (int)(U.safeAbs(Thread.currentThread().getId()) % connectionsPerNode);
-                }
-            };
+            connPlc = new RoundRobinConnectionPolicy();
```
slukyano: Why is it in this patch?

modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioMetricsListener.java
alamar: Because we have exclusion from toString() for EventListener's

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
alamar: Because they will look better in toString() with proper names. Instead of $1 and $2

PR #4330, opened 2018-07-08
IGNITE-8828

ilantukh:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -1225,6 +1232,82 @@ private void distributedExchange() throws IgniteCheckedException {
...
+                            "sending check request to coordinator");
+
+                        cctx.kernalContext().io().sendToGridTopic(
+                            crd, GridTopic.TOPIC_EXCHANGE, new PartitionsExchangeFinishedCheckRequest(), SYSTEM_POOL);
+
+                        cctx.time().schedule(() -> {
+                            if (!isDone())
+                                cctx.kernalContext().failure().process(
+                                    new FailureContext(FailureType.CRITICAL_ERROR,
+                                        new TimeoutException("Last exchange check response wasn't received.")));
```
mcherkasov: I asked 4 people to read this message, no one did understand what this means. I don't think that regular user would understand this message in logs. Could you please add more details here too?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -1225,6 +1232,82 @@ private void distributedExchange() throws IgniteCheckedException {
...
+    }
+
+    /**
+     * @param lastFinishedVer Last finished topology version.
+     * @param pendingVer Pending topology version.
+     * @param receivedSingleMsg Received single message flag.
+     */
+    public void onCrdLastFinishedVersionReceived(AffinityTopologyVersion lastFinishedVer,
+        AffinityTopologyVersion pendingVer, boolean receivedSingleMsg) {
+        log.warning("Received check response from coordinator: " +
```
mcherkasov: The warning means that something bad happens, but from a user point of view, I can't understand what this message means and what actions are expected of me. I think we should move this to INFO level.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -1225,6 +1232,82 @@ private void distributedExchange() throws IgniteCheckedException {
...
+            failLocal = true;
+        else if (pendingVer.equals(exchId.topologyVersion()) && !receivedSingleMsg)
+            failLocal = true;
+        else if (pendingVer.equals(exchId.topologyVersion()) && receivedSingleMsg)
+            return; // Do nothing, this case should be resolved by coordinator.
+        else
+            failLocal = false;
+
+        if (failLocal)
+            cctx.kernalContext().failure().process(new FailureContext(FailureType.CRITICAL_ERROR, new IgniteException(
```
mcherkasov: "Coordinator topology version is greater than local.",  I think it should sound like: "Exchange hard timeout is reached, but coordinator topology version is greater than local."

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -1225,6 +1232,82 @@ private void distributedExchange() throws IgniteCheckedException {
...
+    private void onDistributedExchangeTimeout() {
+        if (!isDone()) {
+            synchronized (mux) {
+                if (crd.isLocal()) {
+                    Set<UUID> remaining0 = new HashSet<>(this.remaining);
+
+                    log.warning("Exchange timeout reached on coordinator node, kicking pending nodes: " +
+                        S.arrayToString(UUID.class, remaining0.toArray()));
+
+                    for (UUID remainingUuid : remaining0)
```
mcherkasov: ticket in jira says: " it should check that stopping these nodes won't lead to data loss and then stop them forcibly."
I don't see how you check this here.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -1215,6 +1220,8 @@ private void distributedExchange() throws IgniteCheckedException {
 
         changeWalModeIfNeeded();
 
+        cctx.time().schedule(this::onDistributedExchangeTimeout, cctx.gridConfig().getExchangeHardTimeout(), -1);
```
sk0x50: Is this supposed behavior that the returned CancelledTask is not re-scheduled in case of the coordinator node left the cluster? Let's assume that the coordinator node left the cluster for some reason and therefore, a new node will be elected. New coordinator has to request single messages from all participated nodes and so, there is a chance onDistributedExchangeTimeout() method will be called. I mean that time required for exchange on new coordinator is less than getExchangeHardTimeout(). Perhaps, it's not an issue at all.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -1607,6 +1618,44 @@ private void processSinglePartitionRequest(ClusterNode node, GridDhtPartitionsSi
         }
     }
 
+    /**
+     * @param nodeId Node id.
+     * @param req Request.
+     */
+    private void processExchangeCheckRequest(UUID nodeId, PartitionsExchangeFinishedCheckRequest req) {
```
dmekhanikov: `req` parameter is not used

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -1607,6 +1618,44 @@ private void processSinglePartitionRequest(ClusterNode node, GridDhtPartitionsSi
...
+        catch (IgniteCheckedException ex) {
+            throw new IgniteException(ex);
+        }
+    }
+
+    /**
+     * @param nodeId Node id.
+     * @param res Response.
+     */
+    private void processExchangeCheckResponse(UUID nodeId, PartitionsExchangeFinishedCheckResponse res) {
```
dmekhanikov: `nodeId` parameter is not used

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -1225,6 +1232,82 @@ private void distributedExchange() throws IgniteCheckedException {
...
+     * @param pendingVer Pending topology version.
+     * @param receivedSingleMsg Received single message flag.
+     */
+    public void onCrdLastFinishedVersionReceived(AffinityTopologyVersion lastFinishedVer,
+        AffinityTopologyVersion pendingVer, boolean receivedSingleMsg) {
+        log.warning("Received check response from coordinator: " +
+            "[lastFinishedVer=" + lastFinishedVer + ", pendingVer=" + pendingVer +
+            ", receivedSingleMsg=" + receivedSingleMsg + ", localExchId=" + exchId + "]");
+
+        boolean failLocal;
```
dmekhanikov: `failLoc`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
ilantukh: It is intended behavior. ExchangeHardTimeout is maximum amount of time that user can afford to spend on exchange. If this timeout is reached, we should do everything to finish exchange ASAP.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -3001,6 +3008,28 @@ public IgniteConfiguration setAuthenticationEnabled(boolean authEnabled) {
...
+        return exchangeHardTimeout;
+    }
+
+    /**
+     * Sets hard timeout for partition map exchange process.
+     *
+     * @param timeout Hard exchange timeout, in millis.
+     * @return {@code this} for chaining.
+     */
+    public IgniteConfiguration setExchangeHardTimeout(long timeout) {
```
dmekhanikov: What should I do, if I don't want this functionality? Should I set this timeout to Long.MAX_VALUE? It will make us create a lot of schedules, that will never be executed. I think, -1 should be a special value, that disables this feature.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -1607,6 +1618,44 @@ private void processSinglePartitionRequest(ClusterNode node, GridDhtPartitionsSi
         }
     }
 
+    /**
+     * @param nodeId Node id.
+     * @param req Request.
+     */
+    private void processExchangeCheckRequest(UUID nodeId, PartitionsExchangeFinishedCheckRequest req) {
+        log.warning("Received check request from node: " + nodeId);
```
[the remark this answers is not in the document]
ilantukh: @mcherkasov what message would you consider meaningful in this and other cases?

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/GridIoMessageFactory.java
```diff
@@ -927,6 +929,16 @@ public GridIoMessageFactory(MessageFactory[] ext) {
...
+                msg = new PartitionsExchangeFinishedCheckRequest();
+
+                break;
+
+            case 137:
+                msg = new PartitionsExchangeFinishedCheckResponse();
+
+                break;
+
             // [-3..119] [124..129] [-23..-27] [-36..-55]- this
```
dmekhanikov: 124..129 should be replaced with 124..137

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
sk0x50: Thank you, Ilya. Got it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -1225,6 +1232,82 @@ private void distributedExchange() throws IgniteCheckedException {
...
+
+        if (lastFinishedVer.compareTo(exchId.topologyVersion()) >= 0)
+            failLocal = true;
+        else if (pendingVer.equals(exchId.topologyVersion()) && !receivedSingleMsg)
+            failLocal = true;
+        else if (pendingVer.equals(exchId.topologyVersion()) && receivedSingleMsg)
+            return; // Do nothing, this case should be resolved by coordinator.
+        else
+            failLocal = false;
```
dmekhanikov: Consider a situation, when a non-coordinator node didn't receive one of the messages, that change topology version, due to a bug in discovery. Let's say, coordinator has pending exchange with `topVer=5` and the remaining node – `topVer=6`. In this case the remaining node will decide, that coordinator should die. After coordinator re-election this situation may repeat until the problematic node is left the only one in the cluster.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
dmekhanikov: If a coordinator didn't send a full message and stopped processing check requests, then all other nodes will die. I doubt, that this is a good outcome.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
ilantukh: OK, good point.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
ilantukh: This is intentional. I wanted all process***Message methods to have similar signature.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -1225,6 +1232,82 @@ private void distributedExchange() throws IgniteCheckedException {
...
+                else {
+                    try {
+                        log.warning("Exchange timeout reached on non-coordinator node, " +
+                            "sending check request to coordinator");
+
+                        cctx.kernalContext().io().sendToGridTopic(
+                            crd, GridTopic.TOPIC_EXCHANGE, new PartitionsExchangeFinishedCheckRequest(), SYSTEM_POOL);
+
+                        cctx.time().schedule(() -> {
+                            if (!isDone())
```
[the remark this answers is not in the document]
ilantukh: This rule applies to if block with multiple statements. In this case, it has a single statement split in multiple lines.

PR #4334, opened 2018-07-09
IGNITE-8957: fix assert when histMap is empty

andrewmed:


andrewmed:
Fixed. Thanks!

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -3202,19 +3202,23 @@ private void doCheckpoint() {
 
                 if (chp.hasDelta() || destroyedPartitionsCnt > 0) {
                     if (printCheckpointStats) {
-                        if (log.isInfoEnabled())
+                        if (log.isInfoEnabled()) {
+
+
                             log.info(String.format("Checkpoint finished [cpId=%s, pages=%d, markPos=%s, " +
-                                    "walSegmentsCleared=%d, walSegmentsCovered=%s, markDuration=%dms, pagesWrite=%dms, fsync=%dms, " +
+                                    "walSegmentsCleared=%d, walSegmentsCovered=[%s - %s], markDuration=%dms, pagesWrite=%dms, fsync=%dms, " +
```
Mmuzaf: Will we print `[0 - 0]` here if no wal segmetns covered?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -3920,10 +3924,10 @@ public void walFilesDeleted(int walFilesDeleted) {
         }
 
         /**
-         * @param walSegmentsCovered WAL segments fully covered by this checkpoint.
+         * @param walSegsCoveredRange WAL segments fully covered by this checkpoint.
          */
-        public void walSegmentsCovered(final List<Long> walSegmentsCovered) {
-            this.walSegmentsCovered = walSegmentsCovered;
+        public void walSegsCoveredRange(final IgniteBiTuple<Long, Long> walSegsCoveredRange) {
```
Mmuzaf: I think it can be sipmlified to `(final long begin, final long end)`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointHistory.java
```diff
@@ -245,12 +222,46 @@ public boolean hasSpace() {
...
-        return removed;
+        return rmv;
+    }
+
+    /**
+     * Calculates indexes of WAL segments covered by last checkpoint.
+     *
+     * @return list of indexes or empty list if there are no checkpoints.
+     */
+    private IgniteBiTuple<Long, Long> calculateWalSegmentsCovered() {
```
Mmuzaf: From my point, if we just setting out only first and last values of covered segments this method can by simplified to 4 - 5 lines.
So, whole class can be much shorter.

sergey-chugunov-1985: Most of code in the method covers corner cases: there is no checkpoint in history, no WAL segments were covered by the latest checkpoint.
So I doubt the code could be simplified significantly.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
sergey-chugunov-1985: I improved implementation so it will handle corner cases as follows:

- if there is no segments covered, empty square brackets will be printed out: []

- if only one segment covered, it won't be duplicated: [6]

- range will be printed if more than one segment if covered: [0 - 1]

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
sergey-chugunov-1985: It is questionable: from point of view of method signature it is simpler but more code should be added to Checkpoint entity: getters and setters for two fields instead of one.

Also BiTuple represents here a concept of "range" while two separate parameters don't capture it as good as single one.

PR #4338, opened 2018-07-10
IGNITE-8968 Failed to shutdown node due to "Error saving backup value"

pvinokurov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearGetFuture.java
```diff
@@ -1016,17 +1013,28 @@ void onResult(final GridNearGetResponse res) {
...
+         *
+         * @param res Response.
+         */
+        private void postProcessResultAndDone(final GridNearGetResponse res){
+            try {
                 postProcessResult(res);
 
+                // It is critical to call onDone after adding futures to compound list.
                 onDone(loadEntries(node.id(), keys.keySet(), res.entries(), savedEntries, topVer));
+            }catch (Exception ex){
```
alamar: Missing whitespace

PR #4342, opened 2018-07-10
IGNITE-8971 GridRestProcessor should propagate error message

andrewmed:


modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -17,27 +17,7 @@
...
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.ObjectInput;
-import java.io.ObjectOutput;
-import java.io.OutputStream;
-import java.io.PrintStream;
-import java.io.Reader;
-import java.io.StringWriter;
-import java.io.Writer;
+import java.io.*;
```
akuznetsov-gridgain: We do not use wildcard imports by Ignite code style.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
```diff
@@ -349,7 +351,7 @@
                             .a(", params=").a(tskReq.params());
                     }
 
-                    sb.a(", err=").a(e.getMessage() != null ? e.getMessage() : e.getClass().getName()).a(']');
+                    sb.a(", err=").a(getErrorMessage(e)).a(']');
```
akuznetsov-gridgain: I think this can break Web Console, because it is also relay on this output.

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -551,6 +531,14 @@
     private static boolean devOnlyLogDisabled =
         IgniteSystemProperties.getBoolean(IgniteSystemProperties.IGNITE_DEV_ONLY_LOGGING_DISABLED);
 
+    /** List of OS specific system messages for disk full error. */
+    private static final String[] ERR_DISK_FULL = {
+            "No space left on device", // Linux
+            "There is not enough space on the disk", // Windows
+            "Not enough space", // Solaris
```
akuznetsov-gridgain: What about MacOS?

andrewmed: Anyone with MacOsX out there? it is Darwin, so must be unix message, but would be nice to  check, cant google it

andrewmed: Checked: on macos it is 'Linux' message

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
andrewmed: fixed

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
andrewmed: fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
```diff
@@ -366,6 +372,23 @@
         });
     }
 
+
+    /**
+     * @param th Th.
+     * @return Stack trace
+     */
+    private String getErrorMessage(Throwable th) {
```
Mmuzaf: Method is not suitable `X#getFullStackTrace` for this case?

dspavlov:

- to use GitHub UI (committer  can associate accounts here: https://gitbox.apache.org/setup/ )

PR #4381, opened 2018-07-18
Adding experimental support for Intel Optane DC Persistent Memory

mulugetam:
Ignite, when persistence mode is enabled, stores data and indexes on disk. To minimize the latency of disks, several tuning options can be applied. Setting the page size of a memory region to match the page size of the underlying storage, using a separate disk for the WAL, and using production-level SSDs are just a few of them [https://apacheignite.readme.io/docs/durable-memory-tuning#section-native-persistence-related-tuning].

A persistent memory store with low latency and high capacity offers a viable alternative to disks. In light of this, this patch introduces an experimental support for Intel Optane DC Persistent Memory (aka 3DXPoint, AEP) based on our Low Level Persistent Library (LLPL).

The patch eliminates the checkpoint process and the WAL to propagate changes to disk and maintain durability. Instead, we rely on a transactional memory region/block defined by LLPL to store both data and indexes into a persistent heap that is restored back upon application reboots.

LLPL: https://github.com/pmem/LLPL/
Examples: examples/src/main/java/org/apache/ignite/examples/aep/

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/Ignite.java
+++ b/modules/core/src/main/java/org/apache/ignite/Ignite.java
@@ -31,6 +31,7 @@ import org.apache.ignite.configuration.DataRegionConfiguration;
 import org.apache.ignite.configuration.DataStorageConfiguration;
 import org.apache.ignite.configuration.IgniteConfiguration;
 import org.apache.ignite.configuration.NearCacheConfiguration;
+import org.apache.ignite.internal.GridKernalContext;
 import org.apache.ignite.internal.util.typedef.G;
 import org.apache.ignite.lang.IgniteProductVersion;
 import org.apache.ignite.plugin.IgnitePlugin;
@@ -722,4 +723,11 @@ public interface Ignite extends AutoCloseable {
      * @return {@link DataStorageMetrics} snapshot.
      */
     public DataStorageMetrics dataStorageMetrics();
+
+    /**
+     *
+     * @return
+     */
+    public GridKernalContext context();
+
 }
--- a/modules/core/src/main/java/org/apache/ignite/internal/binary/BinarySchema.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/binary/BinarySchema.java
@@ -25,10 +25,7 @@ import java.io.Externalizable;
 import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.List;
+import java.util.*;
 
 /**
  * Schema describing binary object content. We rely on the following assumptions:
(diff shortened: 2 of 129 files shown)


svladykin:
Hi!

It seems that the LLPL is not productized at all: no releases, no Maven, no docs. 

In Apache Ignite we add dependencies using Maven. I don't think it makes sense to proceed with this PR until we have the binary release of LLPL in Maven available.

examples/src/main/java/org/apache/ignite/examples/datagrid/CacheDataStreamerExample.java
```diff
@@ -56,7 +56,7 @@
     public static void main(String[] args) throws IgniteException {
         ExamplesUtils.checkMinMemory(MIN_MEMORY);
 
-        try (Ignite ignite = Ignition.start("examples/config/example-ignite.xml")) {
+        try (Ignite ignite = Ignition.start("ekxamples/config/example-ignite.xml")) {
```
AMashenkov: Is it a typo?

modules/core/src/main/java/org/apache/ignite/Ignite.java
```diff
@@ -722,4 +723,11 @@ public IgniteLock reentrantLock(String name, boolean failoverSafe, boolean fair,
      * @return {@link DataStorageMetrics} snapshot.
      */
     public DataStorageMetrics dataStorageMetrics();
+
+    /**
+     *
+     * @return
+     */
+    public GridKernalContext context();
```
AMashenkov: GridKernalContext is internal class and resides in 'internal' package. 
Ignite is a public API interface.
Exposing internal class via public API method doesn't look like a good idea.

modules/core/src/main/java/org/apache/ignite/internal/binary/BinarySchema.java
```diff
@@ -25,10 +25,7 @@
 import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.List;
+import java.util.*;
```
AMashenkov: Ignite code convention doesn't allow asterisk in imports.

https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines#CodingGuidelines-PackageImporting

modules/core/src/main/java/org/apache/ignite/internal/binary/BinarySchemaRegistry.java
```diff
@@ -17,6 +17,11 @@
 
 package org.apache.ignite.internal.binary;
 
+import lib.llpl.Heap;
+import lib.llpl.MemoryBlock;
+import lib.llpl.Transactional;
+import org.apache.ignite.Ignition;
+import org.apache.ignite.internal.util.*;
```
AMashenkov: Ignite code convention doesn't allow asterisk in imports.

PR #4385, opened 2018-07-19
IGNITE-9021: Refactor vectors/matrix classes

zaleslaw:
- Fixed tests
- Leave classes used in MLP/DT/another algorithms only
- Fixed code-style for many places


examples/src/main/java/org/apache/ignite/examples/ml/svm/multiclass/SVMMultiClassClassificationExample.java
```diff
@@ -109,8 +109,8 @@ public static void main(String[] args) throws InterruptedException {
                         double[] inputs = Arrays.copyOfRange(val, 1, val.length);
                         double groundTruth = val[0];
 
-                        double prediction = mdl.apply(new DenseLocalOnHeapVector(inputs));
-                        double predictionWithNormalization = mdlWithNormalization.apply(new DenseLocalOnHeapVector(inputs));
+                        double prediction = mdl.apply(new DenseVector(inputs));
+                        double predictionWithNormalization = mdlWithNormalization.apply(new DenseVector(inputs));
```
avplatonov: Maybe we can use "VectorUtils.of" instead of using new-operator directly?

PR #4403, opened 2018-07-23
IGNITE-9009 Don't process backup updates for local continuous queries.

dmekhanikov:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/ContinuousQueryReassignmentTest.java
```diff
@@ -0,0 +1,170 @@
...
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+
+import static org.apache.ignite.testframework.GridTestUtils.waitForCondition;
+
+/**
+ *
+ */
+public class ContinuousQueryReassignmentTest extends GridCommonAbstractTest {
+    /** Ip finder. */
+    private static final TcpDiscoveryIpFinder IP_FINDER = new TcpDiscoveryVmIpFinder(true);
```
mcherkasov: Please use LOCAL_IP_FINDER instead.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryHandler.java
```diff
@@ -144,7 +144,7 @@
     private transient boolean skipPrimaryCheck;
 
     /** */
-    private boolean locCache;
+    private transient boolean locOnly;
```
mcherkasov: Are you sure that it should be transient?

dmekhanikov: `transient` modifier has only semantic meaning here, since this class implements `Externalizable`.
All fields in this class, that are not marked as `transient`, are used in `writeExternal()` and `readExternal()` methods. The same rule should apply for this field as well.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/ContinuousQueryReassignmentTest.java
dmekhanikov: There are only 9 usages of this IP finder in the project. Shared static IP finder is used in most tests.
I started a discussion on developers list on this matter: http://apache-ignite-developers.2346864.n4.nabble.com/IP-finder-in-tests-td33322.html
This boilerplate pattern won't be needed soon.

PR #4427, opened 2018-07-25
IGNITE-8546 Permission denied while copying OPTION_LIBS on Openshift

gromtech:
* Added permissions to copy optional libs, create work dir and execute entry point.
* Changed run.sh: `cp libs/optional/$lib/*` replaced  by `cp libs/optional/$lib` to avoid overwriting files (licenses, README, etc).

### DIFF
--- a/docker/apache-ignite/run.sh
+++ b/docker/apache-ignite/run.sh
@@ -20,7 +20,7 @@ if [ ! -z "$OPTION_LIBS" ]; then
   IFS=, LIBS_LIST=("$OPTION_LIBS")
 
   for lib in ${LIBS_LIST[@]}; do
-    cp -r $IGNITE_HOME/libs/optional/"$lib"/* \
+    cp -r $IGNITE_HOME/libs/optional/"$lib" \
         $IGNITE_HOME/libs/
   done
 fi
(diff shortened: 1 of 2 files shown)


docker/apache-ignite/run.sh
```diff
@@ -20,7 +20,7 @@ if [ ! -z "$OPTION_LIBS" ]; then
   IFS=, LIBS_LIST=("$OPTION_LIBS")
 
   for lib in ${LIBS_LIST[@]}; do
-    cp -r $IGNITE_HOME/libs/optional/"$lib"/* \
+    cp -r $IGNITE_HOME/libs/optional/"$lib" \
```
vveider: Won't be better to write it down as `cp -r "${IGNITE_HOME}/libs/optional/$lib"`? `${IGNITE_HOME}` potentially can have spaces or other special symbols as well.

gromtech: Since `$IGNITE_HOME` specified in `Dockerfile`, we can assure it doesn't contain any spaces or other special symbols.

vveider: Agree.

PR #4434, opened 2018-07-26
IGNITE-9607 Service Grid redesign - phase 1

daradurvs:


modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceDeploymentFuture.java
```diff
@@ -28,22 +29,34 @@
...
+    public GridServiceDeploymentFuture(ServiceConfiguration cfg, IgniteUuid srvcId) {
         this.cfg = cfg;
+        this.srvcId = srvcId;
     }
 
     /**
      * @return Service configuration.
      */
-    ServiceConfiguration configuration() {
+    public ServiceConfiguration configuration() {
```
nizhikov: Why this becomes public?

modules/core/src/main/java/org/apache/ignite/internal/processors/service/LazyServiceConfiguration.java
```diff
@@ -64,6 +64,7 @@ public LazyServiceConfiguration(ServiceConfiguration cfg, byte[] srvcBytes) {
         this.srvcBytes = srvcBytes;
         srvc = cfg.getService();
         srvcClsName = srvc.getClass().getName();
+        plc = cfg.getPolicy();
```
nizhikov: Is this some kind of bug?
Should we fix it in separate(quickly merged) PR?

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceInfo.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override public UUID originNodeId() {
+        return nodeId;
+    }
+
+    /** {@inheritDoc} */
+    @Override public Map<UUID, Integer> topologySnapshot() {
+        return top;
```
nizhikov: Do we want to share a reference to this map? 
Should we share a reference to read only view of it?

modules/core/src/test/java/org/apache/ignite/testsuites/IgniteServiceConfigVariationsFullApiTestSuite.java
```diff
@@ -89,4 +89,4 @@ public static TestSuite suite() throws Exception {
 
         return suite;
     }
-}
\ No newline at end of file
+}
```
nizhikov: Let's revert to master whole file

modules/core/src/test/java/org/apache/ignite/internal/processors/service/ServicePredicateAccessCacheTest.java
```diff
@@ -75,36 +76,39 @@
     public void testPredicateAccessCache() throws Exception {
         final Ignite ignite0 = startGrid(0);
 
-        ignite0.getOrCreateCache(new CacheConfiguration<>(DEFAULT_CACHE_NAME)
-            .setName("testCache")
-            .setAtomicityMode(ATOMIC)
-            .setCacheMode(REPLICATED)
-            .setWriteSynchronizationMode(FULL_SYNC));
+        CacheConfiguration<String, String> cacheCfg  = new CacheConfiguration<>();
```
nizhikov: Why this change?

modules/core/src/test/java/org/apache/ignite/internal/processors/service/ServicePredicateAccessCacheTest.java
```diff
@@ -75,36 +76,39 @@
...
+        IgniteCache<String, String> cache = ignite0.getOrCreateCache(cacheCfg);
+
+        cache.put(ignite0.cluster().localNode().id().toString(), "val");
 
         latch = new CountDownLatch(1);
 
-        final ClusterGroup grp = ignite0.cluster().forPredicate(new IgnitePredicate<ClusterNode>() {
-            @Override public boolean apply(ClusterNode node) {
-                System.out.println("Predicated started [thread=" + Thread.currentThread().getName() + ']');
+        final ClusterGroup grp = ignite0.cluster().forPredicate((IgnitePredicate<ClusterNode>) node -> {
```
nizhikov: Why this change?

daradurvs: The reason is serialization. This predicate will be used in ServiceConfiguration and send over a network.

modules/core/src/test/java/org/apache/ignite/internal/processors/service/ServicePredicateAccessCacheTest.java
daradurvs: I changed the test to get created cache for putting a value for predicate:
`cache.put(ignite0.cluster().localNode().id().toString(), "val");`

This based on deprecated logic, when service processor use internal cache and contained a value for predicate.

modules/core/src/test/java/org/apache/ignite/testsuites/IgniteServiceConfigVariationsFullApiTestSuite.java
daradurvs: Reverted.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceInfo.java
daradurvs: Wrapped by `Collections#unmodifiableMap`

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceDeploymentFuture.java
daradurvs: Legacy change. Reverted.

modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/GridDiscoveryManager.java
```diff
@@ -715,6 +715,9 @@ else if (customMsg instanceof ChangeGlobalStateFinishMessage) {
                     ctx.cache().onDiscoveryEvent(type, customMsg, node, nextTopVer, ctx.state().clusterState());
                 }
 
+                if (type == EVT_DISCOVERY_CUSTOM_EVT)
```
dmekhanikov: Just include it into the next `if` clause.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/DynamicServiceChangeRequest.java
```diff
@@ -0,0 +1,139 @@
...
+import java.io.Serializable;
+import org.apache.ignite.internal.util.typedef.internal.S;
+import org.apache.ignite.lang.IgniteUuid;
+import org.apache.ignite.services.ServiceConfiguration;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Service change request.
+ */
+public class DynamicServiceChangeRequest implements Serializable {
```
dmekhanikov: Why don't we make separate classes for deploy and undeploy actions?
The flags thing seems over-complicated.

modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/GridDiscoveryManager.java
daradurvs: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/service/DynamicServiceChangeRequest.java
daradurvs: I want to avoid creating multiple classes with similar fields, which will be processed in a unified flow. The main reason to maintain messages compatibility between releases. Also, we can use this class in the future for adding a new action "redeploy" by the new flag.

The flags give us the flexibility to add actions across versions without breaking serialization/deserialization.

BTW, this approach uses widely in Ignite and should be well known for the community, see `DynamicCacheChangeRequest`, `GridMemcachedMessage`, `GridDhtPartitionsAbstractMessage` etc.

nizhikov: @dmekhanikov 

Actually, we take main principles from cache creation flow.
As you can see `DynamicCacheChangeRequest` shares similar struscture.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
```diff
@@ -28,188 +27,127 @@
...
-                CU.UTILITY_CACHE_NAME, new ServiceEntriesListener(), null, null
-            );
-        }
-    }
-
     /** {@inheritDoc} */
     @Override public void start() throws IgniteCheckedException {
         ctx.addNodeAttribute(ATTR_SERVICES_COMPATIBILITY_MODE, srvcCompatibilitySysProp);
 
-        if (ctx.isDaemon())
```
dmekhanikov: Why did you remove this check?

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
```diff
@@ -220,217 +158,220 @@ public void onContinuousProcessorStarted(GridKernalContext ctx) throws IgniteChe
...
-        // Will not release it.
-        if (busyLock != null) {
-            busyLock.block();
+        dataBag.addGridCommonData(SERVICE_PROC.ordinal(), clusterData);
+    }
 
-            this.busyLock = null;
-        }
+    /** {@inheritDoc} */
+    @Override public void onGridDataReceived(DiscoveryDataBag.GridDiscoveryData data) {
```
dmekhanikov: What if there is a conflict between local and received services? Will an error be logged?

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
```diff
@@ -220,217 +158,220 @@ public void onContinuousProcessorStarted(GridKernalContext ctx) throws IgniteChe
...
 
-            if (svc != null)
-                try {
-                    svc.cancel(ctx);
-                }
-                catch (Throwable e) {
-                    log.error("Failed to cancel service (ignoring) [name=" + ctx.name() +
-                        ", execId=" + ctx.executionId() + ']', e);
+            if (oldDesc != null) { // In case of a collision of IgniteUuid.randomUuid() (almost impossible case)
+                U.warn(log, "Failed to register service configuration received from joining node : " +
```
dmekhanikov: I think, debug level is enough here. This is a common situation, when every node has static service configuration on them. It would lead to a lot of warnings.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
```diff
@@ -610,178 +568,95 @@ private PreparedConfigurations prepareServiceConfigurations(Collection<ServiceCo
...
      */
-    private IgniteInternalFuture<?> deployAll(Collection<ServiceConfiguration> cfgs,
+    private IgniteInternalFuture<?> deployAll(@NotNull Collection<ServiceConfiguration> cfgs,
         @Nullable IgnitePredicate<ClusterNode> dfltNodeFilter) {
-        assert cfgs != null;
-
-        PreparedConfigurations srvCfg = prepareServiceConfigurations(cfgs, dfltNodeFilter);
-
-        List<ServiceConfiguration> cfgsCp = srvCfg.cfgs;
+        opsLock.readLock().lock();
```
dmekhanikov: Why not `enterBusy()`?

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
```diff
@@ -610,178 +568,95 @@ private PreparedConfigurations prepareServiceConfigurations(Collection<ServiceCo
...
 
-                if (fut != null)
-                    fut.onDone(err);
-            }
+            GridServiceDeploymentCompoundFuture res = new GridServiceDeploymentCompoundFuture();
 
-            return new GridFinishedFuture<>(err);
-        }
+            if (!cfgsCp.isEmpty()) {
+                cfgsCp.sort(Comparator.comparing(ServiceConfiguration::getName));
```
dmekhanikov: There is no point in sorting the configurations. We did it before to avoid deadlocks. This is not needed now, since we don't run transactions, so there cannot be any deadlocks.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
```diff
@@ -1013,19 +829,16 @@ private CancelResult removeServiceFromCache(String name) throws IgniteCheckedExc
...
 
-        synchronized (locSvcs) {
-            ctxs = locSvcs.get(name);
-        }
+        try {
+            ctx.security().authorize(name, SecurityPermission.SERVICE_INVOKE, null);
 
-        if (ctxs == null)
-            return null;
+            Collection<ServiceContextImpl> ctxs = Optional.ofNullable(lookupDeployedServiceId(name))
```
dmekhanikov: This two-liner is evil. I need to spend more time understanding it rather than if it were a long code block with `if` statements.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
```diff
@@ -1120,18 +936,18 @@ private boolean hasLocalNode(ClusterGroup prj) {
...
+            return null;
 
-        Collection<ServiceContextImpl> ctxs;
+        try {
+            ctx.security().authorize(name, SecurityPermission.SERVICE_INVOKE, null);
 
-        synchronized (locSvcs) {
-            ctxs = locSvcs.get(name);
-        }
+            Collection<ServiceContextImpl> ctxs = Optional.ofNullable(lookupDeployedServiceId(name))
```
dmekhanikov: Evil.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
```diff
@@ -1037,24 +850,24 @@ private CancelResult removeServiceFromCache(String name) throws IgniteCheckedExc
...
-        }
-
-        if (ctxs == null)
+        if (!enterBusy())
             return null;
 
-        synchronized (ctxs) {
-            if (ctxs.isEmpty())
+        try {
+            Collection<ServiceContextImpl> ctxs = Optional.ofNullable(lookupDeployedServiceId(name))
```
dmekhanikov: Evil

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
```diff
@@ -1469,727 +1242,543 @@ private Service copyAndInject(ServiceConfiguration cfg) throws IgniteCheckedExce
...
-        boolean rollback;
-
-        /**
-         * @param fut Future.
-         * @param rollback {@code True} if service was cancelled during current call.
-         */
-        CancelResult(IgniteInternalFuture<?> fut, boolean rollback) {
-            this.fut = fut;
-            this.rollback = rollback;
+    @Nullable private ServiceInfo lookupInRegisteredServices(String name) {
```
dmekhanikov: This is a possible point of optimization. This lookup is done in a discovery thread and may take significant time, if there are thousands of services deployed. This place should be benchmarked. I don't think, this is a blocker for the PR, but it would be good to investigate, how it affects the performance.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceInfo.java
```diff
@@ -0,0 +1,170 @@
...
+        return (K)cfg.getAffinityKey();
+    }
+
+    /** {@inheritDoc} */
+    @Override public UUID originNodeId() {
+        return originNodeId;
+    }
+
+    /** {@inheritDoc} */
+    @Override public Map<UUID, Integer> topologySnapshot() {
```
dmekhanikov: Topology snapshot has a different meaning in Ignite.
If you use it here, it will overload and confuse its meaning.
I think, "service assignments" would be fine.

daradurvs: `ServiceInfo` implements `ServiceDescriptor` and `#topologySnapshot` relates to the interface which I'd not changed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
daradurvs: `GridServiceProcessor` has been annotated `@SkipDaemon`, so no any additional checks are needed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
daradurvs: Any conflicts can't be here.
Joining node sends static configs in the cluster and receives verified and full data from the cluster, which will contain these static configs too if they pass validation.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
daradurvs: The logging level has been chosen to be similar with IgniteCache.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
daradurvs: In case of using `enterBusy()` we should duplicate the checks below because of returning value of the method.
The main reason to use lock here to synchronize method which changing state (onDisconnect, onKernalStop), that means we should throw several types of exceptions.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
daradurvs: Removed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/DynamicServiceChangeRequest.java
dmekhanikov: The only common property you need to share between `deploy` and `undeploy` actions is `serviceId`. You can make an abstract class with this property.
`cfg` field is always `null` for the undeployment requests, so it doesn't need to be shared. And flags `deploy` and `undeploy` look like a bad design, because they can be set to true at the same time.
I don't see, how adding new message types violate compatibility. If you need to add a new action, then you'll just add another message type.

modules/core/src/test/java/org/apache/ignite/internal/processors/service/GridServiceDeploymentExceptionPropagationTest.java
```diff
@@ -31,15 +30,15 @@
...
+
+                    fail("Deployment exception has been expected.");
                 }
-                catch (IgniteException ignored) {
-                    assertTrue(log.toString().contains("ServiceImpl init exception"));
+                catch (ServiceDeploymentException ex) {
+                    assertTrue(ex.getSuppressed()[0].getMessage().contains("ServiceImpl init exception"));
 
                     return; // Exception is what we expect.
                 }
```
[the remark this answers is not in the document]
daradurvs: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProxy.java
```diff
@@ -198,7 +198,7 @@ else if (U.isToStringMethod(mtd))
                             true).get();
                     }
                 }
-                catch (GridServiceNotFoundException | ClusterTopologyCheckedException e) {
+                catch (ClusterTopologyCheckedException e) {
                     if (log.isDebugEnabled())
                         log.debug("Service was not found or topology changed (will retry): " + e.getMessage());
```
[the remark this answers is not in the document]
daradurvs: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesCommonDiscoveryData.java
```diff
@@ -0,0 +1,55 @@
...
+     * @param registeredServices Clusters registered services descriptors.
+     */
+    public ServicesCommonDiscoveryData(@NotNull ArrayList<ServiceInfo> registeredServices) {
+        this.registeredServices = registeredServices;
+    }
+
+    /**
+     * Returns clusters registered services descriptors.
+     *
+     * @return Deployed services descriptors.
```
dmekhanikov: I'm confused. Is it a collection of deployed or registered services?

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentTask.java
```diff
@@ -0,0 +1,856 @@
...
+    }
+
+    /**
+     * Handles when this task is being added in deployment queue.
+     * <p/>
+     * Introduced to avoid overhead on calling of {@link Collection#contains(Object)}}.
+     *
+     * @return {@code true} if task is has not been added previously, otherwise {@code false}.
+     */
+    protected boolean onAdditionInQueue() {
```
dmekhanikov: `onEnqueued()` maybe?

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesCommonDiscoveryData.java
```diff
@@ -0,0 +1,55 @@
...
+
+package org.apache.ignite.internal.processors.service;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import org.apache.ignite.internal.util.typedef.internal.S;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Initial data container to send on joined node.
```
dmekhanikov: Send **to** or send **from**?

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentTask.java
```diff
@@ -0,0 +1,856 @@
...
+            finally {
+                srvcProc.deployment().deployerBlockingSectionEnd();
+            }
+        });
+
+        depActions.servicesToDeploy().forEach((srvcId, desc) -> {
+            try {
+                ServiceConfiguration cfg = desc.configuration();
+
+                Map<UUID, Integer> top = reassign(srvcId, cfg, evtTopVer, filterDeadNodes(desc.topologySnapshot()));
```
dmekhanikov: This code is executed in `ServicesDeploymentWorker` thread, so discovery events may occur concurrently with it.
So, it's possible for different nodes to get different `oldTop` and calculate different service assignments.

It would be great to cover such cases with tests. Try reproducing this bug in a test.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentManager.java
```diff
@@ -0,0 +1,553 @@
...
+     */
+    private void addTask(@NotNull DiscoveryEvent evt, @NotNull AffinityTopologyVersion topVer,
+        @Nullable ServicesDeploymentActions depActions) {
+        final ServicesDeploymentProcessId depId = deploymentId(evt, topVer);
+
+        ServicesDeploymentTask task = tasks.computeIfAbsent(depId,
+            t -> new ServicesDeploymentTask(ctx, depId));
+
+        if (!task.onAdditionInQueue()) {
+            log.warning("Do not start service deployment process for event: " + evt);
```
dmekhanikov: This is not informative. What is the reason? What should a user do?

daradurvs: This is not useful for the user. I changed the level to 'debug' and enrol description.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesCommonDiscoveryData.java
daradurvs: Send to joining node. I changed the comment.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentTask.java
daradurvs: I like your idea. Changed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesCommonDiscoveryData.java
daradurvs: Fixed comment.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProxy.java
```diff
@@ -212,7 +208,7 @@ else if (U.isToStringMethod(mtd))
                     if (ignorableCause == null)
                         ignorableCause = X.cause(e, ClusterTopologyCheckedException.class);
 
-                    if (ignorableCause != null) {
+                    if (ignorableCause != null && !ctx.service().eventDrivenServiceProcessorEnabled()) {
```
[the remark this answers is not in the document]
daradurvs: Makes sense to me, fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceDeploymentFuture.java
```diff
@@ -19,20 +19,33 @@
 
 import org.apache.ignite.internal.util.future.GridFutureAdapter;
 import org.apache.ignite.internal.util.typedef.internal.S;
+import org.apache.ignite.lang.IgniteUuid;
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
```diff
@@ -814,13 +792,9 @@ private void writeServiceToCache(GridServiceDeploymentCompoundFuture res, Servic
...
 
-    /**
-     * @param svcNames Name of service to deploy.
-     * @return Future.
-     */
-    @SuppressWarnings("unchecked")
-    public IgniteInternalFuture<?> cancelAll(Collection<String> svcNames) {
-        List<String> svcNamesCp = new ArrayList<>(svcNames);
+    /** {@inheritDoc} */
+    @Override public IgniteInternalFuture<?> cancelAll(Collection<String> servicesNames) {
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceDeploymentFuture.java
```diff
@@ -19,20 +19,33 @@
...
 import org.apache.ignite.services.ServiceConfiguration;
 
 /**
  * Service deployment future.
  */
-public class GridServiceDeploymentFuture extends GridFutureAdapter<Object> {
+public class GridServiceDeploymentFuture<T> extends GridFutureAdapter<Object> {
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1691 @@
...
+     *
+     * @param evt Discovery event.
+     * @param discoCache Discovery cache.
+     */
+    public void onLocalJoin(DiscoveryEvent evt, DiscoCache discoCache) {
+        assert ctx.localNodeId().equals(evt.eventNode().id());
+        assert evt.type() == EVT_NODE_JOINED;
+
+        if (isLocalNodeCoordinator()) {
+            // First node start, {@link #onGridDataReceived(DiscoveryDataBag.GridDiscoveryData)} has not been called
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1691 @@
...
+            DynamicServicesChangeRequestBatchMessage msg0 = ((DynamicServicesChangeRequestBatchMessage)msg);
+
+            if (!state.active() || state.transition()) {
+                for (DynamicServiceChangeRequest req : msg0.requests()) {
+                    GridFutureAdapter<?> fut = req.deploy() ?
+                        depFuts.remove(req.serviceId()) :
+                        undepFuts.remove(req.serviceId());
+
+                    if (fut != null) {
+                        fut.onDone(new IgniteCheckedException("Operation has been cancelled " +
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/GridDiscoveryManager.java
```diff
@@ -712,6 +715,9 @@ else if (customMsg instanceof ChangeGlobalStateFinishMessage) {
                 }
 
                 if (type == EVT_DISCOVERY_CUSTOM_EVT) {
+                    if (ctx.service().eventDrivenServiceProcessorEnabled())
+                        ((IgniteServiceProcessor)ctx.service()).onCustomEvent(customMsg, node, ctx.state().clusterState());
```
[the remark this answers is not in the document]
daradurvs: Good suggestion. I will test this refactoring.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1691 @@
...
+
+    /**
+     * Callback invoked from discovery thread when discovery custom message is received, before discovery listeners
+     * notification.
+     *
+     * @param msg Discovery custom message.
+     * @param node Event node.
+     * @param state Current cluster state.
+     */
+    public void onCustomEvent(DiscoveryCustomMessage msg, ClusterNode node, DiscoveryDataClusterState state) {
```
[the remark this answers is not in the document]
daradurvs: Reworked via CustomEventListener.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentManager.java
```diff
@@ -0,0 +1,553 @@
...
+    /**
+     * Services deployment worker.
+     */
+    private class ServicesDeploymentWorker extends GridWorker {
+        /** Queue to process. */
+        private final LinkedBlockingQueue<ServicesDeploymentTask> tasksQueue = new LinkedBlockingQueue<>();
+
+        /** {@inheritDoc} */
+        private ServicesDeploymentWorker() {
+            super(ctx.igniteInstanceName(), "services-deployment-deployer",
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1691 @@
...
+
+                                used.add(e.getKey());
+
+                                if (--remainder == 0)
+                                    break;
+                            }
+                        }
+
+                        if (remainder > 0) {
+                            List<Map.Entry<UUID, Integer>> entries = new ArrayList<>(cnts.entrySet());
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentTask.java
```diff
@@ -0,0 +1,856 @@
...
+     * @param msg Full services map message.
+     */
+    protected void onReceiveFullDeploymentsMessage(ServicesFullDeploymentsMessage msg) {
+        assert depId.equals(msg.deploymentId()) : "Wrong message's deployment process id, msg=" + msg;
+
+        initTaskFut.listen((IgniteInClosure<IgniteInternalFuture<?>>)fut -> {
+            if (isCompleted())
+                return;
+
+            ctx.closure().runLocalSafe(() -> {
```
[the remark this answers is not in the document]
daradurvs: No, this method is called once services deployments message received. The closure engine is used to avoid long-running handling by discovery thread.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentManager.java
```diff
@@ -0,0 +1,553 @@
...
+
+                            if (task != null) // May be null in case of double delivering
+                                task.onReceiveFullDeploymentsMessage(msg0);
+                        }
+                        else if (msg instanceof CacheAffinityChangeMessage)
+                            addTask(copyIfNeeded((DiscoveryCustomEvent)evt), discoCache.version(), null);
+                        else {
+                            ServicesDeploymentActions depActions = null;
+
+                            if (msg instanceof ChangeGlobalStateMessage)
```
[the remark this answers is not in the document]
daradurvs: I think that's not necessary, because we have only one specific implementation on all services deployment infrastructure.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentManager.java
```diff
@@ -0,0 +1,553 @@
...
+import static org.apache.ignite.events.EventType.EVT_NODE_LEFT;
+import static org.apache.ignite.failure.FailureType.CRITICAL_ERROR;
+import static org.apache.ignite.failure.FailureType.SYSTEM_WORKER_TERMINATION;
+import static org.apache.ignite.internal.GridTopic.TOPIC_SERVICES;
+import static org.apache.ignite.internal.events.DiscoveryCustomEvent.EVT_DISCOVERY_CUSTOM_EVT;
+
+/**
+ * Services deployment manager.
+ */
+public class ServicesDeploymentManager {
```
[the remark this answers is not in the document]
daradurvs: `ServicesDeploymentManager` holds discovery and communication listeners, manages tasks queue and worker.
`IgniteServiceProcessor` listens only custom messages of specific types, not all discovery events, and manages services lifecycle.
I'd not merge them in one class, because this makes it harder to understand and support.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentManager.java
```diff
@@ -0,0 +1,553 @@
...
+     *
+     * @param evt Discovery event.
+     * @param topVer Topology version.
+     * @return Services deployment process id.
+     */
+    private ServicesDeploymentProcessId deploymentId(@NotNull DiscoveryEvent evt,
+        @NotNull AffinityTopologyVersion topVer) {
+        return evt instanceof DiscoveryCustomEvent ?
+            new ServicesDeploymentProcessId(((DiscoveryCustomEvent)evt).customMessage().id()) :
+            new ServicesDeploymentProcessId(topVer);
```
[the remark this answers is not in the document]
daradurvs: All events reached 'addTask' method.
The common thing that the deployment process on the whole cluster may be triggered by:
1. The custom event, that has a unique message id (the same across cluster nodes)
2. Topology change 
So, we using this approach we are able to identify all messages related to the service deployment process across the cluster.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentTask.java
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1691 @@
...
+     */
+    protected void redeploy(IgniteUuid srvcId, ServiceConfiguration cfg,
+        Map<UUID, Integer> top) throws IgniteCheckedException {
+        String name = cfg.getName();
+        String cacheName = cfg.getCacheName();
+        Object affKey = cfg.getAffinityKey();
+
+        int assignCnt = top.getOrDefault(ctx.localNodeId(), 0);
+
+        Collection<ServiceContextImpl> ctxs = locServices.computeIfAbsent(srvcId, c -> new ArrayList<>());
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1691 @@
...
+     * Processes deployment result.
+     *
+     * @param fullTops Deployment topologies.
+     */
+    protected void updateServicesTopologies(@NotNull final Map<IgniteUuid, HashMap<UUID, Integer>> fullTops) {
+        if (!enterBusy())
+            return;
+
+        try {
+            updateServicesMap(deployedServices, fullTops);
```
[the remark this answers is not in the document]
daradurvs: Because we update deployed services here. Notifications are needed only for registered services collection.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1691 @@
...
+     */
+    protected void updateDeployedServices(final ServicesDeploymentActions depActions) {
+        if (!enterBusy())
+            return;
+
+        try {
+            depActions.servicesToDeploy().forEach((srvcId, desc) -> {
+                ServiceInfo old = deployedServices.putIfAbsent(srvcId, desc);
+
+                assert old == desc || old == null : "Concurrent map modification.";
```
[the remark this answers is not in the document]
daradurvs: Legacy. Removed check.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1691 @@
...
+     * @param staticCfgs {@code true} to check statically defined services during node startup process, in this case
+     * {@link JdkMarshaller} (common marshaller for discovery layer) will be used to check services serialization
+     * because of {@link GridMarshallerMappingProcessor} is not started at this point, otherwise {@code false}.
+     * @return Configurations to deploy.
+     */
+    private PreparedConfigurations prepareServiceConfigurations(Collection<ServiceConfiguration> cfgs,
+        IgnitePredicate<ClusterNode> dfltNodeFilter, boolean staticCfgs) {
+        List<ServiceConfiguration> cfgsCp = new ArrayList<>(cfgs.size());
+
+        Marshaller marsh = !staticCfgs ? ctx.config().getMarshaller() : new JdkMarshaller();
```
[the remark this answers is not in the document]
daradurvs: Reworked.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1691 @@
...
+
+            if (err == null)
+                err = checkPermissions(cfg.getName(), SecurityPermission.SERVICE_DEPLOY);
+
+            if (err == null) {
+                try {
+                    byte[] srvcBytes = U.marshal(marsh, cfg.getService());
+
+                    if (!staticCfgs)
+                        cfgsCp.add(new LazyServiceConfiguration(cfg, srvcBytes));
```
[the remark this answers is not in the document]
daradurvs: Fixed,

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1691 @@
...
+ * Event driven implementation of service processor. Services deployment are managed by messages across discovery spi
+ * and communication spi.
+ */
+@SkipDaemon
+public class IgniteServiceProcessor extends ServiceProcessorAdapter implements IgniteChangeGlobalStateSupport {
+    /** Local service instances. */
+    private final ConcurrentMap<IgniteUuid, Collection<ServiceContextImpl>> locServices = new ConcurrentHashMap<>();
+
+    /** Cluster services info <b>updated from discovery thread</b>. */
+    private final ConcurrentMap<IgniteUuid, ServiceInfo> registeredServices = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
daradurvs: Improved javadoc.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1691 @@
...
+            }
+        }
+    }
+
+    /**
+     * @param name Service name.
+     * @return Service topology.
+     */
+    private Map<UUID, Integer> serviceTopology(String name) {
+        for (ServiceInfo desc : registeredServices.values()) {
```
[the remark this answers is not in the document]
daradurvs: Use of `registeredServices` allows to newly joining node use a proxy for working with services already deployed in the cluster and avoid waiting for the finishing of the deployment process on joining topVer.
Also `registeredServices ` contains actual services assignments received in `ServicesFullDeploymentsMessage`.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1691 @@
...
+                if (!toDeploy.isEmpty())
+                    depActions.servicesToDeploy(toDeploy);
+
+                if (!toUndeploy.isEmpty())
+                    depActions.servicesToUndeploy(toUndeploy);
+
+                msg0.servicesDeploymentActions(depActions);
+            }
+        }
+        else if (msg instanceof ChangeGlobalStateMessage) {
```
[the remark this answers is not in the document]
daradurvs: Activation and deactivation actions are processed in a similar way as other service's state change request. At this point, we just have prepared `ServicesDeplpoymentActions` to process them in deployment worker in order of queue.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1691 @@
...
+        }
+    }
+
+    /**
+     * @param ctxs Contexts to cancel.
+     * @param cancelCnt Number of contexts to cancel.
+     */
+    private void cancel(Iterable<ServiceContextImpl> ctxs, int cancelCnt) {
+        for (Iterator<ServiceContextImpl> it = ctxs.iterator(); it.hasNext(); ) {
+            cancel(it.next());
```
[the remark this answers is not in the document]
daradurvs: Yes, it's expected. It's done to avoid a hang in case of node stopping or on disconnect during deployment. Now it considers behaviour as it was before, to avoid breaking existing tests. It is covered by `GridServiceProcessorStopSelfTest`.
Also, here is a task to improve this in near future:
https://issues.apache.org/jira/browse/IGNITE-9076

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1691 @@
...
+    }
+
+    /**
+     * @param cfgs Service configurations.
+     * @param dfltNodeFilter Default NodeFilter.
+     * @return Future for deployment.
+     */
+    private IgniteInternalFuture<?> deployAll(@NotNull Collection<ServiceConfiguration> cfgs,
+        @Nullable IgnitePredicate<ClusterNode> dfltNodeFilter) {
+        opsLock.readLock().lock();
```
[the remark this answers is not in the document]
daradurvs: If we can't enter to the busy state it means that cluster is in a transition state. But we should throw different types of exception in different cases: {"node stopping", "disconnected"}. It's node about exception's message, it's about returning "reconnect-feature" wrapped by exception. Users requests and cluster change requests are processed in different threads with different entrances depend on causes, so we need synchronize deploy/cancel methods with `onKernalStop` and `onDisconneced`.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentManager.java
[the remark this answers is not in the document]
dmekhanikov: There are four message types, that have this method: `ChangeGlobalStateMessage`, `DynamicServicesChangeRequestBatchMessage`, `DynamicCacheChangeBatch` and `ServicesFullDeploymentsMessage`. This piece of code would look simpler, if they shared the same interface, responsible for the deployment actions.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentTask.java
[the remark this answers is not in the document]
dmekhanikov: Consider a situation, when there is a service deployed in a cluster, and several nodes leave the cluster consequently.
So, is it possible to receive two `ServicesFullDeploymentsMessage`s regarding the same service, one after another? In this case two tasks will be executed concurrently in different threads.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceDeploymentCompoundFuture.java
```diff
@@ -23,6 +23,7 @@
 import org.apache.ignite.IgniteCheckedException;
 import org.apache.ignite.internal.IgniteInternalFuture;
 import org.apache.ignite.internal.util.future.GridCompoundFuture;
+import org.apache.ignite.lang.IgniteUuid;
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentTask.java
```diff
@@ -0,0 +1,868 @@
...
+            initTaskFut.onDone(err);
+
+        if (!initCrdFut.isDone())
+            initCrdFut.onDone(err);
+    }
+
+    /**
+     * Releases resources to reduce memory usages.
+     */
+    protected void clear() {
```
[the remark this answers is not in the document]
daradurvs: The main reason is to release objects for GC as soon as possible.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentTask.java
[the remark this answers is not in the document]
daradurvs: It's impossible in the current design. If N nodes fail, N tasks will be added in the deployment queue, each task will be processed one by one.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentTask.java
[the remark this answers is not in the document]
dmekhanikov: But you call it right after removing the task from `ServicesDeploymentManager#tasks` map. The entire task is ready for GC after that.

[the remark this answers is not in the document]
daradurvs: `singleDepsMsgs` may be cached in communications spi;
Link on an entry of `depErrors` may be cached in `ServicesSingleDeploymentsMessage` also.

This depends on an algorithm of GC very much. For example, G1 splits memory on equal chunks and clean them separately, but objects can be placed in different chunks, e.g `singleDepsMsgs` in chunk `A` and its entry in `B`, and an object can't be cleaned from `B` while `singleDepsMsgs` from `A` is not cleaned.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentProcessId.java
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Services deployment process id.
+ */
+public class ServicesDeploymentProcessId implements Message {
+    /** */
+    private static final long serialVersionUID = 0L;
+
+    /** Topology version. */
+    @Nullable private AffinityTopologyVersion topVer;
```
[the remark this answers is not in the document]
daradurvs: Some of service deployment processes are triggered by `DiscoveryEvent` not `DiscoveryCustomEvent`, this kind of event does not have `DiscoveryCustomMessage` to getting request id. Topology version is the unique identifier for this events.

modules/core/src/test/java/org/apache/ignite/internal/processors/service/GridServiceProcessorMultiNodeConfigSelfTest.java
```diff
@@ -23,6 +23,7 @@
 import org.apache.ignite.Ignite;
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1764 @@
...
+    @Override public <T> T service(String name) {
+        if (!enterBusy())
+            return null;
+
+        try {
+            ctx.security().authorize(name, SecurityPermission.SERVICE_INVOKE, null);
+
+            Collection<ServiceContextImpl> ctxs = serviceContexts(name);
+
+            if (F.isEmpty(ctxs))
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1764 @@
...
+
+    /** {@inheritDoc} */
+    @Override public ServiceContextImpl serviceContext(String name) {
+        if (!enterBusy())
+            return null;
+
+        try {
+            Collection<ServiceContextImpl> ctxs = serviceContexts(name);
+
+            if (F.isEmpty(ctxs))
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1764 @@
...
+    @Override public <T> Collection<T> services(String name) {
+        if (!enterBusy())
+            return null;
+
+        try {
+            ctx.security().authorize(name, SecurityPermission.SERVICE_INVOKE, null);
+
+            Collection<ServiceContextImpl> ctxs = serviceContexts(name);
+
+            if (F.isEmpty(ctxs))
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1764 @@
...
+    }
+
+    /**
+     * @param srvcId Service id.
+     * @return Count of locally deployed service with given id.
+     */
+    protected int localInstancesCount(IgniteUuid srvcId) {
+        Collection<ServiceContextImpl> ctxs = locServices.get(srvcId);
+
+        return ctxs != null ? ctxs.size() : 0;
```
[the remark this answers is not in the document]
daradurvs: Synchronization is not needed here since `locServices` may be changed only from the same thread (deployment worker).

modules/core/src/test/java/org/apache/ignite/internal/processors/service/DynamicServiceChangeRequestTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests of {@link DynamicServiceChangeRequest}.
+ */
+public class DynamicServiceChangeRequestTest {
```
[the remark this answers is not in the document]
daradurvs: Changed the requests structure and removed the test.

modules/core/src/test/java/org/apache/ignite/internal/processors/service/ServiceInfoSelfTest.java
```diff
@@ -0,0 +1,137 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertSame;
+
+/**
+ * Tests of {@link ServiceInfo} class.
+ */
+public class ServiceInfoSelfTest {
```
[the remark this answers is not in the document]
daradurvs: Classic unit test to protect the code in future refactoring.

modules/core/src/test/java/org/apache/ignite/internal/processors/service/ServicesDeploymentProcessIdSelfTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests of {@link ServicesDeploymentProcessId}.
+ */
+@RunWith(Parameterized.class)
+public class ServicesDeploymentProcessIdSelfTest {
```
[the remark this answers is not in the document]
daradurvs: Classic unit test to protect the code in future refactoring.

modules/core/src/test/java/org/apache/ignite/internal/processors/service/ServiceDeploymentProcessingOnCoordinatorChangeTest.java
```diff
@@ -0,0 +1,134 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * Tests that requests of change service's state won't be missed and will be handled correctly on a coordinator change.
+ *
+ * It uses {@link LongInitializedTestService} with long running #init method to delay requests processing.
+ */
+@RunWith(JUnit4.class)
+public class ServiceDeploymentProcessingOnCoordinatorChangeTest extends GridCommonAbstractTest {
```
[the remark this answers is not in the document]
daradurvs: Added `BlockingTcpDiscoverySpi` to the tests.

As for non-coordinator tests, we have a lot of such tests, e.g. `GridServiceProcessorMultiNodeConfigSelfTest` and `GridServiceReassignmentSelfTest` which tests redeployments (deployment processes) in unstable topology, so it looks like we don't need more tests for this case.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentTask.java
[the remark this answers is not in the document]
daradurvs: `DiscoveryCustomEvent#nullifyingCustomMsgLock` is used to avoid nullified messages which may be nullified by other systems.

For example, we listen to the following messages:
- `ChangeGlobalStateMessage` - to perform activation/deactivation actions; 
- `DynamicCacheChangeBatch` - to handle caches stopping to undeploy 
related affinity services; 
- `CacheAffinityChangeMessage `- to recalculate assignments for affinity 
services in case of late affinity. 

These messages may be nullified in PME infrastructure (`GridDhtPartitionsExchangeFuture#onDone`)

I've already faced with such case:
http://apache-ignite-developers.2346864.n4.nabble.com/Danger-change-of-DiscoveryCustomEvent-in-GridDhtPartitionsExchangeFuture-onDone-td35946.html

modules/core/src/test/java/org/apache/ignite/internal/processors/service/ServiceInfoSelfTest.java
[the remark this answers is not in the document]
daradurvs: The test is intended to test the implementation of `ServiceDescriptor` interface.
I mean `ServiceInfo` implements `ServiceDescriptor` interface and initialized by service configuration, interface methods should return equal values in comparison with used service configuration.

modules/core/src/test/java/org/apache/ignite/internal/processors/service/ServiceDeploymentProcessingOnCoordinatorChangeTest.java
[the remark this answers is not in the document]
daradurvs: Class does not have a getter for `ServiceConfiguration` passed in the constructor.

modules/core/src/test/java/org/apache/ignite/internal/processors/service/ServiceInfoSelfTest.java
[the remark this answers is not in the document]
daradurvs: Class does not have a getter for ServiceConfiguration passed in the constructor.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServicesDeploymentTask.java
```diff
@@ -0,0 +1,868 @@
...
+     *
+     * @param evtTopNodes Ids being used to filter.
+     * @param top Service topology snapshot.
+     * @return Filtered service topology snapshot.
+     */
+    private Map<UUID, Integer> filterDeadNodes(Collection<UUID> evtTopNodes, Map<UUID, Integer> top) {
+        if (F.isEmpty(top))
+            return top;
+
+        Map<UUID, Integer> filtered = new HashMap<>();
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/service/ServiceDeploymentProcessingOnCoordinatorChangeTest.java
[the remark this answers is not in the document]
daradurvs: Agree. Tests have been reworked.
See:
`ServiceDeploymentProcessingOnCoordinatorLeftTest`
`ServiceDeploymentProcessingOnCoordinatorFailTest`
`ServiceDeploymentProcessingOnNodesLeftTest`
`ServiceDeploymentProcessingOnNodesFailTest`

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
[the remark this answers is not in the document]
dmekhanikov: Actually it can be called from a system pool. Take a look at the following method: `ServicesDeploymentTask#onReceiveFullDeploymentsMessage`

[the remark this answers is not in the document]
dmekhanikov: Actually this method may be called from another thread, e.g. from system thread pool. Take a look at the following method: `ServicesDeploymentTask#onReceiveFullDeploymentsMessage`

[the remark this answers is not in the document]
daradurvs: Yes, but only one `ServicesDeploymentTask` can be processed at any moment, it means when method `localInstancesCount` will be called nobody changes `locServices`.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1764 @@
...
+            oldDesc = lookupInRegisteredServices(desc.name());
+
+            if (oldDesc == null) {
+                registeredServices.put(desc.serviceId(), desc);
+
+                continue;
+            }
+
+            if (oldDesc.configuration().equalsIgnoreNodeFilter(desc.configuration())) {
+                U.warn(log, "Ignore service configuration received from joining node : " +
```
dmekhanikov: Log level of this message should be changed to debug. There is nothing bad in matching configurations. It may happen if some nodes have the same services in their static configuration.
The same situation with cache configurations doesn't trigger any warnings.

daradurvs: Changed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/DynamicServiceChangeRequest.java
daradurvs: Reworked.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceProcessorAdapter.java
```diff
@@ -0,0 +1,155 @@
...
+     * @param timeout If greater than 0 limits task execution time. Cannot be negative.
+     * @return Service topology.
+     * @throws IgniteCheckedException On error.
+     */
+    public abstract Map<UUID, Integer> serviceTopology(String name, long timeout) throws IgniteCheckedException;
+
+    /**
+     * @return {@code true} if event driven service processor enabled, otherwise {@code false};
+     */
+    public boolean eventDrivenServiceProcessorEnabled() {
```
slukyano: So, this method is supposed to be removed together with the old `GridServiceProcessor`, right? So it's kind of deprecated and doomed?

I think it would be better to remove the method and all related checks. If you need to do something only in the new processor (e.g. `localNodeJoin`), just and a no-op implementation to the old one. If you need to do something only in the old one, just check with `srvcProc instanceof GridServiceProcessor` - the ugliness of the check will match the deprecated status of the old processor.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1788 @@
...
+    @Override public void onActivate(GridKernalContext kctx) {
+        // No-op.
+    }
+
+    /**
+     * Invokes from services deployment worker.
+     * <p/>
+     * {@inheritDoc}
+     */
+    @Override public void onDeActivate(GridKernalContext kctx) {
```
slukyano: Is this actually called? I see `onDeActivate` being called exclusively for `GridServiceProcessor` in some other places.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1788 @@
...
+        }
+    }
+
+    /**
+     * @param deploy {@code true} if complete deployment requests, otherwise complete undeployment request will be
+     * completed.
+     * @param reqSrvcId Request's service id.
+     * @param err Error to complete with. If {@code null} a future will be completed successfully.
+     */
+    protected void completeInitiatingFuture(boolean deploy, IgniteUuid reqSrvcId, Throwable err) {
```
slukyano: Why `protected`?
I actually see a lot of `protected` in the patch, and one usually needs that only when 3rd party code (or test) is supposed to extend one's classes. Could you please review the `protected` usages and replace them with `private`/package-private where possible?

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceClusterDeploymentResultBatch.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /** {@inheritDoc} */
+    @Override public boolean stopProcess() {
+        return false;
+    }
+
+    /** {@inheritDoc} */
+    @Override public DiscoCache createDiscoCache(GridDiscoveryManager mgr, AffinityTopologyVersion topVer,
+        DiscoCache discoCache) {
+        // No-op.
```
slukyano: No need in this comment.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceDeploymentManager.java
```diff
@@ -0,0 +1,534 @@
...
+            depWorker.tasksQueue.clear();
+
+            pendingEvts.clear();
+
+            tasks.values().forEach(t -> t.completeError(stopErr));
+
+            tasks.clear();
+        }
+        catch (Exception e) {
+            log.error("Error occurred during stopping deployment worker.", e);
```
slukyano: I think we would want to propagate the exception upwards, probably killing the node. Swallowing an exception doesn't seem to be good.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1788 @@
+/*
```
slukyano: Can we avoid adding yet another 2k-lines class? I'd be happy if this would be decomposed into 4-5 smaller ones. Some of the methods (like `coordinator`) could be moved to utility classes.

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1051,6 +1053,23 @@
...
+     * in cluster must have identical value of this property.
+     * <p/>
+     * If the property is {@code true} then even driven architecture based implementation of service processor will be
+     * used.
+     * <p/>
+     * If the property is {@code false} then internal cache based implementation of service processor will be used.
+     * <p/>
+     * Default is {@code true}.
+     *
+     * @see GridServiceProcessor
```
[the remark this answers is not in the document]
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceDeploymentManager.java
daradurvs: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -0,0 +1,1788 @@
...
+ * Event driven implementation of service processor. Services deployment are managed by messages across discovery spi
+ * and communication spi.
+ *
+ * @see ServiceDeploymentManager
+ * @see ServiceDeploymentTask
+ * @see ServiceDeploymentActions
+ * @see ServiceChangeBatchRequest
+ */
+@SkipDaemon
+@SuppressWarnings("SynchronizationOnLocalVariableOrMethodParameter")
```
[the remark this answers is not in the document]
daradurvs: In this case, we should mark many methods, I'd like to avoid this.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
daradurvs: Yes, it's called from own deployment manager.
`GridServiceProcessor` is called in PME flow, because it is based on an internal cache, that means it has to be deactivated with caches.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceClusterDeploymentResultBatch.java
daradurvs: It has been added to be similar to Ignite's codebase where it is used widely.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceProcessorAdapter.java
daradurvs: This method has been introduced to be called in heavy loaded place to avoid performance issues with `instance of` operator. But these places have been reworked and the remained places are not hot, so it can be replaced `instance of`.
I removed this method.
Thanks!

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
daradurvs: The implementation is based on `GridServiceProcessor` but incompatible with it. Implementation has been detached to be able to use both implementations.
I'd like to avoid complete reworking of the class, in phase 1, to not complicate review.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceClusterDeploymentResultBatch.java
slukyano: I thought we only use it in `void` methods (to have at least some body), but OK, I don't mind having that here either.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
slukyano: When will this method be called if not during PME? I'd expect all services to be cancelled by the time `cluster().active(false)` returns. Is it the case?

daradurvs: Sure, it is. `ServiceDeploymentManager` listens to `ChangeGlobalStateMessage` to perform activation/deactivation.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceProcessorAdapter.java
slukyano: Can you elaborate on the "performance issues with `instanceof`"? Never heard of it being a thing. I'd say that JIT should optimize that to `true`/`false` quite easily, especially since you're going to only have instances of one of the service processor implementations in the JVM.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
slukyano: This kind of implies that reworking of the class is scheduled for phase 2 :) If that's the case then great!

PR #4436, opened 2018-07-26
IGNITE-9064: Decision tree optimization

avplatonov:
@dmitrievanthony 
@artemmalykh 
Please review this code


modules/ml/src/main/java/org/apache/ignite/ml/tree/DecisionTree.java
```diff
@@ -62,21 +65,22 @@
      * @param decisionTreeLeafBuilder Decision tree leaf builder.
      */
     DecisionTree(int maxDeep, double minImpurityDecrease, StepFunctionCompressor<T> compressor,
-        DecisionTreeLeafBuilder decisionTreeLeafBuilder) {
+        DecisionTreeLeafBuilder decisionTreeLeafBuilder, boolean useIndex) {
```
zaleslaw: I think it shouldn't be a constructor parameter. This is a not a part of algorithm user settings. This is a performance improvement. Design it via setter or something else. And set the default value (true, for example). Or provide two the same realization like DT and DT on steroids.

examples/src/main/java/org/apache/ignite/examples/ml/tree/DecisionTreeClassificationTrainerExample.java
```diff
@@ -50,7 +50,7 @@ public static void main(String... args) throws InterruptedException {
                 // Create cache with training data.
                 CacheConfiguration<Integer, LabeledPoint> trainingSetCfg = new CacheConfiguration<>();
                 trainingSetCfg.setName("TRAINING_SET");
-                trainingSetCfg.setAffinity(new RendezvousAffinityFunction(false, 10));
+                trainingSetCfg.setAffinity(new RendezvousAffinityFunction(false, 1));
```
dmitrievanthony: Why did we change it on 1? It doesn't work with multiple partitions?

examples/src/main/java/org/apache/ignite/examples/ml/tree/DecisionTreeClassificationTrainerExample.java
```diff
@@ -61,7 +61,7 @@ public static void main(String... args) throws InterruptedException {
                     trainingSet.put(i, generatePoint(rnd));
 
                 // Create classification trainer.
-                DecisionTreeClassificationTrainer trainer = new DecisionTreeClassificationTrainer(4, 0);
+                DecisionTreeClassificationTrainer trainer = new DecisionTreeClassificationTrainer(4, 0, false);
```
dmitrievanthony: Looks like it makes sense to add some default behavior (use index for example).

examples/src/main/java/org/apache/ignite/examples/ml/tree/DecisionTreeRegressionTrainerExample.java
```diff
@@ -49,15 +49,15 @@ public static void main(String... args) throws InterruptedException {
                 // Create cache with training data.
                 CacheConfiguration<Integer, Point> trainingSetCfg = new CacheConfiguration<>();
                 trainingSetCfg.setName("TRAINING_SET");
-                trainingSetCfg.setAffinity(new RendezvousAffinityFunction(false, 10));
+                trainingSetCfg.setAffinity(new RendezvousAffinityFunction(false, 1));
```
dmitrievanthony: And again, why 1? Doesn't work with multiple partitions?

modules/ml/src/main/java/org/apache/ignite/ml/composition/boosting/GDBTrainer.java
```diff
@@ -165,7 +165,7 @@ public GDBTrainer(double gradStepSize, Integer cntOfIterations,
 
         try (Dataset<EmptyContext, DecisionTreeData> dataset = builder.build(
             new EmptyContextBuilder<>(),
-            new DecisionTreeDataBuilder<>(featureExtractor, lbExtractor)
+            new DecisionTreeDataBuilder<>(featureExtractor, lbExtractor, false)
```
dmitrievanthony: How do you decide when to use `false` and when `true` in these example? Is it a random?

modules/ml/src/main/java/org/apache/ignite/ml/tree/DecisionTree.java
dmitrievanthony: I like constructor to be honest. Would be great ho have constructor with default value as well.

modules/ml/src/main/java/org/apache/ignite/ml/tree/data/TreeDataIndex.java
```diff
@@ -0,0 +1,179 @@
...
+package org.apache.ignite.ml.tree.data;
+
+import org.apache.ignite.ml.tree.TreeFilter;
+
+/**
+ * Index for representing sorted dataset rows for each features.
+ * It may be reused while decision tree learning at several levels through filter method.
+ */
+public class TreeDataIndex {
+    /** Index containing IDs of rows as if they is sorted by feature values. */
```
dmitrievanthony: Spaces between fields is amazing approach to make your PR bigger :)

modules/ml/src/main/java/org/apache/ignite/ml/tree/data/TreeDataIndex.java
```diff
@@ -0,0 +1,179 @@
...
+     */
+    public TreeDataIndex(double[][] features, double[] labels) {
+        this.features = features;
+        this.labels = labels;
+
+        int rows = features.length;
+        int cols = features.length == 0 ? 0 : features[0].length;
+
+        double[][] featuresCopy = new double[rows][cols];
+        index = new int[rows][cols];
```
dmitrievanthony: I think it makes sense to use here `Arrays.copyOf` for `featuresCopy` and `Arrays.fill` for `index`. What do you think?
And do we actually need copy of features here?

modules/ml/src/main/java/org/apache/ignite/ml/tree/data/TreeDataIndex.java
```diff
@@ -0,0 +1,179 @@
...
+     * @param featureId Feature id.
+     */
+    public double featureInSortedOrder(int k, int featureId) {
+        return featuresInSortedOrder(k, featureId)[featureId];
+    }
+
+    /**
+     * Creates projection of current index in according to {@link TreeFilter}.
+     *
+     * @param filter Filter.
```
dmitrievanthony: No `@return` javadoc here and in methods above. Maybe it's fine, but looks strange.

modules/ml/src/main/java/org/apache/ignite/ml/tree/impurity/ImpurityMeasureCalculator.java
```diff
@@ -18,21 +18,71 @@
...
+import org.apache.ignite.ml.tree.data.TreeDataIndex;
 import org.apache.ignite.ml.tree.impurity.util.StepFunction;
 
 /**
  * Base interface for impurity measure calculators that calculates all impurity measures required to find a best split.
  *
  * @param <T> Type of impurity measure.
  */
-public interface ImpurityMeasureCalculator<T extends ImpurityMeasure<T>> extends Serializable {
+public abstract class ImpurityMeasureCalculator<T extends ImpurityMeasure<T>> implements Serializable {
```
dmitrievanthony: Do you think it's a good idea to remove high-level abstraction like `ImpurityMeasureCalculator` and replace it with something that has index? What about calculators without index?

It's not so important, but from design perspective looks not good.

modules/ml/src/main/java/org/apache/ignite/ml/tree/impurity/gini/GiniImpurityMeasureCalculator.java
```diff
@@ -38,54 +41,70 @@
...
-                double[] x = new double[features.length + 1];
-                GiniImpurityMeasure[] y = new GiniImpurityMeasure[features.length + 1];
+            StepFunction<GiniImpurityMeasure>[] res = new StepFunction[colsCnt];
 
-                int xPtr = 0, yPtr = 0;
+            for (int col = 0; col < res.length; col++) {
+                if(!useIndex)
+                    data.sort(col);
+
+                ArrayList<Double> x = new ArrayList<>();
```
dmitrievanthony: We know size of these arrays, don't we?

modules/ml/src/main/java/org/apache/ignite/ml/tree/impurity/util/StepFunction.java
```diff
@@ -47,6 +49,19 @@ public StepFunction(double[] x, T[] y) {
         sort(x, y, 0, x.length - 1);
     }
 
+    @SuppressWarnings("unchecked")
+    public StepFunction(ArrayList<Double> x, ArrayList<T> y, Class<T> yClass) {
+        assert x.size() == y.size() : "Argument and value arrays have to be the same length";
+
+        this.x = new double[x.size()];
+        this.y = (T[])Array.newInstance(yClass, y.size());
```
dmitrievanthony: Does it make sense to update constructor signature so that it accepts arrays and make this list to array transformation before the construction? Will look better from my perspective.

modules/ml/src/main/java/org/apache/ignite/ml/tree/randomforest/RandomForestTrainer.java
```diff
@@ -28,6 +28,8 @@
     protected final int maxDeep;
     /** Min impurity decrease. */
     protected final double minImpurityDecrease;
+    /** Use index structure instead of using sorting while decision tree learning. */
+    protected boolean useIndex = false;
```
dmitrievanthony: Somewhere it's final, somewhere it's not. Would be great to use the same approach everywhere. Or you assumed builder approach here?

zaleslaw: My 5 cents: All trainer's fields in future should be not final due to Parameter Grid feature (it requires setters calls for each tuning field).

modules/ml/src/main/java/org/apache/ignite/ml/tree/DecisionTree.java
avplatonov: okay...
I prefer Builder-like variant

zaleslaw: The story with constructors make a problem for injecting new values into trainer fields.
Of course we can pass parameters via constructors and duplicates with setters, but it would be 2-approach problem for our users.

modules/ml/src/main/java/org/apache/ignite/ml/tree/randomforest/RandomForestTrainer.java
dmitrievanthony: Wouldn't it be better to create new trainer for each Parameter Grid iteration?

examples/src/main/java/org/apache/ignite/examples/ml/tree/DecisionTreeClassificationTrainerExample.java
avplatonov: I forgot to revert after debugging. Thanks

examples/src/main/java/org/apache/ignite/examples/ml/tree/DecisionTreeClassificationTrainerExample.java
avplatonov: Yes, it is. It may be fixed after Builder-like methods and "true"-value by default.

examples/src/main/java/org/apache/ignite/examples/ml/tree/DecisionTreeRegressionTrainerExample.java
avplatonov: see above

modules/ml/src/main/java/org/apache/ignite/ml/composition/boosting/GDBTrainer.java
avplatonov: Yes, in random manner)

modules/ml/src/main/java/org/apache/ignite/ml/tree/impurity/ImpurityMeasureCalculator.java
avplatonov: leave it as is

modules/ml/src/main/java/org/apache/ignite/ml/tree/impurity/gini/GiniImpurityMeasureCalculator.java
avplatonov: thanks
I forgot revert it after first iteration of indexes

modules/ml/src/main/java/org/apache/ignite/ml/tree/impurity/util/StepFunction.java
avplatonov: I delete this, because of  "We know size of these arrays, don't we?"

modules/ml/src/main/java/org/apache/ignite/ml/tree/DecisionTree.java
```diff
@@ -89,7 +92,7 @@
      * @param dataset Dataset.
      * @return Impurity measure calculator.
      */
-    abstract ImpurityMeasureCalculator<T> getImpurityMeasureCalculator(Dataset<EmptyContext, DecisionTreeData> dataset);
+    abstract ImpurityMeasureCalculator<T> getImpurityMeasureCalculator(Dataset<EmptyContext, DecisionTreeData> dataset, boolean useIndex);
```
dmitrievanthony: We already have `useIndex` as field, do we need to specify it as parameter as well?

modules/ml/src/main/java/org/apache/ignite/ml/tree/boosting/GDBRegressionOnTreesTrainer.java
```diff
@@ -32,6 +32,8 @@
     private final int maxDepth;
     /** Min impurity decrease. */
     private final double minImpurityDecrease;
+    /** Use index structure instead of using sorting while learning. */
```
dmitrievanthony: \+ New line between field declarations.

modules/ml/src/main/java/org/apache/ignite/ml/tree/randomforest/RandomForestClassifierTrainer.java
```diff
@@ -72,6 +72,17 @@ public RandomForestClassifierTrainer(int featureVectorSize,
...
     @Override protected DatasetTrainer<DecisionTreeNode, Double> buildDatasetTrainerForModel() {
-        return new DecisionTreeClassificationTrainer(maxDeep, minImpurityDecrease);
+        return new DecisionTreeClassificationTrainer(maxDeep, minImpurityDecrease).withUseIndex(useIndex);
+    }
+
+    /**
+     * Sets useIndex parameter and returns trainer instance.
+     *
+     * @param useIndex Use index.
+     * @return Decision tree trainer
```
dmitrievanthony: \+ dot in the end.

modules/ml/src/main/java/org/apache/ignite/ml/tree/randomforest/RandomForestRegressionTrainer.java
```diff
@@ -73,6 +73,17 @@ public RandomForestRegressionTrainer(int featureVectorSize,
...
     @Override protected DatasetTrainer<DecisionTreeNode, Double> buildDatasetTrainerForModel() {
-        return new DecisionTreeRegressionTrainer(maxDeep, minImpurityDecrease);
+        return new DecisionTreeRegressionTrainer(maxDeep, minImpurityDecrease).withUseIndex(useIndex);
+    }
+
+    /**
+     * Sets useIndex parameter and returns trainer instance.
+     *
+     * @param useIndex Use index.
+     * @return Decision tree trainer
```
dmitrievanthony: \+ dot in the end.

modules/ml/src/test/java/org/apache/ignite/ml/tree/DecisionTreeClassificationTrainerTest.java
```diff
@@ -40,15 +40,20 @@
...
-    @Parameterized.Parameter
+    @Parameterized.Parameter(0)
     public int parts;
 
+    /** Use index [= 1 if true]. */
+    @Parameterized.Parameter(1)
+    public int useIndex;
 
-    @Parameterized.Parameters(name = "Data divided on {0} partitions")
+    @Parameterized.Parameters(name = "Data divided on {0} partitions. Use index = {1}.")
```
dmitrievanthony: \+ Comment.

modules/ml/src/test/java/org/apache/ignite/ml/tree/DecisionTreeRegressionTrainerTest.java
```diff
@@ -40,14 +40,20 @@
...
-    @Parameterized.Parameter
+    @Parameterized.Parameter(0)
     public int parts;
 
-    @Parameterized.Parameters(name = "Data divided on {0} partitions")
+    /** Use index [= 1 if true]. */
+    @Parameterized.Parameter(1)
+    public int useIndex;
+
+    @Parameterized.Parameters(name = "Data divided on {0} partitions. Use index = {1}.")
```
dmitrievanthony: \+ Comment.

modules/ml/src/main/java/org/apache/ignite/ml/tree/data/DecisionTreeData.java
```diff
@@ -29,17 +31,28 @@
...
+    private final boolean buildIndex;
+
     /**
      * Constructs a new instance of decision tree data.
      *
      * @param features Matrix with features.
      * @param labels Vector with labels.
      */
-    public DecisionTreeData(double[][] features, double[] labels) {
+    public DecisionTreeData(double[][] features, double[] labels, boolean buildIndex) {
```
dmitrievanthony: `buildIndex` not in list of parameters.

PR #4440, opened 2018-07-26
IGNITE-6167: Ability to enabled TLS protocols and cipher suites

mcherkasov:


modules/core/src/main/java/org/apache/ignite/ssl/SslContextFactory.java
```diff
@@ -89,6 +90,12 @@
     /** Trust managers. */
     private TrustManager[] trustMgrs;
 
+    /** Enabled cipher suites. */
+    private String[] cipherSuites;
+
+    /** Enabled cipher suites. */
```
alamar: We already have sslContextFactory in igniteConfiguration which has setProtocol().
Having protocols here is redundant and it is not obvious which one of two will be used. I suggest dropping it, using new String[] {sslContextFactory.create().getProtocol()} in its place

modules/core/src/main/java/org/apache/ignite/ssl/SslContextFactory.java
```diff
@@ -310,6 +349,18 @@ private SSLContext createSslContext() throws SSLException {
 
             SSLContext ctx = SSLContext.getInstance(proto);
 
+            if(cipherSuites != null || protocols != null) {
```
alamar: A whitespace is desired

modules/core/src/main/java/org/apache/ignite/ssl/SslContextFactory.java
```diff
@@ -310,6 +349,18 @@ private SSLContext createSslContext() throws SSLException {
...
+                SSLParameters sslParameters = new SSLParameters();
+
+                if(cipherSuites != null)
```
alamar: A whitespace is desired

modules/core/src/main/java/org/apache/ignite/ssl/SslContextFactory.java
```diff
@@ -310,6 +349,18 @@ private SSLContext createSslContext() throws SSLException {
...
+                    sslParameters.setCipherSuites(cipherSuites);
+
+                if(protocols != null)
```
alamar: A whitespace is desired

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoverySslParametersTest.java
```diff
@@ -0,0 +1,221 @@
...
+    private volatile String[] cipherSuites;
+
+    /** */
+    private volatile String[] protocols;
+
+    /** {@inheritDoc} */
+    @Override protected IgniteConfiguration getConfiguration(String gridName) throws Exception {
+        IgniteConfiguration cfg = super.getConfiguration(gridName);
+
+        SslContextFactory factory = (SslContextFactory)GridTestUtils.sslTrustedFactory("node01", "trustone");
```
alamar: Split into two lines is desired

modules/core/src/main/java/org/apache/ignite/ssl/SSLSocketFactoryWrapper.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /** */
+    private final SSLSocketFactory delegate;
+
+    /** */
+    private final SSLParameters parameters;
+
+    /** */
+    SSLSocketFactoryWrapper(SSLSocketFactory delegate,
+        SSLParameters parameters) {
```
alamar: Extra line feed

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoverySslParametersTest.java
```diff
@@ -0,0 +1,221 @@
...
+import org.apache.ignite.IgniteCheckedException;
+import org.apache.ignite.configuration.IgniteConfiguration;
+import org.apache.ignite.ssl.SslContextFactory;
+import org.apache.ignite.testframework.GridTestUtils;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+
+/**
+ * Tests cases when node connects to cluster with different set of cipher suites.
+ */
+public class TcpDiscoverySslParametersTest extends GridCommonAbstractTest {
```
alamar: Would be nice if nonexistent ciphers and protocols were tested also.

modules/core/src/main/java/org/apache/ignite/ssl/SslContextFactory.java
mcherkasov: it's not the same, the protocol field allows to set only one particular
protocol, while protocols allow you to set several different versions, for
example with protocols you can set:
TLSv1.1 and TLSv1.2, but exclude TLSv1.0

modules/core/src/main/java/org/apache/ignite/ssl/SslContextFactory.java
mcherkasov: fixed

modules/core/src/main/java/org/apache/ignite/ssl/SslContextFactory.java
mcherkasov: fixed

modules/core/src/main/java/org/apache/ignite/ssl/SslContextFactory.java
mcherkasov: fixed

modules/core/src/main/java/org/apache/ignite/ssl/SSLSocketFactoryWrapper.java
mcherkasov: fixed

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoverySslParametersTest.java
mcherkasov: added

PR #4464, opened 2018-08-01
IGNITE-8158 Wrap the method afterTestsStopped() call with try/catch b…

zzzadruga:
…lock


modules/core/src/test/java/org/apache/ignite/testframework/junits/GridAbstractTest.java
```diff
@@ -1710,11 +1712,21 @@ protected static ClassLoader getExternalClassLoader() {
...
+                    err = e;
+                }
 
-                if(isSafeTopology())
+                if (isSafeTopology()) {
                     stopAllGrids(false);
 
+                    if (stopGridErr)
+                        err = new RuntimeException("Not all Ignite instances has been stopped. " +
+                            "Please, see log for details.", err);
```
Mmuzaf: `,err` don't need here.

PR #4505, opened 2018-08-09
IGNITE-9196 Fix memory leak in MapNodeResults

dmekhanikov:


modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/CacheQueryMemoryLeakTest.java
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.ignite.cache.query.annotations.QuerySqlField;
+import org.apache.ignite.configuration.CacheConfiguration;
+import org.apache.ignite.configuration.IgniteConfiguration;
+import org.apache.ignite.internal.IgniteKernal;
+import org.apache.ignite.spi.discovery.tcp.TcpDiscoverySpi;
+import org.apache.ignite.spi.discovery.tcp.ipfinder.vm.TcpDiscoveryVmIpFinder;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+
+/** */
+public class CacheQueryMemoryLeakTest extends GridCommonAbstractTest {
```
alamar: You haven't add this test to any of suites.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/CacheQueryMemoryLeakTest.java
```diff
@@ -0,0 +1,125 @@
...
+            if (i % 1000 == 0)
+                System.out.println("Select iteration #" + i);
+
+            Query<List<?>> qry = new SqlFieldsQuery("select * from people");
+            qry.setPageSize(pageSize);
+            QueryCursor<List<?>> cursor = cache.query(qry);
+            cursor.getAll();
+            cursor.close();
+
+            assertTrue("Long JVM pause detected", srv.getLongJVMPausesTotalDuration() < 5000);
```
alamar: Here I think that peeking into some internal memory structure will be better than relying on GC, since we do not know how much RAM is available on e.g. TC agents.

PR #4529, opened 2018-08-13
IGNITE-9054 Avoid using OptimizedMarshaller with ScanQuery initial.

alamar:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/ContinuousQueryMarshallerTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+        for (Cache.Entry<Integer, MarshallerCheckingEntry> entry : cache1.query(qry)) {
+            latch.countDown();
+            System.out.println(">> Initial entry " + entry);
+        }
+
+        for (int i = 10; i < 20; i++)
+            cache1.put(i, new MarshallerCheckingEntry(i));
+
+        assert latch.await(5, TimeUnit.SECONDS) : latch.getCount();
```
slukyano: Let's use `assertTrue()` instead of `assert`.

PR #4535, opened 2018-08-14
IGNITE-8919 Wrong documentation of exec methods in StartNodeCallableImpl class

1vanan:


modules/ssh/src/main/java/org/apache/ignite/internal/util/nodestart/StartNodeCallableImpl.java
```diff
@@ -395,12 +395,12 @@ private String exec(Session ses, String cmd) throws JSchException, IOException {
     }
 
     /**
-     * Gets the value of the specified environment variable.
+     * Executes command using {@code exec} channel with setting encoding.
```
SomeFire: "a command"

modules/ssh/src/main/java/org/apache/ignite/internal/util/nodestart/StartNodeCallableImpl.java
```diff
@@ -382,11 +382,11 @@ private String env(Session ses, String name, String dflt, String encoding) throw
     }
 
     /**
-     * Gets the value of the specified environment variable.
+     * Executes command using {@code exec} channel.
```
SomeFire: "a command"

PR #4551, opened 2018-08-15
IGNITE-9228 Spark SQL Table Schema Specification

stuartmacd:


modules/spark/src/test/scala/org/apache/ignite/spark/IgniteCatalogSpec.scala
```diff
@@ -39,9 +39,9 @@ class IgniteCatalogSpec extends AbstractDataFrameSpec {
         it("Should observe all available SQL tables") {
             val tables = igniteSession.catalog.listTables.collect()
 
-            tables.length should equal (3)
+            tables.length should equal(6)
 
-            tables.map(_.name).sorted should equal (Array("CITY", "EMPLOYEE", "PERSON"))
+            tables.map(_.name).sorted should equal(Array("CITY", "EMPLOYEE", "PERSON", "PUBLIC.CITY", "PUBLIC.PERSON", "cache3.EMPLOYEE"))
```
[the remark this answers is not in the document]
stuartmacd: CITY and PERSON tables are present in the PUBLIC schema because they were created with the CREATE TABLE syntax which [according to the docs](https://apacheignite-sql.readme.io/docs/create-table) always uses the public schema, whereas the EMPLOYEE table is present in the cache3 schema because it was created as a cache using CacheConfiguration and so resides in the schema defaulting to the cache name. I'll add a comment as such in the test.

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
```diff
@@ -88,42 +89,51 @@ package object impl {
...
-            val queryEntities = ccfg.getQueryEntities
-
-            queryEntities.find(_.getTableName.equalsIgnoreCase(tabName)).map(qe ⇒ (ccfg, qe))
-        }.find(_.isDefined).flatten
+    def sqlTableInfo[K, V](ignite: Ignite, tabName: String,
+        schemaName: Option[String]): Option[(CacheConfiguration[K, V], QueryEntity)] =
+        ignite.cacheNames()
+            .map(cacheName => ignite.cache[K, V](cacheName).getConfiguration(classOf[CacheConfiguration[K, V]]))
+            .filter(
+                ccfg => schemaName.isEmpty || schemaName.get.equalsIgnoreCase(normalizeSchemaName(ccfg.getName, ccfg.getSqlSchema)))
```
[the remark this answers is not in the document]
stuartmacd: I have changed the `employeeCache2` declaration in `IgniteSQLDataFrameSpec` to specify its own schema to test this case fully.

stuartmacd:
@nizhikov thanks for the review, I have addressed your comments now, please let me know if there's anything further

modules/spark/src/test/scala/org/apache/ignite/spark/IgniteSQLDataFrameSpec.scala
```diff
@@ -284,5 +294,27 @@ class IgniteSQLDataFrameSpec extends AbstractDataFrameSpec {
             .load()
 
         personDataFrame.createOrReplaceTempView("person")
+
+        createEmployeeCache(client, "employeeCache1")
+
+        spark.read
```
nizhikov: Le'ts move creation of this data frame inside the test.
It will simplify understanding.

modules/spark/src/test/scala/org/apache/ignite/spark/IgniteSQLDataFrameSpec.scala
```diff
@@ -284,5 +294,27 @@ class IgniteSQLDataFrameSpec extends AbstractDataFrameSpec {
...
+            .format(FORMAT_IGNITE)
+            .option(OPTION_CONFIG_FILE, TEST_CONFIG_FILE)
+            .option(OPTION_TABLE, "employee")
+            .option(OPTION_SCHEMA, "employeeCache1")
+            .load()
+            .createOrReplaceTempView("employeeWithSchema")
+
+        createEmployeeCache(client, "employeeCache2", Some("employeeSchema"))
+
+        client.cache("employeeCache2").remove("key1")
```
nizhikov: Why we need this line?

modules/spark/src/test/scala/org/apache/ignite/spark/IgniteCatalogSpec.scala
```diff
@@ -39,9 +39,12 @@ class IgniteCatalogSpec extends AbstractDataFrameSpec {
         it("Should observe all available SQL tables") {
             val tables = igniteSession.catalog.listTables.collect()
 
-            tables.length should equal (3)
+            tables.length should equal(6)
 
-            tables.map(_.name).sorted should equal (Array("CITY", "EMPLOYEE", "PERSON"))
+            // each table is accessible both with and without schema; tables which were created via CREATE TABLE
+            // expressions reside in the PUBLIC schema whereas cache-created tables reside in their own schema
+            tables.map(_.name).sorted should equal(
```
nizhikov: I think we should return each table only once.
Seems, we should use `schema.name` form to be sure we can use it in all cases.

modules/spark/src/main/scala/org/apache/spark/sql/ignite/IgniteExternalCatalog.scala
```diff
@@ -144,7 +161,9 @@ private[ignite] class IgniteExternalCatalog(defaultIgniteContext: IgniteContext)
 
             val ccfg = cache.getConfiguration(classOf[CacheConfiguration[Any, Any]])
 
-            ccfg.getQueryEntities.map(_.getTableName)
+            ccfg.getQueryEntities.flatMap(
+                e => Array(e.getTableName, normalizeSchemaName(ccfg.getName, ccfg.getSqlSchema)
```
nizhikov: Let's add a test for a cache with special characters - `cache-for-table` or something similar.

modules/spark/src/test/scala/org/apache/ignite/spark/IgniteCatalogSpec.scala
stuartmacd: If we just use the `schema.name` form then we will break anyone who is currently using the catalog and depending on just the `name` form being in the tables list. The current change is backwards-compatible with current assumptions.

I'll take your advice on the best course of action here - do you want to break backwards-compatibility?

modules/spark/src/main/scala/org/apache/spark/sql/ignite/IgniteExternalCatalog.scala
stuartmacd: Looking at adding this test I realised there's an existing bug which means that table names with special characters don't work in the current codebase. I have a few changes to fix this but I need to go through most of the tests and change the SQL query assertions. Suggest I add this as part of a separate ticket otherwise this set of changes will balloon. I've created [IGNITE-9317](https://issues.apache.org/jira/browse/IGNITE-9317) for this purpose.

stuartmacd:
@nizhikov I've pushed test refactoring as requested; please see my other two comments and let me know how you want to proceed with regards to backwards compatibility of the table catalog.

modules/spark/src/main/scala/org/apache/spark/sql/ignite/IgniteExternalCatalog.scala
nizhikov: > table names with special characters don't work in the current codebase.

Thanks for bringing this up.
Please, attach a reproducer to a ticket.
Or write a test in  pull request, if you decide to contribute solution.
Let's fix it!

modules/spark/src/test/scala/org/apache/ignite/spark/IgniteCatalogSpec.scala
nizhikov: > do you want to break backwards-compatibility?
> we will break anyone who is currently using the catalog and depending on just the `name`

Yes, I understand that we will break compatibility with this change.
Personally, I like to fix it now and live with the correct solution.

Let's discuss it on dev-list, so anyone can share an opinion.

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
```diff
@@ -88,42 +89,48 @@ package object impl {
...
-            val queryEntities = ccfg.getQueryEntities
-
-            queryEntities.find(_.getTableName.equalsIgnoreCase(tabName)).map(qe ⇒ (ccfg, qe))
-        }.find(_.isDefined).flatten
+    def sqlTableInfo[K, V](ignite: Ignite, tabName: String,
+        schemaName: Option[String]): Option[(CacheConfiguration[K, V], QueryEntity)] =
+        ignite.cacheNames()
+            .map(cacheName => ignite.cache[K, V](cacheName).getConfiguration(classOf[CacheConfiguration[K, V]]))
+            .filter(
+                ccfg => schemaName.isEmpty
```
nizhikov: can be simplified to

`schemName.forall(_.equalsIgnoreCase(normalizeSchemaName(ccfg.getName, ccfg.getSqlSchema))))`

modules/spark/src/test/scala/org/apache/ignite/spark/IgniteSQLDataFrameSpec.scala
```diff
@@ -270,6 +270,42 @@ class IgniteSQLDataFrameSpec extends AbstractDataFrameSpec {
...
+                .option(OPTION_SCHEMA, "employeeCache1")
+                .load()
+                .createOrReplaceTempView("employeeWithSchema")
+
+            // `employeeCache2` is created with a custom schema of `employeeSchema`
+            createEmployeeCache(client, "employeeCache2", Some("employeeSchema"))
+
+            // remove a value from `employeeCache2` so that we know whether the select statement picks up the
+            // correct cache, ie. it should now have 2 values compared to 3 in `employeeCache1`
+            client.cache("employeeCache2").remove("key1")
```
nizhikov: Let's move this line to specific test where it results are used.
Also, let's check table size before and after removing a key.

modules/spark/src/test/scala/org/apache/ignite/spark/AbstractDataFrameSpec.scala
```diff
@@ -117,9 +117,12 @@ abstract class AbstractDataFrameSpec extends FunSpec with Matchers with BeforeAn
         cache.query(qry.setArgs(4L.asInstanceOf[JLong], "St. Petersburg")).getAll
     }
 
-    def createEmployeeCache(client: Ignite, cacheName: String): Unit = {
+    def createEmployeeCache(client: Ignite, cacheName: String, schemaName: Option[String] = None): Unit = {
         val ccfg = AbstractDataFrameSpec.cacheConfiguration[String, Employee](cacheName)
 
+        if (schemaName.isDefined)
```
nizhikov: Let's change it to `schemaName.foreach(ccfg.setSqlSchema(_))`

modules/spark/src/test/scala/org/apache/ignite/spark/IgniteCatalogSpec.scala
```diff
@@ -39,9 +39,12 @@ class IgniteCatalogSpec extends AbstractDataFrameSpec {
         it("Should observe all available SQL tables") {
             val tables = igniteSession.catalog.listTables.collect()
 
-            tables.length should equal (3)
+            tables.length should equal(6)
 
-            tables.map(_.name).sorted should equal (Array("CITY", "EMPLOYEE", "PERSON"))
+            // each table is accessible both with and without schema; tables which were created via CREATE TABLE
```
nizhikov: Comments should start with capital letter and end with a dot.

nizhikov:
@stuartmacd Overall looks good. I made some minor comments.

The only thing we should discuss on the dev-list is how we should form tables list.
Please, send a mail with your proposal to the ticket topic.

Also, when PR be a fully ready, please, run tests on [Ignite Team City server](https://ci.ignite.apache.org/project.html?projectId=IgniteTests24Java8) and provide a link to it in the ticket.
You may want to run following test plans:
  * [RDD] (https://ci.ignite.apache.org/viewType.html?buildTypeId=IgniteTests24Java8_Rdd)
  * [Scala examples](https://ci.ignite.apache.org/viewType.html?buildTypeId=IgniteTests24Java8_ScalaExamples)

modules/spark/src/main/scala/org/apache/spark/sql/ignite/IgniteExternalCatalog.scala
stuartmacd: I'll fix it - I have assigned the ticket to myself.

stuartmacd:
@nizhikov Thanks, I have sent a note to the dev list - please do add your opinion there!

stuartmacd:
@nizhikov all tables created through CREATE TABLE need to be in the PUBLIC schema (which is default on CREATE TABLE), Ignite throws an exception if you try to do anything else. See [the Ignite docs](https://apacheignite-sql.readme.io/docs/create-table)

nizhikov:
> CREATE TABLE need to be in the PUBLIC schema

Yes, I checked it and this is true. Sorry for bad advice.

Looks like in `save` mode we should allow to specify `OPTION_SCHEMA` only for an `Append` or when table already created. 
If `OPTION_SCHEMA` is defined and table doesn't exists or mode is `Overwrite` we should throw an exception.

Also, we need to add proper explanation to the scaladoc and ignite documentation.

What do you think?

modules/spark/src/test/scala/org/apache/ignite/spark/IgniteCatalogSpec.scala
```diff
@@ -39,9 +40,30 @@ class IgniteCatalogSpec extends AbstractDataFrameSpec {
...
 
-            tables.length should equal (3)
+            tables.length should equal(3)
 
-            tables.map(_.name).sorted should equal (Array("CITY", "EMPLOYEE", "PERSON"))
+            tables.map(_.name).sorted should equal(Array("CITY", "EMPLOYEE", "PERSON"))
+        }
+
+        it("Should use the database context when providing tables")
+        {
```
nizhikov: This brace should be on the same line as method declaration.

nizhikov:
Hello, @stuartmacd 

Overall looks good.

1. Why we use `DEFAULT_DATABASE` as a default name in the External catalog?
If it schema name we should use default schema - `QueryUtils.DFLT_SCHEMA`.
Is it makes sense for you?

2. Let's add some tests for a new "change DB" in catalog feature.
The test should list all tables from one schema and then from another.

3. Should we add some warning or even throw an exception if the user tries to specify `OPTION_SCHEMA` in write mode? I propose to throw an exception in that case.
We need a test for that too.

stuartmacd:
@nizhikov thanks for the review. Please find my follow-ups below.

> Why we use DEFAULT_DATABASE as a default name in the External catalog?
> If it schema name we should use default schema - QueryUtils.DFLT_SCHEMA.
> Is it makes sense for you?

We use `DEFAULT_DATABASE` as a signifier that by default we will search all schemas for the given table unless a schema other than `DEFAULT_DATABASE` has been specified. If we were to default to `PUBLIC` then we would only look for tables in that schema (or if we treated `PUBLIC` like we do `DEFAULT_DATABASE` then there would be no way for a caller to specifically ask for a table from `PUBLIC`). Let me know if that doesn't make sense.

> Let's add some tests for a new "change DB" in catalog feature.
> The test should list all tables from one schema and then from another.

Agreed - I have expanded my test in `IgniteCatalogSpec` to switch between two different databases.

> Should we add some warning or even throw an exception if the user tries to specify OPTION_SCHEMA in write mode? I propose to throw an exception in that case.
> We need a test for that too.

I think this should only be an error if we are creating a new table in non-PUBLIC schema; I have added a check and test for this. If we are in write mode and not creating the table or if we are creating a table and the schema has been specified as PUBLIC, this should be valid as far as I can see - do you agree?

nizhikov:
> I think this should only be an error if we are creating a new table in non-PUBLIC schema

Yes. Thanks for the clarification.

nizhikov:
> We use DEFAULT_DATABASE as a signifier that by default we will search all schemas for the given table unless a schema other than DEFAULT_DATABASE has been specified.

OK. I agree. 

I make a review in a day.

Thank you, @stuartmacd !

nizhikov:
@stuartmacd 

> We use DEFAULT_DATABASE as a signifier that by default we will search all schemas for the given table unless a schema other than DEFAULT_DATABASE has been specified.

I thought more about your proposal.
What if a user wants to browse tables from schema `default` which is equal to constant `DEFAULT_DATABASE`. Looks like we should use an empty string to denote "all schemas" case.

What do you think?

modules/spark/src/test/scala/org/apache/ignite/spark/IgniteSQLDataFrameWriteSpec.scala
```diff
@@ -332,6 +332,22 @@ class IgniteSQLDataFrameWriteSpec extends AbstractDataFrameSpec {
...
+                personDataFrame.write
+                    .format(FORMAT_IGNITE)
+                    .option(OPTION_CONFIG_FILE, TEST_CONFIG_FILE)
+                    .option(OPTION_TABLE, "nonexistant-table-name")
+                    .option(OPTION_CREATE_TABLE_PRIMARY_KEY_FIELDS, "id")
+                    .option(OPTION_SCHEMA, "mySchema")
+                    .save()
+            }
+
+            assert(ex.getMessage ==
```
nizhikov: Let's use `assertEquals` here.

modules/spark/src/test/scala/org/apache/ignite/spark/IgniteSQLDataFrameSpec.scala
```diff
@@ -270,6 +271,44 @@ class IgniteSQLDataFrameSpec extends AbstractDataFrameSpec {
...
+                .load()
+                .createOrReplaceTempView("employeeWithSchema")
+
+            // `employeeCache2` is created with a custom schema of `employeeSchema`.
+            createEmployeeCache(client, "employeeCache2", Some("employeeSchema"))
+
+            // Remove a value from `employeeCache2` so that we know whether the select statement picks up the
+            // correct cache, ie. it should now have 2 values compared to 3 in `employeeCache1`.
+            Iterators.size(client.cache("employeeCache2").iterator()) should equal(3)
+            val removed = client.cache("employeeCache2").remove("key1")
```
nizhikov: nit: empty line here.

modules/spark/src/test/scala/org/apache/ignite/spark/IgniteSQLDataFrameSpec.scala
```diff
@@ -270,6 +271,44 @@ class IgniteSQLDataFrameSpec extends AbstractDataFrameSpec {
...
+            removed shouldBe true
```
nizhikov: Can we just make it
`client.cache("employeeCache2").remove("key1") shouldBe true`?

nizhikov:
@stuartmacd Overall looks very good.
Let's come to a consensus with the default database name.

Just a friendly reminder: October 10 is code freeze date for a 2.7 release. 
Please, ping me when you address my comments.

modules/spark/src/main/scala/org/apache/ignite/spark/impl/QueryHelper.scala
```diff
@@ -105,6 +111,7 @@ private[apache] object QueryHelper {
       */
     def saveTable(data: DataFrame,
         tblName: String,
+        schemaName: Option[String],
```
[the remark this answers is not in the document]
stuartmacd: The schema name isn't necessary in the Ignite `INSERT` statement. The schema name gets passed through to `savePartition()` and then is applied to the data streamer separately from the SQL statement. So yes it will work if we have two tables with the same name in different schemas.

stuartmacd:
@nizhikov thanks for continuing the review. I don't think we have a lot of option with the `DEFAULT_DATABASE` naming, as it is used internally by Spark whenever you do not specify the name of the schema in Spark SQL. Ie. you can run the first test case in `IgniteCatalogSpec` and see that Spark passes `default` as the schema name through to the `listTables()` operation - so I believe this is a Spark limitation as I'm not aware of any way to change the value of `DEFAULT_DATABASE`.

modules/spark/src/main/scala/org/apache/spark/sql/ignite/IgniteExternalCatalog.scala
```diff
@@ -271,9 +259,7 @@ private[ignite] class IgniteExternalCatalog(defaultIgniteContext: IgniteContext)
 
     /** @inheritdoc */
     override protected def doCreateTable(tableDefinition: CatalogTable, ignoreIfExists: Boolean): Unit = {
-        val ignite = igniteOrDefault(tableDefinition.identifier.database.getOrElse(DEFAULT_DATABASE), default)
-
-        igniteSQLTable(ignite, tableDefinition.identifier.table) match {
+        igniteSQLTable(ignite, tableDefinition.identifier.table, tableDefinition.identifier.database) match {
```
nizhikov: We should check the schema name in `tableDefinition` here.
If the schema name is not `PUBLIC` we should throw.

stuartmacd: Good point - that's done now, with appropriate tests added.

PR #4558, opened 2018-08-16
Ignite 9053

anton-vinogradov:


modules/core/src/main/java/org/apache/ignite/internal/processors/continuous/GridContinuousProcessor.java
```diff
@@ -1165,7 +1168,35 @@ public void addNotification(UUID nodeId,
...
+
+                            break;
+                        }
+
+                        LT.warn(log, "Failed to wait for ack message. [node=" + nodeId +
+                            "routine=" + routineId + "]");
+                    }
+                }
+
+                assert fut.isDone();
```
EdShangGG: Please, add a message for assert, at least print the future itself.

modules/core/src/main/java/org/apache/ignite/internal/processors/continuous/GridContinuousProcessor.java
```diff
@@ -1165,7 +1168,35 @@ public void addNotification(UUID nodeId,
...
+                                    "Node left grid after receiving, but before processing the message [node=" +
+                                        nodeId + "]");
+
+                                fut0.onDone(err);
+                            }
+
+                            break;
+                        }
+
+                        LT.warn(log, "Failed to wait for ack message. [node=" + nodeId +
```
EdShangGG: GridDhtPartitionsExchangeFuture#nextDumpTimeout

Please, use the same approach (or generalize this one) to not spam with the log.

PR #4570, opened 2018-08-18
IGNITE-9301: Support method compute WithNoResultCache in .Net

alexzaitzev:
Realized method WithNoResultCache in .NET client


alexzaitzev:
@alamar it's possible yes, but I'm a bit concerned that we are writing component tests for every minor change. It will lead to the situation with tests as in the Java code. Moreover .NET code just calls java function, which is tested on their side.

alexzaitzev:
I absolutely agree with you, that features must be covered with the tests, and, as you can see, I already wrote a test for this feature. But my point is that writing both unit and component tests for the every code change is not right. Especially in this case.

PR #4577, opened 2018-08-20
IGNITE-8251 Reduce testPageEviction run time

udaykale:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/GridCachePreloadingEvictionsSelfTest.java
```diff
@@ -219,22 +218,14 @@ private void sleepUntilCashesEqualize(final Ignite ignite1, final Ignite ignite2
...
-            if (cache1.size(ALL_PEEK_MODES) != cache2.size(ALL_PEEK_MODES)) {
-                U.warn(log, "Sizes do not match (will retry in 1000 ms) [s1=" + cache1.size(ALL_PEEK_MODES) +
-                    ", s2=" + cache2.size(ALL_PEEK_MODES) + ']');
-
-                U.sleep(1000);
-            }
-            else
-                break;
-        }
+        // Sleeping to allow the cache sizes to settle down.
```
dspavlov: I'm not sure I agree here about the replacement of conditional iterative sleep with unconditional. Why will it be faster?

udaykale: The initial code was causing this test to fail sometimes. After debugging I figured out that when the eviction was only partially completed, the conditional  check may return false and exit the loop. This happens because the cache sizes may match even when the eviction was not completed. So when the code reaches the assertion part and the eviction is still partially done the test would fail. Hence, I made the test sleep for 3s to allow sufficient time for eviction to complete to avoid false positives during the conditional check.

PR #4592, opened 2018-08-22
IGNITE-9054 Avoid using OptimizedMarshaller with initial ScanQuery.

alamar:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryResponseEntry.java
```diff
@@ -27,6 +27,8 @@
 /**
  * Class to store query results returned by remote nodes. It's required to fully
  * control serialization process. Local entries can be returned to user as is.
+ * <p>
+ * @deprecated Should be removed in Apache Ignite 3.0.
  */
 public class GridCacheQueryResponseEntry<K, V> implements Map.Entry<K, V>, Externalizable {
```
slukyano: Please add `@Deprecated` annotation.

alamar:
@slukyano done!

PR #4600, opened 2018-08-23
Ignite 9274

ygerzhedovich:


modules/core/src/main/java/org/apache/ignite/events/CacheEvent.java
```diff
@@ -90,10 +90,14 @@
...
     private Object key;
 
-    /** Event ID. */
+    /** Transaction Id. */
     @GridToStringInclude
     private final IgniteUuid xid;
 
+    /** Transaction label. */
+    @GridToStringInclude
+    private String txLb;
```
AMashenkov: Let's rename this to "txLbl".

ygerzhedovich: You are right. Thanks. Fixed

PR #4602, opened 2018-08-23
Ignite 9340

dspavlov:


modules/rest-http/src/main/java/org/apache/ignite/internal/processors/rest/protocols/http/jetty/GridJettyRestProtocol.java
```diff
@@ -65,11 +65,11 @@
...
 
-            p.setProperty("org.eclipse.jetty.LEVEL", "WARN");
-            p.setProperty("org.eclipse.jetty.util.log.LEVEL", "OFF");
-            p.setProperty("org.eclipse.jetty.util.component.LEVEL", "OFF");
+            System.setProperty("org.eclipse.jetty.LEVEL", "WARN");
+            System.setProperty("org.eclipse.jetty.util.log.LEVEL", "OFF");
+            System.setProperty("org.eclipse.jetty.util.component.LEVEL", "OFF");
 
-            StdErrLog.setProperties(p);
+            //todo StdErrLog.setProperties(p);
```
slukyano: I guess we don't have a rule for that, but I believe that it is a good practice for TODOs to have JIRA issue numbers.

On the other hand, I've went over Jetty docs (https://www.eclipse.org/jetty/documentation/9.4.x/configuring-logging.html) and it seems that using system properties should be fine. If someone have issues with how Ignite handles Jetty logging, they can set IGNITE_JETTY_LOG_NO_OVERRIDE and configure it however they want.

I suggest to remove the `p` and the TODO, just leave the system properties setting.

There is also the log4j reflection voodoo below which looks suspicious to me, but let's not fix what isn't broken.

dspavlov:
@slukyano I've updated code according to your suggestions. Thank you for your efforts. Is patch looking good now?

PR #4604, opened 2018-08-23
IGNITE-8971: make GridRestProcessor propagate error message

macrergate:


modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
```diff
@@ -349,7 +351,11 @@
                             .a(", params=").a(tskReq.params());
                     }
 
-                    sb.a(", err=").a(e.getMessage() != null ? e.getMessage() : e.getClass().getName()).a(']');
+                    sb.a(", err=")
```
dspavlov: Unconditional error propagation is not the best solution from the security point of view.

It provides remote attacker more information about the internals of code. It should be covered by option not enabled for production mode.

PR #4629, opened 2018-08-28
IGNITE-9382 Fixed data handling for multiple chunks of data

tyabuki:


modules/platforms/nodejs/lib/internal/ClientSocket.js
```diff
@@ -196,27 +197,40 @@ class ClientSocket {
...
+            }
+            else {
+                this._buffer = MessageBuffer.from(message, 0);
+            }
+
             // Response length
-            const length = buffer.readInteger();
+            const length = this._buffer.readInteger();
             offset += length + BinaryUtils.getSize(BinaryUtils.TYPE_CODE.INTEGER);
+            if (this._buffer.length !== offset) {
```
ekaterina-nbl: In fact, "message" may contain several responses. In this case this condition is always true and we will not process the responses.
Let me please think and implement the full solution.

tyabuki: I'm thinking of changing `this._buffer.length !== offset` to  `this._buffer.length < offset` .
Is it right?

ekaterina-nbl: That's not enough

ekaterina-nbl: I suppose to support several different responses in the message
- local "offset" variable can not be used, new "_offset" property should be added to ClientSocket class instead
- _buffer initialization (or concatenation) should be placed before while loop

So _processResponse implementation could look something like below.
Could you please review and test it?

    async _processResponse(message) {
        if (this._state === STATE.DISCONNECTED) {
            return;
        }
        if (this._buffer) {
            this._buffer.concat(message);
            this._buffer.position = this._offset;
        }
        else {
            this._buffer = MessageBuffer.from(message, 0);
        }
        while (this._buffer && this._offset < this._buffer.length) {
            // Response length
            const length = this._buffer.readInteger() + BinaryUtils.getSize(BinaryUtils.TYPE_CODE.INTEGER);
            if (this._buffer.length < this._offset + length) {
                break;
            }
            this._offset += length;

            let requestId, isSuccess;
            const isHandshake = this._state === STATE.HANDSHAKE;

            if (isHandshake) {
                // Handshake status
                isSuccess = (this._buffer.readByte() === HANDSHAKE_SUCCESS_STATUS_CODE)
                requestId = this._handshakeRequestId.toString();
            }
            else {
                // Request id
                requestId = this._buffer.readLong().toString();
                // Status code
                isSuccess = (this._buffer.readInteger() === REQUEST_SUCCESS_STATUS_CODE);
            }

            this._logMessage(requestId, false, this._buffer.data);

            const buffer = this._buffer;
            if (this._offset === this._buffer.length) {
                this._buffer = null;
                this._offset = 0;
            }

            if (this._requests.has(requestId)) {
                const request = this._requests.get(requestId);
                this._requests.delete(requestId);
                if (isHandshake) {
                    await this._finalizeHandshake(buffer, request, isSuccess);
                }
                else {
                    await this._finalizeResponse(buffer, request, isSuccess);
                }
            }
            else {
                throw Errors.IgniteClientError.internalError('Invalid response id: ' + requestId);
            }
        }
    }

tyabuki: @ekaterina-nbl 
Thanks!
I tested with your code, and it works good with large data as well.
I added this changes.

ekaterina-nbl: Thanks! Let's merge it.

PR #4635, opened 2018-08-28
IGNITE-9149 Get rid of logging remaining supplier nodes rebalance time

ololo3000:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1002,7 +995,7 @@ private String demandRoutineInfo(int topicId, UUID supplier, GridDhtPartitionSup
         private final IgniteLogger log;
 
         /** Remaining. T2: startTime, partitions */
```
Mmuzaf: @ololo3000 I think the JavaDoc should also be fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1123,7 +1116,7 @@ private void cancel(UUID nodeId) {
 
                 U.log(log, ("Cancelled rebalancing [grp=" + grp.cacheOrGroupName() +
                     ", supplier=" + nodeId + ", topVer=" + topologyVersion() +
-                    ", time=" + (U.currentTimeMillis() - remaining.get(nodeId).get1()) + " ms]"));
+                    "]"));
```
Mmuzaf: @ololo3000 I think we don't need line break here. Also, according to [LineBreaking](https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines#CodingGuidelines-LineBreaking) we should use single quotes for the `]` character.

PR #4639, opened 2018-08-29
IGNITE-9388: mesos IgniteProvider tries to access obsolete ignite.run…

shroman:
… or downloads from slow archive.


modules/mesos/src/main/java/org/apache/ignite/mesos/resource/IgniteProvider.java
```diff
@@ -17,29 +17,52 @@
...
 
+    // This constants are set by maven-ant-plugin.
     /** */
     private static final String DOWNLOAD_URL_PATTERN = "https://archive.apache.org/dist/ignite/%s/apache-ignite-fabric-%s-bin.zip";
 
+    /** URL for request Ignite latest version. */
+    private final static String IGNITE_LATEST_VERSION_URL = "https://ignite.apache.org/latest";
+
+    /** Mirrors. */
+    private static final String APACHE_MIRROR_URL = "https://www.apache.org/dyn/closer.cgi?as_json=1";
```
alamar: I really don't like the duplication of this link here & in the pom.xml. Can we just have placeholders here for everything other than DOWNLOAD_URL_PATTERN? Also we don't need IGNITE_LATEST_VERSION_URL, we need *current* Ignite version. If we're on 2.5 we don't need to download 2.7. It should come from pom.xml.

modules/mesos/src/main/java/org/apache/ignite/mesos/resource/IgniteProvider.java
```diff
@@ -52,40 +75,93 @@ public IgniteProvider(String downloadFolder) {
...
      */
     public String downloadIgnite(String ver) throws IOException {
         assert ver != null;
 
         URL url;
 
+        // get the latest version.
         if (ver.equals(ClusterProperties.DEFAULT_IGNITE_VERSION)) {
-            URL updateUrl = new URL(DOWNLOAD_LINK);
+            ver = findLatestVersion();
```
alamar: We really need try {} around all this logic, if catch() then log error then use DOWNLOAD_URL_PATTERN.

modules/mesos/pom.xml
```diff
@@ -36,8 +37,10 @@
 
     <properties>
         <mesos.version>0.22.0</mesos.version>
-        <ignite.latest.url>http://ignite.run/download_ignite.php</ignite.latest.url>
-        <ignite.direct.url>https://archive.apache.org/dist/ignite/%s/apache-ignite-fabric-%s-bin.zip</ignite.direct.url>
+        <ignite.archive.url>https://archive.apache.org/dist/ignite/%s/apache-ignite-fabric-%s-bin.zip</ignite.archive.url>
+        <ignite.version.url>https://ignite.apache.org/latest</ignite.version.url>
```
alamar: Should use build version and not the latest one. Don't know what to do for snapshot builds.

modules/mesos/src/main/java/org/apache/ignite/mesos/resource/IgniteProvider.java
shroman: Agree on the duplication. Will remove it from the code (leaving only in pom.xml).

Sorry, what do you mean by "we need current Ignite version"? There's no such a thing...
At start, the framework will download the latest Ignite version by default. If a user needs a different version, he/she can specify it (and it'll be downloaded from the archives).

modules/mesos/src/main/java/org/apache/ignite/mesos/resource/IgniteProvider.java
shroman: @alamar Agree, good point.

modules/mesos/pom.xml
shroman: @alamar Sorry, I don't quite understand your concern. https://ignite.apache.org/latest will not give you the snapshot or nightly builds, just the latest stable one, AFAIK.

Can you please explain?

modules/mesos/src/main/java/org/apache/ignite/mesos/resource/IgniteProvider.java
alamar: How does it work currently?

modules/mesos/pom.xml
```diff
@@ -20,7 +20,8 @@
 <!--
     POM file.
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
alamar: Do you have to change XML header?

modules/mesos/src/main/java/org/apache/ignite/mesos/resource/IgniteProvider.java
```diff
@@ -17,28 +17,51 @@
...
  * Class downloads and stores Ignite.
  */
 public class IgniteProvider {
+    /** Logger. */
+    private static final Logger log = Logger.getLogger(IgniteProvider.class.getSimpleName());
+
     // This constants are set by maven-ant-plugin.
     /** */
-    private static final String DOWNLOAD_LINK = "http://ignite.run/download_ignite.php";
+    private static final String DOWNLOAD_URL_PATTERN = "";
```
alamar: Still, maybe I will keep this one as https://archive.apache.org/dist/ignite/%s/apache-ignite-fabric-%s-bin.zip, access it in case that it's not possible to detect latest version - if something changes on Apache side. We should be able to access current version as a version of ignite-mesos JAR. I'm just worried that we're changing one fragile thing (ignite.run) to another fragile thing (apache mirror JSON API).

modules/mesos/src/main/java/org/apache/ignite/mesos/resource/IgniteProvider.java
shroman: @alamar Currently (before the fix), 
1. if a user specifies the latest version (which is default) it tries to download the latest stable version from http://ignite.run/, which was discarded several versions ago (and therefore fails.)
2. Otherwise it uses the version specified by the user from the archives.
Please note, that a user can specify a URL for his/her custom builds with IGNITE_PACKAGE_URL.

Basically, I preserve this behavior, just fixed the download path.

modules/mesos/pom.xml
shroman: @alamar Well, ignite-mesos 2.x has no dependencies on Ignite, therefore it doesn't care what version to run -- nothing will be broken. It's just a driver, and the user decides what version to run.
Looking at ignite-mesos 2.x may be misleading ;)

modules/mesos/pom.xml
shroman: @alamar No, but Intellij simplified it and it works. Do you think I have to revert it?

modules/mesos/src/main/java/org/apache/ignite/mesos/resource/IgniteProvider.java
shroman: @alamar Totally share your concerns on the fragility.
But anyway, user will see errors if Apache changes something, and can specify the version (from the "latest" to "2.7.0" for example) to download it from the archives. The url is in pom.

shroman: I'll put it back, in case it's needed for development.

modules/mesos/src/main/java/org/apache/ignite/mesos/resource/IgniteProvider.java
```diff
@@ -94,10 +94,15 @@ public String downloadIgnite(String ver) throws IOException {
...
+            try {
+                ver = findLatestVersion();
 
-            // and try to retrieve from a mirror.
-            url = new URL(String.format(findMirror() + IGNITE_PATH, ver, ver));
+                // and try to retrieve from a mirror.
+                url = new URL(String.format(findMirror() + IGNITE_PATH, ver, ver));
+            }
+            catch (Exception e) {
+                throw new RuntimeException("Failed to find URL for " + ClusterProperties.DEFAULT_IGNITE_VERSION, e);
```
alamar: Can we still perhaps do a warning and `url = new URL(String.format(DOWNLOAD_URL_PATTERN, ver.replace("-incubating", ""), LAST_KNOWN_VERSION));` here?

modules/mesos/pom.xml
alamar: I'd revert it since all other pom.xmls seem to have same header line.

modules/mesos/pom.xml
```diff
@@ -76,23 +85,39 @@
...
+                                <replaceregexp byline="true" encoding="UTF-8">
+                                    <regexp pattern="(.*IGNITE_LATEST_VERSION_URL = &quot;)(.*)(&quot;.*)"/>
+                                    <substitution expression="\1${ignite.version.url}\3"/>
+                                    <fileset dir="${basedir}/">
+                                        <include name="**/IgniteProvider.java"/>
+                                    </fileset>
+                                </replaceregexp>
+
+                                <replaceregexp byline="true" encoding="UTF-8">
+                                    <regexp pattern="(.*DOWNLOAD_URL_PATTERN = &quot;)(.*)(&quot;.*)"/>
```
alamar: We can remove this substitution now, I guess. Maybe with ${ignite.archive.url} along. Let's DRY.

shroman: @alamar Agree. Done.

modules/mesos/pom.xml
shroman: @alamar Agree. Done.

modules/mesos/src/main/java/org/apache/ignite/mesos/resource/IgniteProvider.java
shroman: @alamar The exception will be caught by `ResourceProvider#init` and its contents will go to logs as SEVERE level -- I think we don't need another warning here.

As for 'incubating', the project is graduated from the incubator many releases ago. I don't think we need it anymore ;)

alamar:
Can you please also rebase it?

modules/mesos/src/main/java/org/apache/ignite/mesos/resource/IgniteProvider.java
```diff
@@ -52,40 +75,99 @@ public IgniteProvider(String downloadFolder) {
...
+            try {
+                ver = findLatestVersion();
 
-            HttpURLConnection conn = (HttpURLConnection)updateUrl.openConnection();
+                // and try to retrieve from a mirror.
+                url = new URL(String.format(findMirror() + IGNITE_PATH, ver, ver));
+            }
+            catch (Exception e) {
+                // fallback to archive.
+                url = new URL(String.format(DOWNLOAD_URL_PATTERN, ver, ver));
```
alamar: If findLatestVersion() fails then we will try to download https://archive.apache.org/dist/ignite/latest/apache-ignite-fabric-latest-bin.zip which is 404. We should somehow push the latest known version from pom.xml to source in order to fall back to it.

modules/mesos/src/main/java/org/apache/ignite/mesos/resource/IgniteProvider.java
```diff
@@ -52,40 +75,98 @@ public IgniteProvider(String downloadFolder) {
...
+                // and try to retrieve from a mirror.
+                url = new URL(String.format(findMirror() + IGNITE_PATH, ver, ver));
+            }
+            catch (Exception e) {
+                throw new RuntimeException("Failed to find URL for " + ClusterProperties.DEFAULT_IGNITE_VERSION, e);
+            }
+        }
+        else {
+            // or from archive.
+            url = new URL(String.format(DOWNLOAD_URL_PATTERN, ver, ver));
```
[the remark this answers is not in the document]
shroman: @alamar Unfortunately no, because mirrors usually have only the latest version.

modules/mesos/src/main/java/org/apache/ignite/mesos/resource/IgniteProvider.java
shroman: @alamar I doubt fetching from pom.xml is very clean and improve much.

We can do
```
          catch (Exception e) {
                if (ver.equalsIgnoreCase(ClusterProperties.DEFAULT_IGNITE_VERSION))
                    throw new RuntimeException("Failed to identify the latest version.");
                
                // fallback to archive.
                url = new URL(String.format(DOWNLOAD_URL_PATTERN, ver, ver));
            }
```
Just let's fail (it's not critical) if nothing else can be done -- the user will notice and specify the version explicitly to download from the archive or from a custom URL.

PR #4664, opened 2018-09-03
IGNITE-8987 Ignite hangs during getting of atomic structure after…

gromtech:
…autoactivation

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/IgnitePersistentStoreDataStructuresTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/IgnitePersistentStoreDataStructuresTest.java
@@ -17,6 +17,7 @@
 
 package org.apache.ignite.internal.processors.cache.persistence;
 
+import java.util.concurrent.TimeUnit;
 import org.apache.ignite.Ignite;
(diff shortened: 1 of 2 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/IgnitePersistentStoreDataStructuresTest.java
```diff
@@ -43,6 +46,9 @@
     /** */
     private static final TcpDiscoveryIpFinder ipFinder = new TcpDiscoveryVmIpFinder(true);
 
+    /** */
+    private static volatile boolean autoActivationEnabled = false;
```
alamar: I don't think you really need volatile here. Nodes are started from single thread aren't they?

gromtech: @alamar , thank you for the review. `beforeTest` and `getConfiguration` methods can be called by different threads.

alamar: If it's the case it should be a major eye-opener since I'm sure I've seen a few tests which may fail due to this.

PR #4684, opened 2018-09-05
IGNITE-9298

deviljelly:
Adds support for SSL to CommandHandler (control.sh)

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
@@ -49,6 +49,7 @@ import org.apache.ignite.internal.client.GridClientHandshakeException;
 import org.apache.ignite.internal.client.GridClientNode;
 import org.apache.ignite.internal.client.GridServerUnreachableException;
 import org.apache.ignite.internal.client.impl.connection.GridClientConnectionResetException;
+import org.apache.ignite.internal.client.ssl.GridSslBasicContextFactory;
 import org.apache.ignite.internal.commandline.cache.CacheArguments;
 import org.apache.ignite.internal.commandline.cache.CacheCommand;
 import org.apache.ignite.internal.processors.cache.verify.CacheInfo;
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -1882,6 +2002,39 @@ public int execute(List<String> rawArgs) {
                 clientCfg.setSecurityCredentialsProvider(
                     new SecurityCredentialsBasicProvider(new SecurityCredentials(args.user(), args.password())));
             }
+            
+            if (args.ssl()) {
+
+                 
+				GridSslBasicContextFactory factory = new GridSslBasicContextFactory();
```
alamar: bad formatting

modules/core/src/test/java/org/apache/ignite/internal/commandline/CommandHandlerParsingTest.java
```diff
@@ -61,6 +61,7 @@
      * validate_indexes command arguments parsing and validation
      */
     public void testValidateIndexArguments() {
+
```
alamar: not needed here, also see below

modules/core/src/test/java/org/apache/ignite/internal/commandline/CommandHandlerParsingTest.java
```diff
@@ -166,10 +168,112 @@ public void testExperimentalCommandIsDisabled() {
...
+            try {
+                hnd.parseAndValidate(asList("--truststore-type"));
+
+                fail("expected exception: Expected truststore type");
+            }
+            catch (IllegalArgumentException e) {
+                e.printStackTrace();
+            }
+            
+            Arguments args = hnd.parseAndValidate(asList("--keystore", "testKeystore", "--keystore-password", "testKeystorePassword", "--keystore-type", "testKeystoreType",
```
alamar: I don't think you should be testing inner workings that much

PR #4687, opened 2018-09-05
Revert TOPIC_MONGO removal in "IGNITE-9361 Remove IgniteInternalCache…

alamar:
….isMongo*Cache() and other remnants."

This partially reverts commit de30a86e740067ed239602fb5dc28396144fd446.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/GridTopic.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/GridTopic.java
@@ -81,6 +81,12 @@ public enum GridTopic {
     /** */
     TOPIC_CONTINUOUS,
 
+    /**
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/ignite/internal/GridTopic.java
```diff
@@ -81,6 +81,12 @@
     /** */
     TOPIC_CONTINUOUS,
 
+    /**
+     * @deprecated Should be removed in Apache Ignite 3.0.
```
AMashenkov: "Not used anymore. Must not be removed until Apache Ignite 3.0 release to preserve binary protocol compatibility."

PR #4697, opened 2018-09-06
IGNITE-9298 control.sh does not support SSL

a-polyakov:
Signed-off-by: a-polyakov <polyakov.alexandr.alexandrovich@gmail.com>

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
@@ -49,6 +49,7 @@ import org.apache.ignite.internal.client.GridClientHandshakeException;
 import org.apache.ignite.internal.client.GridClientNode;
 import org.apache.ignite.internal.client.GridServerUnreachableException;
 import org.apache.ignite.internal.client.impl.connection.GridClientConnectionResetException;
+import org.apache.ignite.internal.client.ssl.GridSslBasicContextFactory;
 import org.apache.ignite.internal.commandline.cache.CacheArguments;
 import org.apache.ignite.internal.commandline.cache.CacheCommand;
 import org.apache.ignite.internal.processors.cache.verify.CacheInfo;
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -1861,6 +1974,36 @@ public int execute(List<String> rawArgs) {
                     new SecurityCredentialsBasicProvider(new SecurityCredentials(args.user(), args.password())));
             }
 
+            if (args.isSslEnable()){
```
alamar: whitespace desired

modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -1861,6 +1974,36 @@ public int execute(List<String> rawArgs) {
...
+                GridSslBasicContextFactory factory = new GridSslBasicContextFactory();
+
+                factory.setProtocol(args.getSslProtocol());
+                factory.setKeyAlgorithm(args.getSslAlgorithm());
+
+                if (args.getSslKeyStorePath()==null)
```
alamar: whitespace desired before and after ==, also see below

modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
alamar: Can you move all the following into a separate method? It becomes too long otherwise.

modules/core/src/test/java/org/apache/ignite/util/GridCommandHandlerSslTest.java
```diff
@@ -0,0 +1,90 @@
...
+ * Command line handler test.
+ */
+public class GridCommandHandlerSslTest extends GridCommonAbstractTest {
+
+    /** {@inheritDoc} */
+    @Override protected void beforeTest() throws Exception {
+        cleanPersistenceDir();
+
+        stopAllGrids();
+
```
alamar: line feed not needed, also  see below

modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -161,6 +162,35 @@
...
+    // SSL configuration section
+
+    /** */
+    protected static final String CMD_SSL_ENABLED = "--ssl_enabled";
+
+    /** */
+    protected static final String CMD_SSL_PROTOCOL = "--ssl_protocol";
+
+    /** */
+    protected static final String CMD_SSL_ALGORITHM = "--ssl_algorithm";
```
alamar: Other parameters use "-" as words delimiter which is preferred to "_". It's also more UNIX convenient.

PR #4763, opened 2018-09-14
Ignite 9599 mx bean

a-polyakov:


modules/core/src/main/java/org/apache/ignite/configuration/DataStorageConfiguration.java
```diff
@@ -957,6 +969,20 @@ public DataStorageConfiguration setWalCompactionEnabled(boolean walCompactionEna
...
+     */
+    public int getWalCompactionLevel() {
+        return walCompactionLevel;
+    }
+
+    /**
+     * @param walCompactionLevel New ZIP level to WAL compaction.
+     */
+    public void setWalCompactionLevel(int walCompactionLevel) {
+        this.walCompactionLevel = walCompactionLevel;
```
andrey-kuznetsov: I'm a bit late to the party, but I'm curious: why don't your return `this` like other setters?

PR #4789, opened 2018-09-19
Add TypeScript support to web console frontend

Klaster1:


Klaster1:
@akuznetsov-gridgain @nva can you take a look at this?

PR #4808, opened 2018-09-21
IGNITE-9665: fix for RPM build

nizhikov:


nizhikov:
@pavel-kuznetsov Thanks! It's a typo. IGNITE-96*6*5 is correct ticket ID.

PR #4820, opened 2018-09-24
IGNITE-9676 Spring Session integration implemented.

antkr:


ambition-consulting:
Looks like a really useful extension. Is there a reason why this does not get merged?

thaingo:
May I know if there is any progress with this integration?

PR #4822, opened 2018-09-25
IGNITE-9675: Deadlock on Ignite:active() and stopping grid simultaneously calling

avplatonov:


modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskProcessor.java
```diff
@@ -187,7 +188,16 @@ private IgniteClientDisconnectedCheckedException disconnectedError(@Nullable Ign
     /** {@inheritDoc} */
     @SuppressWarnings("TooBroadScope")
     @Override public void onKernalStop(boolean cancel) {
-        lock.writeLock();
+        while(true) {
+            try {
+                if (lock.tryWriteLock(1, TimeUnit.SECONDS))
+                    break;
+                else
+                    Thread.sleep(1000);
```
akuznetsov-gridgain: May be it make sense to U.warn("Await on lock while node stopping") ?

modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskProcessor.java
```diff
@@ -187,7 +188,16 @@ private IgniteClientDisconnectedCheckedException disconnectedError(@Nullable Ign
...
+            } catch (InterruptedException e) {
+                U.warn(log, e);
```
akuznetsov-gridgain: May be make sense to print a warning "Await for lock on node stop was interrupted"?

modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskProcessor.java
avplatonov: maybe you're right
thanks

modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskProcessor.java
avplatonov: maybe you're right
thanks

PR #4827, opened 2018-09-25
IGNITE-9418 initialize cache persistent storage concurrently.

zstan:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStoreManager.java
```diff
@@ -1010,6 +1017,38 @@ private IgniteCheckedException shutdown(PageStore store, boolean cleanFile, Igni
...
+                Integer cacheId = F.firstValue(gDesc.caches());
+
+                if (cacheId != null) {
+                    DynamicCacheDescriptor cacheDesc = cctx.cache().cacheDescriptor(cacheId);
+
+                    if (CU.isPersistentCache(cacheDesc.cacheConfiguration(), cctx.gridConfig().getDataStorageConfiguration())) {
+                        try {
+                            holder0 = initForCache(cacheDesc.groupDescriptor(), cacheDesc.cacheConfiguration());
+                        } catch (IgniteCheckedException e) {
+                            throw new IgniteException(e);
```
sergey-chugunov-1985: What will happen if exception is thrown here? Previously when we initialized everything beforehand we had a clear path: node fails to restore memory and shuts down.

Now with this change we don't even have a checked exception to catch, it is wrapped into unchecked.
Should we call FailureHandler directly in this case if something goes wrong during lazy initialization of cache store?

zstan: I  change a bit FilePageStoreManager#initDir cached exceptions, hope it would be enough.

PR #4859, opened 2018-09-27
IGNITE-9726: GridCacheAbstractFailoverSelfTest may lock all suite on put/remove cache operations

avplatonov:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheAbstractFailoverSelfTest.java
```diff
@@ -228,8 +228,9 @@ protected void testConstantTopologyChange(@Nullable final TransactionConcurrency
             }
         }, TOP_CHANGE_THREAD_CNT, "topology-change-thread");
 
+        boolean isInterrupted = false;
```
SomeFire: Add empty line between variable declaration and try block, because they are different semantic units.

avplatonov: Thanks a lot

PR #4872, opened 2018-09-28
IGNITE-1436: Build C++ API on the Mac

sdarlington:
Creating PR to 'advertise' that it's being worked on. Tasks still to complete:

* Test on Linux / Windows to make sure I've not broken anything
* Tests and ODBC driver don't currently build

Test will have to wait until Boost libraries have been updated to something more modern (IGNITE-8310).


sdarlington:
Updated to build against (current) master.

I'm not completely sure how to run tests... but I can run the Ignite node successfully.

One bit that might break on Linux/Windows: in a number of places it refers to `libboost-thread` but only `libboost-thread-mt` exists on my machine. I'm unclear whether this is something unique to the Mac, old versions of Boost or something else.

sdarlington:
Closing. We don't need this now we have CMake build scripts (#7956)

PR #4905, opened 2018-10-03
IGNITE-9719 Extra rebalanceThreadPoolSize check on client node

luchnikovbsk:


modules/core/src/test/java/org/apache/ignite/client/ClientConfigurationTest.java
```diff
@@ -73,4 +82,49 @@ public void testSerialization() throws IOException, ClassNotFoundException {
 
         assertTrue(Comparers.equal(target, desTarget));
     }
+
+    /**
+     * Test check the case when {@link IgniteConfiguration#getRebalanceThreadPoolSize()} is equal
+     * to {@link IgniteConfiguration#getSystemThreadPoolSize()}
+     */
+    @Test
```
zstan: need to check log message correcty printed, check GridStringLogger usage.

PR #4906, opened 2018-10-03
IGNITE-9606: JDBC getPrimaryKeys() returns wrong value for COLUMN_NAME

pavel-kuznetsov:


modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
```diff
@@ -571,8 +571,9 @@ public static void processBinaryMeta(GridKernalContext ctx, QueryEntity qryEntit
         for (Map.Entry<String, String> entry : fields.entrySet()) {
             Boolean isKeyField;
 
-            if (isKeyClsSqlType) // We don't care about keyFields in this case - it might be null, or empty, or anything
-                isKeyField = false;
+            if (isKeyClsSqlType)
+                // In case of field is neither composite nor wrapped, it is a key, if it equal to keyFieldName.
+                isKeyField = entry.getKey().equals(qryEntity.getKeyFieldName());
             else
                 isKeyField = (hasKeyFields ? keyFields.contains(entry.getKey()) : null);
```
pavel-kuznetsov: Could someone explain, when we need to set `isKeyField` to `null`

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
```diff
@@ -571,8 +571,9 @@ public static void processBinaryMeta(GridKernalContext ctx, QueryEntity qryEntit
         for (Map.Entry<String, String> entry : fields.entrySet()) {
             Boolean isKeyField;
 
-            if (isKeyClsSqlType) // We don't care about keyFields in this case - it might be null, or empty, or anything
-                isKeyField = false;
+            if (isKeyClsSqlType)
+                // In case of field is neither composite nor wrapped, it is a key, if it equal to keyFieldName.
+                isKeyField = entry.getKey().equals(qryEntity.getKeyFieldName());
```
ygerzhedovich: is it correct that you check it case sensitive?

pavel-kuznetsov:
Afrer review, decieded to fix problem in another way 
see https://github.com/apache/ignite/pull/4948

PR #4972, opened 2018-10-12
IGNITE-9809 Web console: more minor fixes

Klaster1:
1. Update confirmation E2E selectors so these work when multiple modals are opened.
2. Update TestCafe to 0.22.0.
3. Update reduxDevtoolsIntegration.js so it doesn't crash when there's a Promise in action payload.
4. Fix form-field postfix position in browsers that don't support [place-self](https://developer.mozilla.org/en-US/docs/Web/CSS/place-self) (Edge and Safari).


Klaster1:
@nva please review.

PR #4996, opened 2018-10-15
IGNITE-9830

ascherbakoff:


modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryUtils.java
```diff
@@ -1036,8 +1056,14 @@ public static BinaryMetadata mergeMetadata(@Nullable BinaryMetadata oldMeta, Bin
             Collection<BinarySchema> mergedSchemas = new HashSet<>(oldMeta.schemas());
 
             for (BinarySchema newSchema : newMeta.schemas()) {
-                if (mergedSchemas.add(newSchema))
+                if (mergedSchemas.add(newSchema)) {
                     changed = true;
+
+                    if (changedSchemas != null) {
+                        for (BinarySchema schema : newMeta.schemas())
+                            changedSchemas.add(schema.schemaId());
```
sergey-chugunov-1985: As far as I understand semantics of a `changedSchemas` set we want it to contain only schemas newMeta has but oldMeta doesn't.

But in this piece of code I see that all schemas from newMeta will be added to the set anyway.

To me it looks like a copy-paste bug (the same code was added above where it is fine to do such thing).

PR #5008, opened 2018-10-17
adding hibernate-5.3 module

scottmf:
IGNITE-9893 IGNITE-1757 adding hibernate-5.3 module

This change is based on the hibernate-5.1 example with fixes to tests
that I ran into during developement

updates to the module:
1. added ability to allow for caches to be lazily loaded and created as a result
       of post-bootstrap scanning.
2. added org.apache.ignite.hibernate.verify_atomicity - relaxes the cache
       atomicMode check
3. HibernateCacheProxy - now accepts a Supplier which lazily loads the
       Ignite caches after the SPI is initialized
4. added org.apache.ignite.hibernate.cache_prefix - prepends all cache names
       when a cache is looked up in ignite
5. removed dflt cache as ignite 2.x does not allow mixed classes in one cache

All tests now pass as a result of fixing a couple bugs including IGNITE-1757


scottmf:
fyi - I'm seeing flakey tests after rebasing from ignite origin master on HibernateL2CacheTransactionSelfTest.testQueryCache().  It looks like it is only failing during tests using TRANSACTIONAL atomicity.  It is very difficult to isolate the error.

PR #5010, opened 2018-10-17
IGNITE-5565 replace Cron4J with Spring task scheduler

macrergate:


modules/schedule/src/main/java/org/apache/ignite/internal/processors/schedule/CronExpression.java
```diff
@@ -0,0 +1,69 @@
+/*
+ *
+ *  * Licensed to the Apache Software Foundation (ASF) under one or more
```
akuznetsov-gridgain: It seems a copy-pase bug with license header. you have extra "*" chars.

modules/schedule/src/main/java/org/apache/ignite/internal/processors/schedule/SpringScheduler.java
```diff
@@ -0,0 +1,129 @@
+/*
+ *
+ *  * Licensed to the Apache Software Foundation (ASF) under one or more
+ *  * contributor license agreements.  See the NOTICE file distributed with
```
akuznetsov-gridgain: Same issue with "*".

modules/schedule/src/test/java/org/apache/ignite/internal/processors/schedule/GridScheduleSelfTest.java
```diff
@@ -120,7 +121,7 @@ public void testScheduleRunnable() throws Exception {
 
         SchedulerFuture<?> fut = null;
 
-        long freq = 60; // 1 minute frequency.
+        long freq = 2; // 2 seconds frequency.
```
akuznetsov-gridgain: May be it will be better to stay old tests as-is to make sure new implementation works the same way?
And add new tests?

modules/schedule/src/main/java/org/apache/ignite/internal/processors/schedule/ScheduleFutureImpl.java
```diff
@@ -90,7 +86,7 @@
     private int callCnt;
 
     /** De-schedule flag. */
-    private final AtomicBoolean descheduled = new AtomicBoolean(false);
+    private final AtomicBoolean descheduled = new AtomicBoolean(true);
```
akuznetsov-gridgain: Could you explain why this flag is 'true' now?

parent/pom.xml
```diff
@@ -59,7 +59,6 @@
         <commons.codec.version>1.11</commons.codec.version>
         <commons.collections.version>3.2.2</commons.collections.version>
         <commons.lang.version>2.6</commons.lang.version>
-        <cron4j.version>2.2.5</cron4j.version>
```
akuznetsov-gridgain: You should also change ignite/pom.xml 
and move <module>modules/schedule</module> from lgpl profile
and add to <modules>

modules/schedule/src/test/java/org/apache/ignite/internal/processors/schedule/GridScheduleSelfTest.java
macrergate: Agree. It looks we don't need a new test here.

modules/schedule/src/main/java/org/apache/ignite/internal/processors/schedule/ScheduleFutureImpl.java
macrergate: it was a bug, I believe. This flag reflects was the task actually scheduled by the underlying Scheduler or not. Initially task is not scheduled and it will be scheduled in org.apache.ignite.internal.processors.schedule.ScheduleFutureImpl#schedule0

And it can be delayed as we are supporting delayed scheduling (
org.apache.ignite.internal.processors.schedule.ScheduleFutureImpl#delay)

Now I've renamed flag to reverse name scheduled, it's a bit easier to understand I  suppose.

assembly/dependencies-apache-ignite.xml
```diff
@@ -130,7 +130,6 @@
                 <exclude>org.apache.ignite:ignite-hibernate-core</exclude>
                 <exclude>org.apache.ignite:ignite-hibernate_4.2</exclude>
                 <exclude>org.apache.ignite:ignite-hibernate_5.1</exclude>
-                <exclude>org.apache.ignite:ignite-schedule</exclude>
```
alamar: Why?

examples/README.txt
```diff
@@ -27,7 +27,7 @@ LGPL
 =====
 LGPL examples can be activated by turning lgpl profile on.
 
-lgpl profile required some lgpl-based libs, for example: ignite-hibernate & ignite-schedule.
```
alamar: Can you please replace it with " & ignite-geospatial"?

modules/core/src/main/java/org/apache/ignite/IgniteScheduler.java
```diff
@@ -40,11 +40,11 @@
  * to all nodes five times, once every minute, with initial delay of two seconds:
  * <pre name="code" class="java">
  * Ignition.ignite().scheduler().scheduleLocal(
- *     GridSchedulerFuture&lt;?&gt; = Ignition.ignite().scheduler().scheduleLocal(new Callable&lt;Object&gt;() {
+ *     SchedulerFuture&lt;?&gt; = Ignition.ignite().scheduler().scheduleLocal(new Callable&lt;Object&gt;() {
  *         &#64;Override public Object call() throws IgniteCheckedException {
  *             g.broadcast(new GridCallable() {...}).get();
  *         }
- *     }, "{2, 5} * * * * *" // 2 seconds delay with 5 executions only.
+ *     }, "{2, 5} * * * * * ?" // 2 seconds delay with 5 executions only.
```
alamar: What does this question mark mean? Can you explain it in javadoc?

modules/osgi-karaf/src/main/resources/features.xml
```diff
@@ -21,9 +21,9 @@
     <feature name="ignite-all" version="${project.version}" description="Apache Ignite :: All">
         <details>
             <![CDATA[Aggregate feature for installing all Apache Ignite module + their dependencies.
-            
+
```
alamar: Can you please revert all changes in this file other than the relevant one, avoid showing large diff?

modules/osgi-karaf/src/main/resources/features.xml
```diff
@@ -237,7 +239,6 @@
             <![CDATA[The Apache Ignite Schedule module + dependencies. This module is a fragment of ignite-core.]]>
         </details>
         <feature prerequisite="true">wrap</feature>
-        <bundle start="true" dependency="true">wrap:mvn:it.sauronsoftware.cron4j/cron4j/${cron4j.version}$Bundle-SymbolicName=cron4j&amp;Bundle-Version=${cron4j.version}</bundle>
```
alamar: This is the only relevant diff line I guess

modules/schedule/src/test/java/org/apache/ignite/internal/processors/schedule/GridScheduleSelfTest.java
```diff
@@ -243,9 +244,11 @@ public void testScheduleRunnable() throws Exception {
     public void testScheduleCallable() throws Exception {
         SchedulerFuture<Integer> fut = null;
 
-        long freq = 60; // 1 minute frequency.
+        long freq = 1; // 1 second frequency.
```
alamar: Can you avoid changing existing test, add a new test instead, to be able to check both formats?

macrergate: I changed this just to speed up the test as with Spring we can schedule with seconds.
There is another test with a minute frequency.

assembly/dependencies-apache-ignite.xml
macrergate: It is needed to add ignite-schedule module to nolgpl buid. Peter Ivanov suggested this change.

modules/core/src/main/java/org/apache/ignite/IgniteScheduler.java
macrergate: Actually it is equivalent to '*' (no specific meaning except it is only allowed in the day of month and day of week fields), so I better to change it back to '*'.

macrergate: Also I fixed expression to satisfy description and add a line that seconds can be omitted.

modules/osgi-karaf/src/main/resources/features.xml
macrergate: Sure.

parent/pom.xml
macrergate: done.

modules/osgi-karaf/src/main/resources/features.xml
alamar: It's OK, I'll do the rest before merge to master

PR #5019, opened 2018-10-18
IGNITE-9308 Added baseline topology command to REST API.

nva:


modules/clients/src/test/java/org/apache/ignite/internal/processors/rest/JettyRestProcessorBaselineSelfTest.java
```diff
@@ -0,0 +1,196 @@
...
+     * @return Collection of consistentIds.
+     */
+    private static Collection<String> consistentIdMapper(@Nullable Collection<? extends BaselineNode> nodes) {
+        return Optional.ofNullable(nodes).orElseGet(Collections::emptyList)
+            .stream().map(n -> String.valueOf(n.consistentId()))
+            .collect(toSet());
+    }
+
+    /**
+     * @param content Content to check.
```
akuznetsov-gridgain: baselineSz & srvSz are missing in javadoc

modules/clients/src/test/java/org/apache/ignite/internal/processors/rest/JettyRestProcessorBaselineSelfTest.java
```diff
@@ -0,0 +1,196 @@
...
+        assertEqualsCollections(consistentIdMapper(grid(0).cluster().nodes()), new HashSet<>(baseline.getServers()));
+    }
+
+    /**
+     * @throws Exception If failed.
+     */
+    public void testBaseline() throws Exception {
+        assertBaseline(content(null, GridRestCommand.BASELINE_CURRENT_STATE), gridCount(), gridCount());
+
+        stopGrid(gridCount() - 1);
```
akuznetsov-gridgain: May be add a comment like this:
// Stop one node. It will stay in baseline.

modules/clients/src/test/java/org/apache/ignite/internal/processors/rest/JettyRestProcessorBaselineSelfTest.java
```diff
@@ -0,0 +1,196 @@
...
+        assertEqualsCollections(consistentIdMapper(grid(0).cluster().currentBaselineTopology()), new HashSet<>(baseline.getBaseline()));
+        assertEquals(srvsSz, baseline.getServers().size());
+        assertEqualsCollections(consistentIdMapper(grid(0).cluster().nodes()), new HashSet<>(baseline.getServers()));
+    }
+
+    /**
+     * @throws Exception If failed.
+     */
+    public void testBaseline() throws Exception {
+        assertBaseline(content(null, GridRestCommand.BASELINE_CURRENT_STATE), gridCount(), gridCount());
```
akuznetsov-gridgain: May be add local variable: sz = gridCount() ? for readability ?

modules/clients/src/test/java/org/apache/ignite/internal/processors/rest/JettyRestProcessorBaselineSelfTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+        stopGrid(gridCount() - 1);
+        assertBaseline(content(null, GridRestCommand.BASELINE_CURRENT_STATE), gridCount(), gridCount() - 1);
+
+        startGrid(gridCount() - 1);
```
akuznetsov-gridgain: May be add a comment like this:
// Start one node. It is expected that it will be auto-added to base line.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandHandler.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    /**
+     * Set new baseline.
+     *
+     * @param baselineTop Initial collection of nodes.
+     * @param consistentIds Collection of consistent IDs to set.
+     */
+    private void set(Set<BaselineNode> baselineTop, List<Object> consistentIds) {
+        Map<Object, BaselineNode> srvrs = ctx.grid().cluster().forServers().nodes().stream()
```
akuznetsov-gridgain: Let's refactor to local variable "cluster = ctx.grid().cluster()"

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandHandler.java
```diff
@@ -0,0 +1,199 @@
...
+        for (Object consistentId : consistentIds) {
+            BaselineNode node = srvrs.get(consistentId);
+
+            if (node == null)
+                throw new IllegalStateException("Node not found for consistent ID: " + consistentId);
+
+            baselineTop.add(node);
+        }
+
+        ctx.grid().cluster().setBaselineTopology(baselineTop);
```
akuznetsov-gridgain: use local variable "cluster" from here too.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandResponse.java
```diff
@@ -0,0 +1,160 @@
...
+    private Collection<String> baseline;
+
+    /** Current server nodes. */
+    private Collection<String> srvs;
+
+    /**
+     * @param nodes Nodes to process.
+     * @return Collection of consistentIds.
+     */
+    private static Collection<String> mapper(Collection<? extends BaselineNode> nodes) {
```
akuznetsov-gridgain: let's rename this methid from "mapper" to "consistentIds" ?

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/request/GridRestBaselineRequest.java
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public void consistentIds(List<Object> consistentIds) {
+        this.consistentIds = consistentIds;
+    }
+
+    /** {@inheritDoc} */
+    @Override public String toString() {
+        return S.toString(GridRestBaselineRequest.class, this, super.toString());
+    }
+}
```
akuznetsov-gridgain: missing empty line in the end of file

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestCommand.java
```diff
@@ -176,6 +176,18 @@
     /** */
     CLUSTER_CURRENT_STATE("currentstate"),
 
+    /** */
+    BASELINE_CURRENT_STATE("baseline"),
+
+    /** */
+    BASELINE_SET("baselineset"),
```
ibessonov: According to what I see in this class we should probably use names like "baselinestate", "setbaseline", "addbaseline" and "removebaseline".

modules/rest-http/src/main/java/org/apache/ignite/internal/processors/rest/protocols/http/jetty/GridJettyRestHandler.java
```diff
@@ -759,6 +761,20 @@ else if (cmd == CLUSTER_ACTIVE || cmd == CLUSTER_ACTIVATE)
...
             }
 
+            case BASELINE_CURRENT_STATE:
+            case BASELINE_SET:
+            case BASELINE_ADD:
+            case BASELINE_REMOVE: {
+                GridRestBaselineRequest restReq0 = new GridRestBaselineRequest();
+
+                restReq0.topologyVersion(longValue("topVer", params, null));
+                restReq0.consistentIds(values(null, "consistentId", params));
```
ibessonov: This looks confusing. Param named "consistentId" is actually the collection of ids. I propose to rename it.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandResponse.java
```diff
@@ -0,0 +1,160 @@
...
+     * @param active Cluster state.
+     * @param topVer Current topology version.
+     * @param baseline Current baseline nodes.
+     * @param srvs Current server nodes.
+     */
+    public GridBaselineCommandResponse(
+        boolean active,
+        long topVer,
+        Collection<? extends BaselineNode> baseline,
+        Collection<? extends BaselineNode> srvs) {
```
ibessonov: Minor code style comment - following formatting would look better:
```
    ...
    Collection<? extends BaselineNode> srvs
) {
    this.active = active;
    ...
```

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandResponse.java
```diff
@@ -0,0 +1,160 @@
...
+
+    /** Current server nodes. */
+    private Collection<String> srvs;
+
+    /**
+     * @param nodes Nodes to process.
+     * @return Collection of consistentIds.
+     */
+    private static Collection<String> consistentIds(Collection<? extends BaselineNode> nodes) {
+        return nodes.stream().map(n -> String.valueOf(n.consistentId())).collect(toList());
```
ibessonov: How likely "n.consistentId()" will be null? I doubt that it's possible.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandHandler.java
```diff
@@ -0,0 +1,201 @@
...
+                throw new IllegalStateException("Node not found for consistent ID: " + consistentId);
+
+            baselineTop.add(node);
+        }
+
+        cluster.setBaselineTopology(baselineTop);
+    }
+
+    /** {@inheritDoc} */
+    private GridBaselineCommandResponse handleAsyncUnsafe(GridRestRequest req) throws IgniteCheckedException {
```
ibessonov: /** {@inheritDoc} */ can't be applied to method that doesn't override anything.
Anyway, I don't understand the name of the method:
- it's not async, as it's suggested;
- why it's unsafe?

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandHandler.java
```diff
@@ -0,0 +1,201 @@
...
+     * Set new baseline.
+     *
+     * @param baselineTop Initial collection of nodes.
+     * @param consistentIds Collection of consistent IDs to set.
+     */
+    private void set(Set<BaselineNode> baselineTop, List<Object> consistentIds) {
+        IgniteClusterEx cluster = ctx.grid().cluster();
+        
+        Map<Object, BaselineNode> srvrs = cluster.forServers().nodes().stream()
+            .collect(toMap(n -> String.valueOf(n.consistentId()), identity()));
```
ibessonov: Same question about nullability of consistent ids. "valueOf" returns "null" as a String for empty strings, I think it's not what we want.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandHandler.java
```diff
@@ -0,0 +1,201 @@
...
+
+        for (Object consistentId : consistentIds) {
```
ibessonov: Input collection should be immutable as I proposed earlier. This is the place where you should create new HashSet and fill it with nodes.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandHandler.java
```diff
@@ -0,0 +1,201 @@
...
+            case BASELINE_SET: {
+                Long topVer = req0.topologyVersion();
+
+                if (topVer == null && consistentIds == null)
+                    throw new IgniteCheckedException("Failed to handle request (either topVer or consistentIds should be specified).");
+
+                if (topVer != null)
+                    cluster.setBaselineTopology(topVer);
+                else
+                    set(new HashSet<>(), consistentIds);
```
ibessonov: Please use emptySet() here.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandResponse.java
akuznetsov-gridgain: @ibessonov  It is not to avoid "null", but because consistentId declared as Object, but REST can operate only with Strings.

modules/rest-http/src/main/java/org/apache/ignite/internal/processors/rest/protocols/http/jetty/GridJettyRestHandler.java
ibessonov: I believe that you actually used "consistentIds" in other places, like in missingParameter("consistentIds").

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandHandler.java
```diff
@@ -0,0 +1,201 @@
...
+                set(currentBaseLine(), consistentIds);
+
+                break;
+            }
+
+            case BASELINE_REMOVE: {
+                if (consistentIds == null)
+                    throw new IgniteCheckedException(missingParameter("consistentIds"));
+
+                Map<Object, BaselineNode> baseline = currentBaseLine().stream()
```
ibessonov: BASELINE_ADD invoked separate method. It would look convenient to make the same thing for all operations.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandHandler.java
```diff
@@ -0,0 +1,201 @@
...
+                    .collect(toMap(n -> String.valueOf(n.consistentId()), identity()));
+
+                for (Object consistentId : consistentIds) {
+                    BaselineNode node = baseline.remove(consistentId);
```
ibessonov: This code is dangerous. Does Collectors.toMap guarantees you a mutable object by default?
I mean, it does return HashMap as far as I know, but is it specified in specification or just an artifact of the implementation?

ibessonov: Another weird thing is that you use "String.valueOf(n.consistentId())" to create map but use "Object consistentId" to search in that map. Doesn't look consistent even if it works fine.

modules/clients/src/test/java/org/apache/ignite/internal/processors/rest/JettyRestProcessorBaselineSelfTest.java
```diff
@@ -0,0 +1,202 @@
...
+        assertEqualsCollections(consistentIdMapper(grid(0).cluster().currentBaselineTopology()), new HashSet<>(baseline.getBaseline()));
+        assertEquals(srvsSz, baseline.getServers().size());
+        assertEqualsCollections(consistentIdMapper(grid(0).cluster().nodes()), new HashSet<>(baseline.getServers()));
+    }
+
+    /**
+     * @throws Exception If failed.
+     */
+    public void testBaseline() throws Exception {
+        int sz = gridCount();
```
ibessonov: This variable is cool, please use it in other tests here as well!

modules/clients/src/test/java/org/apache/ignite/internal/processors/rest/JettyRestProcessorBaselineSelfTest.java
```diff
@@ -0,0 +1,202 @@
...
+        JsonNode res = node.get("response");
+
+        assertFalse(res.isNull());
+
+        GridBaselineCommandResponse baseline = JSON_MAPPER.treeToValue(res, GridBaselineCommandResponse.class);
+
+        assertTrue(baseline.isActive());
+        assertEquals(grid(0).cluster().topologyVersion(), baseline.getTopologyVersion());
+        assertEquals(baselineSz, baseline.getBaseline().size());
+        assertEqualsCollections(consistentIdMapper(grid(0).cluster().currentBaselineTopology()), new HashSet<>(baseline.getBaseline()));
```
ibessonov: Will it work without additional "new HashSet"? If it will then "new HashSet" should be removed to make code simpler.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandResponse.java
ibessonov: I see that String is required. I was wondering about "String.valueOf" usage instead of simple "*.toString()".

modules/rest-http/src/main/java/org/apache/ignite/internal/processors/rest/protocols/http/jetty/GridJettyRestHandler.java
nva: "consistent If" it's not a parameter name, its parameter prefix for passing multiple values through REST. Look at the "values" method.

ibessonov: Now I see, thank you!

modules/clients/src/test/java/org/apache/ignite/internal/processors/rest/JettyRestProcessorBaselineSelfTest.java
nva: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandHandler.java
```diff
@@ -0,0 +1,201 @@
...
+        return new GridBaselineCommandResponse(cluster.active(), cluster.topologyVersion(), currentBaseLine(), srvrs);
+    }
+
+    /**
+     * Set new baseline.
+     *
+     * @param baselineTop Initial collection of nodes.
+     * @param consistentIds Collection of consistent IDs to set.
+     */
+    private void set(Set<BaselineNode> baselineTop, List<Object> consistentIds) {
```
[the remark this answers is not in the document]
nva: This method was removed.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandHandler.java
```diff
@@ -0,0 +1,201 @@
...
+            if (log.isDebugEnabled())
+                log.debug("Handled task REST request: " + req);
+        }
+    }
+
+    /**
+     * @return Current baseline.
+     */
+    private Set<BaselineNode> currentBaseLine() {
+        return Optional.ofNullable(ctx.grid().cluster().currentBaselineTopology())
```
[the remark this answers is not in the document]
nva: I agree.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandHandler.java
nva: Removed this method

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cluster/GridBaselineCommandResponse.java
nva: I checked and can not find the case when it can be null.

PR #5027, opened 2018-10-18
IGNITE-9843 If metadata is not available due to reconnect..,

alamar:
…return reconnect future wrapped in CacheException.


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -1878,7 +1878,17 @@ UpdateResult runDistributedUpdate(
                     }
 
                     @Override public Cache.Entry<K, V> next() {
-                        List<?> l = iter0.next();
+                        List<?> l;
+
+                        try {
+                            l = iter0.next();
+                        }
+                        catch (Exception e) {
```
slukyano: A more idiomatic way to do this would be
```suggestion
                        catch (CacheException e) {
                            throw e;
                        }
                        catch (Exception e) {
                            throw new CacheException(e);
                        }
```

PR #5078, opened 2018-10-25
IGNITE-9937 Primary response error can be lost due to unwrapping a key

gromtech:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/atomic/GridNearAtomicAbstractUpdateFuture.java
```diff
@@ -401,7 +401,7 @@ final void onPrimaryError(GridNearAtomicAbstractUpdateRequest req, GridNearAtomi
         Collection<Object> keys = new ArrayList<>(keys0.size());
 
         for (KeyCacheObject key : keys0)
-            keys.add(cctx.cacheObjectContext().unwrapBinaryIfNeeded(key, keepBinary, false));
+            keys.add(cctx.cacheObjectContext().unwrapBinaryIfNeeded(key, true, false));
```
alamar: I suggest `keys.add(cctx.toCacheKeyObject(key));` instead

gromtech: @alamar , thanks for the review. This is a good suggestion. Since we already have a collection of KeyCacheObject probably it makes sense to put that as a parameter of the exception `err.add(keys,... )` (all keys will be copied into `CachePartialUpdateCheckedException.failedKeys`). What do you think?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/atomic/GridDhtAtomicCache.java
```diff
@@ -2734,9 +2735,22 @@ else if (GridDhtCacheEntry.ReaderId.contains(readers, nearNode.id())) {
...
-                    continue;
+                if (storeErr != null) {
+                    Object key = entry.key();
+
+                    try {
+                        key = entry.key().value(ctx.cacheObjectContext(), false);
+                    }
+                    catch (BinaryInvalidTypeException e) {
+                        if (log.isDebugEnabled()) {
+                            log.warning("Failed to get a value of the key: " + key +
```
alamar: Without log throttle it is a real possibility to generate gigabytes of logs per minute.

I don't think that this is sound approach.

PR #5079, opened 2018-10-25
IGNITE-9990 Control.sh utility should request a password if necessary.

a-polyakov:
Signed-off-by: a-polyakov <polyakov.alexandr.alexandrovich@gmail.com>

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
@@ -2495,9 +2495,8 @@ public class CommandHandler {
 
                         tryConnectMaxCount--;
                     }
-                    else {
+                    else
                         throw e;
-                    }
                 }
             }
             return EXIT_CODE_OK;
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -1936,50 +1932,101 @@ public int execute(List<String> rawArgs) {
...
+                            System.out.println("password: ");
+
+                            args.setPassword(scanner.next());
+                        }
+
+                        tryConnectAgain = true;
 
+                        tryConnectMaxCount--;
+                    }
+                    else {
```
voropava: braces

modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -1936,50 +1932,101 @@ public int execute(List<String> rawArgs) {
...
+
+                        case CACHE:
+                            cache(client, args.cacheArgs());
+
+                            break;
+
+                        case WAL:
+                            wal(client, args.walAction(), args.walArguments());
+
+                            break;
```
voropava: please add default throwing exception

a-polyakov: implemented in parseAndValidate

PR #5189, opened 2018-10-28
IGNITE-9828: Continuous query failover.

rkondakov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -4304,7 +4304,7 @@ public void awaitLastFut() {
                             AffinityTopologyVersion awaitVer = new AffinityTopologyVersion(
                                 topVer.topologyVersion() + 1, 0);
 
-                            ctx.shared().exchange().affinityReadyFuture(awaitVer).get();
+                            ctx.shared().exchange().affinityReadyFuture(awaitVer).get(); // TODO ??
```
pavlukhin: Is TODO still relevant?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/PartitionUpdateCounter.java
```diff
@@ -180,14 +181,24 @@ private void offer(Item item) {
...
 
-        if (last != null)
-            update(last.start + last.delta);
+            // Call closure for each counter between current value and next range start.
+            for (long cntr0 = cur; cntr0 < item.start; cntr0++)
+                onGapClose.apply(cntr0 + 1);
 
-        queue.clear();
+            // Close pending ranges.
+            update(item.start + item.delta);
```
pavlukhin: Can we call update only for the last item?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/PartitionUpdateCounter.java
```diff
@@ -180,14 +181,24 @@ private void offer(Item item) {
...
+        Item item = poll();
+
+        while (item != null) {
+            long cur = cntr.get();
 
-        if (last != null)
-            update(last.start + last.delta);
+            // Call closure for each counter between current value and next range start.
+            for (long cntr0 = cur; cntr0 < item.start; cntr0++)
+                onGapClose.apply(cntr0 + 1);
```
pavlukhin: Just for understanding. Why do we need to notify the listener for each counter with step equal to 1? I am thinking of possible hundreds or thousands of notifications (or even more).

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/GridDhtPartitionTopologyImpl.java
```diff
@@ -2679,8 +2681,22 @@ private void removeNode(UUID nodeId) {
...
 
-                if (finalizeCntrsBeforeCollecting)
-                    part.finalizeUpdateCountres();
+                if (finalizeCntrsBeforeCollecting && grp.mvccEnabled()) {
+                    // We need to close all gaps in partition update counters sequence. We assume this finalizing is
+                    // happened on exchange and hence all txs are completed. Therefore each gap in update counters
+                    // sequence is a result of undelivered DhtTxFinishMessage on backup (sequences on primary nodes
+                    // do not have gaps). Here we close these gaps and asynchronously notify continuous query engine
+                    // about the skipped events.
+                    AffinityTopologyVersion topVer = ctx.exchange().readyAffinityVersion();
```
pavlukhin: Just double check. What version is assumed to be used here? Before exchange or after?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxAdapter.java
```diff
@@ -275,13 +275,15 @@
...
     /** Mvcc tx update snapshot. */
+    @GridToStringInclude
     protected volatile MvccSnapshot mvccSnapshot;
 
     /** Rollback finish future. */
     @GridToStringExclude
     private volatile IgniteInternalFuture rollbackFut;
 
     /** */
+    @SuppressWarnings("unused")
```
pavlukhin: `@GridToStringExclude`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/GridDhtPartitionTopologyImpl.java
```diff
@@ -2679,8 +2681,22 @@ private void removeNode(UUID nodeId) {
...
+
+                    part.finalizeUpdateCounters(new IgniteInClosure<Long>() {
+                        @Override public void apply(Long cntr) {
+                            for (GridCacheContext ctx0 : grp.caches()) {
```
pavlukhin: Single line loop body without braces.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
rkondakov: No it's not. I was sure I removed it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/PartitionUpdateCounter.java
rkondakov: No, we can't. We need too iterate over all items in the queue and handle all gaps between them.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/PartitionUpdateCounter.java
rkondakov: The reason ow this action is to fill backup CQ queue with the "fake" entries instead of gaps. Of cause we can think about optimization but it is not necessary here because every listener notification here is a simple put to queue.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/GridDhtPartitionTopologyImpl.java
rkondakov: It is version before exchange.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxAdapter.java
rkondakov: No. This field is changed by AtomicReferenceFieldUpdater, so IDEA thinks it is unused.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/GridDhtPartitionTopologyImpl.java
rkondakov: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/PartitionUpdateCounter.java
pavlukhin: I meant `update` method specifically. As far as I see we need to call `update` only for the greatest interval. I could suggest something like this:
```
if (queue.isEmpty())
    return;

long cur = cntr.get();

for (Item item : queue) {
    for (long c = cur + 1; c <= item.start; c++)
        onGapClose.apply(c);

    cur = item.start;
}

Item last = queue.pollLast();

update(last.start + last.delta);

queue.clear();
```

rkondakov: Yes, your variant is slightly better from the performance point of view. But this code is not on the hot path and even more - this is a very, very cold path. Gaps happens in the very rare cases. So it is not the best place to start optimization from.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxAdapter.java
rkondakov: OK, I get it. I'll fix it in next patch, I think. May be it would be better to override `toString` method in txCounters class.

PR #5199, opened 2018-10-29
ignite-10045 Strict bounds check

SpiderRus:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/reader/StandaloneWalRecordsIterator.java
```diff
@@ -139,6 +143,45 @@
...
+
+                assert desc != null;
+
+                if (desc.idx() == lowBound.index())
+                    break;
+            }
+        }
+
+        if (idx == walFiles.size())
+            throw new IgniteCheckedException("Wal segments not in bounds.");
```
dspavlov: I suggest providing a more informative message about which bounds were broken, how to avoid this message to be generated and how to fix it. At least message may contain min/max bounds we broke here.

This suggestion is related to current message and all following exceptions

PR #5200, opened 2018-10-29
IGNITE-10330 Page compression for Ignite persistent storage.

svladykin:


modules/compress/src/main/java/org/apache/ignite/internal/processors/compress/CompressionProcessorImpl.java
```diff
@@ -0,0 +1,345 @@
...
+        int pageSize,
+        int blockSize,
+        PageCompression compression,
+        int compressLevel
+    ) throws IgniteCheckedException {
+        assert compression != null;
+        assert U.isPow2(pageSize): pageSize;
+        assert page.position() == 0 && page.limit() == pageSize;
+
+        if (!U.isPow2(blockSize))
```
agoncharuk: Can we move these checks to the processor startup? It would be better to fail early on activation rather than at checkpoint.

modules/compress/src/main/java/org/apache/ignite/internal/processors/compress/NativeFileSystemPosix.java
```diff
@@ -0,0 +1,54 @@
...
+import jnr.posix.FileStat;
+import jnr.posix.POSIX;
+import jnr.posix.POSIXFactory;
+
+/**
+ * Posix file system API.
+ */
+public class NativeFileSystemPosix implements NativeFileSystem {
+    /** */
+    private static POSIX posix = POSIXFactory.getPOSIX();
```
agoncharuk: This is a subject to discussion rather than a comment. Would it be possible to get rid of LGPL dependencies in the module? It looks like we only need fstat which we can implement via JNA native calls. My concern is that we cannot ship LGPL libs in Apache Ignite builds, which will be a major downside for community users.

modules/compress/src/test/java/org/apache/ignite/internal/processors/compress/CompressionTestSuite.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.ignite.internal.processors.compress;
+
+import junit.framework.TestSuite;
+
+/**
+ */
+public class CompressionTestSuite extends TestSuite {
```
agoncharuk: We need to add more suites the same way Direct IO sutes are added with one default compression algorithm used to make sure all recovery mechanics works well. Once we introduce nightly builds, we may consider adding all compression algorithms to the nightly suite.

modules/compress/src/main/java/org/apache/ignite/internal/processors/compress/CompressionProcessorImpl.java
```diff
@@ -0,0 +1,345 @@
...
+import org.apache.ignite.internal.processors.cache.persistence.tree.io.PageIO;
+import org.apache.ignite.internal.util.typedef.internal.U;
+
+import static org.apache.ignite.configuration.PageCompression.SKIP_GARBAGE;
+import static org.apache.ignite.internal.util.GridUnsafe.NATIVE_BYTE_ORDER;
+
+/**
+ * Compression processor.
+ */
+public class CompressionProcessorImpl extends CompressionProcessor {
```
agoncharuk: I think we need to add some basic metrics to DataRegionMetrics, otherwise it will be very difficult for us to estimate the benefit of compression during testing as well as users who will be playing with configuration.

modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
```diff
@@ -384,6 +384,12 @@
      */
     private boolean encryptionEnabled;
 
+    /** */
+    private PageCompression pageCompression;
```
agoncharuk: Subject to discuss: given that hole-punching will not be the only compression approach, should we:
1) Introduce some sort of compression strategy?
2) Add an ability to have a pluggable compression algorithm?
3) For hole-punching in-memory caches are not supported. For some other compression approaches we may support in-memory caches, but data from different caches may be placed to the same page. Should we move this to the DataStorageConfiguration?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -1533,6 +1535,7 @@ private GridCacheContext createCache(CacheConfiguration<?, ?> cfg,
 
         boolean nearEnabled = GridCacheUtils.isNearEnabled(cfg);
 
+        CacheCompressionManager compressMgr = new CacheCompressionManager();
```
agoncharuk: Add a check to fail or warn if compression is configured for an in-memory cache

modules/compress/src/main/java/org/apache/ignite/internal/processors/compress/CompressionProcessorImpl.java
svladykin: Agree. Done.

modules/compress/src/test/java/org/apache/ignite/internal/processors/compress/CompressionTestSuite.java
svladykin: Agree. Please provide the list of tests to be added.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
svladykin: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/compress/CompressionProcessor.java
```diff
@@ -0,0 +1,173 @@
...
+    public static final int LZ4_DEFAULT_LEVEL = 0;
+
+    /** */
+    public static final int ZSTD_MIN_LEVEL = -131072;
+
+    /** */
+    public static final int ZSTD_MAX_LEVEL = 22;
+
+    /** */
+    public static final int ZSTD_DEFAULT_LEVEL = 3;
```
alamar: My recommendation is to use 2 since performance is much better with roughly same compression ratio:
http://apache-ignite-developers.2346864.n4.nabble.com/Compression-prototype-tp34199p34649.html

modules/compress/src/main/java/org/apache/ignite/internal/processors/compress/CompressionProcessorImpl.java
```diff
@@ -0,0 +1,399 @@
...
+     * @param cap Capacity.
+     * @return Direct byte buffer.
+     */
+    static ByteBuffer allocateDirectBuffer(int cap) {
+        return ByteBuffer.allocateDirect(cap).order(NATIVE_BYTE_ORDER);
+    }
+
+    /** {@inheritDoc} */
+    @Override public void checkPageCompressionSupported(Path storagePath, int pageSize) throws IgniteCheckedException {
+        if (testMode)
```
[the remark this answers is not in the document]
svladykin: Fixed: this test mode was not really necessary, just dropped it altogether.

modules/compress/src/main/java/org/apache/ignite/internal/processors/compress/CompressionProcessorImpl.java
```diff
@@ -0,0 +1,399 @@
...
+            ByteBuffer dst = compressBuf.get();
+
+            // Position on a part that needs to be decompressed.
+            page.limit(compressedSize)
+                .position(PageIO.COMMON_HEADER_END);
+
+            // LZ4 needs this limit to be exact.
+            dst.limit(compactSize - PageIO.COMMON_HEADER_END);
+
+            if (compressType == ZSTD_COMPRESSED_PAGE) {
```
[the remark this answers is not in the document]
svladykin: Fixed.

modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
```diff
@@ -2297,6 +2309,54 @@ public boolean isEncryptionEnabled() {
         return this;
     }
 
+    /**
```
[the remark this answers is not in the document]
svladykin: Disagree. DataRegion is all about memory shared between some caches. It has nothing to do with disk compression. 
Say we have two regions: 100GB for fast caches and 10GB for less important slow ones. This is unrelated to what must be compressed ant what not.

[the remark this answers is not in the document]
ilantukh: DataRegionConfiguration in persistent mode is about how data from memory pages is moved to a disk - it has properties like pageEvictionMode and checkpointPageBufSize. For Cache diskPageCompression (and even just "disk page") is a too low-level concept.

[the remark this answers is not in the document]
svladykin: Ok, DataRegionConfiguration has some relation to persistence configuration. This is not really important here.

The important thing is that I do not see any reasons why caches with different compression levels/algorithms can not share the same data region. If we will create a separate data region for each page compression configuration, we may end up with ineffective memory utilization.

"Too low level concept" is a kind of unclear argument for me. Configuration can not be too high or too low level, it can only be either correctly or incorrectly placed with respect to granularity of it's effect. And my point is that having page compression configuration per data region is too coarse grained and restrictive.

PR #5208, opened 2018-10-30
IGNITE-10011 Fixed pages leak

agoncharuk:


dspavlov:
is this PR still actual?

PR #5217, opened 2018-10-30
ClusterNodeAttributeAffinityBackupFilter: make constructor public

rynffoll:


DaveWHarvey:
The lack of public was an oversight.   The only testing had been via Spring.

PR #5223, opened 2018-10-31
IGNITE-8719 Store information about starting of index rebuild.

antonovsergey93:


modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/StartBuildIndexRecord.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+package org.apache.ignite.internal.pagemem.wal.record;
+
+import org.apache.ignite.internal.util.tostring.GridToStringInclude;
+import org.apache.ignite.internal.util.typedef.internal.S;
+
+/**
+ * Record
+ */
+public class StartBuildIndexRecord extends TimeStampRecord {
```
zstan: may be more accurate here is to reuse GroupPartitionId class ?

PR #5225, opened 2018-10-31
IGNITE-8896

Rodion4ik:
In TcpCommunicationSpi text of message "greater than" replace on "greater than or equal to" according to conditions


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -2208,7 +2208,7 @@ public int boundPort() {
                 "since may produce significant delays with some scenarios.");
 
         if (slowClientQueueLimit > 0 && msgQueueLimit > 0 && slowClientQueueLimit >= msgQueueLimit) {
-            U.quietAndWarn(log, "Slow client queue limit is set to a value greater than message queue limit " +
+            U.quietAndWarn(log, "Slow client queue limit is set to a value greater than or equal to message queue limit " +
```
NSAmelchev: This change makes the length of the line more than 120 symbols. (According to coding guide it has 120 char limit) 

Also, you should fix the wrong JavaDoc of the **TcpCommunicationSpi.setSlowClientQueueLimit** method that says to set the same value as **getMessageQueueLimit**. (It's confusing with the warning message)

PR #5251, opened 2018-11-05
IGNITE-10130 Option for disable triggering cache interceptor on events by cross datacenter replication was added.

antonovsergey93:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -5835,6 +5843,8 @@ private CacheDataRow checkRowExpired(CacheDataRow row) throws IgniteCheckedExcep
         /** OldRow expiration flag. */
         private boolean oldRowExpiredFlag = false;
 
+        private boolean disableInterceptAfter = false;
```
ascherbakoff: Javadoc is missing.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/version/GridCacheLazyPlainVersionedEntry.java
```diff
@@ -106,6 +106,11 @@ public V value(boolean keepBinary) {
         return val;
     }
 
+    /**
+     * @return Cache context.
+     */
+    public GridCacheContext cacheContext() { return cctx; }
```
ascherbakoff: This is not needed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -6344,6 +6359,25 @@ else if (interceptorVal != updated0) {
...
         }
 
+        /**
+         * Checks, that change was made on another cluster.
+         *
+         * @param conflictCtx Conflict context associated with change.
+         * @return {@code True} If change was made on another cluster (i.e. change was got by DR) and {@code False}
+         * otherwise.
+         */
+        private boolean changesFromAnotherDataCenter(@Nullable GridCacheVersionConflictContext<?, ?> conflictCtx){
```
ascherbakoff: Looks like it's enough to check for `conflictVer != null` to determine if update came from DrReceiver.
Newly introduced method is not needed.
Do it before calling `org.apache.ignite.internal.processors.cache.GridCacheMapEntry.AtomicCacheUpdateClosure#resolveConflict` or the property could be cleared by conflict resolution code.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -5763,6 +5766,11 @@ private CacheDataRow checkRowExpired(CacheDataRow row) throws IgniteCheckedExcep
      *
      */
     private static class AtomicCacheUpdateClosure implements IgniteCacheOffheapManager.OffheapInvokeClosure {
+
+        /** Flag for disabling interceptor on updates/deletes by DR events. */
+        private final boolean disableInterceptorOnDrEvts =
```
ascherbakoff: You don't need another one boolean field in each atomic update closure.

Better move the property to GridCacheDrManager: `org.apache.ignite.internal.processors.cache.dr.GridCacheDrManager#cacheInterceptorDisabled`

antonovsergey93:
@ascherbakoff i apllied your comments. Could you check code again?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -1518,9 +1520,16 @@ else if (res.resultType() == ResultType.LOCKED) {
 
                 CacheLazyEntry e = new CacheLazyEntry(cctx, key, old, keepBinary);
 
-                Object interceptorVal = cctx.config().getInterceptor().onBeforePut(
-                    new CacheLazyEntry(cctx, key, old, keepBinary),
-                    val0);
+                Object interceptorVal = val0;
+
+                if(cctx.dr().cacheInterceptorDisabled() && explicitVer != null && explicitVer.dataCenterId() != cctx.dr().dataCenterId())
```
ascherbakoff: This could be simpified to
`if (!(interceptDisabled = cctx.dr().cacheInterceptorDisabled() && explicitVer != null &&
                    explicitVer.dataCenterId() != cctx.dr().dataCenterId()))
                    interceptorVal = cctx.config().getInterceptor().onBeforePut(
                        new CacheLazyEntry(cctx, key, old, keepBinary),
                        val0);`

Also you should avoid too long lines and keep right formatting.

antonovsergey93:
@ascherbakoff check my code again please.

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1031,6 +1031,13 @@
      */
     public static final String IGNITE_ALLOW_START_CACHES_IN_PARALLEL = "IGNITE_ALLOW_START_CACHES_IN_PARALLEL";
 
+    /**
+     * Name of property for disabling cache triggering of interceptor by DR events.
+     *
+     * Default is {@code false}.
+     */
+    public static final String IGNITE_DISABLE_INTERCEPTOR_INVOCATION_ON_DR_EVENTS = "IGNITE_DISABLE_CACHE_INTERCEPTOR_INVOCATION_ON_DR_EVENTS";
```
ascherbakoff: This should be removed from ignite properties.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -3289,6 +3296,34 @@ protected final boolean hasValueUnlocked() {
...
+    }
+
+    /**
+     * Checks, that cache interceptor should be skipped.
+     * <p>
+     * In current cache interceptor implementation, if changes were got by DR, the following methods will be invoked:
+     * {@link CacheInterceptor#onBeforePut(Cache.Entry, Object)}, {@link CacheInterceptor#onAfterPut(Cache.Entry)},
+     * {@link CacheInterceptor#onBeforeRemove(Cache.Entry)} and {@link CacheInterceptor#onAfterRemove(Cache.Entry)}, but
+     * method {@link CacheInterceptor#onGet(Object, Object)} will not be invoked. This implementation could corrupt
+     * data. For example, cache interceptor wrap data on update/remove and unwrap data back on get from cache. For avoid
```
AMashenkov: Let's fix use case description to smth like this:

It is expects by default behavior that  mentioned Interceptor methods will be called, but onGet(). This can even make DR-update flow broken in case of non-idempotent Interceptor and force users to call onGet manually as the only workaround.
Also, user may want to skip Interceptor to avoid redundant entry transformation for DR updates and exchange with internal data b/w data centres which is a normal case.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -5957,6 +6000,7 @@ private CacheDataRow checkRowExpired(CacheDataRow row) throws IgniteCheckedExcep
             this.conflictResolve = conflictResolve;
             this.intercept = intercept;
             this.updateCntr = updateCntr;
+            this.disableTriggeringCacheInterceptorOnConflict = disableTriggeringCacheInterceptorOnConflict;
```
AMashenkov: Let's use shorter name, e.g. skipInterceptorOnConflict

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -230,6 +231,10 @@
     @GridToStringExclude
     private final Lock listenerLock;
 
+    /** */
+    private final boolean disableTriggeringCacheInterceptorOnConflict =
```
AMashenkov: Let's make this 'static' otherwise we'll get noticeable performance drop due to HashTable lacks.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -3289,6 +3294,31 @@ protected final boolean hasValueUnlocked() {
...
+     * @return {@code true} if changes were got by DR and {@code false} otherwise.
+     */
+    private boolean isRemoteDrUpdate(@Nullable GridCacheVersion explicitVer) {
+        return explicitVer != null && explicitVer.dataCenterId() != cctx.dr().dataCenterId();
+    }
+
+    /**
+     * Checks, that cache interceptor should be skipped.
+     * <p>
+     * It is expects by default behavior that mentioned Interceptor methods will be called, but onGet(). This can even
```
AMashenkov: Now, context is missed here - "mentioned Interceptor methods".

PR #5308, opened 2018-11-06
IGNITE-9989: Fixed jdbc v2 PK metadata

pavel-kuznetsov:


bwzhang2011:
System.out.println() should be removed *MetaData class.

pavel-kuznetsov:
Using another PR :  #5716

PR #5317, opened 2018-11-06
IGNITE-10150: Fixed handling of non-existing cache context in GridRed…

slukyano:
…uceQueryExecutor.


modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/twostep/DisappearedCacheWasNotFoundMessageSelfTest.java
```diff
@@ -61,14 +62,19 @@ public void testDisappearedCacheWasNotFoundMessage() {
             fail("No CacheException emitted.");
         }
         catch (CacheException e) {
-            assertTrue(e.getMessage(), e.getMessage().contains("Cache not found on local node"));
+            boolean exp = e.getMessage().contains("Cache not found on local node");
```
SomeFire: This will not work if the exception is not thrown.

slukyano: If exception is not thrown we hit `fail("No CacheException emitted.");`. Am I missing something?

SomeFire: Yep, sorry, I missed that.

PR #5327, opened 2018-11-06
IGNITE-10085 Compressed files were made more user friendly.

antonovsergey93:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -2146,7 +2146,7 @@ private void compressSegmentToFile(long nextSegment, File raw, File zip)
 
             try (ZipOutputStream zos = new ZipOutputStream(new BufferedOutputStream(new FileOutputStream(zip)))) {
                 zos.setLevel(dsCfg.getWalCompactionLevel());
-                zos.putNextEntry(new ZipEntry(""));
+                zos.putNextEntry(new ZipEntry(cctx.igniteInstanceName() + "-" + nextSegment + ".wal"));
```
voropava: why not just make it zip without .zip suffix, that is commmon constract for single file archives.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -2146,8 +2146,8 @@ private void compressSegmentToFile(long nextSegment, File raw, File zip)
 
             try (ZipOutputStream zos = new ZipOutputStream(new BufferedOutputStream(new FileOutputStream(zip)))) {
                 zos.setLevel(dsCfg.getWalCompactionLevel());
-                zos.putNextEntry(new ZipEntry(""));
-
+                zos.putNextEntry(new ZipEntry(cctx.igniteInstanceName() + "-" + nextSegment + ".wal.zip"));
```
voropava: why not just name entry as zip without ".zip" suffix, this is common contract

PR #5334, opened 2018-11-07
IGNITE-10118: JDBC v2 metadata getSchemas() fix

pavel-kuznetsov:


pavel-kuznetsov:
Patch is declined, because this issue is already fixed by IGNITE-9989 fix.

PR #5391, opened 2018-11-14
IGNITE-5115

NSAmelchev:
Problem was that coordinator fails when process the fail message about itself. Reproducer attached to PR.
I have fixed this issue by segmenting coordinator on getting the fail message about itself.


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -4959,6 +4959,11 @@ private void processNodeFailedMessage(TcpDiscoveryNodeFailedMessage msg) {
 
                     return;
                 }
+                else if (locNodeId.equals(failedNodeId)) {
```
voropava: Why not do this check in the beginning of method?

NSAmelchev: Hi, @voropava 
This fix is for the case when the coordinator fails. In other cases, the failing non-coordinator node doesn't get this message and it'll be redundant check.

voropava: Ok, but why else if? If message is not verified we should fail as well? Don't yuo want to extract method for coordinator's branch with @NotNull failed node parameter

voropava: Ok, it is not really readable approach, but seems this is not scope of this jira.

PR #5394, opened 2018-11-14
IGNITE-10058

xtern:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheAffinitySharedManager.java
```diff
@@ -318,6 +319,35 @@ void checkRebalanceState(GridDhtPartitionTopology top, Integer checkGrpId) {
...
 
+    /**
+     * @param top Topology version.
+     * @param aff Affinity assignment cache.
+     * @param waitNode Node rebalancing data.
+     * @param p Parition id.
+     * @param state Partition state.
+     * @return {@code True} if this partition is moving in LOST state.
+     */
+    private boolean lostPartitionMoving(
```
sergey-chugunov-1985: As I understand, the only purpose of this method is to ensure that partition state on local node is the same as on oldPrimary and equals to LOST.

Do we actually need it? I tried to run new test along with others from the same class with simplified check `state != LOST` and everything seems working.

xtern: Yes. I removed it, because I don't know scenario when this check fails.

PR #5422, opened 2018-11-19
IGNITE-10314 QueryEntity is not updated when column and index added or dropped

ldzhjn:


ldzhjn:
Close this PR due to solution change

PR #5425, opened 2018-11-19
IGNITE-10300 Fixed correct message in case of user/password retries e…

voropava:
…xceeded.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
@@ -2559,7 +2559,7 @@ public class CommandHandler {
                     }
                     else {
                         if (tryConnectMaxCount == 0)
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -2557,8 +2557,12 @@ public int execute(List<String> rawArgs) {
 
                         tryConnectMaxCount--;
                     }
-                    else
+                    else {
+                        if (tryConnectMaxCount == 0)
+                            throw new GridClientException("Authentication error, max retries exceeded");
```
sergey-chugunov-1985: It is better to use more specific `GridClientAuthenticationException` here.

voropava: Fixed

PR #5446, opened 2018-11-20
IGNITE-10193: IgniteBaselineAffinityTopologyActivationTest#testBaselineTopologyHistoryIsDeletedOnBaselineDelete fails

avplatonov:


avplatonov:
We (@avplatonov and @agoncharuk) decided that there is no reason to save this test because it assume and check old behavior.

PR #5473, opened 2018-11-22
IGNITE-9171  Use lazy mode for default

tledkov-gridgain:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridReduceQueryExecutor.java
```diff
@@ -283,11 +291,20 @@ private void onFail(ClusterNode node, GridQueryFailResponse msg) {
      */
     private void fail(ReduceQueryRun r, UUID nodeId, String msg, byte failCode) {
         if (r != null) {
-            CacheException e = new CacheException("Failed to execute map query on remote node [nodeId=" + nodeId +
-                ", errMsg=" + msg + ']');
+            CacheException e;
 
-            if (failCode == GridQueryFailResponse.CANCELLED_BY_ORIGINATOR)
-                e.addSuppressed(new QueryCancelledException());
+            if (failCode == GridQueryFailResponse.CANCELLED_BY_ORIGINATOR) {
```
pavlukhin: I see that now *cause* is used instead of *suppressed* exception. But I sill see usages of `wasCancelled` in current class. Are you sure that new translation scheme works as expected?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridReduceQueryExecutor.java
```diff
@@ -832,6 +851,8 @@ else if (mvccTracker != null)
                 throw resEx;
             }
             finally {
+                detachedConn.recycle();
```
pavlukhin: Why do we need detach-recycle in reducer?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Session.java
```diff
@@ -0,0 +1,86 @@
...
+
+        if (lock.isHeldByCurrentThread())
+            lock.unlock();
+    }
+
+    /**
+     *
+     */
+    public void checkTablesVersions() {
+        if (ses != null)
```
pavlukhin: Could we omit null check here?

pavlukhin:
Also I have not noticed if query originating node failure is handled. We need to cleanup resources left by lazy query execution in that case.

PR #5484, opened 2018-11-22
IGNITE-10272: Inject learning environment into scope of dataset compute task.

artemmalykh:


modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
```diff
@@ -0,0 +1,155 @@
...
+import org.apache.ignite.ml.environment.parallelism.NoParallelismStrategy;
+import org.apache.ignite.ml.environment.parallelism.ParallelismStrategy;
+import org.apache.ignite.ml.math.functions.IgniteSupplier;
+
+/**
+ * Builder for LearningEnvironment.
+ */
+public class DefaultLearningEnvironmentBuilder implements LearningEnvironmentBuilder {
+    /** Parallelism strategy. */
+    private ParallelismStrategy parallelismStgy;
```
dmitrievanthony: Missed empty lines.

modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
```diff
@@ -0,0 +1,155 @@
...
+import org.apache.ignite.ml.environment.logging.NoOpLogger;
+import org.apache.ignite.ml.environment.parallelism.DefaultParallelismStrategy;
+import org.apache.ignite.ml.environment.parallelism.NoParallelismStrategy;
+import org.apache.ignite.ml.environment.parallelism.ParallelismStrategy;
+import org.apache.ignite.ml.math.functions.IgniteSupplier;
+
+/**
+ * Builder for LearningEnvironment.
+ */
+public class DefaultLearningEnvironmentBuilder implements LearningEnvironmentBuilder {
```
dmitrievanthony: I guess `LearningEnvironmentBuilder` should be `Serializable` (we send it to workers) and so that this class should have `serialVersionUUID`.

modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
```diff
@@ -0,0 +1,155 @@
...
+        random.setSeed(seed);
+        return new LearningEnvironmentImpl(part, random, parallelismStgy, loggingFactory);
+    }
+
+    /**
+     * Default LearningEnvironment implementation.
+     */
+    private class LearningEnvironmentImpl implements LearningEnvironment {
+        /** Parallelism strategy. */
+        private final ParallelismStrategy parallelismStgy;
```
dmitrievanthony: Missed empty lines.

modules/ml/src/main/java/org/apache/ignite/ml/environment/LearningEnvironmentBuilder.java
```diff
@@ -17,109 +17,78 @@
...
-/**
- * Builder for LearningEnvironment.
- */
-public class LearningEnvironmentBuilder {
-    /** Parallelism strategy. */
-    private ParallelismStrategy parallelismStgy;
-    /** Logging factory. */
-    private MLLogger.Factory loggingFactory;
-
+public interface LearningEnvironmentBuilder {
```
dmitrievanthony: It should be `Serializable` because we send it to worker nodes.

modules/ml/src/main/java/org/apache/ignite/ml/math/functions/IgniteFunction.java
```diff
@@ -26,5 +27,22 @@
  * @see java.util.function.Function
  */
 public interface IgniteFunction<T, R> extends Function<T, R>, Serializable {
-
+    static <T, R> IgniteFunction<T, R> constant(R r) {
+        return t -> r;
+    }
+//    R apply(T t);
```
dmitrievanthony: Please, remove commented code.

modules/ml/src/main/java/org/apache/ignite/ml/math/functions/IgniteFunction.java
```diff
@@ -26,5 +27,22 @@
  * @see java.util.function.Function
  */
 public interface IgniteFunction<T, R> extends Function<T, R>, Serializable {
-
+    static <T, R> IgniteFunction<T, R> constant(R r) {
```
dmitrievanthony: Do we actually need it?

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/VectorUtils.java
```diff
@@ -39,6 +39,17 @@ public static DenseVector zeroes(int n) {
         return (DenseVector)new DenseVector(n).assign(0.0);
     }
 
+    /**
+     * Create new vector of specified size n with specified value.
+     *
+     * @param val Value.
+     * @param n Size;
+     * @return new vector of specified size n with specified value.
```
dmitrievanthony: Capital letter.

modules/ml/src/main/java/org/apache/ignite/ml/pipeline/Pipeline.java
```diff
@@ -53,6 +54,8 @@
     /** Final trainer stage. */
     private DatasetTrainer finalStage;
 
+    private LearningEnvironmentBuilder envBuilder = LearningEnvironmentBuilder.defaultBuilder();
```
dmitrievanthony: Missed javadoc.

modules/ml/src/main/java/org/apache/ignite/ml/trainers/DatasetTrainer.java
```diff
@@ -38,8 +40,10 @@
  * @param <L> Type of a label.
  */
 public abstract class DatasetTrainer<M extends Model, L> {
+    /** Learning environment builder. */
+    protected LearningEnvironmentBuilder envBuilder = LearningEnvironmentBuilder.defaultBuilder();
```
dmitrievanthony: Missed empty line.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/UpstreamTransformer.java
```diff
@@ -28,15 +27,8 @@
...
-     * Perform transformation of upstream.
-     *
-     * @param rnd Random numbers generator.
-     * @param upstream Upstream.
-     * @return Transformed upstream.
-     */
-    // TODO: IGNITE-10296: Inject capabilities of randomization through learning environment.
-    // TODO: IGNITE-10297: Investigate possibility of API change.
-    public Stream<UpstreamEntry<K, V>> transform(Random rnd, Stream<UpstreamEntry<K, V>> upstream);
+public interface UpstreamTransformer<K, V> extends IgniteFunction<Stream<UpstreamEntry<K, V>>, Stream<UpstreamEntry<K, V>>> {
```
dmitrievanthony: Do you think it's a good idea to inherit from `IgniteFunction`? We're loosing clear `transform` signature, but I don't see any advantages.

modules/ml/src/test/java/org/apache/ignite/ml/TestUtils.java
```diff
@@ -325,4 +326,43 @@ private static boolean equals(final double x, final double y, final int maxUlps)
...
+     * Gets test learning environment builder with a given seed.
+     *
+     * @param seed Seed.
+     * @return test learning environment builder.
+     */
+    public static LearningEnvironmentBuilder testEnvBuilder(long seed) {
+        return LearningEnvironmentBuilder.defaultBuilder().withRNGSeed(seed);
+    }
+
+    public static class DataWrapper<T> implements AutoCloseable {
```
dmitrievanthony: Missed javadoc.

modules/ml/src/test/java/org/apache/ignite/ml/composition/boosting/convergence/ConvergenceCheckerTest.java
```diff
@@ -23,6 +23,7 @@
 import org.apache.ignite.ml.composition.ModelsComposition;
 import org.apache.ignite.ml.composition.boosting.loss.Loss;
 import org.apache.ignite.ml.dataset.impl.local.LocalDatasetBuilder;
+import org.apache.ignite.ml.environment.LearningEnvironmentBuilder;
```
dmitrievanthony: Do we need this import statement?

modules/ml/src/main/java/org/apache/ignite/ml/Model.java
```diff
@@ -39,4 +39,8 @@
     public default String toString(boolean pretty) {
         return getClass().getSimpleName();
     }
+
+    static <T, V> Model<T, V> constantModel(V v) {
```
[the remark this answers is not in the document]
artemmalykh: This (and also identity model which I've added) is needed as standard functional primitive. Ofc, we can always write corresponding lambda, but this approach raises readability. Javadoc fixed.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/UpstreamTransformer.java
```diff
@@ -28,15 +27,8 @@
...
+    default UpstreamTransformer<K, V> then(UpstreamTransformer<K, V> other) {
```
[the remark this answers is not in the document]
artemmalykh: Thanks for javadoc and 'public', as for `then` the point was to
1. save interoperability of `UpstreamTransformer` with plain functions (hence inheritance from `IgniteFunction`);
 2. make combiner which is like `andThen`, but under which `UpstreamTransformers` are closed (`upstreamTransformer.then(upstreamTransformer)` produces `UpstreamTransformer`, not just a `Function`).
If we drop interoperability thing, then sure, right way to go would be `andThen`. Maybe we can drop it, because `UpstreamTransformer` will be, I guess, primarily used with other `UpstreamTransformer`s.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/UpstreamTransformerBuilder.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.ml.dataset;
+
+import org.apache.ignite.ml.environment.LearningEnvironment;
+
+public interface UpstreamTransformerBuilder<K, V> {
```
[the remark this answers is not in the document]
artemmalykh: See my comment above about `then` and `andThen`.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/UpstreamTransformerBuildersChain.java
```diff
@@ -0,0 +1,101 @@
...
+    private List<UpstreamTransformerBuilder<K, V>> list;
+
+    /**
+     * Creates empty upstream transformers chain (basically identity function).
+     *
+     * @param <K> Type of upstream keys.
+     * @param <V> Type of upstream values.
+     * @return Empty upstream transformers chain.
+     */
+    public static <K, V> UpstreamTransformerBuildersChain<K, V> empty() {
```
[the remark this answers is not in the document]
artemmalykh: Primarily readability.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/bootstrapping/BootstrappedDatasetBuilder.java
```diff
@@ -69,12 +70,16 @@ public BootstrappedDatasetBuilder(IgniteBiFunction<K, V, Vector> featureExtracto
...
+    @Override public BootstrappedDatasetPartition build(
+        LearningEnvironment env,
+        Iterator<UpstreamEntry<K, V>> upstreamData,
+        long upstreamDataSize,
         EmptyContext ctx) {
 
         BootstrappedVector[] dataset = new BootstrappedVector[Math.toIntExact(upstreamDataSize)];
 
         int cntr = 0;
+        // TODO: use rng from env.
```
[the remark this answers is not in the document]
artemmalykh: Done.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/cache/util/ComputeUtils.java
```diff
@@ -328,25 +371,27 @@ public static void removeData(Ignite ignite, UUID datasetId) {
...
      * @param qry Cache query.
-     * @param transformersChain Transformers of stream of upstream data.
+     * @param transformer Upstream transformer.
+     * @param idTransformer If upstream transformer is trivial (identity function)
+     * @param <K> Type of a key in {@code upstream} data.
+     * @param <V> Type of a value in {@code upstream} data.
      * @return Number of entries supplied by the iterator.
      */
-    private static <K, V> long computeCount(
+    private static  <K, V> long computeCount(
```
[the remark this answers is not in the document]
artemmalykh: Fixed.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/cache/util/PartitionLearningEnvStorage.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param part Partition index.
+     * @param supplier {@link LearningEnvironment} supplier.
+     * @return {@link LearningEnvironment}.
+     */
+    @SuppressWarnings("unchecked")
+    <D> D computeDataIfAbsent(int part, Supplier<LearningEnvironment> supplier) {
+        LearningEnvironment env = storage.get(part);
+
+        if (env == null) {
+            Lock lock = locks.computeIfAbsent(part, p -> new ReentrantLock());
```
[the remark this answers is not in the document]
artemmalykh: Agree, removed this class entirely, because without locks we can just use ConcurrentMap.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/cache/util/PartitionLearningEnvStorage.java
```diff
@@ -0,0 +1,67 @@
...
+     * Unlike {@link ConcurrentMap#computeIfAbsent(Object, Function)},
+     * this method guarantees that function will be called only once.
+     *
+     * @param <D> Type of data.
+     * @param part Partition index.
+     * @param supplier {@link LearningEnvironment} supplier.
+     * @return {@link LearningEnvironment}.
+     */
+    @SuppressWarnings("unchecked")
+    <D> D computeDataIfAbsent(int part, Supplier<LearningEnvironment> supplier) {
```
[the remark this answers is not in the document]
artemmalykh: Removed this class, see comment about locks in it.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/local/LocalDataset.java
```diff
@@ -32,6 +33,8 @@
  * @param <D> Type of a partition {@code data}.
  */
 public class LocalDataset<C extends Serializable, D extends AutoCloseable> implements Dataset<C, D> {
+    /** Partition {@code data} storage. */
+    private final List<LearningEnvironment> envs;
```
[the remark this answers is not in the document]
artemmalykh: Fixed.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/local/LocalDataset.java
```diff
@@ -45,35 +48,38 @@
      * @param ctx Partition {@code context} storage.
      * @param data Partition {@code data} storage.
      */
-    LocalDataset(List<C> ctx, List<D> data) {
+    LocalDataset(List<LearningEnvironment> envs, List<C> ctx, List<D> data) {
```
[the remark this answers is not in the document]
artemmalykh: Done.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/local/LocalDatasetBuilder.java
```diff
@@ -99,37 +123,43 @@ public LocalDatasetBuilder(Map<K, V> upstreamMap, IgniteBiPredicate<K, V> filter
 
         int ptr = 0;
 
+        List<LearningEnvironment> envs = IntStream.range(0, partitions)
```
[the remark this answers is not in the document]
artemmalykh: Fixed, now it is 2 lines :)

modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
artemmalykh: Fixed.

modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
artemmalykh: Agree, fixed.

modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
artemmalykh: Fixed.

modules/ml/src/main/java/org/apache/ignite/ml/environment/LearningEnvironmentBuilder.java
artemmalykh: Now we have `IgniteFunction<partIdx, X>` fields instead of `X` in this class.

modules/ml/src/main/java/org/apache/ignite/ml/math/functions/IgniteFunction.java
artemmalykh: Sorry for mess, fixed.

modules/ml/src/main/java/org/apache/ignite/ml/math/functions/IgniteFunction.java
artemmalykh: Yes, I guess, because it is standard functional block which raises readability.

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/VectorUtils.java
artemmalykh: Fixed.

modules/ml/src/main/java/org/apache/ignite/ml/pipeline/Pipeline.java
artemmalykh: Fixed.

modules/ml/src/main/java/org/apache/ignite/ml/trainers/DatasetTrainer.java
artemmalykh: Fixed.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/UpstreamTransformerBuildersChain.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.List;
+import org.apache.ignite.ml.environment.LearningEnvironment;
+
+/**
+ * Class representing chain of transformers applied to upstream.
+ *
+ * @param <K> Type of upstream keys.
+ * @param <V> Type of upstream values.
+ */
+public class UpstreamTransformerBuildersChain<K, V> implements Serializable {
```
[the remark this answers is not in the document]
artemmalykh: Yes, actually, we can, refactored `UpstreamTransformerBuilder` and removed this class.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/UpstreamTransformerBuildersChain.java
```diff
@@ -0,0 +1,101 @@
...
+    /** List of upstream transformations. */
+    private List<UpstreamTransformerBuilder<K, V>> list;
```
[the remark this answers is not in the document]
artemmalykh: Removed this class.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/cache/util/ComputeUtils.java
```diff
@@ -328,25 +371,27 @@ public static void removeData(Ignite ignite, UUID datasetId) {
...
         IgniteCache<K, V> cache,
         ScanQuery<K, V> qry,
-        UpstreamTransformerChain<K, V> transformersChain) {
+        UpstreamTransformer<K, V> transformer,
+        boolean idTransformer) {
```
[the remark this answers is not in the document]
artemmalykh: Agree, removed.

modules/ml/src/main/java/org/apache/ignite/ml/composition/boosting/GDBLearningStrategy.java
```diff
@@ -157,10 +161,11 @@
     /**
      * Sets learning environment.
      *
-     * @param environment Learning Environment.
+     * @param envBuilder Learning Environment.
      */
-    public GDBLearningStrategy withEnvironment(LearningEnvironment environment) {
-        this.environment = environment;
+    public GDBLearningStrategy withEnvironmentBuilder(LearningEnvironmentBuilder envBuilder) {
```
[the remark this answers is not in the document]
dmitrievanthony: @artemmalykh, could you please have a look at this problem?

modules/ml/src/main/java/org/apache/ignite/ml/dataset/DatasetBuilder.java
```diff
@@ -21,8 +21,10 @@
 import org.apache.ignite.lang.IgniteBiPredicate;
 import org.apache.ignite.ml.dataset.impl.cache.CacheBasedDatasetBuilder;
 import org.apache.ignite.ml.dataset.impl.local.LocalDatasetBuilder;
+import org.apache.ignite.ml.environment.LearningEnvironmentBuilder;
 import org.apache.ignite.ml.trainers.transformers.BaggingUpstreamTransformer;
 
+
```
dmitrievanthony: There are more empty lines then we actually need.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/UpstreamTransformer.java
```diff
@@ -18,25 +18,34 @@
...
      *
-     * @param rnd Random numbers generator.
-     * @param upstream Upstream.
+     * @param upstream Upstream to transform.
      * @return Transformed upstream.
      */
-    // TODO: IGNITE-10296: Inject capabilities of randomization through learning environment.
-    // TODO: IGNITE-10297: Investigate possibility of API change.
-    public Stream<UpstreamEntry<K, V>> transform(Random rnd, Stream<UpstreamEntry<K, V>> upstream);
+    Stream<UpstreamEntry<K, V>> transform(Stream<UpstreamEntry<K, V>> upstream);
```
dmitrievanthony: Please, don't forget `public`.

modules/ml/src/main/java/org/apache/ignite/ml/environment/LearningEnvironmentBuilder.java
```diff
@@ -17,109 +17,135 @@
...
+    }
+
+
+    /**
+     * Specify dependency (partition -> logging factory).
+     *
+     * @param loggingFactory Function describing (partition -> logging factory).
+     * @return This object.
+     */
+    public LearningEnvironmentBuilder withLoggingFactory(IgniteFunction<Integer, MLLogger.Factory> loggingFactory);
```
avplatonov: Let's suppose that logging with metrics aggregation may be huge process. If we bind factory to partition then such huge processes will be started several times on one node. I think that on this step of architecture this is not problem but we should consider this later.

modules/ml/src/test/java/org/apache/ignite/ml/environment/LearningEnvironmentBuilderTest.java
```diff
@@ -39,7 +39,7 @@
     /** */
     @Test
     public void basic() {
-        LearningEnvironment env = LearningEnvironment.DEFAULT;
+        LearningEnvironment env = LearningEnvironment.DEFAULT_TRAINER_ENV;
```
avplatonov: Why did you rename this object? I think it is not necessary

modules/ml/src/test/java/org/apache/ignite/ml/dataset/impl/local/LocalDatasetBuilderTest.java
```diff
@@ -105,6 +107,7 @@ public void testBuildWithPredicate() {
         };
 
         return builder.build(
+            TestUtils.testEnvBuilder(),
```
avplatonov: builder.withEnvBuilder? why not?

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/cache/util/ComputeUtils.java
```diff
@@ -244,25 +287,24 @@ public static void removeData(Ignite ignite, UUID datasetId) {
             qry.setFilter(filter);
 
             C ctx;
-            UpstreamTransformerChain<K, V> chainCopy = Utils.copy(transformersChain);
-            chainCopy.modifySeed(s -> s + part);
+            UpstreamTransformer<K, V> transformer = transformerBuilder.build(env);
+            UpstreamTransformer<K, V> transformerCopy = Utils.copy(transformer);
```
dmitrievanthony: Could you please clarify why we need a copy here?

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/local/LocalDatasetBuilder.java
```diff
@@ -99,37 +122,28 @@ public LocalDatasetBuilder(Map<K, V> upstreamMap, IgniteBiPredicate<K, V> filter
...
         int ptr = 0;
 
+        List<LearningEnvironment> envs = IntStream.range(0, partitions).boxed().map(envBuilder::buildForWorker)
+            .collect(Collectors.toList());
+
         for (int part = 0; part < partitions; part++) {
             int cnt = part == partitions - 1 ? entriesList.size() - ptr : Math.min(partSize, entriesList.size() - ptr);
+            LearningEnvironment env = envs.get(part);
+            UpstreamTransformer<K, V> transformer1 = upstreamTransformerBuilder.build(env);
+            UpstreamTransformer<K, V> transformer2 = Utils.copy(transformer1);
```
dmitrievanthony: Why do we need a copy here?

modules/ml/src/main/java/org/apache/ignite/ml/trainers/DatasetTrainer.java
```diff
@@ -25,7 +25,9 @@
 import org.apache.ignite.ml.dataset.DatasetBuilder;
 import org.apache.ignite.ml.dataset.impl.cache.CacheBasedDatasetBuilder;
 import org.apache.ignite.ml.dataset.impl.local.LocalDatasetBuilder;
+import org.apache.ignite.ml.environment.LearningEnvironmentBuilder;
 import org.apache.ignite.ml.environment.LearningEnvironment;
+import org.apache.ignite.ml.environment.LearningEnvironmentBuilder;
```
avplatonov: duplicate?

modules/ml/src/main/java/org/apache/ignite/ml/structures/partition/LabeledDatasetPartitionDataBuilderOnHeap.java
```diff
@@ -57,8 +58,10 @@ public LabeledDatasetPartitionDataBuilderOnHeap(IgniteBiFunction<K, V, Vector> x
     }
 
     /** {@inheritDoc} */
-    @Override public LabeledVectorSet<Double, LabeledVector> build(Iterator<UpstreamEntry<K, V>> upstreamData,
-                                                                   long upstreamDataSize, C ctx) {
+    @Override public LabeledVectorSet<Double, LabeledVector> build(
+        LearningEnvironment env,
```
avplatonov: maybe we move set of env through with.. method?

modules/ml/src/main/java/org/apache/ignite/ml/environment/LearningEnvironmentBuilder.java
```diff
@@ -17,109 +17,135 @@
...
-    /** Logging factory. */
-    private MLLogger.Factory loggingFactory;
+public interface LearningEnvironmentBuilder extends Serializable {
+    /**
+     * Builds {@link LearningEnvironment} for worker on given partition.
+     *
+     * @param part Partition.
+     * @return {@link LearningEnvironment} for worker on given partition.
+     */
+    default public LearningEnvironment buildForWorker(int part) {
```
dmitrievanthony: `s/default public/public default`

modules/ml/src/main/java/org/apache/ignite/ml/environment/LearningEnvironmentBuilder.java
```diff
@@ -17,109 +17,135 @@
...
+        return buildForTrainer();
```
dmitrievanthony: Are you sure that it works **correctly**?

modules/ml/src/main/java/org/apache/ignite/ml/environment/LearningEnvironmentBuilder.java
```diff
@@ -17,109 +17,135 @@
...
-            case NO_PARALLELISM:
-                this.parallelismStgy = NoParallelismStrategy.INSTANCE;
-                break;
-            case ON_DEFAULT_POOL:
-                this.parallelismStgy = new DefaultParallelismStrategy();
-                break;
-        }
-        return this;
+    public default LearningEnvironmentBuilder withLoggingFactory(MLLogger.Factory loggingFactory) {
+        return withLoggingFactory(constant(loggingFactory));
```
dmitrievanthony: Doesn't look like it's correct. If factory is not serializable you can't wrap it into function and so that make it serializable.

modules/ml/src/main/java/org/apache/ignite/ml/environment/LearningEnvironmentBuilder.java
avplatonov: In interface you assume infinite recursion but in child class you override this method. Why you provide default realization of both methods? It can be a cause of errors in code.

modules/ml/src/main/java/org/apache/ignite/ml/environment/LearningEnvironmentBuilder.java
```diff
@@ -17,109 +17,135 @@
...
-        /** {@inheritDoc} */
-        @Override public MLLogger logger() {
-            return loggingFactory.create(getClass());
-        }
-
-        /** {@inheritDoc} */
-        @Override public <T> MLLogger logger(Class<T> clazz) {
-            return loggingFactory.create(clazz);
-        }
+    public static LearningEnvironmentBuilder defaultBuilder() {
```
dmitrievanthony: Why not just `new DefaultLearningEnvironmentBuilder()`?

modules/ml/src/main/java/org/apache/ignite/ml/environment/logging/ConsoleLogger.java
```diff
@@ -82,6 +82,9 @@ private void print(VerboseLevel verboseLevel, String line) {
      * ConsoleLogger factory.
      */
     private static class Factory implements MLLogger.Factory {
+        /** Serial version uuid. */
+        private static final long serialVersionUID = 5864605548782107893L;
```
dmitrievanthony: Looks like `MLLogger.Factory` is not serializable, so `serialVersionUID` is redundant here.

modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+
+    /** {@inheritDoc} */
+    @Override public LearningEnvironmentBuilder withRandom(IgniteFunction<Integer, Random> rngSupplier) {
+        this.rngSupplier = rngSupplier;
+
+        return this;
+    }
+
+    /** {@inheritDoc} */
+    @Override public DefaultLearningEnvironmentBuilder withParallelismStrategy(
```
avplatonov: Why you provide functions? You've added Serializable to ParallelismStrategy and you can add this i-face to LoggingFactory. After it you may expect just Map<Integer, ...> instead of functions.

modules/ml/src/main/java/org/apache/ignite/ml/environment/logging/MLLogger.java
```diff
@@ -17,6 +17,7 @@
 
 package org.apache.ignite.ml.environment.logging;
 
+import java.io.Serializable;
```
dmitrievanthony: Unused import.

modules/ml/src/main/java/org/apache/ignite/ml/environment/logging/ConsoleLogger.java
avplatonov: It seems that you forget to extend Serializable i-face in MLLogger.Factory

modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override public LearningEnvironment buildForWorker(int part) {
+        Random random = rngSupplier.apply(part);
+        random.setSeed(seed.apply(part));
+        return new LearningEnvironmentImpl(part, random, parallelismStgy.apply(part), loggingFactory.apply(part));
+    }
+
+    /**
```
avplatonov: As I know this comment should be in one line

modules/ml/src/test/java/org/apache/ignite/ml/selection/scoring/evaluator/EvaluatorTest.java
```diff
@@ -288,19 +289,24 @@ void evaluate(DecisionTreeClassificationTrainer trainer,
                 .withEncoderType(EncoderType.STRING_ENCODER)
                 .withEncodedFeature(1)
                 .withEncodedFeature(6) // <--- Changed index here
-                .fit(ignite,
+                .fit(
+                    testEnvBuilder(123L),
```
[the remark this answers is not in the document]
artemmalykh: Well, there are bunch of `with*` methods in extending classes and for API to be sane, i'd expect that `with` methods commute. If we'll make `PreprocessingTrainer withEnvironmentBuilder(...)` in `PreprocessingTrainer`, commutativity will not hold: `new BinarizationTrainer().withThreshhold(x).withEnvironmentBuilder(b)` will be correct, but `new BinarizationTrainer().withEnvBuilder(b).withThreshold(x)` will not compile because of return type of first `with`. Of course, we could return not just  `PreprocessingTrainer`, but in class signature  add `P extends PreprocessingTrainer`, and return this `P` in `withEnvironmentBuilder` and also make each subclass pass self into this param, but that'll make generics count `5` which is probably a little to much... But I personally not against it :)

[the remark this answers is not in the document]
avplatonov: I think we should discuss such situation
I don't like both versions))) but we should choose between two evils)

modules/ml/src/test/java/org/apache/ignite/ml/environment/LearningEnvironmentBuilderTest.java
artemmalykh: This is exactly default learning environment for trainer, in contrast to workers, so I renamed it for clarity.

modules/ml/src/test/java/org/apache/ignite/ml/TestUtils.java
artemmalykh: Fixed.

modules/ml/src/test/java/org/apache/ignite/ml/composition/boosting/convergence/ConvergenceCheckerTest.java
artemmalykh: Yeah, for `isConverged` method.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/local/LocalDatasetBuilder.java
artemmalykh: Because we can possibly have not a stateless transformer. For now, we have only stateless transformer (`BaggingUpstreamTransformer`), but maybe in the future... On the other hand, we can add some contracts in documentation banning states and get rid of this clumsy copying.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/cache/util/ComputeUtils.java
artemmalykh: See https://github.com/apache/ignite/pull/5484#discussion_r236280053

modules/ml/src/main/java/org/apache/ignite/ml/environment/LearningEnvironmentBuilder.java
artemmalykh: I think here we should give user most precise level of granularity, it is her choice to blur it (maybe even ignore it) in needed ways (for example `part -> Ignition.localIgnite().cluster().nodeLocalMap().computeIfAbsent(LOG_AGGREGATOR_FACTORY_FOR_DATASET, key -> someFactorySupplier)`)

examples/src/main/java/org/apache/ignite/examples/ml/preprocessing/ImputingExample.java
```diff
@@ -59,11 +60,12 @@ public static void main(String[] args) throws Exception {
             );
 
             // Defines second preprocessor that imputing features.
+            LearningEnvironmentBuilder envBuilder = LearningEnvironmentBuilder.defaultBuilder();
             IgniteBiFunction<Integer, Person, Vector> preprocessor = new ImputerTrainer<Integer, Person>()
-                .fit(ignite, persons, featureExtractor);
+                .fit(envBuilder, ignite, persons, featureExtractor);
```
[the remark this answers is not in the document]
artemmalykh: See https://github.com/apache/ignite/pull/5484#discussion_r236242793.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/DatasetBuilder.java
artemmalykh: Fixed.

modules/ml/src/main/java/org/apache/ignite/ml/environment/LearningEnvironmentBuilder.java
artemmalykh: Fixed.

modules/ml/src/main/java/org/apache/ignite/ml/trainers/DatasetTrainer.java
```diff
@@ -290,10 +295,29 @@
...
+        this.envBuilder  = envBuilder;
+        this.environment = envBuilder.buildForTrainer();
+    }
+
+    /**
+     * Get learning environment builder.
+     *
+     * @return Learning environment builder.
+     */
+    public LearningEnvironmentBuilder learningEnvironmentBuilder() {
```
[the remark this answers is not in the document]
artemmalykh: Um... Maybe for now, yes. The idea was to have the ability to, for example, create a family of  `DatasetTrainer`s from one, with same `LearningEnvironmentBuilder`. Maybe we should add it by need.

modules/ml/src/main/java/org/apache/ignite/ml/environment/logging/ConsoleLogger.java
artemmalykh: No, but `LearningEnvironmentBuilder::withLoggingFactory` signature is changed.

modules/ml/src/test/java/org/apache/ignite/ml/dataset/impl/local/LocalDatasetBuilderTest.java
artemmalykh: Yeah, I thought about it too, but decided to go along with pattern with `contextBuilder` (we in majority of cases have EmptyContext() as context, but do not have defaults + `with` for it, and if we opted to change this sad situation, we'd have maybe too much methods in interface). But maybe we should add ticket for this API change:)

modules/ml/src/main/java/org/apache/ignite/ml/structures/partition/LabeledDatasetPartitionDataBuilderOnHeap.java
artemmalykh: No, we can do such things with entry point of environment in the chain of learning, but in the middle of it we should explicitly require `env` as a parameter. Imagine if we passed some learning environment builder in trainer, but forgot to set it in the middle of chain: it will be inconsistent chain.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/UpstreamTransformer.java
artemmalykh: I guess, you're right, fixed.

modules/ml/src/test/java/org/apache/ignite/ml/TestUtils.java
artemmalykh: Fixed.

modules/ml/src/main/java/org/apache/ignite/ml/environment/LearningEnvironmentBuilder.java
artemmalykh: We should give user maximal level of granularity, it is user choice to blur it (in most extreme case, ignore it). With this API we still can make grouping by nodes instead of paritions.

modules/ml/src/main/java/org/apache/ignite/ml/trainers/DatasetTrainer.java
artemmalykh: Yes, fixed.

modules/ml/src/main/java/org/apache/ignite/ml/structures/partition/LabeledDatasetPartitionDataBuilderOnHeap.java
artemmalykh: No, i think. Approach `with` + sensible default is good in the beginning of dependency chain, but in the middle of it we should force explicit passing of parameter to not break dependency because of forgetfulness to call set or sth like this. In our case beginning of chain is `DatasetTrainer`, and there we use this approah.

modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
artemmalykh: `ParallelismStrategy` is now not `Serializable`.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/UpstreamTransformer.java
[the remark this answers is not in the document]
dmitrievanthony: I think we could eliminate inheritance from `IgniteFunction` here. I don't see any benefits of it, but two methods like `then` and `andThen` are very confusing.

modules/ml/src/main/java/org/apache/ignite/ml/Model.java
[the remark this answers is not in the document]
dmitrievanthony: I see that it's used only once. Generally speaking it's not a great idea to make a lot of "standard" util code here, especially if it's used very rare.

modules/ml/src/main/java/org/apache/ignite/ml/composition/boosting/GDBLearningStrategy.java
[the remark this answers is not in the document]
dmitrievanthony: Perhaps.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/UpstreamTransformerBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+     * Create {@link UpstreamTransformer} based on learning environment.
+     *
+     * @param env Learning environment.
+     * @return Upstream transformer.
+     */
+    public UpstreamTransformer<K, V> build(LearningEnvironment env);
+
+    /**
+     * Combunes two builders (this and other respectfully)
+     * env -> transformer1
```
dmitrievanthony: Does it make sense to add something like `<pre>` or `{@code }` here?

modules/ml/src/test/java/org/apache/ignite/ml/composition/boosting/convergence/ConvergenceCheckerTest.java
dmitrievanthony: Not sure, IDEA highlights it as unused import.

modules/ml/src/main/java/org/apache/ignite/ml/Model.java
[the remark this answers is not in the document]
artemmalykh: Okay, for now I'll move it into `TestUtils`, but I ask that if anyone needs it in the code, he'll add it here instead of writing v -> v or t -> v.

modules/ml/src/main/java/org/apache/ignite/ml/composition/boosting/GDBLearningStrategy.java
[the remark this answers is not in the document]
artemmalykh: For now added check, as you asked because otherwise to be consistent, we should probably move all other necessary args into constructor, which I do not want to do in this PR.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/UpstreamTransformer.java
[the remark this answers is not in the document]
artemmalykh: Yeah, already done that.

modules/ml/src/test/java/org/apache/ignite/ml/composition/boosting/convergence/ConvergenceCheckerTest.java
artemmalykh: Fixed, I've looked into `ConvergenceChecker`.

modules/ml/src/main/java/org/apache/ignite/ml/dataset/UpstreamTransformerBuilder.java
artemmalykh: Agree.

PR #5485, opened 2018-11-23
IGNITE-10354 Failing client node due to not receiving metrics updates

gromtech:


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNode.java
```diff
@@ -499,6 +499,13 @@ public void clientAliveTime(long aliveTime) {
         this.aliveCheckTime = U.currentTimeMillis() + aliveTime;
     }
 
+    /**
+     * @return Client alive check time.
+     */
+    public long getAliveCheckTime() {
```
alamar: By CC this method should be called aliveCheckTime()

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -5273,6 +5273,11 @@ private void processMetricsUpdateMessage(TcpDiscoveryMetricsUpdateMessage msg) {
                             if (clientNodeIds.contains(clientNode.id()))
                                 clientNode.clientAliveTime(spi.clientFailureDetectionTimeout());
                             else {
+                                if (clientNode.getAliveCheckTime() == 0L) {
+                                    // Init client alive time.
```
alamar: I don't think this comment is needed.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNode.java
alamar: You should probably call it clientAliveTime() to match setter

gromtech: Agree. Sorry, forgot about this rule.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
gromtech: Agree.

PR #5507, opened 2018-11-27
IGNITE-10287: Add ML inference model storage.

dmitrievanthony:


modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/File.java
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.ml.inference.storage.model;
+
+/**
+ * Implementation of file {@link ModelStorageProvider} works with.
+ */
+class File implements FileOrDirectory {
```
avplatonov: Let's add ZippedFile?

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/LocalModelStorageProvider.java
```diff
@@ -0,0 +1,56 @@
...
+import java.lang.ref.WeakReference;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * Implementation of {@link ModelStorageProvider} based on local {@link ConcurrentHashMap}.
+ */
+public class LocalModelStorageProvider implements ModelStorageProvider {
```
avplatonov: I would add "InMemory" because LocalStorage associates with LocalHDD

modules/ml/src/main/java/org/apache/ignite/ml/inference/plugin/MLInferencePluginProvider.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override public void initExtensions(PluginContext ctx, ExtensionRegistry registry) {
+        IgniteConfiguration igniteCfg = ctx.igniteConfiguration();
+
+        this.ignite = ctx.grid();
+        this.log = ctx.log(this.getClass());
+
+        if (igniteCfg.getPluginConfigurations() != null) {
```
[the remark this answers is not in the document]
dmitrievanthony: We could:
1. Ask user to provide configuration that contains information about our caches manually, but it looks a bit too verbose and error prone.
2. Use the same approach Ignite uses to start SYSTEM_CACHE, but in this case we will have to add ML dependency (or at least some ML code) into Core module. I would prefer to avoid it.
3. Start required caches from static initializer of some ML class so that if ML module in classpath the caches will be started. But in this case we need additional synchronization with Ignite start process and can't pass configuration parameters properly.
4. Use [Apache Ignite Plugins](https://apacheignite.readme.io/docs/plugins). This approach looks good to me because it provides a way to pass configuration, provides entry point to run our code (syncronized with Ignite start) and this approach described in official Ignite docs.

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/DefaultModelStorage.java
```diff
@@ -0,0 +1,278 @@
...
+                    + path + "]");
+
+            // If parent is not a directory throw an exception.
+            if (!parent.isDirectory())
+                throw new IllegalStateException("Cannot create file because parent is not a directory [path="
+                    + path + "]");
+
+            Directory dir = (Directory) parent;
+            // Update parent if it's a new file.
+            if (!dir.getFiles().contains(path)) {
```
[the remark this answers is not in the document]
dmitrievanthony: It's put operation, so it's assumed that if file exists it will be replaced.

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/DefaultModelStorage.java
```diff
@@ -0,0 +1,278 @@
...
+        // Paths are locked in child-first order.
+        Lock pathLock = storageProvider.lock(path);
+        Lock parentPathLock = storageProvider.lock(parentPath);
+
+        pathLock.lock();
+        parentPathLock.lock();
+
+        try {
+            // If a directory associated with specified path exists return.
+            if (isDirectory(path))
```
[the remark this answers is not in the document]
dmitrievanthony: If file/directory doesn't exist `isDirectory` will return `false`, so it's fine.

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/File.java
dmitrievanthony: We could do that on top of this abstraction. Any archive can be presenter as byte array.

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/DefaultModelStorage.java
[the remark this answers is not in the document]
dmitrievanthony: Flag `onlyIfNotExist` added.

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/DefaultModelStorage.java
```diff
@@ -0,0 +1,278 @@
...
+    /** {@inheritDoc} */
+    @Override public void mkdir(String path) {
+        String parentPath = getParent(path);
+
+        // Paths are locked in child-first order.
+        Lock pathLock = storageProvider.lock(path);
+        Lock parentPathLock = storageProvider.lock(parentPath);
+
+        pathLock.lock();
+        parentPathLock.lock();
```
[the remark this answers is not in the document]
dmitrievanthony: Good point, I added util `synchronize` method, please have a look.

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/DefaultModelStorage.java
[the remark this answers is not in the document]
dmitrievanthony: Flag `onlyIfNotExist` added.

examples/src/main/java/org/apache/ignite/examples/ml/inference/ModelStorageExample.java
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * This example demonstrates how to work with {@link ModelStorage}.
+ */
+public class ModelStorageExample {
+    /** Run example. */
+    public static void main(String... args) throws IOException, ClassNotFoundException {
+        try (Ignite ignite = Ignition.start("examples/config/example-ignite-ml.xml")) {
+            System.out.println(">>> Ignite grid started.");
+
+            ModelStorage storage = ModelStorageFactory.getModelStorage(ignite);
```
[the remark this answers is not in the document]
dmitrievanthony: I would actually prefer to have something like `ignite.ml().modelStorage()`. Unfortunately, we can't do it without changing core, so it's postponed for now. This static implementation is the closest solution to the final goal.

In favor of flexibility user can create ModelStorages using their constructors...

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/File.java
avplatonov: Yes, you're right but as user I cannot think about such low processes, I just want to save model and if someone setup filesystem with compression - OK

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/descriptor/ModelDescriptorStorage.java
```diff
@@ -45,4 +47,7 @@
      * @param mdlId Model identifier.
      */
     public void remove(String mdlId);
+
+    /** {@inheritDoc} */
+    public Iterator<IgniteBiTuple<String, ModelDescriptor>> iterator();
```
artemmalykh: Maybe we should clarify what entries iterator produces instead of just `@inheritDoc`?

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/ModelStorageProvider.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Removes file or directory associated with the specified path.
+     *
+     * @param path Path to the file or directory.
+     */
+    public void remove(String path);
+
+    /**
+     * Locks the row associated with the specified path.
```
artemmalykh: I would rather not use "row" here, here we are in upper level of abstraction.

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/descriptor/ModelDescriptorStorage.java
dmitrievanthony: Good point, done.

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/ModelStorageProvider.java
dmitrievanthony: Agree, done.

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/LocalModelStorageProvider.java
dmitrievanthony: This name is chosen to be similar to `LocalModelDescriptorStorage`. This name was discussed in previous PR. We could do it again.

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/DefaultModelStorage.java
```diff
@@ -0,0 +1,278 @@
...
+            FileOrDirectory parent = storageProvider.get(parentPath);
+
+            // If parent doesn't exist throw an exception.
+            if (parent == null)
+                throw new IllegalStateException("Cannot create file because directory doesn't exist [path="
+                    + path + "]");
+
+            // If parent is not a directory throw an exception.
+            if (!parent.isDirectory())
+                throw new IllegalStateException("Cannot create file because parent is not a directory [path="
```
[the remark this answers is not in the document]
dmitrievanthony: I think `IllegalArgumentException` would be better.

modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/File.java
dmitrievanthony: So, you suggest to implement compression of file-system level? I think we shell do it in another task.

PR #5514, opened 2018-11-27
IGNITE-10429: Wrap Scanner in DirectorySerializerTest into try-with-resources.

dmitrievanthony:


oignatenko:
I checked this fix: code diffs look right and involved tests don't fail anymore (these failed prior to fix)

PR #5520, opened 2018-11-28
IGNITE-10376

1vanan:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/CacheObjectBinaryProcessorImpl.java
```diff
@@ -214,6 +214,8 @@ public CacheObjectBinaryProcessorImpl(GridKernalContext ctx) {
                     assert newMeta != null;
                     assert newMeta instanceof BinaryTypeImpl;
 
+                    BinaryMetadata newMeta0 = ((BinaryTypeImpl)newMeta).metadata();
```
sergey-chugunov-1985: Don't see a reason for this change.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/CacheObjectBinaryProcessorImpl.java
```diff
@@ -535,8 +535,14 @@ public GridBinaryMarshaller marshaller() {
 
             BinaryMetadata mergedMeta = BinaryUtils.mergeMetadata(oldMeta, newMeta0, changedSchemas);
 
-            if (oldMeta != null && mergedMeta == oldMeta && metaHolder.pendingVersion() == metaHolder.acceptedVersion())
-                return; // Safe to use existing schemas.
+            if (mergedMeta == oldMeta)
+                // Metadata locally is up-to-date. Waiting for updating metadata in an entire cluster, if necessary.
+                while (true) {
+                    if (metaHolder.pendingVersion() == metaHolder.acceptedVersion())
```
sergey-chugunov-1985: Instead of spinning in loop we should use `BinaryMetadataTransport::awaitMetadataUpdate` method and wait when metaHolder.pendingVersion is accepted by the cluster.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/CacheObjectBinaryProcessorImpl.java
sergey-chugunov-1985: Don't see a reason for this change.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/CacheObjectBinaryProcessorImpl.java
```diff
@@ -535,8 +535,14 @@ public GridBinaryMarshaller marshaller() {
 
             BinaryMetadata mergedMeta = BinaryUtils.mergeMetadata(oldMeta, newMeta0, changedSchemas);
 
-            if (oldMeta != null && mergedMeta == oldMeta && metaHolder.pendingVersion() == metaHolder.acceptedVersion())
-                return; // Safe to use existing schemas.
+            if (mergedMeta == oldMeta)
+                // Metadata locally is up-to-date. Waiting for updating metadata in an entire cluster, if necessary.
+                while (true) {
```
sergey-chugunov-1985: Instead of spinning in loop we should use BinaryMetadataTransport::awaitMetadataUpdate method and wait when metaHolder.pendingVersion is accepted by the cluster.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/CacheObjectBinaryProcessorImpl.java
```diff
@@ -535,8 +535,16 @@ public GridBinaryMarshaller marshaller() {
 
             BinaryMetadata mergedMeta = BinaryUtils.mergeMetadata(oldMeta, newMeta0, changedSchemas);
 
-            if (oldMeta != null && mergedMeta == oldMeta && metaHolder.pendingVersion() == metaHolder.acceptedVersion())
-                return; // Safe to use existing schemas.
+            if (mergedMeta == oldMeta) {
+                // Metadata locally is up-to-date. Waiting for updating metadata in an entire cluster, if necessary.
+                if (metaHolder.pendingVersion() == metaHolder.acceptedVersion()) {
```
sergey-chugunov-1985: This condition is incorrect: if pending version and accepted version are the same it means metadata is stable, no on-going updates in the cluster, so thread doesn't need to wait for anything and is allowed to proceed right away.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/CacheObjectBinaryProcessorImpl.java
1vanan: Ok, I removed it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/CacheObjectBinaryProcessorImpl.java
1vanan: I agree it should be vice-versa condition. Thank you for the remark.

PR #5574, opened 2018-12-05
Ignite-10534

ygerzhedovich:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/ddl/DdlStatementsProcessor.java
```diff
@@ -527,6 +528,10 @@ private static void isDdlOnSchemaSupported(String schemaName) {
      * @param tbl Table.
      */
     private static void isDdlSupported(GridH2Table tbl) {
+        //Local cache always started, so if cache is not started it's not local.
```
AMashenkov: I'd expect 'isXXX' method is a getter and should return boolean. Name 'ensureDdlSupported' looks more suitable. 

It is not clear whether you mean 'local' cache mode or current node is not a data node for cache. 
Is it possible transform comment into assertion?

ygerzhedovich: It's old method, I don't change semantic the method and keep previous name.
About local, please look below - there is check on local type of cache. Seems I can modify comment to like 'Cache with local mode always started, so if cache is not started it's not local.'
I don't see how the comment can be transform to assertion, it just clarification.

PR #5581, opened 2018-12-05
IGNITE-10552 Web Agent: Improve logging when cluster topology changed

akuznetsov-gridgain:


modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/handlers/ClusterListener.java
```diff
@@ -368,6 +369,14 @@ public IgniteProductVersion clusterVersion() {
         boolean differentCluster(TopologySnapshot prev) {
             return prev == null || F.isEmpty(prev.nids) || Collections.disjoint(nids, prev.nids);
         }
+
+        /**
+         * @param prev Previous topology.
+         * @return {@code true} in case if current topology is the same cluster, but topology changed.
+         */
+        boolean topologyChanged(TopologySnapshot prev) {
+            return prev != null && !prev.nids.equals(nids) && !Collections.disjoint(nids, prev.nids);
```
nva:
```suggestion
            return prev != null && !prev.nids.equals(nids);
```

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/rest/RestExecutor.java
```diff
@@ -70,6 +71,9 @@
     /** Index of alive node URI. */
     private Map<List<String>, Integer> startIdxs = U.newHashMap(2);
 
+    /** Cache for parsed URLs. */
+    private Map<String, HttpUrl> urlCache = new ConcurrentHashMap<>();
```
nva: I think lazy cache of parsed url is not good idea. Why we don't map nodeURIs to Collection<HttpUrl> before any sendRequest?

akuznetsov-gridgain: Will rollback, as it is not that easy to refactor.

PR #5595, opened 2018-12-06
IGNITE-10463: MVCC: Create "PDS 3" test suite for MVCC mode.

rkondakov:


modules/core/src/test/java/org/apache/ignite/testframework/MvccFeatureChecker.java
```diff
@@ -120,9 +120,9 @@ public static boolean isSupported(CacheMode mode) {
     public static void assertMvccWriteConflict(Exception e) {
         IgniteSQLException sqlEx = X.cause(e, IgniteSQLException.class);
 
-        assertNotNull(sqlEx);
+        assertNotNull("Unexpected exception:" + X.getFullStackTrace(e), sqlEx);
```
AMashenkov: Can we simplyfy this with:
``` 
if (sqlEx == null ||  sqlEx.statusCode != TRANSACTION_SERIALIZATION_ERROR)
   throw e;
```

PR #5598, opened 2018-12-07
IGNITE-10314 Spark dataframe will get wrong schema if user executes add/drop column DDL

ldzhjn:


modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
```diff
@@ -124,10 +130,63 @@ package object impl {
...
+    /**
+      * @param ignite Ignite instance.
+      * @param ccfg Cache Configuration.
+      * @param tabName Table name.
+      * @tparam K Key class.
+      * @tparam V Value class.
+      * @return HashMap consists of fields and fields type for a given table.
+      */
+    def refreshFields[K, V](ignite: Ignite, ccfg : CacheConfiguration[K, V], tabName: String): util.LinkedHashMap[String, String] = {
+        val connection = connect(ignite)
```
nizhikov: Seems, we can do this without jdbc connection, just using internal Ignite API.
Please, take a look at 

`org.apache.ignite.internal.processors.odbc.jdbc.JdbcRequestHandler#getColumnsMeta`

```
            for (String cacheName : ctx.cache().publicCacheNames()) {
                for (GridQueryTypeDescriptor table : ctx.query().types(cacheName)) {
                    if (!matches(table.schemaName(), req.schemaName()))
                        continue;

                    if (!matches(table.tableName(), req.tableName()))
                        continue;
        //Process column here
```

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
```diff
@@ -124,10 +127,41 @@ package object impl {
...
+
+    /**
+      * @param ignite Ignite instance.
+      * @param ccfg Cache Configuration.
+      * @param tabName Table name.
+      * @tparam K Key class.
+      * @tparam V Value class.
+      * @return HashMap consists of fields and fields type for a given table.
+      */
+	def refreshFields[K, V](ignite: Ignite, ccfg: CacheConfiguration[K, V], tabName: String): util.LinkedHashMap[String, String] = {
```
nizhikov: We shouldn't modify `QueryEntity` content.
Seems we can completely replace usages of `QueryEntity` with `GridQueryTypeDescriptor`.
Let's do it

What do you think?

ldzhjn: Ok, I'll implement using GridQueryTypeDescriptor to replace QueryEntity.

One question here, do we still need CacheConfiguration?
Seems the only usage for CacheConfiguration is to get schema from it, but now GridQueryTypeDescriptor has schema field.

nizhikov: > do we still need CacheConfiguration?

It seems that we don't.
If it really not necessary, we can remove it from this code.

ldzhjn: During implantation, I found out that GridQueryTypeDescriptor doesn't have keyfields set like QueryEntity.
The only way to get key fields is to iterate props map of QueryTypeDescriptorImpl, based on value of GridQueryProperty.key to determine if that field is a key field.
And this approach does not work with composite object key field if the table is created by cache API.

Does this approach to get key fields makes sense to you?

nizhikov:
@ldzhjn 

Seems, we can iterate over `GridQueryTypeDescriptor#fields` and call `GridQueryTypeDescriptor#property` -> `GridQueryProperty#key` to get all key fields.

This calls doesn't require to use implementation specific methods, all methods are belong to interfaces.

Does it makes sense for you?

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
nizhikov: @ldzhjn

Seems, we can iterate over GridQueryTypeDescriptor#fields and call GridQueryTypeDescriptor#property -> GridQueryProperty#key to get all key fields.

This calls doesn't require to use implementation specific methods, all methods are belong to interfaces.

Does it makes sense for you?

nizhikov: @ldzhjn

Seems, we can iterate over GridQueryTypeDescriptor#fields and call GridQueryTypeDescriptor#property -> GridQueryProperty#key to get all key fields.

This calls doesn't require to use implementation specific methods, all methods are belong to interfaces.

Does it makes sense for you?

ldzhjn: Thanks for the heads up, I'll use this way then.

One more question here since you mentioned not to use implementation specific methods.
In line 123 of IgniteSQLRelation, we need to get alias for field name.
Currently my implementation is 
name = table.asInstanceOf[QueryTypeDescriptorImpl].aliases.getOrDefault(name, name),

Do I have to change the implementation here?

nizhikov: Currently, we doesn't have another option to get aliases, so I propose to do following:

1. Use `table.asInstanceOf[QueryTypeDescriptorImpl]`
2. Fill the ticket to bring `aliases` to `GridQueryTypeDescriptor` interface.

ldzhjn: I have implemented the fix, please review and comment.
The tests in teamcity is all green.
I attached the link in jira.

modules/spark/src/test/scala/org/apache/ignite/spark/IgniteDataFrameSchemaSpec.scala
```diff
@@ -88,6 +111,24 @@ class IgniteDataFrameSchemaSpec extends AbstractDataFrameSpec {
 
         createPersonTable(client, DEFAULT_CACHE)
 
+        createMetaTestTable(client, DEFAULT_CACHE)
```
nizhikov: Let's move this calls into the one test. 

I think the test should be the following:

1. Read `test` table. Check the initial fields list.
2. Execute `addColumnForTable`. Read `test` table. Check the fields list.
3. Execute `dropColumnForTable`. Read `test` table. Check the fields list.

modules/spark/src/test/scala/org/apache/ignite/spark/AbstractDataFrameSpec.scala
```diff
@@ -129,6 +129,27 @@ abstract class AbstractDataFrameSpec extends FunSpec with Matchers with BeforeAn
         cache.put("key3", Employee(3, "Arnold Schwarzenegger", 27, 1000))
     }
 
+    def createMetaTestTable(client: Ignite, cacheName: String): Unit = {
```
nizhikov: Let's move these methods to the `IgniteDataFrameSchemaSpec` because it used only there.

modules/spark/src/main/scala/org/apache/spark/sql/ignite/IgniteExternalCatalog.scala
```diff
@@ -97,9 +97,9 @@ private[ignite] class IgniteExternalCatalog(igniteContext: IgniteContext)
 
         val schemaName = schemaOrDefault(db, currentSchema)
 
-        igniteSQLTable(ignite, tabName, Some(db)) match {
+        sqlTableInfo(ignite, tabName, Some(db)) match {
```
nizhikov: Why this renaming? Let's keep the existing name.

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
```diff
@@ -117,24 +113,37 @@ package object impl {
...
+            cacheName => ignite.asInstanceOf[IgniteEx].context.query.types(cacheName))
+            .find(table => table.tableName.equalsIgnoreCase(tabName) && isValidSchema(table, schemaName))
+
+    /**
+      * @param table GridQueryTypeDescriptor for a given table.
+      * @param schemaName Optional schema name.
+      * @return `True` if schema is valid.
+      */
+    def isValidSchema(table: GridQueryTypeDescriptor, schemaName: Option[String]): Boolean =
+        if (schemaName.isDefined) schemaName.get.equalsIgnoreCase(table.schemaName) ||
```
nizhikov: Let's use pattern matching of map here:

```
schemaName match {
    case Some(schema) => schema.equalsIgnoreCase(table.schemaName)
    case None => true
}
```

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
```diff
@@ -117,24 +113,37 @@ package object impl {
...
+            schemaName.contains(SessionCatalog.DEFAULT_DATABASE)
```
nizhikov: Can we use `equals` here?

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
```diff
@@ -71,29 +74,22 @@ package object impl {
...
-
     /**
       * @param ignite Ignite instance.
       * @param tabName Table name.
       * @param schemaName Optional schema name.
       * @return Cache name for given table.
       */
     def sqlCacheName(ignite: Ignite, tabName: String, schemaName: Option[String]): Option[String] =
-        sqlTableInfo[Any, Any](ignite, tabName, schemaName).map(_._1.getName)
+        if (sqlTableInfo(ignite, tabName, schemaName).isDefined)
```
nizhikov: Let's use `map` here, like in previous implementation.

modules/spark/src/test/scala/org/apache/ignite/spark/IgniteDataFrameSchemaSpec.scala
ldzhjn: Fixed

modules/spark/src/test/scala/org/apache/ignite/spark/AbstractDataFrameSpec.scala
ldzhjn: Moved

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
ldzhjn: Fixed

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
ldzhjn: Fixed

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
```diff
@@ -71,29 +73,23 @@ package object impl {
...
-
     /**
       * @param ignite Ignite instance.
       * @param tabName Table name.
       * @param schemaName Optional schema name.
       * @return Cache name for given table.
       */
     def sqlCacheName(ignite: Ignite, tabName: String, schemaName: Option[String]): Option[String] =
-        sqlTableInfo[Any, Any](ignite, tabName, schemaName).map(_._1.getName)
+		sqlTableInfo(ignite, tabName, schemaName) match {
```
nizhikov: Let's use `map` here.

```
    def sqlCacheName(ignite: Ignite, tabName: String, schemaName: Option[String]): Option[String] =
        sqlTableInfo(ignite, tabName, schemaName).map(_.asInstanceOf[QueryTypeDescriptorImpl].cacheName)
```

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
```diff
@@ -117,24 +113,39 @@ package object impl {
...
+      * @return GridQueryTypeDescriptor for a given table.
       */
-    def sqlTableInfo[K, V](ignite: Ignite, tabName: String,
-        schemaName: Option[String]): Option[(CacheConfiguration[K, V], QueryEntity)] =
-        cachesForSchema[K,V](ignite, schemaName)
-            .map(ccfg ⇒ ccfg.getQueryEntities.find(_.getTableName.equalsIgnoreCase(tabName)).map(qe ⇒ (ccfg, qe)))
-            .find(_.isDefined)
-            .flatten
+    def sqlTableInfo(ignite: Ignite, tabName: String, schemaName: Option[String]): Option[GridQueryTypeDescriptor] =
+        ignite.asInstanceOf[IgniteEx].context.cache.publicCacheNames.flatMap(
```
nizhikov: Let's use a bit different formatting?

```
        ignite.asInstanceOf[IgniteEx].context.cache.publicCacheNames
            .flatMap(cacheName => ignite.asInstanceOf[IgniteEx].context.query.types(cacheName))
            .find(table => table.tableName.equalsIgnoreCase(tabName) && isSameSchema(table, schemaName))
```

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
```diff
@@ -117,24 +113,39 @@ package object impl {
...
+            cacheName => ignite.asInstanceOf[IgniteEx].context.query.types(cacheName))
+            .find(table => table.tableName.equalsIgnoreCase(tabName) && isValidSchema(table, schemaName))
+
+    /**
+      * @param table GridQueryTypeDescriptor for a given table.
+      * @param schemaName Optional schema name.
+      * @return `True` if schema is valid.
+      */
+    def isValidSchema(table: GridQueryTypeDescriptor, schemaName: Option[String]): Boolean =
```
nizhikov: Let's use a bit different formatting?

```
    def isSameSchema(table: GridQueryTypeDescriptor, schemaName: Option[String]): Boolean =
        schemaName match {
            case Some(schema) => 
                schema.equalsIgnoreCase(table.schemaName) || schema.equals(DEFAULT_DATABASE)
            case None => 
                true
        }
```

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
ldzhjn: Fixed

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
ldzhjn: Fixed

modules/spark/src/main/scala/org/apache/ignite/spark/impl/package.scala
ldzhjn: Fixed

PR #5606, opened 2018-12-07
IGNITE-10225 fix starvation check

DirectXceriD:


modules/core/src/main/java/org/apache/ignite/internal/util/StripedExecutor.java
```diff
@@ -48,15 +48,20 @@
...
 import org.jetbrains.annotations.NotNull;
 
+import static org.apache.ignite.IgniteSystemProperties.IGNITE_STRIPED_POOL_STARVATION_THRESHOLD;
+
 /**
  * Striped executor.
  */
 public class StripedExecutor implements ExecutorService {
+    /** Default threshold for starvation check */
+    private static final long DEFAULT_THRESHOLD = 30 * 1000;
```
alamar: Why not use Failure Detection Timeout? We're doing this for all other thread blocking detection.

modules/core/src/main/java/org/apache/ignite/internal/util/StripedExecutor.java
```diff
@@ -139,20 +143,28 @@ public StripedExecutor(
     }
 
     /**
-     * Checks starvation in striped pool. Maybe too verbose
-     * but this is needed to faster debug possible issues.
+     * Checks starvation in striped pool.
      */
     public void checkStarvation() {
```
alamar: Why do we need to create new method? Note this is an internal package, so we can just change return type.

modules/core/src/main/java/org/apache/ignite/internal/util/StripedExecutor.java
```diff
@@ -139,20 +143,28 @@ public StripedExecutor(
...
+        for (Stripe stripe : stripes) {
             boolean active = stripe.active;
 
-            if (completedCntrs[i] != -1 &&
-                completedCntrs[i] == completedCnt &&
-                active) {
+            long lastStartedTs = stripe.lastStartedTs;
+
+            if (lastStartedTs + threshold < U.currentTimeMillis() && active) {
+                retVal = true;
```
alamar: can you rename it to something descriptive, such as 'starvation'?

alamar: can you also put `active` boolean check first? I.e. if (active && ...

modules/core/src/test/java/org/apache/ignite/internal/util/StripedExecutorTest.java
```diff
@@ -39,6 +48,8 @@
     /** {@inheritDoc} */
     @Override public void afterTest() {
         stripedExecSvc.shutdown();
+        if (origStarvationThreshold != null)
```
alamar: line feed is desired

modules/core/src/test/java/org/apache/ignite/internal/util/StripedExecutorTest.java
```diff
@@ -19,17 +19,26 @@
...
  *
  */
 public class StripedExecutorTest extends GridCommonAbstractTest {
     /** */
     private StripedExecutor stripedExecSvc;
 
+    /** */
+    private String origStarvationThreshold;
+
     /** {@inheritDoc} */
```
alamar: Can you add JUnit 4-style annotations?

modules/core/src/main/java/org/apache/ignite/internal/util/StripedExecutor.java
```diff
@@ -104,9 +109,8 @@ public StripedExecutor(
 
         stripes = new Stripe[cnt];
 
-        completedCntrs = new long[cnt];
-
-        Arrays.fill(completedCntrs, -1);
+        final String thresholdStr = IgniteSystemProperties.getString(IGNITE_STRIPED_POOL_STARVATION_THRESHOLD);
```
alamar: Can you drop this 'final' here?

modules/core/src/main/java/org/apache/ignite/internal/util/StripedExecutor.java
DirectXceriD: I wanted it to be shown via JMX, it seems more convenient for a user to see a result without watching logs, for instance if he disabled automatic check. But I was afraid of removing the old method as it's also JMX-exposed.

DirectXceriD: @alamar you mean that I shouldn't have added new MX method in addition to the old one?

modules/core/src/main/java/org/apache/ignite/internal/util/StripedExecutor.java
DirectXceriD: Yeah, sure, good idea, I'll fix.

modules/core/src/test/java/org/apache/ignite/internal/util/StripedExecutorTest.java
DirectXceriD: sure

modules/core/src/main/java/org/apache/ignite/internal/util/StripedExecutor.java
alamar: It's OK that you have added new method to MX bean, but you should reuse existing method in StripedExecutor class, or remove it outright when adding a new one.

modules/core/src/main/java/org/apache/ignite/internal/util/StripedExecutor.java
DirectXceriD: ok

modules/core/src/main/java/org/apache/ignite/internal/util/StripedExecutor.java
DirectXceriD: ok

modules/core/src/main/java/org/apache/ignite/internal/util/StripedExecutor.java
```diff
@@ -52,11 +51,12 @@
  * Striped executor.
  */
 public class StripedExecutor implements ExecutorService {
+
```
alamar: As per CC, no line feed here

modules/core/src/test/java/org/apache/ignite/internal/util/StripedExecutorTest.java
```diff
@@ -37,7 +38,7 @@
         stripedExecSvc = new StripedExecutor(3, "foo name", "pool name", new JavaLogger(),
             new IgniteInClosure<Throwable>() {
                 @Override public void apply(Throwable throwable) {}
-            }, null);
+            }, null, 2000);
     }
 
     /** {@inheritDoc} */
```
alamar: needs @After

modules/core/src/test/java/org/apache/ignite/internal/util/StripedExecutorTest.java
```diff
@@ -37,7 +38,7 @@
         stripedExecSvc = new StripedExecutor(3, "foo name", "pool name", new JavaLogger(),
```
alamar: needs @Before
=

modules/core/src/test/java/org/apache/ignite/internal/util/StripedExecutorTest.java
```diff
@@ -140,6 +141,22 @@ public void testQueueSize() throws Exception {
         assertEquals(1, stripedExecSvc.queueSize());
     }
 
+    /**
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testStarvationDetected() throws Exception {
```
alamar: Unfortunately this test is flaky. It will sometimes fail for the 1st time and almost always will fail for 2nd run.

DirectXceriD: I've fixed it, ~200 sequential runs -> pass

modules/core/src/main/java/org/apache/ignite/internal/StripedExecutorMXBeanAdapter.java
```diff
@@ -40,7 +40,12 @@
 
     /** {@inheritDoc} */
     @Override public void checkStarvation() {
-        exec.checkStarvation();
+        exec.isStarvationDetected();
+    }
+
+    /** {@inheritDoc} */
+    @Override public boolean isStarvationDetected() {
```
slukyano: I don't like that the name sounds like a simple getter but is actually supposed to do work with the main effect being writing a message to the log. How about renaming this to `detectStarvation`?

alamar: Aaand please mark checkStarvation @Deprecated so that we can ditch it in 3.0

DirectXceriD: done

PR #5612, opened 2018-12-07
IGNITE-10371: [ML] Add multiple metrics calculation fo Binary Classification Evaluation process

zaleslaw:


modules/ml/src/main/java/org/apache/ignite/ml/selection/scoring/metric/BinaryClassificationMetricValues.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /** Returns the pair of metric name and metric value. */
+    public Map<String, Double> toMap() {
+        Map<String, Double> metricValues = new HashMap<>();
+        for (Field field : this.getClass().getDeclaredFields())
+            try {
+                metricValues.put(field.getName(), field.getDouble(this));
+            } catch (IllegalAccessException e) {
+                e.printStackTrace();
```
avplatonov: I don't think that exception hiding is valid behavior for this class

modules/ml/src/main/java/org/apache/ignite/ml/selection/scoring/metric/BinaryClassificationMetricValues.java
```diff
@@ -0,0 +1,221 @@
...
+    private double tn;
+
+    /** False Positive (FP). */
+    private double fp;
+
+    /** False Negative (FN). */
+    private double fn;
+
+    /** Sensitivity or True Positive Rate (TPR). */
+    private double recall;
```
avplatonov: I thought that you just add tp,tn,fp and fn values and calculate other metrics just-in-time by methods like accuracy() 
I realize that it counter-productive but computing of such metrics is a cheap procedure. IMHO a lot of fields are non-readable especially with setters.

modules/ml/src/main/java/org/apache/ignite/ml/selection/scoring/metric/BinaryClassificationMetricValues.java
```diff
@@ -0,0 +1,221 @@
...
+        this.fn = fn;
+    }
+
+    /** */
+    public double recall() {
+        return recall;
+    }
+
+    /** */
+    public void setRecall(double recall) {
```
avplatonov: Moreover setter for field dependent on other fields can break state of class

modules/ml/src/main/java/org/apache/ignite/ml/selection/scoring/metric/BinaryClassificationMetricValues.java
zaleslaw: Good point, will think about that

modules/ml/src/main/java/org/apache/ignite/ml/selection/scoring/metric/BinaryClassificationMetricValues.java
zaleslaw: Should we rethrow something IgniteML related do yu think?

PR #5614, opened 2018-12-08
Replace explicit generic type with diamond operator

igor-suhorukov:


nizhikov:
Hello @igor-suhorukov 
Thanks for the patch!
There are several issues that has to be resolved prior merge:

1. Please, edit PR description. Use format IGNITE-XXXX: Ticket description.
Where: "IGNITE-XXXX" is the Jira ticket you trying to resolve with this patch.
"Ticket description" = Replace explicit generic type with diamond operator

2. Please, resolve conflicts with master

3. You should edit `~/src/ignite/idea/ignite_codeStyle.xml`, `~/src/ignite/idea/ignite_inspections_teamcity.xml` to enable corresponding inspection. So we would check future patches with it.

PR #5662, opened 2018-12-13
IGNITE-8227 Research possibility and implement JUnit test failure handler for TeamCity

SomeFire:


modules/core/src/test/java/org/apache/ignite/failure/TestFailingFailureHandler.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override public boolean handle(Ignite ignite, FailureContext failureCtx) {
+        if (!testIsRunning.getAndSet(false)) {
+            ignite.log().info("Critical issue detected after test finished. Test failure handler ignore it.");
+
+            return true;
+        }
+
+        boolean nodeStopped = super.handle(ignite, failureCtx);
+
+        TestCase.fail(failureCtx.toString());
```
dspavlov: why this line is needed here? This will result in Assertion failed error and no handling will be performed later.

SomeFire: Leftover from previous version. I removed this line and restarted tests - they looks good enough.

modules/core/src/test/java/org/apache/ignite/failure/ExpectThrowableFailureHandler.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.ignite.internal.util.typedef.X;
+import org.apache.ignite.internal.util.typedef.internal.S;
+import org.apache.ignite.testframework.junits.GridAbstractTest;
+
+/**
+ * Stops node and fails test.
+ */
+public class ExpectThrowableFailureHandler extends TestFailingFailureHandler {
+    /** Expected throwables. */
+    private final List<Class<? extends Throwable>> errs = new ArrayList<>();
```
dspavlov: we can use self-descriptive name expectedErrs instead of errs. It will make code more readable

modules/core/src/test/java/org/apache/ignite/failure/TestFailingFailureHandler.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * @param test Test.
+     */
+    public TestFailingFailureHandler(GridAbstractTest test) {
+        this.test = test;
+    }
+
+    /** {@inheritDoc} */
+    @Override public boolean handle(Ignite ignite, FailureContext failureCtx) {
+        if (!testIsRunning.getAndSet(false)) {
```
dspavlov: Is it possible that a failure handler will be called by different threads with different exceptions? Will these exceptions be visible in a resulting test failure?

SomeFire: Yes, they will be visible, because in most cases error is printed before moving to failure handler. In other cases error is thrown to the upper code, where it is logged with additional actions.

modules/core/src/test/java/org/apache/ignite/failure/ExpectThrowableFailureHandler.java
SomeFire: Done.

PR #5691, opened 2018-12-18
IGNITE-10718: Merge XGBoost and Ignite ML trees together

dmitrievanthony:


modules/ml/src/main/java/org/apache/ignite/ml/preprocessing/minmaxscaling/MinMaxScalerPreprocessor.java
```diff
@@ -71,8 +71,13 @@ public MinMaxScalerPreprocessor(double[] min, double[] max, IgniteBiFunction<K,
         assert res.size() == min.length;
         assert res.size() == max.length;
 
-        for (int i = 0; i < res.size(); i++)
-            res.set(i, (res.get(i) - min[i]) / (max[i] - min[i]));
+        for (int i = 0; i < res.size(); i++) {
+            double scaled = (res.get(i) - min[i]) / (max[i] - min[i]);
+            if (Double.isNaN(scaled))
+                res.set(i, 0);
```
avplatonov: Such behavior can break logic of min-max scaler. In my opinion we should write like this:
double denom = max[i] - min[i]
(res.get(i) - min[i]) / (denom > EPS ? denom : 1.0);

modules/ml/src/main/java/org/apache/ignite/ml/tree/DecisionTreeConditionalNode.java
```diff
@@ -45,17 +48,29 @@
...
         this.elseNode = elseNode;
+        this.missingNode = missingNode;
     }
 
     /** {@inheritDoc} */
     @Override public Double apply(Vector features) {
-        return features.get(col) > threshold ? thenNode.apply(features) : elseNode.apply(features);
+        double val = features.get(col);
+
+        if (Double.isNaN(val)) {
```
avplatonov: maybe we save old logic of NaN processing?
like: 

if (Double.isNaN(val) && missingNode != null) 
   return missingNode.apply(features);
else 
   return val > threshold ? thenNode.apply(features) : elseNode.apply(features);

avplatonov: in terms of compatibility with old version?

modules/ml/src/main/java/org/apache/ignite/ml/Model.java
```diff
@@ -17,11 +17,13 @@
...
 package org.apache.ignite.ml;
 
+import java.io.Serializable;
 import java.util.function.BiFunction;
+import org.apache.ignite.ml.inference.InfModel;
 import org.apache.ignite.ml.math.functions.IgniteFunction;
 
 /** Basic interface for all models. */
-public interface Model<T, V> extends IgniteFunction<T, V> {
+public interface Model<T, V> extends InfModel<T, V>, IgniteFunction<T, V>, Serializable {
```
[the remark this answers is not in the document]
dmitrievanthony: Agree, removed.

modules/ml/src/main/java/org/apache/ignite/ml/inference/InfModel.java
```diff
@@ -17,20 +17,22 @@
...
      * Make a prediction for the specified input arguments.
      *
      * @param input Input arguments.
      * @return Prediction result.
      */
-    public O predict(I input);
+    public O apply(I input);
 
     /** {@inheritDoc} */
     @Override public void close();
```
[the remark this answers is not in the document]
dmitrievanthony: Agree, removed.

modules/ml/src/main/java/org/apache/ignite/ml/preprocessing/minmaxscaling/MinMaxScalerPreprocessor.java
dmitrievanthony: Disagree. The min max scaler is a function that maps on [0, 1] interval. In you example the result value is not necessary in this interval. Regarding choosing the exact value from this interval, I reproduced sklearn behavior.

modules/ml/src/main/java/org/apache/ignite/ml/tree/DecisionTreeConditionalNode.java
dmitrievanthony: The result of old logic was a lot of hidden mistakes (NaN values). Current solution allows to eliminate them. Do you think that it's better to hide such mistakes?

modules/ml/src/main/java/org/apache/ignite/ml/preprocessing/minmaxscaling/MinMaxScalerPreprocessor.java
dmitrievanthony: Okay, new version keeps information when it's possible.

modules/ml/src/main/java/org/apache/ignite/ml/inference/InfModel.java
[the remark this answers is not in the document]
dmitrievanthony: I had to return it back because without it we assume that `close()` throws Exception.

PR #5709, opened 2018-12-20
IGNITE-10076 Web Console: Use slow 'json-bigint' parser only for queries

akuznetsov-gridgain:


modules/web-console/frontend/app/modules/agent/AgentManager.service.js
```diff
@@ -485,8 +485,13 @@ export default class AgentManager {
                         if (cluster.secured)
                             this.clustersSecrets.get(cluster.id).sessionToken = res.sessionToken;
 
-                        if (res.zipped)
-                            return this.pool.postMessage(res.data);
+                        if (res.zipped) {
+                            const taskId = _.get(params, 'taskId', '');
+
+                            const useBigIntJson = taskId.startsWith('querySql') || taskId.startsWith('queryFetch');
```
nva: Just check for 'query'

PR #5710, opened 2018-12-20
IGNITE-10648 Ignite hang to stop if node wasn't started completely. G…

vldpyatkov:
…ridTcpRestNioListener hangs on latch.


modules/core/src/main/java/org/apache/ignite/internal/processors/rest/protocols/tcp/GridTcpRestNioListener.java
```diff
@@ -188,95 +189,102 @@ void marshallers(Map<Byte, GridClientMarshaller> marshMap) {
...
+                    } catch (IgniteInterruptedCheckedException e) {
+                        U.error(log, "Marshaller does not initialized.", e);
 
                         ses.close();
-                    }
-                    else {
-                        ses.addMeta(MARSHALLER.ordinal(), marsh);
 
-                        ses.send(GridClientHandshakeResponse.OK);
+                        Thread.currentThread().interrupt();
```
voropava: You don't need to interrupt thread, it is already interrupted.

voropava: Actually you need to return; here.

PR #5711, opened 2018-12-20
IGNITE-10748: remove dead code

Mmuzaf:


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/HandshakeTimeoutException.java
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.spi.communication.tcp.internal;
+
+import org.apache.ignite.IgniteCheckedException;
+import org.apache.ignite.spi.IgniteSpiOperationTimeoutException;
```
sergey-chugunov-1985: Import is unused.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -3274,7 +3313,7 @@ protected GridCommunicationClient createTcpClient(ClusterNode node, int connIdx)
 
             long needWaitDelay0 = 200;
```
sergey-chugunov-1985: Initial wait delay and max delay below (60_000) are hardcoded but deserve their own (documented) constants.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -3650,34 +3693,33 @@ private void safeShmemHandshake(
...
 
-        long rcvCnt = 0;
+        long rcvCnt;
 
         try {
             BlockingSslHandler sslHnd = null;
 
             ByteBuffer buf;
 
+            // Step 1. Get remote node respose with remote nodeId.
```
sergey-chugunov-1985: Typo in word `response`.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -3753,134 +3795,98 @@ else if (log.isDebugEnabled())
...
-                else {
-                    msg = new HandshakeMessage(locNode.id(),
-                        recovery.incrementConnectCount(),
-                        recovery.received());
-                }
-
-                if (log.isDebugEnabled())
-                    log.debug("Writing handshake message [locNodeId=" + locNode.id() +
-                        ", rmtNode=" + rmtNodeId + ", msg=" + msg + ']');
+            // Step 2. Prepare Handshake message to send remote node.
```
sergey-chugunov-1985: Is it easy to separate each step into a private method? In that case code of the `safeTcpHandshake` would become self-documented.

Mmuzaf: Sergey, I've tried to avoid huge changes and focused only on removing not used code.
In general, I agree with you here - better to have separate methods on each step of handshaking process, but in fact, this method used only in the signle flow and single time and we can leave it as it is.
WDYT?

PR #5725, opened 2018-12-21
IGNITE-10732 Force -Dfile.encoding=UTF-8 

alamar:
in linux/windows batch runners, C# and C++.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
@@ -29,6 +29,7 @@ import java.io.Serializable;
 import java.lang.management.ManagementFactory;
 import java.lang.management.RuntimeMXBean;
 import java.lang.reflect.Constructor;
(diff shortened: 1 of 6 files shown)


modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -1387,6 +1392,20 @@ private void validateCommon(IgniteConfiguration cfg) {
...
     }
 
+    /**
+     * Check whether UTF-8 is the default character encoding.
+     * Differing character encodings across cluster may lead to erratic behavior.
+     */
+    private void checkFileEncoding() {
+        String encodingDisplayName = Charset.defaultCharset().displayName(Locale.ENGLISH);
+
+        if (!"UTF-8".equals(encodingDisplayName)) {
```
dspavlov: Probably we can warn user only if we detect different encoding are applied cluster-wide. I'm not sure we should warn here if we did not validate overall settings.

PR #5727, opened 2018-12-21
IGNITE-10543: Test/train sample generator

avplatonov:


modules/ml/src/main/java/org/apache/ignite/ml/util/generators/primitives/scalar/UniformRandomProducer.java
```diff
@@ -0,0 +1,66 @@
...
+     * Creates an instance of UniformRandomProducer.
+     *
+     * @param from Generate values from this value.
+     * @param to Generate values to this value.
+     * @param seed Seed.
+     */
+    public UniformRandomProducer(double from, double to, long seed) {
+        super(seed);
+
+        A.ensure(to >= from, "from > to");
```
dmitrievanthony: Doesn't look like conditions are same.

modules/ml/src/main/java/org/apache/ignite/ml/util/generators/primitives/vector/VectorGeneratorPrimitives.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Returns vector generator of vectors from multidimension gauss distribution.
+     *
+     * @param means Mean values per dimension.
+     * @param variances Variance values per dimension.
+     * @param seed Seed.
+     * @return Generator.
+     */
+    public static VectorGenerator gauss(Vector means, Vector variances, Long seed) {
+        A.notEmpty(means.asArray(), "mean.size() != 0");
```
dmitrievanthony: Do we actually need to convert vector into array just to check that it's not empty?

modules/ml/src/main/java/org/apache/ignite/ml/util/generators/primitives/vector/VectorGeneratorPrimitives.java
```diff
@@ -0,0 +1,154 @@
...
+            .vectorize(2)
+            .filter(v -> Math.sqrt(v.getLengthSquared()) <= radius);
+    }
+
+    /**
+     * @param size Vector size.
+     * @return Generator of constant vector = zero.
+     */
+    public static VectorGenerator zero(int size) {
+        return constant(VectorUtils.of(new double[size]));
```
dmitrievanthony: I guess we could use `VectorUtils.zeroes(size)` here.

modules/ml/src/main/java/org/apache/ignite/ml/util/generators/primitives/scalar/DiscreteRandomProducer.java
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * Creates a producer of random values from uniform discrete distribution.
+     *
+     * @param numberOfValues Number of distinct values.
+     * @param seed Seed.
+     * @return Producer.
+     */
+    public static DiscreteRandomProducer uniform(int numberOfValues, long seed) {
+        return new DiscreteRandomProducer(seed, IntStream.range(0, numberOfValues)
```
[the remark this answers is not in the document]
avplatonov: yes, you're right

modules/ml/src/main/java/org/apache/ignite/ml/util/generators/primitives/vector/VectorGeneratorPrimitives.java
avplatonov: thank you

PR #5730, opened 2018-12-21
IGNITE-10316 control.sh --baseline remove outputs wrong error message…

vldpyatkov:
… when trying to remove node from baseline when cluster is inactive.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/visor/baseline/VisorBaselineTask.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/visor/baseline/VisorBaselineTask.java
@@ -32,6 +32,7 @@ import org.apache.ignite.internal.util.typedef.F;
 import org.apache.ignite.internal.util.typedef.internal.S;
 import org.apache.ignite.internal.visor.VisorJob;
 import org.apache.ignite.internal.visor.VisorOneNodeTask;
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 /**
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/ignite/internal/visor/baseline/VisorBaselineTask.java
```diff
@@ -32,6 +32,7 @@
 import org.apache.ignite.internal.util.typedef.internal.S;
 import org.apache.ignite.internal.visor.VisorJob;
 import org.apache.ignite.internal.visor.VisorOneNodeTask;
+import org.jetbrains.annotations.NotNull;
```
antonovsergey93: We don't use NotNull annotation https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines#CodingGuidelines-@Annotations

modules/core/src/main/java/org/apache/ignite/internal/visor/baseline/VisorBaselineTask.java
```diff
@@ -173,11 +174,13 @@ private VisorBaselineTaskResult add(List<String> consistentIds) {
         private VisorBaselineTaskResult remove(List<String> consistentIds) {
             Map<String, BaselineNode> baseline = currentBaseLine();
 
-            for (String consistentId : consistentIds) {
-                BaselineNode node = baseline.remove(consistentId);
+            if (!baseline.isEmpty()) {
```
antonovsergey93: Please, use F.isEmpty() instead of

PR #5744, opened 2018-12-25
IGNITE-10803:[ML] Add prototype LogReg loading from PMML format

zaleslaw:


examples/pom.xml
```diff
@@ -122,6 +122,30 @@
             <version>${javassist.version}</version>
             <scope>test</scope>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/org.jpmml/pmml-evaluator -->
+        <dependency>
+            <groupId>org.jpmml</groupId>
+            <artifactId>pmml-evaluator</artifactId>
```
vruusmann: Consider directly depending on `org.jpmml:pmml-model(:1.4.7)` and `org.jpmml:pmml-model-metro(:1.4.7)`. Also, the PMML stuff really doesn't need any of the listed Jackson dependencies.

zaleslaw: @vruusmann Great, thank you for the comment.
But without Jackson dependencies I've got a problem with JavaDoc generation.

zaleslaw: Warning:java: unknown enum constant com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.ANY
  reason: class file for com.fasterxml.jackson.annotation.JsonAutoDetect$Visibility not found

Warning:java: unknown enum constant com.fasterxml.jackson.annotation.JsonInclude.Include.NON_EMPTY
  reason: class file for com.fasterxml.jackson.annotation.JsonInclude$Include not found

Warning:java: unknown enum constant com.fasterxml.jackson.annotation.JsonTypeInfo.Id.NAME
  reason: class file for com.fasterxml.jackson.annotation.JsonTypeInfo$Id not found

For example

nizhikov: Hello @zaleslaw

Please, clarify, why you add jackson dependencies to examples module?
It brokes Spark Data Frame examples.

zaleslaw: @nizhikov Sorry for the delay, currently, did you fixed that or I should fix this moment? Yes, I test with disabled scala profile probably and this is my mistake and I've ready to fix it

nizhikov:
Hello, @zaleslaw.

Thanks for the response!

Yes, I fixed that, already.

You can see my commit, if interested - https://github.com/apache/ignite/commit/678e4455246cd6d000e4bd874524355afdf12806

Also, It's not only your fault :)
There was some misconfiguration on TC, so tests you broke with this change weren't tested.

examples/pom.xml
nizhikov: Hello, @zaleslaw.

Thanks for the response!

Yes, I fixed that, already.

You can see my commit, if interested - 678e445

Also, It's not only your fault :)
There was some misconfiguration on TC, so tests you broke with this change weren't tested.

PR #5778, opened 2019-01-06
IGNITE-5983: Enable obtaining memory policy metrics via REST.

shroman:


modules/clients/src/test/java/org/apache/ignite/internal/processors/rest/JettyRestProcessorAbstractSelfTest.java
```diff
@@ -1975,6 +1979,54 @@ public void testVisorGateway() throws Exception {
...
+    }
+
+    /**
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testDataStorageMetricsEnabled() throws Exception {
+        stopAllGrids();
+
+        memoryMetricsEnabled = true;
```
AMashenkov: Test execution order is not defined usually.
Let's add additional check and restart the grid if this option has to be changed.
Same change need to be done in testDataStorageMetricsDisabled().

PR #5800, opened 2019-01-10
IGNITE-10866: Add an example of LogRegression model loading from Spark via parquet file

zaleslaw:
…k model saved in parquet


zaleslaw:
Dropped the piece of code         /**
         * Prints parquet group of data.
         * NOTE: Recursive procedure.
         *
         * @param g Group for printing.
         */
        private static void printGroup(Group g) {
            int fieldCnt = g.getType().getFieldCount();
            for (int field = 0; field < fieldCnt; field++) {
                int valCnt = g.getFieldRepetitionCount(field);

                Type fieldType = g.getType().getType(field);
                String fieldName = fieldType.getName();

                for (int idx = 0; idx < valCnt; idx++) {
                    if (fieldType.isPrimitive())
                        System.out.println(fieldName + " " + g.getValueToString(field, idx));
                    else
                        printGroup(g.getGroup(field, idx));
                }
            }
            System.out.println();
        }

PR #5852, opened 2019-01-17
IGNITE-10933 Node may hang on join to topology

ascherbakoff:


yzhdanov:
Changes look good to me

yzhdanov:
Looks good to me

PR #5869, opened 2019-01-21
IGNITE-10993 Web console cancel query

vsisko:


modules/web-console/frontend/app/components/page-queries/components/queries-notebook/controller.ts
```diff
@@ -1904,6 +2057,22 @@ export class NotebookCtrl {
                 $modal({scope, templateUrl: messageTemplateUrl, show: true});
             }
         };
+
+        $window.addEventListener('beforeunload', () => {
```
Klaster1: 1. Every subsequent page enter will add new event listener. You should at least remove it when leaving the state.
2. I'm not sure this should be a thing at all. Maybe consider cancelling queries when client session is over (but this might cause issues with faulty network connection). Or just leave queries running, we don't cancel any other requests or long running operations (or do we?).

modules/web-console/frontend/app/modules/agent/AgentManager.service.js
```diff
@@ -762,6 +762,21 @@ export default class AgentManager {
             });
     }
 
+    /**
+     * @param {String} nid Node id.
+     * @param {String} queryId Query ID.
+     * @param {Number} pageSize
+     * @returns {Promise}
```
Klaster1: `Promise` of what? Please add task return type, like this:
```
/**
* @typedef {Object} QueryFetchFirstPageResult
* @prop {string} valueA
* @prop {number} valueB
*/
/*
* @returns {Promise<QueryFetchFirstPageResult>}
*/
```

akuznetsov-gridgain:
@vsisko  I think we also should show "You have running queries. Are you sure you want to cancel them?" on leaving Queries screen.

modules/web-console/frontend/app/components/page-queries/components/queries-notebook/controller.ts
Klaster1: This listener will most likely leak references.

PR #5887, opened 2019-01-22
IGNITE-11023 Processing data bag on GridMarshallerMappingProcessor co…

vldpyatkov:
…nsume many time


modules/core/src/main/java/org/apache/ignite/internal/MarshallerContextImpl.java
```diff
@@ -196,6 +197,11 @@ public void onMappingDataReceived(byte platformId, Map<Integer, MappedName> mars
             int typeId = e.getKey();
             String clsName = e.getValue().className();
 
+            MappedName mappedName = platformCache.get(typeId);
+
+            if (mappedName != null && !F.isEmpty(clsName) && clsName.equals(mappedName.className()))
```
ascherbakoff: There is a logic in mergeAndWriteMapping for checking metadata discrepancies. You should return it back if cached metadata do not match received metadata.

PR #5891, opened 2019-01-22
Ignite-10921

arturmuradimov77:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/filename/IgniteUidAsConsistentIdMigrationTest.java
```diff
@@ -587,6 +589,47 @@ public void testStartOldStyleNoPortsNodesByCompatibleProperty() throws Exception
...
     }
 
+    /**
+     * Test start node with defined consistent id through system property.
+     * Expected to be 1 folder with specified name.
+     *
+     * @throws Exception if failed.
+     */
+    @Test
+    public void testStartNodeWithDefinedConsistentIdBySystemProperty() throws Exception {
```
slukyano: To be honest, I don't like having the test here.
It is a migration test comparing the old and new consistent ID generation when it is not set which is different from testing how it can be set explicitly.

Can you add a separate test with the following cases
- consistentID is not set in config, not set as a property
- consistentID is set in config, not set as a property
- consistentID is not set in config, set as a property
- consistentID is set in config, set as a property

In each case check that the resulting consistentID is as expected. In the first case it will be just a random UUID (you can check it via regexp, for example). In the second case it will be the one from the config. In the third and fourth it will be the one from the property.

modules/core/src/test/java/org/apache/ignite/internal/ConsistentIdImplicitlyExplicitlyTest.java
```diff
@@ -0,0 +1,189 @@
...
+     *
+     * @throws Exception if failed.
+     */
+    @Test
+    public void testConsistentIdNotConfigured() throws Exception {
+        Ignite ignite = startGrid(0);
+
+        assertNull(ignite.configuration().getConsistentId());
+
+        assertTrue(ADDR_WITH_PORT_PATTERN.matcher(ignite.cluster().localNode().consistentId().toString()).find());
```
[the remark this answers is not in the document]
arturmuradimov77: I prefer to use pattern, because they will by compile statically.

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
```diff
@@ -2189,6 +2190,11 @@ private IgniteConfiguration initializeConfiguration(IgniteConfiguration cfg)
 
             myCfg.setNodeId(nodeId);
 
+            String predefineConsistentId = IgniteSystemProperties.getString(IGNITE_CONSISTENT_ID);
+
+            if (!F.isEmpty(predefineConsistentId))
```
alamar: predefineConsistentId should not take precedence over the one specified in configuration. Please only use predefineConsistentId if no consistentId is supplied in cfg.

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -639,6 +639,14 @@
      */
     public static final String IGNITE_CONSISTENT_ID_BY_HOST_WITHOUT_PORT = "IGNITE_CONSISTENT_ID_BY_HOST_WITHOUT_PORT";
 
+    /**
+     * System property to specify consistent id of Ignite node.
+     * <p>
+     * Value of the system property will overwrite matched property
+     * {@link org.apache.ignite.configuration.IgniteConfiguration#setConsistentId(Serializable)} in configuration.
+     */
+    public static final String IGNITE_CONSISTENT_ID = " IGNITE_CONSISTENT_ID";
```
alamar: Please remove whitespace  in constant (did it ever work?)

modules/core/src/test/java/org/apache/ignite/testsuites/IgniteBasicTestSuite.java
```diff
@@ -208,6 +209,7 @@
     ListeningTestLoggerTest.class,
 
     CacheLocalGetSerializationTest.class,
+    ConsistentIdImplicitlyExplicitlyTest.class
```
alamar: Looks like whitespace is expected before this line.

modules/core/src/test/java/org/apache/ignite/internal/ConsistentIdImplicitlyExplicitlyTest.java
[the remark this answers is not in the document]
alamar: This is irrelevant since this test will be only ran once. Please change as adviced, save a few lines.

modules/core/src/test/java/org/apache/ignite/testsuites/IgniteBasicTestSuite.java
arturmuradimov77: @alamar But Ignite code guide speak other.
https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines#CodingGuidelines-MethodArguments

alamar: This is not really a method argument, anyway it is preferred when neighboring  lines do not conflict

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
arturmuradimov77: I am sure that I mean the overwriting property.
I think, it is general approach in Ignite (see IGNITE_OVERRIDE_MCAST_GRP, IGNITE_REFLECTION_CACHE_SIZE, IGNITE_LOCAL_HOST, IGNITE_DEP_MODE_OVERRIDE).

modules/core/src/test/java/org/apache/ignite/internal/ConsistentIdImplicitlyExplicitlyTest.java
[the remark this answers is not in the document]
arturmuradimov77: However I think it correct using static regular expression. Is any reason why I need to using plain regexp here?
Or you are insist?

modules/core/src/test/java/org/apache/ignite/testsuites/IgniteBasicTestSuite.java
zstan: whitespace or new line? anyway no one need here, code correct.

modules/core/src/test/java/org/apache/ignite/internal/ConsistentIdImplicitlyExplicitlyTest.java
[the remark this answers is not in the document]
zstan: i think it`s correct too.

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
alamar: Please rename it to IGNITE_OVERRIDE_CONSISTENT_ID so that behavior is apparent

modules/core/src/test/java/org/apache/ignite/testsuites/IgniteBasicTestSuite.java
alamar: @zstan it is obvious from above code that there is logical grouping. However, ConsistentIdImplicitlyExplicitlyTest does not group logically with ConsistentIdImplicitlyExplicitlyTest. I think it should be either separated by extra newline or grouped logically with some other classes about system properties or consistent id.

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
vldpyatkov: Done.

arturmuradimov77: Renamed

modules/core/src/test/java/org/apache/ignite/testsuites/IgniteBasicTestSuite.java
arturmuradimov77: Added new empty line.

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -639,6 +639,14 @@
      */
     public static final String IGNITE_CONSISTENT_ID_BY_HOST_WITHOUT_PORT = "IGNITE_CONSISTENT_ID_BY_HOST_WITHOUT_PORT";
 
+    /**
+     * System property to specify consistent id of Ignite node.
+     * <p>
+     * Value of the system property will overwrite matched property
+     * {@link org.apache.ignite.configuration.IgniteConfiguration#setConsistentId(Serializable)} in configuration.
+     */
+    public static final String IGNITE_OVERRIDE_CONSISTENT_ID = "IGNITE_OVERRIDE_CONSISTENT_ID ";
```
alamar: Please remove whitespace from constant value (last char)

PR #5897, opened 2019-01-23
Ignite 9913 for TC tests

NSAmelchev:


modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -34,7 +34,10 @@
     TCP_COMMUNICATION_SPI_HANDSHAKE_WAIT_MESSAGE(0),
 
     /** Cache metrics v2 support. */
-    CACHE_METRICS_V2(1);
+    CACHE_METRICS_V2(1),
+
+    /** Local affinity exchange support. */
+    LOCAL_AFFINITY_RECALCULATION_EXCHANGE(2);
```
x-kreator: Current master already contains features with ids 2 & 3.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/CacheDataLossOnPartitionMoveTest.java
```diff
@@ -65,7 +65,7 @@
     public static final int GRIDS_CNT = 2;
 
     /** */
-    public static final String EVEN_GRP = "event";
+    public static final String EVEN_GRP = "even";
```
x-kreator: seems it typo?

PR #5904, opened 2019-01-23
IGNITE-8728: Set dynamic index name in QueryIndex.

slukyano:


modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/IgniteIndexNamesTest.java
```diff
@@ -0,0 +1,206 @@
...
+
+        // Check that repeated queries fail.
+        for (String sql : createIndexQueries) {
+            GridTestUtils.assertThrows(
+                log,
+                () -> {
+                    runSql(sql);
+                    return null;
+                },
+                useJdbc ? SQLException.class : IgniteSQLException.class,
```
AMashenkov: Should we add error code check here?
E.g. IgniteQueryErrorCode.INDEX_ALREADY_EXISTS

slukyano: Agree, fixed.

PR #5914, opened 2019-01-24
IGNITE-9927: Fix flaky failures in CacheContinuousQueryOperationFromCallbackTest.

rkondakov:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryOperationFromCallbackTest.java
```diff
@@ -445,20 +574,46 @@ public CacheTestRemoteFilterAsync(String cacheName) {
...
+                                cache.invoke(key, new IncrementTestEntryProcessor());
+                        }
+
+                        if (tx != null)
+                            tx.commit();
+
+                        committed = true;
+                    }
+                    catch (Exception ex) {
+                        assertTrue(ex.getMessage(), ex.getMessage() != null &&
```
AMashenkov: Why do we expect TransactionRollback here?
Let's check cause for TransactionSerializationException.

PR #5932, opened 2019-01-25
IGNITE-7883 add check property "key configuration" in cache configurations

a-polyakov:


dspavlov:
IGNITE-7883 | Resolved
-- | --

PR #5935, opened 2019-01-25
IGNITE-11059 Log tx lock timeout.

ivandasch:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/verify/ContentionClosure.java
```diff
@@ -105,6 +105,9 @@ public ContentionClosure(int minQueueSize, int maxPrint) {
                         catch (GridCacheEntryRemovedException ignored) {
                             cached = entry.context().cache().entryEx(entry.key());
                         }
+                        finally {
+                            cached.touch();
```
ascherbakoff: It seems I was wrong and touch is not needed here, because it could trigger various exceptions on tx prepare/lock acquisition. Touch will be called by code handling the specific transaction. Please remove.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtLockFuture.java
```diff
@@ -1154,6 +1157,11 @@ private void loadMissingFromStore() {
      * Lock request timeout object.
      */
     private class LockTimeoutObject extends GridTimeoutObjectAdapter {
+        /** */
+        final long longOpDumpTimeout = IgniteSystemProperties.getLong(
```
ascherbakoff: Adding long field overhead to every timeout object holding a constant looks wrong.
Better add new property to IgniteTxManager and create getter/setter. For example look at org.apache.ignite.internal.processors.cache.transactions.IgniteTxManager#slowTxWarnTimeout.
DFLT_LONG_OPERATIONS_DUMP_TIMEOUT will be not needed in such case (could be defined in IgniteTxManager).

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -320,6 +320,9 @@
     /** Separator for formatted coordinator properties. */
     public static final String COORDINATOR_PROPERTIES_SEPARATOR = ",";
 
+    /** Default long operations dump timeout. */
+    public static final long DFLT_LONG_OPERATIONS_DUMP_TIMEOUT = 60_000L;
```
ascherbakoff: Currently used only in LockTimeoutObject , but should be used in IgniteKernal too:
final long longOpDumpTimeout =
            IgniteSystemProperties.getLong(IgniteSystemProperties.IGNITE_LONG_OPERATIONS_DUMP_TIMEOUT, 60_000);

PR #5949, opened 2019-01-28
IGNITE-8841

gvvinblade:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/mvcc/PreviousQueries.java
```diff
@@ -0,0 +1,186 @@
...
+        /** */
+        @GridToStringInclude
+        boolean init;
+
+        /** */
+        @GridToStringInclude
+        Set<Long> cntrs;
+
+        /** */
+        boolean done() {
```
AMashenkov: Would name isDone clearer?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/mvcc/PreviousQueries.java
```diff
@@ -0,0 +1,186 @@
...
+
+            if (init && !hasQueries)
+                removeAndCheckDone(nodeId);
+            else
+                node.init = true;
+        }
+    }
+
+    /** */
+    private void removeAndCheckDone(@NotNull UUID nodeId) {
```
AMashenkov: 'check' look misleading here as we actually 'mark' as done, but not make a check.
Would e.g. 'finish(UUID)' clearer?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/mvcc/CacheMvccAbstractTest.java
```diff
@@ -1716,12 +1716,19 @@ protected MvccProcessorImpl mvccProcessor(Ignite node) {
      * @throws Exception If failed.
      */
     protected final void checkActiveQueriesCleanup(Ignite node) throws Exception {
-        final MvccProcessorImpl crd = mvccProcessor(node);
+        final MvccProcessorImpl prc = mvccProcessor(node);
```
AMashenkov: Let's rename to 'proc'

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearTxLocal.java
```diff
@@ -3336,6 +3346,102 @@ else if (mappings.single())
...
+    public IgniteInternalFuture<MvccSnapshot> requestSnapshot() {
+        if (isRollbackOnly())
+            return new GridFinishedFuture<>(rollbackException());
+
+        MvccSnapshot mvccSnapshot0 = mvccSnapshot;
+
+        if (mvccSnapshot0 != null)
+            return new GridFinishedFuture<>(mvccSnapshot0);
+
+        MvccProcessor prc = cctx.coordinators();
```
AMashenkov: Let's rename to 'proc'

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/mvcc/PreviousQueries.java
gvvinblade: possibly, will change

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/mvcc/PreviousQueries.java
gvvinblade: not sure, it is just a couple of doings which is being done in common. It's almost always difficult to come up with right name for such methods. There are two actions here: 1) removing the node from the active list, 2) checking whether the done flag can be set and setting it if it is so. I think the current name is more or less descriptive.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/mvcc/CacheMvccAbstractTest.java
gvvinblade: Not sure if it's really needed here. There are too many places where this name is used. We may easily rename all of them but in separate commit.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearTxLocal.java
gvvinblade: See my comment above

PR #5961, opened 2019-01-29
IGNITE-11222

artemmalykh:


modules/ml/src/main/java/org/apache/ignite/ml/naivebayes/discrete/DiscreteNaiveBayesTrainer.java
```diff
@@ -116,58 +116,58 @@
...
-
-                if (mdl != null && isUpdateable(mdl)) {
-                    if (checkSumsHolder(sumsHolder, mdl.getSumsHolder()))
-                        sumsHolder = sumsHolder.merge(mdl.getSumsHolder());
-                }
-
-                List<Double> sortedLabels = new ArrayList<>(sumsHolder.featureCountersPerLbl.keySet());
-                sortedLabels.sort(Double::compareTo);
-                assert !sortedLabels.isEmpty() : "The dataset should contain at least one feature";
+            DiscreteNaiveBayesSumsHolder sumsHolder = dataset.compute(t -> t, (a, b) -> {
```
dmitrievanthony: I don't understand what have you changed here?

modules/ml/src/main/java/org/apache/ignite/ml/naivebayes/gaussian/GaussianNaiveBayesTrainer.java
```diff
@@ -41,17 +43,10 @@
...
-     * @param datasetBuilder Dataset builder.
-     * @param featureExtractor Feature extractor.
-     * @param lbExtractor Label extractor.
-     * @return Model.
-     */
+    /** {@inheritDoc} */
     @Override public <K, V> GaussianNaiveBayesModel fit(DatasetBuilder<K, V> datasetBuilder,
-        IgniteBiFunction<K, V, Vector> featureExtractor, IgniteBiFunction<K, V, Double> lbExtractor) {
-        return updateModel(null, datasetBuilder, featureExtractor, lbExtractor);
+        IgniteBiFunction<K, V, SimpleLabeledVector<Double>> extractor) {
```
dmitrievanthony: Does it make sense to define a dedicated class like `FeatureLabelExtractor` with methods like `extract()`, `extractFeatures()` and `extractLabel()`.

modules/ml/src/main/java/org/apache/ignite/ml/clustering/kmeans/KMeansTrainer.java
```diff
@@ -82,14 +76,18 @@
 
     /** {@inheritDoc} */
     @Override protected <K, V> KMeansModel updateModel(KMeansModel mdl, DatasetBuilder<K, V> datasetBuilder,
-        IgniteBiFunction<K, V, Vector> featureExtractor, IgniteBiFunction<K, V, Double> lbExtractor) {
+        IgniteBiFunction<K, V, SimpleLabeledVector<Double>> extractor) {
```
zaleslaw: Why did you change this signature?

zaleslaw: Did you change examples according new signatures?

artemmalykh: `SimpleLabeledVector` is a `LabeledVector` with fixed first type parameter. It was created just not to write both parameters in place of usage, from the other hand after your comment I've rechecked the code usages of `LabeledVector` and found out that `V` type parameter is never used in the sense that we nowhere care what exact `Vector` subclass it is, we treat it just a `Vector`, so maybe I will refactor `LabeledVector` and remove this type parameter and after that remove `SimpleLabeledVector` which will become a full duplicate?

As for API. Separable extractors are expressed via inseparable but not vice versa in common case. That's why now the base method using inseparable extractors. From the other hand, you are right, all current trainers except boosting are fine with inseparable ones. Maybe we can add class like `SimpleDatasetTrainer` which has finalized methods with inseparable extractors expressed through abstract methods with separable extractors and subclass all our trainers from them? Idk, just didn't want to change class hierarchy...

PR #5995, opened 2019-01-31
IGNITE-10261 MVCC: cache operation may hang during late affinity assignment.

rkondakov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtGetFuture.java
```diff
@@ -340,7 +342,14 @@ private boolean map(KeyCacheObject key, boolean forceKeys) {
...
                 if (part.reserve()) {
-                    if (forceKeys || (part.state() == OWNING || part.state() == LOST)) {
+                    boolean mvccNoOwners = false;
+
+                    // Force key request is disabled for MVCC. So if there are no partition owners for the given key
+                    // we need to set flag forceKeys to true to avoid useless remapping to other non-owning partitions.
+                    if (cctx.mvccEnabled())
+                        mvccNoOwners = cctx.topology().owners(cctx.affinity().partition(key), topVer).isEmpty();
+
+                    if (forceKeys || (part.state() == OWNING || part.state() == LOST) || mvccNoOwners) {
```
AMashenkov: Should partition state be LOST if it has no owners?
Is additional flag really needed here?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtGetSingleFuture.java
```diff
@@ -311,7 +313,14 @@ private boolean map(KeyCacheObject key, boolean forceKeys) {
 
             // By reserving, we make sure that partition won't be unloaded while processed.
             if (part.reserve()) {
-                if (forceKeys || (part.state() == OWNING || part.state() == LOST)) {
+                boolean mvccNoOwners = false;
+
+                // Force key request is disabled for MVCC. So if there are no partition owners for the given key
+                // we need to set flag forceKeys to true to avoid useless remapping to other non-owning partitions.
+                if (cctx.mvccEnabled())
```
AMashenkov: Should partition state be LOST if it has no owners?
Is additional flag really needed here?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/IgnitePdsCacheStartStopWithFreqCheckpointTest.java
```diff
@@ -84,7 +83,7 @@ private CacheConfiguration cacheConfiguration(int cacheIdx) {
             .setAtomicityMode(CacheAtomicityMode.TRANSACTIONAL)
             .setCacheMode(CacheMode.REPLICATED)
             .setBackups(0)
-            .setRebalanceMode(CacheRebalanceMode.NONE);
+            .setRebalanceDelay(Long.MAX_VALUE);
```
AMashenkov: RebalanceMode.None deprecated in Mvcc mode only.
So, change in tests should not affect non-mvcc scenarios.

If is possible to set proper mode if MvccFeatureChecker.forceMvcc() == true?
Or rewrite rebalance mode in MvccProcessor like we do for Transactional atomicity mode?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtGetSingleFuture.java
```diff
@@ -311,7 +313,14 @@ private boolean map(KeyCacheObject key, boolean forceKeys) {
...
+                    mvccNoOwners = cctx.topology().owners(cctx.affinity().partition(key), topVer).isEmpty();
+
+                if (forceKeys || (part.state() == OWNING || part.state() == LOST) || mvccNoOwners) {
```
AMashenkov: Is additionl flag really needed here?
If partition has no owner and it is safe for READ operation, it should be in LOST state, isn't is?

PR #5999, opened 2019-02-01
IGNITE-10793: Examples for DataSteamGenerators

avplatonov:


modules/ml/src/main/resources/org/apache/ignite/ml/math/d3-dataset-template.html
```diff
@@ -0,0 +1,112 @@
...
+        width: 25px;
+        height: 25px;
+        margin-left: 5px;
+        vertical-align: bottom;
+    }
+
+</style>
+<body>
+<img style="margin-top: 15px" width="100px" src="https://ignite.apache.org/images/logo3.png">
+<script src="https://cdnjs.cloudflare.com/ajax/libs/d3/4.5.0/d3.min.js"></script>
```
dmitrievanthony: Does it make sense to put the library into local folder? Otherwise these examples won't work without internet connection.

avplatonov: just copy-paste from other tracers)

PR #6001, opened 2019-02-01
IGNITE-11137: IgniteModelStorage improvements for IgniteModel and SQL functionality

dmitrievanthony:


modules/ml/src/main/java/org/apache/ignite/ml/inference/IgniteModelStorageUtil.java
```diff
@@ -112,7 +135,20 @@ private static void saveModelStorage(byte[] serializedMdl, UUID mdlId) {
...
+    /**
+     * Removes model with specified identifier from model storage.
+     *
+     * @param mdlId Model identifier.
+     */
+    private static void removeModelStorage(UUID mdlId) {
+        Ignite ignite = Ignition.ignite();
+
+        ModelStorage storage = new ModelStorageFactory().getModelStorage(ignite);
+        storage.mkdirs(IGNITE_MDL_FOLDER);
```
avplatonov: there is no need to "mkdirs" here
you have added directories after first save

modules/ml/src/main/java/org/apache/ignite/ml/inference/IgniteModelStorageUtil.java
```diff
@@ -125,13 +161,30 @@ private static void saveModelDescriptorStorage(String name, UUID mdlId) {
...
+
+        boolean saved = descStorage.putIfAbsent(name, new ModelDescriptor(
+            mdlId.toString(),
             null,
             new ModelSignature(null, null, null),
             new ModelStorageModelReader(IGNITE_MDL_FOLDER + "/" + mdlId),
             new IgniteModelParser<>()
         ));
+
+        if (!saved)
```
avplatonov: What about flag force for put or another method "putIfAbsent"?
In my opinion such behavior is implicit. OR you should write about exception in javadoc

modules/ml/src/main/java/org/apache/ignite/ml/sql/SQLFunctions.java
```diff
@@ -17,18 +17,29 @@
...
  * SQL functions that should be defined and passed into cache configuration to extend list of functions available
  * in SQL interface.
  */
 public class SQLFunctions {
+    /** Default LRU model cache size. */
+    private static final int LRU_CACHE_SIZE = 10;
+
+    /** Default LRU model cache. */
+    // TODO: IGNITE-11163: Add hart beat tracker to DistributedInfModel.
+    private static final Map<String, Model<Vector, Double>> cache = Collections.synchronizedMap(
```
avplatonov: what about models that was changed after you locally save older model?

modules/ml/src/main/java/org/apache/ignite/ml/inference/IgniteModelStorageUtil.java
dmitrievanthony: I've updated javadoc.

modules/ml/src/main/java/org/apache/ignite/ml/inference/IgniteModelStorageUtil.java
dmitrievanthony: Removed.

modules/ml/src/main/java/org/apache/ignite/ml/sql/SQLFunctions.java
dmitrievanthony: Added thread that clears the cache.

modules/ml/src/main/java/org/apache/ignite/ml/sql/SQLFunctions.java
```diff
@@ -39,10 +70,15 @@
      */
     @QuerySqlFunction
     public static double predict(String mdl, Double... x) {
-        System.out.println("Prediction for " + Arrays.toString(x));
+        Model<Vector, Double> infMdl;
 
-        try (Model<Vector, Double> infMdl = IgniteModelStorageUtil.getModel(mdl)) {
-            return infMdl.predict(VectorUtils.of(x));
+        synchronized (cache) {
```
avplatonov: may be we can use double-check?
synchronization on every predict can be expensive I think

PR #6010, opened 2019-02-04
IGNITE-11183 Fixed proxy in Web Console agent.

nva:


modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentLauncher.java
```diff
@@ -375,13 +374,14 @@ public static void main(String[] args) throws Exception {
...
 
-            OkHttpClient sslFactory = builder.build();
-
-            opts.callFactory = sslFactory;
-            opts.webSocketFactory = sslFactory;
             opts.secure = true;
         }
 
+        OkHttpClient okHttpClient = builder.build();
+
```
akuznetsov-gridgain: It seems that after this change SSL will not work.

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/ProxyAuthenticator.java
```diff
@@ -0,0 +1,81 @@
...
+            Request req = res.request();
+            HttpUrl url = req.url();
+            Proxy proxy = route.proxy();
+
+            InetSocketAddress proxyAddr = (InetSocketAddress)proxy.address();
+
+            PasswordAuthentication auth = java.net.Authenticator.requestPasswordAuthentication(
+                proxyAddr.getHostName(), proxyAddr.getAddress(), proxyAddr.getPort(),
+                url.scheme(), challenge.realm(), challenge.scheme(), url.url(),
+                java.net.Authenticator.RequestorType.PROXY);
```
akuznetsov-gridgain: Refactor this const to static import.

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentLauncher.java
nva: All works fine. This was incorrect name of variable in original code.
`OkHttpClient sslFactory = builder.build()`

PR #6026, opened 2019-02-05
IGNITE-10444: MVCC: fix CacheMvccTxFastFinishTest

pavlukhin:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheMvccTxFastFinishTest.java
```diff
@@ -31,6 +33,8 @@
 public class CacheMvccTxFastFinishTest extends CacheTxFastFinishTest {
```
AMashenkov: Actually, fixed test is muted.
Let's fix this.

pavlukhin: Thanks for noting! Fixed.

PR #6037, opened 2019-02-06
IGNITE-5569 history of node ids to prevent failed nodes to rejoin the…

sergey-chugunov-1985:
… cluster


dspavlov:
IGNITE-5569 | Resolved
-- | --

PR #6048, opened 2019-02-06
IGNITE-11199 Add extra logging for client-server connections in TCP discovery.

6uest:


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -433,9 +433,30 @@
...
                 else {
-                    if (log.isInfoEnabled())
+                    if (log.isInfoEnabled()) {
+                        String serverNodeId = "";
+                        String consistentId = "";
+
+                        try {
+                            Collection<ClusterNode> remoteNodes = getSpiContext().remoteNodes();
+
+                            for (ClusterNode node : remoteNodes) {
```
antonovsergey93: Please merge two lines in single. i.e. `for (ClusterNode node : getSpiContext().remoteNodes())`

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -433,9 +433,30 @@
...
+                                if (U.toInetAddresses(node).contains(ses.remoteAddress().getAddress()) && !node.isClient()) {
```
antonovsergey93: We could get not all possible remote nodes, if we will have a few server nodes on one host.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -433,9 +433,30 @@
...
+                                    consistentId = node.consistentId().toString();
+
+                                    break;
+                                }
+                            }
+                        } catch (IgniteCheckedException e) {
+                            log.warning("Unable to acquire remote node ID from SpiContext: " + e, e);
+                        }
+
                         log.info("Established outgoing communication connection [locAddr=" + ses.localAddress() +
```
antonovsergey93: Please wrap `log.info()` to `if(log.isInfoEnabled())` condition.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -6014,7 +6014,8 @@ public void onInterruption() {
 
             if (log.isInfoEnabled())
                 log.info("Started serving remote node connection [rmtAddr=" + rmtAddr +
-                    ", rmtPort=" + sock.getPort() + ']');
+                    ", localNodeId=" + locNodeId +
```
antonovsergey93: We could try to detect which node is connecting to this node and print info about remote node.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
6uest: But `log.info()` already wrapped

antonovsergey93: Ok. I look over it.

PR #6050, opened 2019-02-06
IGNITE-11149 Update default value for IGNITE_DISABLE_WAL_DURING_REBAL…

6uest:
…ANCING in javadoc.


modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -930,9 +930,8 @@
     public static final String IGNITE_PART_DISTRIBUTION_WARN_THRESHOLD = "IGNITE_PART_DISTRIBUTION_WARN_THRESHOLD";
 
     /**
-     * When set to {@code true}, WAL will be automatically disabled during rebalancing if there is no partition in
```
antonovsergey93: Condition about absent partitions in OWNING state is important for understanding. Please leave it javadoc.

6uest: OK, did it

PR #6072, opened 2019-02-09
IGNITE-11276: Solve NPE problem, uses exchCtx from parent if state is…

ingvard:
… merged. mergeWith sets inside mux.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
@@ -478,7 +478,17 @@ public class GridDhtPartitionsExchangeFuture extends GridDhtTopologyFutureAdapte
         if (isDone())
             return result();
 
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -475,7 +475,12 @@ public void affinityChangeMessage(CacheAffinityChangeMessage affChangeMsg) {
          */
         assert exchangeDone() : "Should not be called before exchange is finished";
 
-        return isDone() ? result() : exchCtx.events().topologyVersion();
+        if (isDone())
+            return result();
+
+        ExchangeContext exchCtx0 = state == ExchangeLocalState.MERGED ? mergedWith.exchCtx : exchCtx;
```
agoncharuk: state read should be done in a synchronized section

ingvard: I see. I fixed it.

PR #6079, opened 2019-02-11
IGNITE-11234: Add fillCache method to DataStreamGenerator

avplatonov:


modules/ml/src/test/java/org/apache/ignite/ml/util/generators/DataStreamGeneratorTest.java
```diff
@@ -158,6 +171,39 @@ public void testAsDatasetBuilder() throws Exception {
...
+    @Test
+    public void testCacheFilling() {
+        IgniteConfiguration configuration = new IgniteConfiguration().setPeerClassLoadingEnabled(true);
+        String cacheName = "TEST_CACHE";
+        CacheConfiguration<UUID, LabeledVector<Double>> cacheConfiguration =
+            new CacheConfiguration<UUID, LabeledVector<Double>>(cacheName)
+                .setAffinity(new RendezvousAffinityFunction(false, 10));
+        int datasetSize = 5000;
+
+        try (Ignite ignite = Ignition.start(configuration)) {
```
dmitrievanthony: If it's okay to start Ignite manually in unit tests? I just thought that unit tests might be run in parallel and in this case we'll get unpredictable behavior.

modules/ml/src/main/java/org/apache/ignite/ml/util/generators/DataStreamGenerator.java
```diff
@@ -125,10 +135,54 @@ public default DataStreamGenerator blur(RandomProducer rnd) {
...
+     * @param datasetSize Rows count to put.
+     * @param cache Cache.
+     * @param keyMapper Mapping from vectors to keys.
+     * @param <K> Key type.
+     */
+    public default <K> void fillCacheWithCustomKey(int datasetSize, IgniteCache<K, LabeledVector<Double>> cache,
+        Function<LabeledVector<Double>, K> keyMapper) {
+
+        Map<K, LabeledVector<Double>> batch = new HashMap<>();
+        labeled().limit(datasetSize).forEach(vec -> {
```
dmitrievanthony: Does it make sense to make in in parallel?

modules/ml/src/test/java/org/apache/ignite/ml/util/generators/DataStreamGeneratorTest.java
avplatonov: Hm, maybe you're right. 
Okay, I'll move this test to "integration" test and use GridAbstracTest class.

modules/ml/src/main/java/org/apache/ignite/ml/util/generators/DataStreamGenerator.java
avplatonov: I don't know. I think benchmarks using this code will help to solve this problem.

modules/ml/src/main/java/org/apache/ignite/ml/util/generators/DataStreamGenerator.java
```diff
@@ -125,10 +135,54 @@ public default DataStreamGenerator blur(RandomProducer rnd) {
...
+            batch.put(keyMapper.apply(vec), vec);
+            if (batch.size() == FILL_CACHE_BATCH_SIZE) {
+                cache.putAll(batch);
```
thatb96:
```suggestion
                

- cache.putAll(batch);

```

avplatonov: What is it? On review page I just see that you deleted tab-character and put additional empty lines after batch size checking. Is it valid suggestion?

PR #6082, opened 2019-02-11
IGNITE-6563 Implement GetSizeLong

ivandasch:


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/CacheImpl.cs
```diff
@@ -817,36 +817,129 @@ public Task RemoveAllAsync()
...
             var modes0 = IgniteUtils.EncodePeekModes(modes);
 
-            var op = loc ? CacheOp.SizeLoc : CacheOp.Size;
+            var op = loc ? 
+                (typeof(T) == typeof(long) ? CacheOp.SizeLongLoc : CacheOp.SizeLoc) : 
+                (typeof(T) == typeof(long) ? CacheOp.SizeLong : CacheOp.Size);
+
+            long ret;
+            
+            if (typeof(T) == typeof(long))
```
ptupitsyn: Too complicated - we have to check for types, then convert the type.
Let's have two separate methods for `long` and `int`, they have very little in common.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/cache/PlatformCache.java
```diff
@@ -1069,6 +1089,27 @@ private void loadCache0(BinaryRawReaderEx reader, boolean loc) {
         return super.processOutObject(type);
     }
 
+    /** {@inheritDoc} */
+    @Override public long processInStreamOutLong(int type, BinaryRawReaderEx reader) throws IgniteCheckedException {
```
ptupitsyn: I think we can put `OP_SIZE_LONG` and `OP_SIZE_LONG_LOC` into `processInStreamOutLong` above as well, to avoid overriding one more method.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/CacheImpl.cs
ivandasch: Yes, it's seems to be a little bit overengineered.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/cache/PlatformCache.java
ivandasch: Yes, you are right.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/CacheImpl.cs
```diff
@@ -846,7 +880,67 @@ private int Size0(bool loc, params CachePeekMode[] modes)
...
+        /// Internal async size routine.
+        /// </summary>
+        /// <param name="part">Partition number</param>
+        /// <param name="modes">peek modes</param>
+        /// <returns>Size.</returns>
+        private Task<T> SizeAsync0<T>(int? part, params CachePeekMode[] modes)
+        {
+            var modes0 = IgniteUtils.EncodePeekModes(modes);
+
+            var op = typeof(T) == typeof(long) ? CacheOp.SizeLongAsync : CacheOp.SizeAsync;
```
ptupitsyn: Let's split this method in two as well to keep it simple and avoid type checks.

PR #6086, opened 2019-02-12
IGNITE-5569: prohibit join of node hidden behind firewall, prohibit join of node which ID is already presented in node IDs history

sergey-chugunov-1985:


dspavlov:
IGNITE-5569 | Resolved
-- | --

PR #6090, opened 2019-02-12
IGNITE-9480  SQL: Introduce H2 LocalResult factory

tledkov-gridgain:


katzyn:
`IgniteH2BaseLocalResult` is a slightly modified copy of [`org.h2.result.LocalResultImpl`](https://github.com/h2database/h2database/blob/master/h2/src/main/org/h2/result/LocalResultImpl.java) that was written by multiple authors and is dual-licensed under MPL 2.0 and EPL 1.0. (See also history of [`LocalResult`](https://github.com/h2database/h2database/blob/master/h2/src/main/org/h2/result/LocalResult.java) that was renamed to `LocalResultImpl` earlier and replaced with an interface.)

Your copy is licenced under Apache License 2.0 and does not contain original copyright headers.

I think you need either to preserve original licence and copyright headers (MPL 2.0 is more or less compatible with Apache License), or rewrite this class from the scratch, or inherit it and override methods that you want to change without copying large pieces of code.

PR #6093, opened 2019-02-12
IGNITE-11299 Avoid busy wait on processWrite during SSL handshake.

alamar:


Mmuzaf:
Hello Ilya,

I'm not an expert of CommunicationSpi :-) your solution looks applicable to me. But I have small questions about your PR, that are plaguing me. Can you clarify them?

I've looked at the method #onSessionWrite(). The `sslSys` variable, from my point, shows us that the stage of the current session interaction is on the `establishing SSL` state. I think we may not to register an OP_WRITE here as we don't need it at all for the SSL handshaking and it has its own queue to process such requests. Will it work?

https://github.com/apache/ignite/blob/master/modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java#L3545

`
boolean sslSys = sslFilter != null && msg instanceof ByteBuffer;

                if (sslSys) {
                    if (!ses0.procWrite.get() && ses0.procWrite.compareAndSet(false, true)) {
                        GridNioWorker worker = ses0.worker();

                        if (worker != null)
                            worker.registerWrite(ses0);
                    }

                    return null;
                }

`
In general, about OP_WRITE operation I think the approach can be: 
- If we have something to write we just try it. 
- If write() returns zero bytes then we will register OP_WRITE (and not OP_READ)
- If write completes successfully then deregister OP_WRITE and register an OP_READ.

alamar:
@Mmuzaf I'm not ready to deep comment TcpCommunicationSpi but I don't see how we will flush data to socket without setting OP_WRITE? We never get notified and never start writing.

SSL has its own handling in BlockingSslHandler, but not sure about GridNioServer.

Otherwise I'm not ready to rewrite TcpCommunicationSpi on the spot as you have suggested. The idea is to fix an obvious bug.

alamar:
Yes, as noted in the ticket, any SSL test such as TcpDiscoverySslTrustedUntrustedTest will hang on Java 11/Windows, but after this patch it would pass normally.

PR #6102, opened 2019-02-13
IGNITE-11313 Cluster hangs on cache invoke with binary objects creation

ibessonov:


modules/core/src/main/java/org/apache/ignite/internal/UnregisteredBinaryTypeException.java
```diff
@@ -28,56 +28,46 @@
     /** */
     private static final long serialVersionUID = 0L;
 
+    /** */
+    private static final String MESSAGE =
+        "Attempted to update binary metadata inside a critical synchronization block (will be " +
```
sergey-chugunov-1985: As I can see from the change this exception is also used when registering marshaller mapping which is slightly different thing as binary metadata.
What do you think about renaming this class to `UnregisteredMetadataException` (because marshaller mappings are also metadata) and creating two different constructors: one is called when binary metadata cannot be registered and another one when the same happens with marshaller mapping?

ibessonov: I'm absolutely positive with these changes, they'll be ready today soon.

PR #6118, opened 2019-02-16
IGNITE-11243 fix NPE if a cluster node has old version of VisorBaseli…

macrergate:
…neNode

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
@@ -1304,9 +1304,9 @@ public class CommandHandler {
         // if task runs on a node with VisorBaselineNode of old version (V1) we'll get order=null for all nodes.
 
         String crdStr = srvs.values().stream()
-            .min(Comparator.comparing(node -> node.getOrder() != null ? node.getOrder() : 0))
             // check for not null
             .filter(node -> node.getOrder() != null)
+            .min(Comparator.comparing(VisorBaselineNode::getOrder))
             // format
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -1302,14 +1301,20 @@ private void baselinePrint0(VisorBaselineTaskResult res) {
...
+        // 2) if we have a node with order == null, then CRD can't be evaluated, otherwise
+        // 3) the node with minimal order is CRD
+
+        String crdStr = srvs.values().stream()
+            // sort: 1, null, 2, 3, ...
+            .min(Comparator.comparing(node1 -> node1.getOrder() != null ? node1.getOrder() << 1 : 3))
+            // check for not null
+            .filter(node1 -> node1.getOrder() != null)
+            // format
+            .map(crd -> " (Coordinator: ConsistentId=" + crd.getConsistentId() + ", Order=" + crd.getOrder() + ")")
```
vldpyatkov: This place of code is difficult to understanding.
1) What mean doubled order (node1.getOrder() << 1)?
2) Why null order in third position? (Integer.MAX_VALUE why not?)
3) After that you filtered all nodes with null order, but min() return only one... may by with nulled order node.
Need to rewrite.

macrergate: thanks @vldpyatkov. I've simplified code. Actually we can have 2 cases only
1)  all nodes  order != null  in case task launched on a new node =>  crd = min node
2) all nodes orde == null in case task launched on old versioned node  => crd can't be specified.

PR #6119, opened 2019-02-17
IGNITE-11277 Added maven-checkstyle-plugin config and code style fixes

Mmuzaf:


nizhikov:
Hello, Maxim.

Thank you for the PR!

1. Should we remove idea inspection file? At lease, the one to run on TC?
2. What about other already enabled inspections? `SizeReplaceableByIsEmpty`, `RedundantSuppression`, `MissingOverrideAnnotation`?

Mmuzaf:
Hello, Nikolay.

Thank you for your comment.

1. After the PR will be merged (if it will) we will get Inspection Suite and Checkstyle working both simultaneously. After we make sure that checkstyle work as expected and community adopted to new build procedure (I expect a day or two) I will create a new issue to disable the Inspection Suite and delete corresponding inspections configuration. 
Is it a plan you expect?

2. The short answer is - there is no analogue of a checkstyle module to previously configured inspections. I've mentioned everything about it in my comment in JIRA. See: [IGNITE-11277?focusedCommentId=16771200](https://issues.apache.org/jira/browse/IGNITE-11277?focusedCommentId=16771200&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16771200)

parent/pom.xml
```diff
@@ -622,6 +622,37 @@
                         </execution>
                     </executions>
                 </plugin>
+
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-checkstyle-plugin</artifactId>
+                    <version>3.0.0</version>
```
vveider: I would recommend moving plugin's version to parent/pom.xml with `maven.checkstyle.plugin.version` name and use variable here instead.

parent/pom.xml
```diff
@@ -1278,5 +1309,25 @@
             </build>
         </profile>
 
+        <profile>
+            <id>checkstyle</id>
+
+            <activation>
+                <activeByDefault>true</activeByDefault>
```
vveider: Is it really have to be activated by default? That can significantly increase  build times. Let's activate it only in corresponding cases like intentional checkstyle run.

parent/pom.xml
Mmuzaf: Pert, thank you for your comments. 
Agree here. Fixed

parent/pom.xml
Mmuzaf: I've measured the increasing build time - it's about +10% with the enabled checkstyle plugin. 
Fixed too.

parent/pom.xml
```diff
@@ -1313,8 +1314,6 @@
             <id>checkstyle</id>
 
             <activation>
-                <activeByDefault>true</activeByDefault>
-
                 <property>
                     <name>skipTests</name>
```
vveider: Does this property definition means that when `-DskipTests` is called, `checkstyle` profile will be activated?
And if so - what's the point of this action? By calling `-DskipTests` we usually trying to skip as much of non build related steps as possible.

Mmuzaf: Yes, you are right. It's a discussable thing, but from my point, the style checking should be related to the build procedure.

vveider: Then I would insists to use it like `<name>!skipTests</name>`
In that case it will be activated every time when `-DskipTests` is not specified (because that flag used to quickly build project without tests, docs, and other stuff).

Mmuzaf: Petr,
Actually, it's not always true. 
`-DskipTests` - will compile source code, javadoc, skip test execution
`-Dmaven.javadoc.skip=true` - additionally will skip javadoc
`-Dmaven.test.skip=true` -- will skip tests compiling

So, I'd like to keep it as it is (enabled by default) and if we need a quick build - skip the checkstyle plugin by disabling it expliciy. For instance,

`mvn clean package -DskipTests -P !checkstyle`

http://maven.apache.org/plugins-archives/maven-surefire-plugin-2.12.4/examples/skipping-test.html

vveider: I cannot agree.
Checkstyle is another test test that count towards PR merge eligibility. With -DskipTests it should be skipped as well.

Mmuzaf: It's not a test, it's a code validation step and run before source code compilation. `skipTests` is related to the JUnit tests execution only and nothing more. 

As I've previously mentioned it's a discussable thing. Let's discuss it on the appropriate topic on dev-list.

PR #6120, opened 2019-02-18
IGNITE-11332 Add jmx ability to exclude node from topology.

zstan:


modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoverySpiMBeanTest.java
```diff
@@ -81,4 +86,62 @@ public void testMBean() throws Exception {
...
+
+            assertEquals(grid0.cluster().forClients().nodes().size(), 0);
+
+            bean.excludeNode(new UUID(0, 0).toString());
+
+            bean.excludeNode("fakeUUID");
+
+            assertEquals(grid0.cluster().forServers().nodes().size(), srvCnt);
+
+            ClusterNode node = grid0.cluster().forServers().nodes().stream().filter(n -> n.id() != grid0.localNode().id())
```
vldpyatkov: grid0.cluster().forServers().forRandom().node();
I believe it will be more simple.

modules/zookeeper/src/main/java/org/apache/ignite/spi/discovery/zk/ZookeeperDiscoverySpi.java
```diff
@@ -620,6 +620,11 @@ public ZookeeperDiscoverySpiMBeanImpl(IgniteSpiAdapter spiAdapter) {
             return String.valueOf(getLocalNode());
         }
 
+        /** {@inheritDoc} */
+        @Override public void excludeNode(String nodeId) {
+            impl.failNode(UUID.fromString(nodeId), "Node failed from JMX interface, initiator=" + getLocalNodeId());
```
vldpyatkov: You need to choose one term, like in TcpDiscovery.

PR #6130, opened 2019-02-18
IGNITE-11245 Replace unused IGNITE_BINARY_META_UPDATE_TIMEOUT parameter.

6uest:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/CacheObjectBinaryProcessorImpl.java
```diff
@@ -748,7 +748,7 @@ else if (holder == null || !holder.metadata().hasSchema(schemaId)) {
                 // Last resort waiting.
                 U.warn(log,
                     "Schema is missing while no metadata updates are in progress " +
-                        "(will wait for schema update within timeout defined by IGNITE_BINARY_META_UPDATE_TIMEOUT system property)" +
+                        "(will wait for schema update within timeout defined by IGNITE_WAIT_SCHEMA_UPDATE system property)" +
```
zstan: it would be more clear if you append real variable IGNITE_WAIT_SCHEMA_UPDATE into message, further code update will prevent from similar errors.

6uest: Done.

PR #6165, opened 2019-02-21
IGNITE-11382 Stopped cache's managers before caches

akalash:


dspavlov:
IGNITE-11382 | Resolved
-- | --

PR #6168, opened 2019-02-21
IGNITE-6578 Too many diagnostic: Found long running cache future

gromtech:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -1935,41 +1935,61 @@ private boolean dumpLongRunningOperations0(long timeout) {
         final IgniteDiagnosticPrepareContext diagCtx = cctx.kernalContext().cluster().diagnosticEnabled() ?
             new IgniteDiagnosticPrepareContext(cctx.localNodeId()) : null;
 
+        WarningsGroup warnings = new WarningsGroup(diagnosticLog, 10);
```
alamar: Can we please extract this number to IGNITE_foo system property?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -1935,41 +1935,61 @@ private boolean dumpLongRunningOperations0(long timeout) {
...
+                        warnings.add(">>> Transaction [startTime=" + formatTime(tx.startTime()) +
+                            ", curTime=" + formatTime(curTime) + ", tx=" + tx + ']');
+                    }
+                    else
+                        warnings.incTotal();
                 }
             }
         }
 
+        warnings.flush("First %d long running transactions [total=%d]");
```
alamar: It would be nicer if we won't reuse warnings, but create new object each time (of smallest visibility scope possible). This way you won't need cleanup code, just discard your object.

gromtech: The smallest visibility scope possible is a good approach. But wouldn't it be better to avoid multiple objects creation? Especially if they won't be used in most cases.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
gromtech: Yes, this makes sense. However other parts of this class don't make a similar parameter configurable.

https://github.com/apache/ignite/blob/26a15264ae69e4127f9e0b3e88830bf3edc56ece/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java#L1859

https://github.com/apache/ignite/blob/26a15264ae69e4127f9e0b3e88830bf3edc56ece/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java#L2729

alamar: I would propose to introduce a single sys var to configure all three of them. Use default values (5, 10) if it is not present.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
alamar: Nope, this code runs once per tens of seconds, correctness and brevity is much more important here

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
gromtech: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
gromtech: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -153,6 +154,9 @@
     private final long IGNITE_EXCHANGE_MERGE_DELAY =
         IgniteSystemProperties.getLong(IgniteSystemProperties.IGNITE_EXCHANGE_MERGE_DELAY, 0);
 
+    /** */
+    private final int DIAGNOSTIC_WARN_LIMIT = IgniteSystemProperties.getInteger(IGNITE_DIAGNOSTIC_WARN_LIMIT, 10);
```
alamar: Private final fields should not be in ALL_CAPS and should go after private static final. My suggestion is to compute it in dumpDebugInfo() every time. We are trying to avoid static fields since they have side-effects when running a lot of tests (can't change mid-flight).

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -1936,40 +1944,62 @@ private boolean dumpLongRunningOperations0(long timeout) {
...
+                    }
+                    else
+                        warnings.incTotal();
 
                     if (diagCtx != null && fut instanceof IgniteDiagnosticAware)
                         ((IgniteDiagnosticAware)fut).addDiagnosticRequest(diagCtx);
                 }
             }
 
+            warnings.flush("First %d long running cache futures [total=%d]", true);
```
alamar: My suggestion is to have two Warnings instances activeWarnings/atomicWarnings instead, removing the need of clear=true.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -3366,4 +3396,99 @@ private AffinityReadyFuture(AffinityTopologyVersion topVer) {
...
+        private void incTotal() {
+            warningsTotal++;
+        }
+
+        /**
+         * Print warnings block title and messages.
+         *
+         * @param title Title template.
+         */
+        private void flush(String title) {
```
alamar: Thus you can only keep one flush() method and simplify it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
gromtech: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
gromtech: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -1856,14 +1860,17 @@ public void dumpDebugInfo(@Nullable GridDhtPartitionsExchangeFuture exchFut) thr
...
+            U.warn(diagnosticLog, "First " + warningsLimit + " pending affinity ready futures [total=" +
+                readyFuts.size() + ']');
 
             int cnt = 0;
 
             for (AffinityReadyFuture fut : readyFuts.values()) {
                 U.warn(diagnosticLog, ">>> " + fut);
 
-                if (++cnt == 5)
+                if (++cnt == warningsLimit)
```
alamar: It looks like we can't have IGNITE_DIAGNOSTIC_WARN_LIMIT = 0 since at least one message will always be printed. Can we move this check up (while avoiding off-by-one error) so that we could completely disable those messages with IGNITE_DIAGNOSTIC_WARN_LIMIT=0? Here and below, everywhere where it is checked.

gromtech: That's a good point. Fixed.

PR #6174, opened 2019-02-23
IGNITE-11397 Binary mode for Ignite Set

udaykale:


modules/core/src/main/java/org/apache/ignite/internal/processors/datastructures/GridCacheSetProxy.java
```diff
@@ -405,6 +411,33 @@ public void blockOnRemove() {
...
+            null,
+            true,
+            null,
+            false,
+            null,
+            false,
+            DFLT_ALLOW_ATOMIC_OPS_IN_TX)
+            : opCtx.keepBinary();
+
+        CacheOperationContext prev = gate.enter(opCtx);
```
ibessonov: I don't understand why we need this gate around constructor invocation.

modules/core/src/main/java/org/apache/ignite/internal/processors/datastructures/GridCacheSetProxy.java
```diff
@@ -405,6 +411,33 @@ public void blockOnRemove() {
...
+
+        try {
+            return new GridCacheSetProxy<>(cctx, (GridCacheSetImpl<V1>)delegate, opCtx);
+        } finally {
```
ibessonov: btw, this violates code style rules, "finally" should be on a new line.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/datastructures/GridCacheSetAbstractSelfTest.java
```diff
@@ -67,11 +69,29 @@
...
     /** {@inheritDoc} */
     @Override protected int gridCount() {
         return 4;
     }
 
+    /**
+     * Intialize Ignite set
+     */
+    protected <T> IgniteSet<T> initSet(int idx, String name, @Nullable CollectionConfiguration cfg) {
+        IgniteSet<T> set = grid(idx).set(SET_NAME, cfg);
```
ibessonov: I think you should pass "name" parameter instead of "SET_NAME" constant.

modules/core/src/main/java/org/apache/ignite/IgniteSet.java
```diff
@@ -127,4 +127,12 @@
...
      */
     public <R> R affinityCall(IgniteCallable<R> job) throws IgniteException;
+
+    /**
+     * Returns set that will operate with binary objects. This is similar to {@link IgniteCache#withKeepBinary()} but
+     * for sets.
+     *
+     * @return New set instance for binary objects.
+     */
+    public <V1> IgniteSet<V1> withKeepBinary();
```
ibessonov: I'll leave my main comment here. I see no specific test that checks that binary mode for set actually gives you binary values. Can you please add it somewhere?

PR #6175, opened 2019-02-23
IGNITE-11398: Cleanup leftover @RunWith(JUnit4.class)

pavlukhin:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/IgnitePdsSporadicDataRecordsOnBackupTest.java
```diff
@@ -1,247 +1,244 @@
-/*
```
pavlukhin: Converted CRLF -> LF in this file.

PR #6177, opened 2019-02-24
IGNITE-11380 Add support of JAR files to UriDeploymentSpi

dmekhanikov:


modules/extdata/uri/pom.xml
```diff
@@ -128,6 +220,32 @@
...
+                            <target>
+                                <mkdir dir="${basedir}/target/file_tmp/classes/org/apache/ignite/spi/deployment/uri/tasks/" />
+                                <copy file="${basedir}/target/classes/org/apache/ignite/spi/deployment/uri/tasks/GridUriDeploymentTestTask11.class" todir="${basedir}/target/file_tmp/classes/org/apache/ignite/spi/deployment/uri/tasks/" />
+                                <copy file="${basedir}/target/classes/org/apache/ignite/spi/deployment/uri/tasks/GridUriDeploymentTestTask11.class" tofile="${basedir}/target/file_tmp/classes/org/apache/ignite/spi/deployment/uri/tasks/GridUriDeploymentTestWithNameTask11.class" />
+
+                                <jar destfile="${basedir}/target/file/bad-signed-deployfile.jar" basedir="${basedir}/target/file_tmp/classes" />
+
+                                <signjar jar="${basedir}/target/file/bad-signed-deployfile.jar" keystore="${basedir}/config/signeddeploy/keystore" storepass="abc123" keypass="abc123" alias="business" />
+
+                                <sleep seconds="2" />
```
alamar: Why?

modules/extdata/uri/src/main/java/org/apache/ignite/spi/deployment/uri/tasks/GridUriDeploymentTestWithNameTask10.java
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.ignite.compute.ComputeTaskSplitAdapter;
+
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * URI deployment test task with name.
+ */
+@ComputeTaskName("GridUriDeploymentTestWithNameTask10")
+public class GridUriDeploymentTestWithNameTask10 extends ComputeTaskSplitAdapter<Object, Object> {
```
alamar: What's the point of these three empty tasks? Maybe log something at least?

modules/urideploy/src/main/java/org/apache/ignite/spi/deployment/uri/GridUriDeploymentUnitDescriptor.java
```diff
@@ -142,17 +143,17 @@ public Type getType() {
...
+     * @return {@code null} if tasks were deployed directly and reference to the package file URI if tasks
+     * were deployed from the file.
      */
     public String getUri() {
         return uri;
     }
 
     /**
-     * Tasks GAR file.
+     * Tasks package file.
```
alamar: Task's? Tasks'?

modules/extdata/uri/src/main/java/org/apache/ignite/spi/deployment/uri/tasks/GridUriDeploymentTestWithNameTask10.java
dmekhanikov: We don't even run them. We only check if we can load these classes. So, it doesn't make much sense to put anything inside.

modules/extdata/uri/pom.xml
dmekhanikov: If JAR assembly and running this Ant script happens within one second, then the `zip` task won't detect any changes and leave the JAR not updated.
This `sleep` and `touch` bellow update the timestamp of the file, so the `zip` task doesn't ignore the file.
The same thing happens in the code bellow, where a similar GAR file is assembled.

dmekhanikov:
@alamar This behaviour is changed actually. If you have JARs lying around your repository, then they will be also loaded. 
We could introduce a new configuration property to make people enable this new feature and acknowledge the fact, that their JARs will also be loaded now. But I don't think, it's too much of a problem, so another piece of configuration seems to be an overkill to me.

PR #6180, opened 2019-02-25
IGNITE-11322 [USABILITY] Extend Node FAILED message by add consistent…

6uest:
…Id if it exist


6uest:
@antonovsergey93 Your solution looks more complex. Thank you for advice. Will try it.

PR #6230, opened 2019-03-05
IGNITE-11419 Memory leak after multiple restarts of server node withi…

pvinokurov:
…n the same thread


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/ConnectionManager.java
```diff
@@ -379,6 +379,7 @@ private void closeConnections() {
 
         threadConns.clear();
         detachedConns.clear();
+        threadConn.remove();
```
alamar: Hello!

Can we add a test for that please? I guess it should be pretty straightforward.

PR #6235, opened 2019-03-05
IGNITE-8376 Cluster activation event added

ibelyakov:


modules/core/src/main/java/org/apache/ignite/events/ClusterActivationEvent.java
```diff
@@ -0,0 +1,72 @@
...
+ * It is <b>highly recommended</b> to enable only those events that your application logic requires
+ * by using {@link IgniteConfiguration#getIncludeEventTypes()} method in Ignite configuration. Note that certain
+ * events are required for Ignite's internal operations and such events will still be generated but not stored by
+ * event storage SPI if they are disabled in Ignite configuration.
+ * @see EventType#EVT_CLUSTER_ACTIVATED
+ * @see EventType#EVT_CLUSTER_DEACTIVATED
+ * @see EventType#EVT_CLUSTER_ACTIVATION_FAILED
+ */
+public class ClusterActivationEvent extends EventAdapter {
+
```
alamar: Empty line is not needed after class opens.

modules/core/src/main/java/org/apache/ignite/internal/cluster/IgniteClusterImpl.java
```diff
@@ -315,15 +317,27 @@ public IgniteClusterImpl(GridKernalContext ctx) {
...
         }
         finally {
             unguard();
         }
     }
 
+    /** */
+    private void recordEvent(int evtType) {
+        if (ctx.event().isRecordable(evtType)) {
+            ClusterActivationEvent evt = new ClusterActivationEvent(null, null, evtType);
```
alamar: Please provide local node (ctx.discovery().localNode()) and human readable message describing event. We do that for all the other events so it makes sense here. Also please add empty line after this one.

modules/core/src/main/java/org/apache/ignite/internal/cluster/IgniteClusterImpl.java
```diff
@@ -315,15 +317,27 @@ public IgniteClusterImpl(GridKernalContext ctx) {
 
         try {
             ctx.state().changeGlobalState(active, baselineNodes(), false).get();
+
+            recordEvent(active ? EventType.EVT_CLUSTER_ACTIVATED : EventType.EVT_CLUSTER_DEACTIVATED);
         }
         catch (IgniteCheckedException e) {
+            recordEvent(EventType.EVT_CLUSTER_ACTIVATION_FAILED);
```
alamar: I'm not sure that we need this one. I can't find any other unhandled failure events.

modules/core/src/main/java/org/apache/ignite/internal/cluster/IgniteClusterImpl.java
```diff
@@ -315,15 +317,27 @@ public IgniteClusterImpl(GridKernalContext ctx) {
 
         try {
             ctx.state().changeGlobalState(active, baselineNodes(), false).get();
```
alamar: Can we add baseline nodes collection to ClusterActivationEvent as a field?

modules/core/src/main/java/org/apache/ignite/events/EventType.java
```diff
@@ -1076,6 +1100,18 @@
...
     };
 
+    /**
+     * All cluster activation events. This array can be directly passed into
+     * {@link IgniteEvents#localListen(IgnitePredicate, int...)} method to
+     * subscribe to all cloud events.
+     *
+     * @see ClusterActivationEvent
+     */
+    public static final int[] EVTS_CLUSTER_ACTIVATION = {
```
alamar: Maybe also add to EVTS_DISCOVERY_ALL? Need to discuss.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -160,6 +164,9 @@
     /** */
     private static final IgniteProductVersion EXCHANGE_PROTOCOL_2_SINCE = IgniteProductVersion.fromString("2.1.4");
 
+    /** Stripe id for cluster activation event. */
+    private static final int CLUSTER_ACTIVATION_EVT_STRIPE_ID = 1;
```
alamar: Why 1? Why not 0?

Come to think of it, I will put something like (2<<31) - 1 here so we always land in the last stripe, which is probably least busy.

PR #6240, opened 2019-03-06
IGNITE-11434 SQL: Create a view with list of existing COLUMNS

tledkov-gridgain:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sys/view/SqlSystemViewColumns.java
```diff
@@ -0,0 +1,142 @@
...
+     *
+     * @param ctx Kernal context.
+     * @param schemaMgr Schema manager.
+     */
+    public SqlSystemViewColumns(GridKernalContext ctx, SchemaManager schemaMgr) {
+        super("COLUMNS", "Ignite columns", ctx, new String[] {TABLE_NAME, COLUMN_NAME},
+            newColumn("SCHEMA_NAME"),
+            newColumn(TABLE_NAME),
+            newColumn(COLUMN_NAME),
+            newColumn("COLUMN_ID", Value.INT),
```
joooger: For other views we placed ID's columns before name's column. I propose to keep the order for all system views

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sys/view/SqlSystemViewColumns.java
```diff
@@ -0,0 +1,142 @@
...
+            newColumn("DATA_TYPE"),
```
joooger: Let's consider to add column DEFAULT_VALUE

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sys/view/SqlSystemViewColumns.java
```diff
@@ -0,0 +1,142 @@
...
+            newColumn("SQL_TYPE", Value.INT),
+            newColumn("IS_NULLABLE", Value.BOOLEAN),
+            newColumn("DISPLAY_SIZE", Value.INT),
```
joooger: What does it mean?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sys/view/SqlSystemViewColumns.java
```diff
@@ -0,0 +1,142 @@
...
+            newColumn("PRECISION", Value.INT),
+            newColumn("SCALE", Value.INT),
+            newColumn("AFFINITY_KEY", Value.BOOLEAN)
```
joooger: Also let's consider to add column with ordinal position of column like a col.getColumnId() - QueryUtils.DEFAULT_COLUMNS_COUNT. It can be useful for indexes view

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sys/view/SqlSystemViewColumns.java
joooger: I'm not sure that ordinal position should be a internal columnId, may be it will be better to use col.getColumnId() - QueryUtils.DEFAULT_COLUMNS_COUNT.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sys/view/SqlSystemViewColumns.java
```diff
@@ -0,0 +1,142 @@
...
+                Column col = tbl.getColumns()[i];
+
+                if (!colFilter.test(col))
+                    continue;
+
+                Object[] data = new Object[] {
+                    col.getTable().getSchema().getName(),
+                    col.getTable().getName(),
+                    col.getName(),
+                    col.getColumnId(),
```
joooger: I'm not sure that we should use here internal column id, may be will be better to use col.getColumnId() - QueryUtils.DEFAULT_COLUMNS_COUNT as value.

PR #6252, opened 2019-03-11
IGNITE-11521: Add a lifecycle event before joining the cluster

lukipuki:
For example, this allows the user to register a local listener before
the node joins the cluster. Without this change, the listener might have
missed some events such as 'EVT_CACHE_OBJECT_*'.

The new lifecycle event is called BEFORE_CLUSTER_CONNNECTION.

See the discussion in
http://apache-ignite-users.70518.x6.nabble.com/Register-listeners-before-joining-the-cluster-td25944.html#none


lukipuki:
Closing, see https://issues.apache.org/jira/browse/IGNITE-11521

PR #6266, opened 2019-03-13
IGNITE-11432: Add ability to specify auto-generated consistent ID in …

dspavlov:
…IgniteConfiguration


dspavlov:
most likely the issue is outdated, for a while users don't ask about this parameter.

PR #6282, opened 2019-03-15
IGNITE-11553 Ignite metrics to Prometheus.io

sdarlington:
More a proof-of-concept than a finished PR at this stage...


sdarlington:
With the addition of the OpenCensus module, this is no longer the correct approach (if it ever was).

PR #6305, opened 2019-03-19
IGNITE-11569 Enable baseline auto-adjust by default only for empty cl…

akalash:
…uster


modules/core/src/test/java/org/apache/ignite/internal/processors/cluster/ChangeTopologyWatcherTest.java
```diff
@@ -344,4 +345,43 @@ public void testBaselineAutoAdjustIgnoreClientNodes() throws Exception {
...
+        }
+        finally {
+            System.clearProperty(IGNITE_BASELINE_AUTO_ADJUST_ENABLED);
+        }
+
+        Ignite ignite0 = startGrids(2);
+
+        ignite0.cluster().active(true);
+
+        Set<Object> initBaseline = ignite0.cluster().currentBaselineTopology().stream()
```
ibessonov: I think that it would be much easier to read if you compared size of baseline nodes set, not their ids. Is that ok?

PR #6346, opened 2019-03-26
IGNITE-11609 Add support of authentication and SSL in yardstick IgniteThinClient benchmark

qvad:


modules/yardstick/src/main/java/org/apache/ignite/yardstick/IgniteThinClient.java
```diff
@@ -41,6 +45,49 @@
  * Thin client.
  */
 public class IgniteThinClient {
+
```
dspavlov: Not needed empty line

qvad: Fixed

PR #6359, opened 2019-03-27
IGNITE-8588 .NET: Serialization issue when derived type hides base type member

ashapkin:
Original exception clarification
@ptupitsyn


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Binary/BinaryReflectiveSerializerInternal.cs
```diff
@@ -142,22 +142,18 @@ bool IBinarySerializerInternal.SupportsHandles
...
             var fields = ReflectionUtils.GetAllFields(type).Where(x => !x.IsNotSerialized).ToList();
 
-            IDictionary<int, string> idMap = new Dictionary<int, string>();
+            var fieldsSet = new HashSet<string>();
 
             foreach (FieldInfo field in fields)
             {
                 string fieldName = BinaryUtils.CleanFieldName(field.Name);
 
-                int fieldId = BinaryUtils.FieldId(typeId, fieldName, converter, idMapper);
```
ashapkin: I removed this one, since it was used only for exception message. Correct me if it's still required.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Binary/BinaryReflectiveSerializerInternal.cs
```diff
@@ -152,9 +152,18 @@ bool IBinarySerializerInternal.SupportsHandles
...
-                                                    type.Name + ", field1=" + idMap[fieldId] + ", field2=" + fieldName +
-                                                    ", fieldId=" + fieldId + ']');
+                    string existingFieldName = idMap[fieldId];
+                    string baseClassName = field.DeclaringType != null ? field.DeclaringType.Name : null;
+                    var msg = string.Format("{0} derives from {1} and hides field {2} from the base class. " +
+                                            "Ignite can not serialize two fields with the same name.", type.Name, baseClassName, fieldName);
+
+                    if (fieldName != existingFieldName)
+                    {
+                        msg = string.Format("Ignite resolved two fields {0} and {1} to the same fieldId {2}. Probably " +
```
ptupitsyn: Let's keep existing error message by default. Only when `fieldName == existingFieldName` provide the new error message.

PR #6378, opened 2019-03-29
IGNITE-11647: ML Vectors should work with all Serializable objects besides double

avplatonov:


modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/AbstractVector.java
```diff
@@ -125,6 +126,20 @@ protected void storageSet(int i, double v) {
         maxElm = minElm = null;
     }
 
+    /**
+     * @param i Index.
+     * @param v Value.
+     */
+    protected <T extends Serializable> void storageSetRaw(int i, T v) {
```
dmitrievanthony: Why not just `protected void storageSetRaw(int i, Serializable v)`?

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/AbstractVector.java
```diff
@@ -133,6 +148,14 @@ protected double storageGet(int i) {
         return sto.get(i);
     }
 
+    /**
```
dmitrievanthony: Javadoc?

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/AbstractVector.java
```diff
@@ -160,6 +183,18 @@ protected void checkIndex(int idx) {
...
 
+    /** {@inheritDoc} */
+    @Override public <T extends Serializable> T getRaw(int idx) {
+        checkIndex(idx);
+
+        return sto.getRaw(idx);
+    }
+
+    /** {@inheritDoc} */
+    @Override public <T extends Serializable> T getRawX(int idx) {
```
dmitrievanthony: Why somewhere you use `checkIndex(idx)`, but somewhere not?

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/AbstractVector.java
```diff
@@ -297,6 +342,22 @@ protected Element makeElement(int idx) {
         return this;
     }
 
+    /** {@inheritDoc} */
+    @Override public <T extends Serializable> Vector setRaw(int idx, T val) {
```
dmitrievanthony: Why not just protected void storageSetRaw(int i, Serializable v)?

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/VectorStorage.java
```diff
@@ -32,24 +33,48 @@
...
     public default double[] data() {
         return null;
     }
+
+    /**
+     * @return Underlying array of serializable objects {@link StorageOpsMetrics#isArrayBased()} returns {@code true}.
+     * Returns {@code null} if in other cases.
+     */
+    public default Serializable[] rawData() {
+        return null;
```
dmitrievanthony: Wouldn't it be better to throw an exception or return empty array?

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/storage/DenseVectorStorage.java
```diff
@@ -56,19 +61,90 @@ public DenseVectorStorage(double[] data) {
...
-        return data == null ? 0 : data.length;
+        if (data == null && rawData == null)
+            return 0;
+        else {
+            if (data != null)
+                return data.length;
+            if (rawData != null)
+                return rawData.length;
+            else
+                throw new IllegalStateException();
```
dmitrievanthony: Unreachable statement?

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/storage/DenseVectorStorage.java
```diff
@@ -56,19 +61,90 @@ public DenseVectorStorage(double[] data) {
...
+    }
+
+    /**
+     * Copies values between internal arrays If toNumericRepresentation = true then method tries cast all serializable
+     * objects to double, initializes double array with converted values and cleans raw data array. If
+     * toNumericRepresentation = false then method copies double array to raw data array and cleans double array.
+     *
+     * @param toNumericRepresentation To numeric representation.
+     */
+    private void swap(boolean toNumericRepresentation) {
```
dmitrievanthony: Lets make two methods `toNumericArray` and `toGenericArray`. Looks like it will be more clear.

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/storage/DenseVectorStorage.java
```diff
@@ -56,19 +61,90 @@ public DenseVectorStorage(double[] data) {
...
 
     /** {@inheritDoc} */
     @Override public double get(int i) {
-        return data[i];
+        if (data != null)
+            return data[i];
+
+        Serializable v = rawData[i];
+        if (v == null)
+            return 0.0;
```
dmitrievanthony: We use `Double.NaN` for missing values. Would it be better to use it here?

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/storage/DenseVectorStorage.java
```diff
@@ -56,19 +61,90 @@ public DenseVectorStorage(double[] data) {
...
+        if (data == null && rawData == null)
+            return;
+
+        if (toNumericRepresentation) {
+            if (data != null)
+                return;
+
+            data = new double[rawData.length];
+            for (int i = 0; i < rawData.length; i++)
+                data[i] = ((Number)rawData[i]).doubleValue();
```
dmitrievanthony: NpE?

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/storage/DenseVectorStorage.java
```diff
@@ -56,19 +61,90 @@ public DenseVectorStorage(double[] data) {
...
-        data[i] = v;
+        if (data != null)
+            data[i] = v;
+        else
+            rawData[i] = v;
+    }
+
+    /** {@inheritDoc} */
+    @Override public <T extends Serializable> void setRaw(int i, T v) {
+        swap(false);
```
dmitrievanthony: Would it be better to keep two keep both arrays "active" and bitset that tells for each index which array contains actual value?

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/storage/SparseVectorStorage.java
```diff
@@ -105,7 +70,16 @@ public int getAccessMode() {
 
     /** {@inheritDoc} */
     @Override public double get(int i) {
-        return sto.getOrDefault(i, 0.0);
+        Serializable obj = sto.get(i);
+        if (obj == null)
+            return 0.0;
```
dmitrievanthony: Why not `Double.NaN`?

modules/ml/src/test/java/org/apache/ignite/ml/math/primitives/vector/storage/DenseVectorStorageTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Tests for DenseVectorStorage.
+ */
+public class DenseVectorStorageTest extends AbstractStorageTest {
+
+    /** {@inheritDoc} */
+    @Override protected boolean isNumericVector(VectorStorage storage) {
+        try {
+            Field f = storage.getClass().getDeclaredField("data");
```
dmitrievanthony: Does it make sense to make protected getter `getData`? It would help it case field is renamed.

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/AbstractVector.java
avplatonov: Please read javadoc for get*X methods:
"Gets the value at specified index without checking for index boundaries"

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/VectorStorage.java
avplatonov: I save such behavior for consistency with old API

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/storage/DenseVectorStorage.java
avplatonov: No, I don't want to save additional states in vector. From my point of view, vector can be numeric and if we try to save some Serializable object - vector change own state to Serializabled representation. Only "double[] data()" can cast vector to numeric representation. These casts won't be so frequently.

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/storage/DenseVectorStorage.java
avplatonov: No, it wasn't documented but old API returned 0.0 and I don't want to change this behavior.

modules/ml/src/test/java/org/apache/ignite/ml/math/primitives/vector/storage/DenseVectorStorageTest.java
avplatonov: Yes, you're right

avplatonov:
@dmitrievanthony thanks for review
please, read my comments

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/storage/DenseVectorStorage.java
```diff
@@ -56,19 +61,85 @@ public DenseVectorStorage(double[] data) {
...
+                data[i] = rawData[i] == null ? 0.0 : ((Number)rawData[i]).doubleValue();
+            rawData = null;
+        }
+    }
+
+    /**
+     * Tries to cast internal representation of data to array of Serializable objects if need.
+     */
+    private void toGenericArray() {
+        A.ensure(data == null || rawData == null, "data == null || rawData == null");
```
zaleslaw: What is A.ensure?

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/storage/DenseVectorStorage.java
zaleslaw: hmmm, support for Double.NaN but not in this ticket, could you make TODO + ticket for this, @avplatonov ?

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/storage/DenseVectorStorage.java
avplatonov: This is a standard way for assertions in Ignite.Core

modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/storage/DenseVectorStorage.java
avplatonov: Yes, sure. But I think we should provide a way to use other user defined values for default values in Vector and use Double.NaN as default way in our code.

dspavlov:
 

PR #6381, opened 2019-04-01
IGNITE-11499 SQL: DML should not use batches by default

tledkov-gridgain:


modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/IgniteSqlNotNullConstraintTest.java
```diff
@@ -920,7 +910,9 @@ public void testNotNullCheckDmlUpdateFromSelect() throws Exception {
 
         executeSql("UPDATE src SET name = 'Ren' WHERE id = 2");
 
-        executeSql("UPDATE dest p SET (name) = (SELECT name FROM src t WHERE p.id = t.id)");
+        System.out.println("+++ " + executeSql("SELECT * FROM src "));
```
AMashenkov: Let's remove debug output from here.

PR #6407, opened 2019-04-04
IGNITE-10344 Speed up cleanupRestoredCaches

denis-chudov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStoreManager.java
```diff
@@ -1083,23 +1145,25 @@ private IgniteCheckedException shutdown(PageStore store, boolean cleanFile, Igni
...
-                            holder0 = initForCache(gDesc, gDesc.config());
-                        } catch (IgniteCheckedException e) {
-                            throw new IgniteException(e);
+            return cleanupAsyncExecutor.afterAsyncCompletion(
+                () -> idxCacheStores.computeIfAbsent(grpId, (key) -> {
+                    CacheGroupDescriptor gDesc = cctx.cache().cacheGroupDescriptors().get(grpId);
+
+                    CacheStoreHolder holder0 = null;
+
+                    if (gDesc != null) {
```
antonovsergey93: Megre both `if` statement into single `if`

PR #6427, opened 2019-04-09
IGNITE-11690 .NET: Unable to locate external assembly with enabled peerAssemblyLoading

ashapkin:
Disable custom request assembly loading for GetAssemblyFunc

### DIFF
--- a/modules/platforms/dotnet/Apache.Ignite.Core/Impl/Deployment/PeerAssemblyResolver.cs
+++ b/modules/platforms/dotnet/Apache.Ignite.Core/Impl/Deployment/PeerAssemblyResolver.cs
@@ -21,7 +21,6 @@ namespace Apache.Ignite.Core.Impl.Deployment
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.Reflection;
-    using System.Runtime.Remoting.Messaging;
     using Apache.Ignite.Core.Cluster;
     using Apache.Ignite.Core.Common;
     using Apache.Ignite.Core.Compute;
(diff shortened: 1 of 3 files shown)


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Deployment/PeerAssemblyResolver.cs
```diff
@@ -36,16 +37,29 @@ internal sealed class PeerAssemblyResolver : IDisposable
...
         /// Initializes a new instance of the <see cref="PeerAssemblyResolver"/> class.
         /// </summary>
         public PeerAssemblyResolver(IIgniteInternal ignite, Guid originNodeId)
         {
             Debug.Assert(ignite != null);
 
-            _handler = (sender, args) => GetAssembly(ignite, args.Name, originNodeId);
+            _handler = (sender, args) =>
+            {
+                object result = CallContext.LogicalGetData(DisablePeerLoadingMark);
```
ashapkin: Probably not a best one towards Net standard, but does reduce amount of code changes for now

ptupitsyn: This does not compile on .NET Core / .NET Standard.
Let's use `ThreadStatic` instead.

Please move the logic to `GetAssembly` method as well to keep the lambda small.

ashapkin: There are multiple calls to `GetAssembly` method , the goal was to reduce an amount of auxiliary checks for all cases except the particular one. I think we can extract a helper method for a better readability. Any thoughts?

PR #6437, opened 2019-04-11
IGNITE-11726 SQL: don't store default precision and scale in the Quer…

tledkov-gridgain:
…yEntity for CHAR/VARCHAR and DECIMAL types


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/CommandProcessor.java
```diff
@@ -725,15 +725,18 @@ private static QueryEntity toQueryEntity(GridSqlCreateTable createTbl) {
                 dfltValues.put(e.getKey(), dfltVal);
 
             if (col.getType() == Value.DECIMAL) {
-                precision.put(e.getKey(), (int)col.getPrecision());
+                if (col.getPrecision() != H2Utils.DECIMAL_DEFAULT_PRECISION)
```
joooger: let's use 'less sign' instead of 'not equals'. Such approach use H2, because precision has type long.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/CommandProcessor.java
```diff
@@ -725,15 +725,18 @@ private static QueryEntity toQueryEntity(GridSqlCreateTable createTbl) {
...
-                scale.put(e.getKey(), col.getScale());
+                if (col.getScale() != H2Utils.DECIMAL_DEFAULT_SCALE)
+                    scale.put(e.getKey(), col.getScale());
             }
 
             if (col.getType() == Value.STRING ||
                 col.getType() == Value.STRING_FIXED ||
                 col.getType() == Value.STRING_IGNORECASE)
-                precision.put(e.getKey(), (int)col.getPrecision());
+                if (col.getPrecision() != H2Utils.STRING_DEFAULT_PRECISION)
```
joooger: May be need to use 'less sign' instead of 'not equals'. Such approach use H2.
''if (col.getPrecision() < H2Utils.STRING_DEFAULT_PRECISION) ''

PR #6438, opened 2019-04-11
IGNITE-11730 Defining zipped message by content header.

akalash:


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/DiscoveryDataPacket.java
```diff
@@ -39,6 +39,8 @@
  * and allows convenient way of converting it to and from {@link DiscoveryDataBag} objects.
  */
 public class DiscoveryDataPacket implements Serializable {
+    /** Local file header signature(read as a little-endian number). */
+    private static int ZIP_HEADER_SIGNATURE = 0x04034b50;
```
alamar: Are you sure that correct uncompressed message cannot be generated which will start from this signature?

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/DiscoveryDataPacket.java
```diff
@@ -108,27 +110,34 @@ public void marshalJoiningNodeData(DiscoveryDataBag bag, Marshaller marsh, boole
         marshalData(bag.joiningNodeData(), joiningNodeData, marsh, isCompressionEnabled, compressionLevel, log);
     }
 
+    /**
+     * @return {@code true} if joining node data was transferred via network in zipped format.
+     */
+    public boolean isJoiningDataZipped(){
```
alamar: Whitespace desired before {

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/DiscoveryDataPacket.java
```diff
@@ -307,6 +308,27 @@ else if (binEntry.getKey() < GridComponent.DiscoveryDataExchangeType.VALUES.leng
...
+        return value != null && value.length > 3 && makeInt(value) == ZIP_HEADER_SIGNATURE;
+    }
+
+    /**
+     * Make int from first 4 bytes in little-endian byte order.
+     *
+     * @param b Source of bytes.
+     * @return Made int.
+     */
+    static private int makeInt(byte[] b) {
```
alamar: Is it not possible to use BinaryPrimitives.readInt() here? Any endianness considerations?

PR #6445, opened 2019-04-12
IGNITE-11735: Safely handle new closures of IGNITE-11392 in mixed cluster environment

denis-chudov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -63,6 +64,7 @@
 import org.apache.ignite.internal.IgniteClientDisconnectedCheckedException;
 import org.apache.ignite.internal.IgniteDiagnosticAware;
 import org.apache.ignite.internal.IgniteDiagnosticPrepareContext;
+import org.apache.ignite.internal.IgniteFeatures;
```
antonovsergey93: Unused import. Remove it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -2099,40 +2103,50 @@ private void dumpLongRunningTransaction(IgniteInternalTx tx) {
...
-                                            "Dumping the near node thread that started transaction [xidVer=%s]\n%s",
-                                            tx.xidVersion().toString(),
-                                            traceDump
-                                        )
-                                    );
+                                    try {
+                                        traceDump = strIgniteFut.get();
+                                    }
+                                    catch (Exception e) {
+                                        U.warn(
```
antonovsergey93: We should log exception in this case. Use `U.error(IgniteLogger, String, Throwable)` for it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -2099,40 +2103,50 @@ private void dumpLongRunningTransaction(IgniteInternalTx tx) {
...
+                                            )
+                                        );
+                                    }
                                 }
-                            }
-                        });
+                            });
+                    }
+                    catch (Exception e) {
+                        U.warn(diagnosticLog, "Could not send dump request to transaction owner near node: " + e.getMessage());
```
antonovsergey93: We should log exception in this case. Use `U.error(IgniteLogger, String, Throwable)` for it.

PR #6448, opened 2019-04-12
Native aligne page

dgovorukhin:


modules/core/src/main/java/org/apache/ignite/internal/mem/unsafe/UnsafeMemoryProvider.java
```diff
@@ -97,10 +103,44 @@ public UnsafeMemoryProvider(IgniteLogger log) {
...
+            if (pageSize < 0)
+                pageSize = GridUnsafe.pageSize();
+
+            if (pageSize < 0){
+                U.warn(log,"Can not get native page size, use 4k as default.");
+
+                pageSize = 4096;
+            }
+
+            int ptrRes = GridUnsafe.posix_memalign(refToPtr, new NativeLong(pageSize), new NativeLong(chunkSize));
```
dspavlov: Availability of this call should be checked. It is not available e.g. on Windows

modules/core/pom.xml
```diff
@@ -68,6 +68,12 @@
             <version>${javax.cache.version}</version>
         </dependency>
 
+        <dependency>
```
dspavlov: We should consider adding JNA, but the decision may depend on our ideas about Unsafe-less client/server implementation. For now, I'm concerned about adding JNA to core dependencies list.

I would suggest allocating memory via plugin, e.g. direct-io

PR #6469, opened 2019-04-17
IGNITE-11753 Improve error message on first connection attempt to sec…

alamar:
…ured cluster.


modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -2857,6 +2860,9 @@ public int execute(List<String> rawArgs) {
                 }
 
                 try (GridClient client = GridClientFactory.start(clientCfg)) {
+                    if (!client.connected())
```
alamar: comment

PR #6475, opened 2019-04-18
IGNITE-11772: [ML] Broken javadoc

ybabak:
fixed javadocs

### DIFF
--- a/modules/ml/src/main/java/org/apache/ignite/ml/composition/boosting/convergence/simple/ConvergenceCheckerStub.java
+++ b/modules/ml/src/main/java/org/apache/ignite/ml/composition/boosting/convergence/simple/ConvergenceCheckerStub.java
@@ -46,7 +46,6 @@ public class ConvergenceCheckerStub<K, V> extends ConvergenceChecker<K, V> {
      * @param externalLbToInternalMapping External label to internal mapping.
      * @param loss Loss function.
      * @param datasetBuilder Dataset builder.
-     * @param vectorizer Upstream vectorizer.
      */
     public ConvergenceCheckerStub(long sampleSize, IgniteFunction externalLbToInternalMapping, Loss loss,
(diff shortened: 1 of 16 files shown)


modules/ml/src/main/java/org/apache/ignite/ml/composition/boosting/convergence/simple/ConvergenceCheckerStub.java
```diff
@@ -46,7 +46,6 @@
      * @param externalLbToInternalMapping External label to internal mapping.
      * @param loss Loss function.
      * @param datasetBuilder Dataset builder.
```
dmitrievanthony: I see `double precision` and `preprocessor` parameter which is not mentioned in javadoc.

modules/ml/src/main/java/org/apache/ignite/ml/genetic/MutateTask.java
```diff
@@ -78,7 +78,7 @@ public MutateTask(GAConfiguration cfg) {
      * @param nodes List of ClusterNode
      * @param chromosomeKeys Primary keys for respective chromosomes
```
dmitrievanthony: Where is the return statement in javadoc?

modules/ml/src/main/java/org/apache/ignite/ml/composition/boosting/convergence/simple/ConvergenceCheckerStub.java
ybabak: fixed

modules/ml/src/main/java/org/apache/ignite/ml/genetic/MutateTask.java
ybabak: fixed

PR #6483, opened 2019-04-19
IGNITE-5714 Implementation of suspend/resume for pessimistic transactions

alex-plekhanov:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
```diff
@@ -0,0 +1,771 @@
...
+     * @return Number of server nodes.
+     */
+    protected int gridCount() {
+        return GRID_CNT;
+    }
+
+    /**
+     * @return Transaction concurrency.
+     */
+    protected abstract TransactionConcurrency transactionConcurrency();
```
Jokser: Minor.
"@Parametrized" can be used instead.

alex-plekhanov: AFAIK "@Paramentrized" can be used if we need to run some set of tests with different parameters, 
but here we have different sets of tests for OPTIMISTIC and PESSIMISTIC transactions.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
```diff
@@ -0,0 +1,771 @@
...
+            catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    /**
+     * Runnable that can throw any exception.
+     */
+    public abstract static class RunnableX implements Runnable {
```
Jokser: I see only one inheritor of this class which contains as runnable content non-throwable command `GridTestUtils.assertThrowsWithCause(txOperation, tx, IgniteException.class);`
Why not use regular Runnable in this case?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
```diff
@@ -0,0 +1,771 @@
...
+            }
+        }
+    }
+
+    /**
+     * Closure that can throw any exception.
+     *
+     * @param <T> Type of closure parameter.
+     */
+    public abstract static class CI1Exc<T> implements CI1<T> {
```
Jokser: I see that all inheritors of this closure don't throw checked exceptions.
It can be removed and lambdas can be used instead.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
alex-plekhanov: This implementation was in class `IgniteOptimisticTxSuspendResumeTest`, the class was renamed to `IgniteAbstractTxSuspendResumeTest`, some OPTIMISTIC tx related methods was extracted to class `IgniteOptimisticTxSuspendResumeTest`.
For sure `Runnable` can be used here. Will fix it.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
alex-plekhanov: This implementation was originally in class `IgniteOptimisticTxSuspendResumeTest`. Will fix it.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
```diff
@@ -0,0 +1,771 @@
...
+
+                    GridTestUtils.assertThrowsWithCause(new Callable<Object>() {
+                        @Override public Object call() throws Exception {
+                            tx.suspend();
+
+                            return null;
+                        }
+                    }, TransactionTimeoutException.class);
+
+                    assertTrue(GridTestUtils.waitForCondition(new PA() {
```
Jokser: Can be replaced with lambda

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
```diff
@@ -0,0 +1,771 @@
...
+
+                    tx.resume();
+
+                    cache.put(1, 1);
+                    cache.put(2, 2);
+                    cache.put(3, 3);
+
+                    U.sleep(TX_TIMEOUT + 100L);
+
+                    GridTestUtils.assertThrowsWithCause(new Callable<Object>() {
```
Jokser: Can be replaced with method reference

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
```diff
@@ -0,0 +1,771 @@
...
+                        @Override public Object call() throws Exception {
+                            tx.resume();
+
+                            return null;
+                        }
+                    }, TransactionTimeoutException.class);
+
+                    assertTrue(GridTestUtils.waitForCondition(new PA() {
```
Jokser: Can be replaced with lambda

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
```diff
@@ -0,0 +1,771 @@
...
+
+                    cache.put(1, 1);
+                    cache.put(2, 2);
+                    cache.put(3, 3);
+
+                    tx.suspend();
+
+                    U.sleep(TX_TIMEOUT + 100L);
+
+                    GridTestUtils.assertThrowsWithCause(new Callable<Object>() {
```
Jokser: Can be replaced with method reference

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
```diff
@@ -0,0 +1,771 @@
...
+            @Override public void applyx(Ignite ignite, final IgniteCache<Integer, Integer> cache) throws Exception {
+                for (final CI1Exc<Transaction> txOperation : SUSPENDED_TX_PROHIBITED_OPS) {
+                    for (TransactionIsolation isolation : TransactionIsolation.values()) {
+                        final Transaction tx = ignite.transactions().txStart(transactionConcurrency(), isolation);
+
+                        cache.put(1, 1);
+
+                        tx.suspend();
+
+                        multithreaded(new RunnableX() {
```
Jokser: Can be replaced with lambda

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
```diff
@@ -0,0 +1,771 @@
...
+                        Transaction tx = ignite.transactions().txStart(transactionConcurrency(), isolation);
+
+                        cache.put(i, i);
+
+                        tx.suspend();
+
+                        clientTxs.add(tx);
+                    }
+
+                    GridTestUtils.runMultiThreaded(new CI1Exc<Integer>() {
```
Jokser: Can be replaced with lambda

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
```diff
@@ -0,0 +1,771 @@
...
+
+                            assertEquals(ACTIVE, tx.state());
+
+                            cache.put(3, 3);
+
+                            tx.rollback();
+                        }
+                    }).get(FUT_TIMEOUT);
+
+                    assertTrue(GridTestUtils.waitForCondition(new PA() {
```
Jokser: Can be replaced with lambda

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
```diff
@@ -0,0 +1,771 @@
...
+                    cache.put(1, 1);
+                    cache.put(2, 2);
+
+                    tx.suspend();
+
+                    assertNull("There is no transaction for current thread", ignite.transactions().tx());
+
+                    assertEquals(SUSPENDED, tx.state());
+
+                    GridTestUtils.runAsync(new Runnable() {
```
Jokser: Can be replaced with lamda

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
Jokser: Tests that not supposed to use such concurrency level can be skipped (I see few counts of them)
Moreover, I see that almost all tests use the following pattern:
peek node, peek cache configuration, peek tx isolation level -> doTest.
Can we somehow replace it with `@Parametrized`?
I see that only one test doesn't match this pattern `testSuspendTxAndResumeAfterTopologyChange`. It can be separated, while others can user test parameters. This will avoid copypaste and every test will be plain.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
```diff
@@ -0,0 +1,771 @@
...
+
+                    final AtomicInteger cntr = new AtomicInteger(0);
+
+                    cache.put(-1, -1);
+                    otherCache.put(-1, -1);
+
+                    tx.suspend();
+
+                    for (int i = 0; i < 10; i++) {
+                        GridTestUtils.runAsync(new Runnable() {
```
Jokser: Can be replaced with lambda.

modules/jta/src/test/java/org/apache/ignite/internal/processors/cache/GridJtaTransactionManagerSelfTest.java
```diff
@@ -66,6 +66,11 @@
         jotm.stop();
     }
 
+    /**
+     * @return Transaction concurrency.
+     */
+    protected abstract TransactionConcurrency txConcurrency();
```
Jokser: `@Parametrized` can be used.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
```diff
@@ -0,0 +1,771 @@
...
+
+                    assertEquals(SUSPENDED, tx.state());
+
+                    assertNull("Thread already have tx", ignite.transactions().tx());
+
+                    assertNull(cache.get(-1));
+                    assertNull(cache.get(cntr.get()));
+
+                    for (int i = 0; i < 10; i++) {
+                        GridTestUtils.runAsync(new Runnable() {
```
Jokser: Can be replaced with lambda.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteAbstractTxSuspendResumeTest.java
alex-plekhanov: `IgniteAbstractTxSuspendResumeTest` is not a new test. The test was renamed from `IgniteOptimisticTxSuspendResumeTest`. It already has some logic which we can accidentally break by deep refactoring, it also has some history on TC, which can be lost.
I see no practical reason to totally rewrite the test only to use some JUnit feature.

Jokser: This is not "some JUnit feature", this is the right way to write tests without boilerplate.
But I agree that the existing test with history is a restrictive factor of refactoring.
Anyway, I recommend writing tests with `@Parametrized` pattern in future tests if applicable.

alex-plekhanov: Ok, I will rewrite `GridJtaTransactionManagerSelfTest` with '@Parametrized', but this test I prefer to left as is.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMvcc.java
```diff
@@ -1331,12 +1331,31 @@ public void remove(GridCacheVersion ver) {
...
             for (GridCacheMvccCandidate c : locs)
-                if (c.nodeId().equals(nodeId) && c.threadId() == threadId)
+                if (c.nodeId().equals(nodeId) && c.isHeldByThread(threadId))
                     return c;
 
         return null;
     }
 
+    /**
+     * Near local candidate.
```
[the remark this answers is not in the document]
alex-plekhanov: It was copied from `localCandidate(UUID nodeId, long threadId)` method. Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMvcc.java
```diff
@@ -1331,12 +1331,31 @@ public void remove(GridCacheVersion ver) {
...
+     *
+     * @param nodeId Node ID.
+     * @param threadId Thread ID.
+     * @param ver Lock version.
```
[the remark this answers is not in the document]
alex-plekhanov: Such a description is commonly used all over the class for this parameter, I don't see a reason to make an exception here.

PR #6486, opened 2019-04-19
IGNITE-11771: adding the option to include pods with a not-ready state

balazspeterfi:


modules/kubernetes/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/kubernetes/TcpDiscoveryKubernetesIpFinder.java
```diff
@@ -227,6 +233,15 @@ public void setAccountToken(String accountToken) {
         this.accountToken = accountToken;
     }
 
+    /**
+     * Determines whether addresses of not-ready pods should be included. Default is false.
+     *
+     * @param includeNotReadyAddresses Flag to include not-ready pods.
+     */
+    public void setIncludeNotReadyAddresses(boolean includeNotReadyAddresses) {
```
dmagda: Please rename to "includeNotReadyAddresses" to follow Ignite-naming convention. Plus, I would have this option enabled by default to avoid any issues with existing deployments. Our discovery SPI does several connection attempts and some of the pods might be ready by that time.

shroman:
Merged 3a8ea330dd5732579097788bba85eae0d1a98f33

PR #6487, opened 2019-04-19
IGNITE-11103 Idle Verify --cache-filter ALL should include system caches.

alamar:
Introduce USER and invisible DEFAULT cache filters.
DEFAULT now outputs all USER caches unless cache names specified, otherwise all that match them, including system.
Fix messages when no caches match filter and --dump supplied.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
@@ -1939,8 +1939,8 @@ public class CommandHandler {
                     "processed by " + IDLE_VERIFY + " command. If cache names are specified, in form of regular " +
                     "expressions, only matching caches will be verified. Caches matched by regexes specified after " +
                     EXCLUDE_CACHES + " parameter will be excluded from verification. Using parameter " + CACHE_FILTER +
-                    " you can verify: only USER caches, only user PERSISTENT caches, only user NOT_PERSISTENT caches, " +
-                    ", only SYSTEM caches, or ALL of the above.";
+                    " you can verify: only " + USER + " caches, only user " + PERSISTENT + " caches, only user " +
+                    NOT_PERSISTENT + " caches, only " + SYSTEM + " caches, or " + ALL + " of the above.";
 
             case VALIDATE_INDEXES:
                 return "Validate indexes on an idle cluster and print out the keys that are missing in the indexes.";
--- a/modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
@@ -864,8 +864,8 @@ public class CommandHandler {
 
         usageCache(LIST, "regexPattern", op(or(GROUP, SEQUENCE)), OP_NODE_ID, op(CONFIG), op(OUTPUT_FORMAT, MULTI_LINE));
         usageCache(CONTENTION, "minQueueSize", OP_NODE_ID, op("maxPrint"));
-        usageCache(IDLE_VERIFY, op(DUMP), op(SKIP_ZEROS), op(CHECK_CRC),
-            op(EXCLUDE_CACHES, CACHES), op(CACHE_FILTER, or(CacheFilterEnum.values())), op(CACHES));
+        usageCache(IDLE_VERIFY, op(DUMP), op(SKIP_ZEROS), op(CHECK_CRC), op(EXCLUDE_CACHES, CACHES),
+            op(CACHE_FILTER, or("ALL|USER|SYSTEM|PERSISTENT|NON_PERSISTENT")), op(CACHES));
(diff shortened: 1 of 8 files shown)


modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -1931,9 +1931,8 @@ private String getCacheSubcommandDesc(CacheCommand cmd) {
             case IDLE_VERIFY:
                 return "Verify counters and hash sums of primary and backup partitions for the specified caches/cache groups on an idle cluster and print out the differences, if any. " +
                     "Cache filtering options configure the set of caches that will be processed by " + IDLE_VERIFY + " command. " +
-                    "Default value for the set of cache names (or cache group names) is all cache groups. Default value for " + EXCLUDE_CACHES + " is empty set. " +
-                    "Default value for " + CACHE_FILTER + " is no filtering. Therefore, the set of all caches is sequently filtered by cache name " +
-                    "regexps, by cache type and after all by exclude regexps.";
+                    "Default value for the set of cache names (or cache group names) is all user cache groups. Default value for " + EXCLUDE_CACHES + " is empty set. " +
+                    "Therefore, the set of all caches is sequently filtered by cache name regexps, by cache type (if specified) and after that by exclude regexps.";
```
antonovsergey93: I think we should define behaviour for all possible `CacheFilterEnum` values

modules/core/src/test/java/org/apache/ignite/util/GridCommandHandlerTest.java
```diff
@@ -1347,6 +1347,34 @@ public void testCacheIdleVerify() throws Exception {
...
+        }
+
+        injectTestSystemOut();
+
+        assertEquals(EXIT_CODE_OK, execute("--cache", "idle_verify", DEFAULT_CACHE_NAME));
+
+        assertContains(testOut.toString(), "no conflicts have been found");
+    }
+
     /**
```
antonovsergey93: Please add explicit test for `DEFAULT` cache filter

alamar: Will add case to testCacheIdleVerifyMultipleCacheFilterOptions

modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
alamar: I have tried to rewrite this messages, including all notable enum values.

modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -1929,11 +1933,14 @@ private String getCacheSubcommandDesc(CacheCommand cmd) {
...
-                    "Default value for " + CACHE_FILTER + " is no filtering. Therefore, the set of all caches is sequently filtered by cache name " +
-                    "regexps, by cache type and after all by exclude regexps.";
+                return "Verify counters and hash sums of primary and backup partitions for the specified caches/cache " +
+                    "groups on an idle cluster and print out the differences, if any. When no parameters are specified, " +
+                    "all user caches are verified. Cache filtering options configure the set of caches that will be " +
+                    "processed by " + IDLE_VERIFY + " command. If cache names are specified, in form of regular " +
+                    "expressions, only matching caches will be verified. Caches matched by regexes specified after " +
+                    EXCLUDE_CACHES + " parameter will be excluded from verification. Using parameter " + CACHE_FILTER +
+                    " you can verify: only USER caches, only user PERSISTENT caches, only user NOT_PERSISTENT caches, " +
+                    ", only SYSTEM caches, or ALL of the above.";
```
antonovsergey93: Please use enum values instead of string. i.e. 
only **USER** caches, only user **PERSISTENT** caches, only user **NOT_PERSISTENT** caches, " +
                    ", only **SYSTEM** caches, or **ALL** of the above.";

modules/core/src/test/java/org/apache/ignite/util/GridCommandHandlerTest.java
```diff
@@ -1347,6 +1347,34 @@ public void testCacheIdleVerify() throws Exception {
...
+        try (IgniteDataStreamer streamer = ignite.dataStreamer(DEFAULT_CACHE_NAME)) {
+            for (int i = 0; i < 100; i++)
+                streamer.addData(i, i);
+        }
+
+        injectTestSystemOut();
+
+        assertEquals(EXIT_CODE_OK, execute("--cache", "idle_verify", DEFAULT_CACHE_NAME));
+
+        assertContains(testOut.toString(), "no conflicts have been found");
```
antonovsergey93: Please add assert on correct number of found partitions.

alamar: Without --dump number of found partitions is not printed so there's nothing to check.

PR #6490, opened 2019-04-20
IGNITE-7285 Add default query timeout

samaitra:


modules/core/src/main/java/org/apache/ignite/internal/processors/platform/utils/PlatformConfigurationUtils.java
```diff
@@ -223,6 +223,7 @@ public static CacheConfiguration readCacheConfiguration(BinaryRawReaderEx in, Cl
         ccfg.setMaxQueryIteratorsCount(in.readInt());
         ccfg.setQueryDetailMetricsSize(in.readInt());
         ccfg.setQueryParallelism(in.readInt());
+        ccfg.setDefaultQueryTimeout(in.readLong());
```
AMashenkov: "writeCacheConfiguration" method should be changed as well.

modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
```diff
@@ -185,6 +185,9 @@
     /** Default query parallelism. */
     public static final int DFLT_QUERY_PARALLELISM = 1;
 
+    /** Default query timeout. */
+    public static final long DFLT_QRY_TIMEOUT = 5000;
```
AMashenkov: I believe we should discuss configuration defaults changes on dev-list at first and preserve default behavior if possible.

Let's make it zero by default and treat zeroed value as Infinite Timeout to preserve previous behavior.
May be we can bound this later after discussion.

modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
```diff
@@ -375,6 +378,9 @@
     /** */
     private int qryParallelism = DFLT_QUERY_PARALLELISM;
 
+    /** Default query timeout. */
+    private long dfltQryTimeout = DFLT_QRY_TIMEOUT;
```
AMashenkov: It make no sense to put query settings into CacheConfiguration as cache is used for querying only as an entry point and we plan to create a new SQL API.

So, you can move this to IgniteConfiguration and mark as deprecated instantly or add a 'Ignite_Default_Query_Timeout' property to IgniteSystemProperties.

samaitra:
@AMashenkov thank you for reviewing the changes and sharing feedback, I have incorporated the review comments. 

Please review and share if any further changes need to be made.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -1032,6 +1039,29 @@ public IgniteConfiguration setSqlQueryHistorySize(int size) {
...
     }
 
+    /**
+     * Defines the default query timeout.
+     *
+     * Defaults to {@link #DFLT_QRY_TIMEOUT}.
+     *
+     * @return Default query timeout.
+     */
+    public long getDfltQryTimeout() {
```
pavlukhin: I suppose according to our conventions a method name should be `getDefaultQueryTimeout`.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -1032,6 +1039,29 @@ public IgniteConfiguration setSqlQueryHistorySize(int size) {
...
+        return dfltQryTimeout;
+    }
+
+    /**
+     * Sets timeout in milliseconds for default query timeout.
```
pavlukhin: It is good to mention that value 0 means that there is no timeout.

samaitra:
@pavlukhin I have updated the PR. Please take a look.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryAdapter.java
```diff
@@ -390,9 +390,10 @@ public int pageSize() {
 
     /** {@inheritDoc} */
     @Override public CacheQuery<T> timeout(long timeout) {
-        A.ensure(timeout >= 0, "timeout >= 0");
-
-        this.timeout = timeout;
+        if (timeout >= 0)
+            this.timeout = timeout;
+        else
+            this.timeout = cctx.gridConfig().getDefaultQueryTimeout();
```
pavlukhin: From the first glance changed behavior of this method looks counter intuitive: if a passed argument is lesser than 0 a default timeout from a configuration is used. For me a check employed before asserting that negative values are invalid.

And does this change really makes queries use a default timeout? I suppose a small test will answer this question.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/utils/PlatformConfigurationUtils.java
```diff
@@ -622,6 +622,8 @@ public static void readIgniteConfiguration(BinaryRawReaderEx in, IgniteConfigura
         ClientListenerProtocolVersion ver) {
         if (in.readBoolean())
             cfg.setClientMode(in.readBoolean());
+        if (in.readBoolean())
```
pavlukhin: As far as I remember, *nullable* properties should be protected with `in.readBoolean()` construct. We need to double check it with .NET contributors. Also, the added property should be added to a .NET configuration mirror `modules/platforms/dotnet/Apache.Ignite.Core/IgniteConfiguration.cs`

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryParser.java
```diff
@@ -134,6 +135,49 @@ public QueryParserResult parse(String schemaName, SqlFieldsQuery qry, boolean re
         return res;
     }
 
+    /**
+     * Create parameters from query.
+     *
+     * @param qry Query.
+     * @return Parameters.
+     */
+    public QueryParameters fromQuery(SqlFieldsQuery qry) {
```
pavlukhin: As this method is now a part of `QueryParser`, a name `queryParameters` for the method looks more suitable.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryParser.java
```diff
@@ -134,6 +135,49 @@ public QueryParserResult parse(String schemaName, SqlFieldsQuery qry, boolean re
...
+
+            if (qry0.getNestedTxMode() != null)
+                nestedTxMode = qry0.getNestedTxMode();
+
+            autoCommit = qry0.isAutoCommit();
+
+            batchedArgs = qry0.batchedArguments();
+        }
+
+        int timeout = 0;
```
pavlukhin: Initialization to 0 is not necessary here.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryParser.java
```diff
@@ -134,6 +135,49 @@ public QueryParserResult parse(String schemaName, SqlFieldsQuery qry, boolean re
...
+
+        if (qry.getTimeout() > 0)
+            timeout = qry.getTimeout();
+        else
+            timeout = (int)idx.kernalContext().config().getDefaultQueryTimeout();
```
pavlukhin: Actually, numeric overflow problems are interesting here. It seems that max Integer value is not so big, do we validate a value when use `setDefaultQueryTimeout`?

samaitra: yes, we validate timeout in `QueryUtils.validateTimeout` which is invoked inside `SqlFieldsQuery.setTimeout`.

I think `int` is mainly used since we check remaining timeout in IgniteH2Indexing.operationTimeout using Math.min/Math.max methods

modules/core/src/main/java/org/apache/ignite/cache/query/SqlFieldsQuery.java
```diff
@@ -66,7 +69,7 @@
     private boolean collocated;
 
     /** Query timeout in millis. */
-    private int timeout;
+    private int timeout = DFLT_QUERY_TIMEOUT;
```
pavlukhin: Also we will have some limitations regarding default timeout usage for other platform clients and ODBC. For them it is impossible to realize if a default timeout should be used if current values is 0. We can add handling of `-1` for them in this PR (similarly to `ClientCacheSqlFieldsQueryRequest`). Changing platforms code can be done later.

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/odbc/OdbcRequestHandler.java
```diff
@@ -338,9 +338,12 @@ private SqlFieldsQueryEx makeQuery(String schema, String sql, Object[] args, int
         SqlFieldsQueryEx qry = makeQuery(schema, sql);
 
         qry.setArgs(args);
-        qry.setTimeout(timeout, TimeUnit.SECONDS);
         qry.setAutoCommit(autoCommit);
 
+        if (timeout >= 0) {
```
pavlukhin: Curly braces are redundant according to code style.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cache/ClientCacheSqlFieldsQueryRequest.java
```diff
@@ -82,8 +82,11 @@ public ClientCacheSqlFieldsQueryRequest(BinaryRawReaderEx reader) {
                 .setReplicatedOnly(replicatedOnly)
                 .setEnforceJoinOrder(enforceJoinOrder)
                 .setCollocated(collocated)
-                .setLazy(lazy)
-                .setTimeout(timeout, TimeUnit.MILLISECONDS);
+                .setLazy(lazy);
+
+        if (timeout >= 0) {
```
pavlukhin: Curly braces are redundant according to code style.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryParser.java
```diff
@@ -134,6 +136,51 @@ public QueryParserResult parse(String schemaName, SqlFieldsQuery qry, boolean re
...
+                nestedTxMode = qry0.getNestedTxMode();
+
+            autoCommit = qry0.isAutoCommit();
+
+            batchedArgs = qry0.batchedArguments();
+        }
+
+        int timeout;
+
+        if (qry.getTimeout() >= 0) {
```
pavlukhin: Curly braces are redundant according to code style.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,200 @@
...
+        cfg.setDefaultQueryTimeout(DEFAULT_QUERY_TIMEOUT);
+
+        return cfg;
+    }
+
+    /** {@inheritDoc} */
+    @Override protected void afterTest() throws Exception {
+        super.afterTest();
+
+        for (Ignite g : G.allGrids())
```
pavlukhin: It is enough to clear caches just from one node, loop is not needed here.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,200 @@
...
+            }
+
+            try (QueryCursor<List<?>> ignored = cursor) {
+                cursor.getAll();
+
+                if (checkCanceled)
+                    fail("Query not canceled");
+            }
+            catch (CacheException ex) {
+                log().error("Got expected exception", ex);
```
pavlukhin: Can be just `error(...)`.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+            checkCleanState();
+        }
+    }
+
+    /**
+     * Validates clean state on all participating nodes after query cancellation.
+     */
+    private void checkCleanState() throws IgniteCheckedException {
+        for (int i = 0; i < GRIDS_CNT; i++) {
```
pavlukhin: `for (Ignite grid : G.allGrids())`?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,200 @@
...
+            IgniteEx grid = grid(i);
+
+            // Validate everything was cleaned up.
+            ConcurrentMap<UUID, ?> map = U.field(((IgniteH2Indexing)U.field((GridProcessor)U.field(
```
pavlukhin: I suggest to use something like `((IgniteH2Indexing)((IgniteEx)grid).context().query().getIndexing()).runningSqlQueries()` for check.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,200 @@
...
+                    fail("Query not canceled");
+            }
+            catch (CacheException ex) {
+                log().error("Got expected exception", ex);
+
+                assertNotNull("Must throw correct exception", X.cause(ex, QueryCancelledException.class));
+            }
+
+            // Give some time to clean up.
+            Thread.sleep(TimeUnit.MILLISECONDS.convert(timeoutUnits, timeUnit) + 3_000);
```
pavlukhin: Here we can use a construct `GridTestUtils.waitForCondition` to reduce wait time.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/local/IgniteCacheLocalQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,153 @@
...
+            cursor = cache.query(qry);
+
+            ignite.scheduler().runLocal(new Runnable() {
+                @Override public void run() {
+                    cursor.close();
+                }
+            }, timeoutUnits, timeUnit);
+        }
+
+        try(QueryCursor<List<?>> ignored = cursor) {
```
pavlukhin: Space between `try` and `(`.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/local/IgniteCacheLocalQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,153 @@
...
+        loadCache(cache);
+
+        SqlFieldsQuery qry = new SqlFieldsQuery(QUERY);
+
+        final QueryCursor<List<?>> cursor;
+        if (timeout) {
+            qry.setTimeout(timeoutUnits, timeUnit);
+
+            cursor = cache.query(qry);
+        } else {
```
pavlukhin: `else` should start from new line.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/local/IgniteCacheLocalQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,153 @@
...
+        super.beforeTestsStarted();
+
+        startGrid(0);
+    }
+
+    /** {@inheritDoc} */
+    @Override protected void afterTest() throws Exception {
+        super.afterTest();
+
+        for (Ignite g : G.allGrids())
```
pavlukhin: It is enough to clear caches just from one node, loop is not needed here.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,200 @@
...
+import org.apache.ignite.internal.util.typedef.G;
+import org.apache.ignite.internal.util.typedef.X;
+import org.apache.ignite.internal.util.typedef.internal.U;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.junit.Test;
+
+/**
+ * Tests distributed SQL queries default timeouts.
+ */
+public class IgniteCacheDistributedQueryDefaultTimeoutSelfTest extends GridCommonAbstractTest {
```
pavlukhin: Do not forget to add test to suite, e.g. `IgniteBinaryCacheQueryTestSuite2`.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/local/IgniteCacheLocalQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,153 @@
...
+import org.apache.ignite.internal.util.typedef.X;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.junit.Test;
+
+import static org.apache.ignite.cache.CacheMode.LOCAL;
+
+/**
+ * Tests local query default timeouts.
+ */
+public class IgniteCacheLocalQueryDefaultTimeoutSelfTest extends GridCommonAbstractTest {
```
pavlukhin: Do not forget to add test to suite, e.g. `IgniteBinaryCacheQueryTestSuite2`.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
samaitra: We are actually using IgniteEx below which is available using grid(i) method of GridAbstractTest

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/local/IgniteCacheLocalQueryDefaultTimeoutSelfTest.java
samaitra: added the test in IgniteBinaryCacheQueryTestSuite2

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
samaitra: added the test in IgniteBinaryCacheQueryTestSuite2

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/local/IgniteCacheLocalQueryDefaultTimeoutSelfTest.java
samaitra: updated the PR

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/local/IgniteCacheLocalQueryDefaultTimeoutSelfTest.java
samaitra: updated the PR

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/local/IgniteCacheLocalQueryDefaultTimeoutSelfTest.java
samaitra: yes, updated the PR

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
samaitra: I have not found similar usage of runningSqlQueries() in any other tests, can you please share an example

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
samaitra: I was reviewing other tests files like IgniteCacheDistributedQueryStopOnCancelOrTimeoutSelfTest and IgniteCacheQueryStopOnCancelOrTimeoutDistributedJoinSelfTest and noticed that similar Thread.sleep method is used for clean up process. 

Can you please share an example for GridTestUtils.waitForCondition that can be used here

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
samaitra: updated the PR

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
samaitra: yes updated the PR

samaitra:
> Additionally to comments for code lines there are some questions regarding testing approach. Actually, I do not see how different are expectations from test with and without configured default timeout. Basically, it is good to test following:
> 
> 1. Only default timeout specified -- query is cancelled after it, it is clear that it was cancelled by timeout.
> 2. Explicit timeout overrides default timeout.


1. My understanding is in the method testQueryCancel when we are passing the 3rd argument timeout as false then only default timeout is specified.
2. When we are passing the 3rd argument timeout as true then explicit timeout overrides default timeout.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -1075,6 +1083,34 @@ public IgniteConfiguration setSqlQueryHistorySize(int size) {
...
+     * {@code 0} means there is no timeout (this
+     * is a default value)
+     *
+     * @param dfltQryTimeout Timeout in milliseconds.
+     * @return {@code this} for chaining.
+     */
+    public IgniteConfiguration setDefaultQueryTimeout(long dfltQryTimeout) {
+        A.ensure(dfltQryTimeout >= 0 && dfltQryTimeout <= Integer.MAX_VALUE, "default query timeout value should be valid Integer.");
```
mcherkasov: please remove dfltQryTimeout <= Integer.MAX_VALUE, it's useless check, because it's always "true".

pavlukhin: @mcherkasov, not quite. `dfltQryTimeout` is long and can be greater than `Integer.MAX_VALUE`

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/local/IgniteCacheLocalQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,153 @@
...
+            }, timeoutUnits, timeUnit);
+        }
+
+        try (QueryCursor<List<?>> ignored = cursor) {
+            cursor.iterator();
+
+            fail("Expecting timeout");
+        }
+        catch (Exception e) {
+            assertNotNull("Must throw correct exception", X.cause(e, QueryCancelledException.class));
```
pavlukhin: I do not  see how do we distinguish _manual cancel_ vs _explicit query timeout_ vs _default query timeout here_.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,199 @@
...
+            try (QueryCursor<List<?>> ignored = cursor) {
+                cursor.getAll();
+
+                if (checkCanceled)
+                    fail("Query not canceled");
+            }
+            catch (CacheException ex) {
+                error("Got expected exception", ex);
+
+                assertNotNull("Must throw correct exception", X.cause(ex, QueryCancelledException.class));
```
pavlukhin: Similarly, I do not  see how do we distinguish _manual cancel_ vs _explicit query timeout_ vs _default query timeout here_.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/local/IgniteCacheLocalQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,153 @@
...
+                p++;
+            }
+        }
+    }
+
+    /**
+     * Tests query execution with default query timeout.
+     */
+    @Test
+    public void testQueryZeroTimeout() {
```
pavlukhin: Can we rename it to `testQueryDefaultTimeout`?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    /** */
+    @Test
+    public void testRemoteQueryExecutionCancel0() throws Exception {
+        testQueryCancel(CACHE_SIZE, VAL_SIZE, QRY_1, 1, TimeUnit.MILLISECONDS, false, true);
+    }
+
+    /** */
+    private void testQueryCancel(int keyCnt, int valSize, String sql, int timeoutUnits, TimeUnit timeUnit,
```
pavlukhin: Should a method name contain _timeout_ instead of _cancel_?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,199 @@
...
+        cfg.setDefaultQueryTimeout(DEFAULT_QUERY_TIMEOUT);
+
+        return cfg;
+    }
+
+    /** {@inheritDoc} */
+    @Override protected void afterTest() throws Exception {
+        super.afterTest();
+
+        G.ignite().cache(DEFAULT_CACHE_NAME).removeAll();
```
pavlukhin: I suppose `grid(0)` should be here.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/local/IgniteCacheLocalQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,153 @@
...
+        super.beforeTestsStarted();
+
+        startGrid(0);
+    }
+
+    /** {@inheritDoc} */
+    @Override protected void afterTest() throws Exception {
+        super.afterTest();
+
+        G.ignite().cache(DEFAULT_CACHE_NAME).removeAll();
```
pavlukhin: I suppose `grid(0)` should be here.

pavlukhin:
> > Additionally to comments for code lines there are some questions regarding testing approach. Actually, I do not see how different are expectations from test with and without configured default timeout. Basically, it is good to test following:
> > 
> > 1. Only default timeout specified -- query is cancelled after it, it is clear that it was cancelled by timeout.
> > 2. Explicit timeout overrides default timeout.
> 
> 1. My understanding is in the method testQueryCancel when we are passing the 3rd argument timeout as false then only default timeout is specified.
> 2. When we are passing the 3rd argument timeout as true then explicit timeout overrides default timeout.

@samaitra I copied `IgniteCacheLocalQueryDefaultTimeoutSelfTest` and executed it on master. And tests passed, so I suppose tests do not distinguish cancellation after _default timeout_ vs other possible ways.

modules/platforms/dotnet/Apache.Ignite.Core/IgniteConfiguration.cs
```diff
@@ -72,6 +72,11 @@ public class IgniteConfiguration
         /// </summary>
         public const int DefaultJvmMaxMem = -1;
 
+        /// <summary>
+        /// Default query timeout.
+        /// </summary>
+        public static readonly TimeSpan DefaultQueryTimeout = TimeSpan.FromMilliseconds(0);
```
ptupitsyn: Same name as the property below - won't compile. When the property is `X`, default value for it is `DefaultX`, so in this case it would be `DefaultDefaultQueryTimeout`. But since it is just `default(TimeSpan)`, I would remove it altogether.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
samaitra: Since the underlying exception thrown is QueryCancelledException I think it is ok to keep it as is. I kept the naming similar to IgniteCacheDistributedQueryStopOnCancelOrTimeoutSelfTest tests

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/local/IgniteCacheLocalQueryDefaultTimeoutSelfTest.java
samaitra: yes sure

samaitra:
> > > Additionally to comments for code lines there are some questions regarding testing approach. Actually, I do not see how different are expectations from test with and without configured default timeout. Basically, it is good to test following:
> > > 
> > > 1. Only default timeout specified -- query is cancelled after it, it is clear that it was cancelled by timeout.
> > > 2. Explicit timeout overrides default timeout.
> > 
> > 
> > 
> > 1. My understanding is in the method testQueryCancel when we are passing the 3rd argument timeout as false then only default timeout is specified.
> > 2. When we are passing the 3rd argument timeout as true then explicit timeout overrides default timeout.
> 
> @samaitra I copied `IgniteCacheLocalQueryDefaultTimeoutSelfTest` and executed it on master. And tests passed, so I suppose tests do not distinguish cancellation after _default timeout_ vs other possible ways.

@pavlukhin Thank you for reviewing and sharing feedback. 

If you note the tests in IgniteCacheDistributedQueryStopOnCancelOrTimeoutSelfTest also, it is similar to Default Query timeout tests. I suppose the reason being we are throwing QueryCancelledException for either manual cancel or query being cancelled due to timeout. I agree throwing a separate exception for timeout vs manual cancel will be better approach to handle different scenarios and help with logging and exception handling.

I am also thinking if that change can be taken up as separate issue and we can close on default query handling change in this PR.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
samaitra: If you note the tests in IgniteCacheDistributedQueryStopOnCancelOrTimeoutSelfTest also, it is similar to Default Query timeout tests. I suppose the reason being we are throwing QueryCancelledException for either manual cancel or query being cancelled due to timeout. I agree throwing a separate exception for timeout vs manual cancel will be better approach to handle different scenarios and help with logging and exception handling.

The only way to differentiate the explicit query timeout vs default query timeout is like in `IgniteCacheLocalQueryDefaultTimeoutSelfTest.testQueryDefaultTimeout() test the method testQuery() method is called with param timeout as false and it will result in query cancellation due to default query timeout.`

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
samaitra: yes updated

modules/platforms/dotnet/Apache.Ignite.Core/IgniteConfiguration.cs
samaitra: @ptupitsyn Thank you for reviewing and sharing your feedback. Can you please clarify if you are asking to rename the parameter from `DefaultQueryTimeout ` to `QueryTimeout ` or to remove this parameter from `IgniteConfiguration.cs`

ptupitsyn: I'm asking to remove `public static readonly TimeSpan DefaultQueryTimeout`, because it is zero.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
```diff
@@ -0,0 +1,198 @@
...
+        for (int i = 0; i < GRIDS_CNT; i++) {
+            IgniteEx grid = grid(i);
+
+            // Validate everything was cleaned up.
+            ConcurrentMap<UUID, ?> map = U.field(((IgniteH2Indexing)U.field((GridProcessor)U.field(
+                grid.context(), "qryProc"), "idx")).mapQueryExecutor(), "qryRess");
+
+            String msg = "Map executor state is not cleared";
+
+            // TODO FIXME Current implementation leaves map entry for each node that's ever executed a query.
```
pavlukhin: Is that this thing really needed? We try to avoid TODOs in new node. Exceptionally there could be a TODO referring to a ticket.

pavlukhin:
>@pavlukhin Thank you for reviewing and sharing feedback.
>
>If you note the tests in IgniteCacheDistributedQueryStopOnCancelOrTimeoutSelfTest also, it is similar to Default Query timeout tests. I suppose the reason being we are throwing QueryCancelledException for either manual cancel or query being cancelled due to timeout. I agree throwing a separate exception for timeout vs manual cancel will be better approach to handle different scenarios and help with logging and exception handling.
>
>I am also thinking if that change can be taken up as separate issue and we can close on default query handling change in this PR.

I can imagine checking exception message (to check that it is timeout) and approximate cancellation time (to check that it was _default_ timeout). What do you think of it?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/IgniteCacheDistributedQueryDefaultTimeoutSelfTest.java
samaitra: I have updated the PR.

samaitra:
> > @pavlukhin Thank you for reviewing and sharing feedback.
> > If you note the tests in IgniteCacheDistributedQueryStopOnCancelOrTimeoutSelfTest also, it is similar to Default Query timeout tests. I suppose the reason being we are throwing QueryCancelledException for either manual cancel or query being cancelled due to timeout. I agree throwing a separate exception for timeout vs manual cancel will be better approach to handle different scenarios and help with logging and exception handling.
> > I am also thinking if that change can be taken up as separate issue and we can close on default query handling change in this PR.
> 
> @samaitra I can imagine checking exception message (to check that it is timeout) and approximate cancellation time (to check that it was _default_ timeout). What do you think of it?

@pavlukhin I looked into the QueryCancelledException class and it has a generic message as "The query was cancelled while executing.". Also my understanding is it is used commonly for both manually cancelled exception cases and also for timeout exception cases. 

We can consider creating  a separate exception when query is cancelled for timeout vs manually cancelled but I think that change can be taken up in separate jira issue and PR considering the scope.

PR #6499, opened 2019-04-24
IGNITE-11671 Thin client: Client may hang when connected to a starting server

NSAmelchev:
Connection id (long) consists of two 32-bits parts:
1. node order
2. Unique identifier (int)

I removed node order and make unique identifier long type.
Node order was used only for assertion on dropping a connection. Moreover, this assertion fails after 2^31 attempts to connect. 

It fixed NPE on a server side and a client hanging.


dspavlov:
IGNITE-11671 | Resolved
-- | --

PR #6500, opened 2019-04-24
IGNITE-10913

denis-chudov:


modules/benchmarks/src/main/java/org/apache/ignite/internal/benchmarks/AbstractOptimizationTestBenchmark.java
```diff
@@ -0,0 +1,279 @@
...
+    /** */
+    protected IgniteConfiguration getConfiguration(String igniteInstanceName) {
+        IgniteConfiguration cfg = new IgniteConfiguration();
+
+        cfg.setIgniteInstanceName(igniteInstanceName);
+
+        cfg.setDiscoverySpi(
+            new TcpDiscoverySpi()
+                .setIpFinder(
+                    new TcpDiscoveryMulticastIpFinder()
```
zstan: no need multicast here

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/Cache64kPartitionsPersistent.java
```diff
@@ -0,0 +1,109 @@
...
+
+            ignite.cluster().active(true);
+
+            IgniteCache<Object, Object> cache = ignite.getOrCreateCache("default");
+
+            for (int i = 0; i < 1000; i++) {
+                cache.put(i, 1);
+
+                System.gc();
+                System.gc();
```
zstan: more gc clean !!!

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStore.java
```diff
@@ -59,7 +61,16 @@
...
+    private volatile File cfgFile;
+
+    /** */
+    private final IgniteOutClosure<File> fileProvider;
+
+    /**
+     * Caches the existance state of storage file. After it is initialized, it will be not set to null
+     * during FilePageStore lifecycle.
+     */
+    private volatile Boolean fileExists;
```
EdShangGG: The same as above. We could calculate it all the time when it would be necessary.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStoreManager.java
```diff
@@ -452,7 +452,10 @@ public FilePageStoreManager(GridKernalContext ctx) {
 
     /** {@inheritDoc} */
     @Override public void onPartitionCreated(int grpId, int partId) throws IgniteCheckedException {
-        // No-op.
+        PageStore store = getStore(grpId, partId);
+
+        if (store instanceof FilePageStore)
+            ((FilePageStore)store).dereferenceFile();
```
EdShangGG: It would be redundant if we remove cfgFile field.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/Cache64kPartitionsPersistent.java
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.ignite.configuration.WALMode;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.junit.Test;
+
+/**
+ *
+ */
+public class Cache64kPartitionsPersistent extends GridCommonAbstractTest {
+    /** */
+    private boolean persistenceEnabled;
```
EdShangGG: When could it be false?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/Cache64kPartitionsPersistent.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * @throws Exception if failed.
+     */
+    @Test
+    public void testManyPartitionsWithPersistence() throws Exception {
+        persistenceEnabled = true;
+
+        checkManyPartitions();
```
EdShangGG: Sorry, but I didn't get when the test could fail.

EdShangGG: Would it be run on TC?

denis-chudov: It seems this test class is not needed since I made a benchmark for this optimization. I removed it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStore.java
denis-chudov: We still need the field caching the flag of file existance to reduce file system calls.

PR #6501, opened 2019-04-24
IGNITE-11805 .NET runnable core executable

ashapkin:


docker/apache-ignite-net/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Build runtime image
+FROM microsoft/dotnet:aspnetcore-runtime
```
ptupitsyn: We don't need ASP.NET Core, please use regular runtime base image.

docker/apache-ignite-net/README.txt
```diff
@@ -0,0 +1,23 @@
...
+        cd modules/docker
+
+3) Copy Apache Ignite.NET's binaries to libs and publish folders
+
+        cp -r ../../modules/platforms/dotnet/bin/libs libs
+        cp -r ../../modules/platforms/dotnet/bin/netcoreapp2.0/publish/ publish
+
+4) Build docker image
+
+        docker build . -t apacheignite-net/ignite[:<version>]
```
ptupitsyn: I believe this should be `apacheignite/ignite-net`, please verify with @vveider .

modules/platforms/dotnet/Apache.Ignite/IgniteCoreRunner.cs
```diff
@@ -0,0 +1,105 @@
...
+    using System.Configuration;
+    using System.Linq;
+    using System.Threading;
+    using Apache.Ignite.Config;
+    using Apache.Ignite.Core;
+
+    /// <summary>
+    /// Runner class.
+    /// </summary>
+    public class IgniteCoreRunner
```
ptupitsyn: Can we make this class `static`?

modules/platforms/dotnet/Apache.Ignite/IgniteCoreRunner.cs
```diff
@@ -0,0 +1,105 @@
...
+                Console.WriteLine("ERROR: " + e);
+
+                Environment.Exit(-1);
+            }
+        }
+
+        /// <summary>
+        /// Prints help.
+        /// </summary>
+        private static void PrintHelp()
```
ptupitsyn: Let's extract this to a separate class and reuse in both runners (add `includeService` argument or similar).

modules/platforms/dotnet/Apache.Ignite/IgniteCoreRunner.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            Console.WriteLine("Usage: Apache.Ignite.exe [-options]");
```
ptupitsyn: There is no exe file on .NET Core.

modules/platforms/dotnet/build.ps1
```diff
@@ -195,6 +195,18 @@ if (!$skipDotNet) {
 	echo "Starting MsBuild: '$msBuildCommand'"
 	cmd /c $msBuildCommand
 
+    # Build core
```
ptupitsyn: Here and below we override the exit code from Classic .NET. Let's move that check higher, so that MSBuild failure exits the script as before.

ptupitsyn:
Please update DEVNOTES.txt: we have a new system requirement for release builds - .NET Core 2.0+.

docker/apache-ignite-net/README.txt
vveider: Yes it is.

modules/platforms/dotnet/build.ps1
```diff
@@ -202,6 +206,21 @@ if (!$skipDotNet) {
...
+    if ($clean) {
+        $cleanCommand = "dotnet clean $targetSolution -c $configuration"
+        echo "Starting dotnet clean: '$cleanCommand'"
+        cmd /c $cleanCommand
+    }
+
+    $publishCommand = "dotnet publish $targetSolution -c $configuration"
+	echo "Starting dotnet publish: '$publishCommand'"
+	cmd /c $publishCommand
+}
```
ptupitsyn: Check exit code here, we don't want to continue if the build fails.

PR #6528, opened 2019-05-06
IGNITE-11346

Maxoid:
Remote client authentication failed for the CommandHandler in the case where it optional on the server

### DIFF
--- a/modules/clients/src/test/java/org/apache/ignite/internal/client/integration/ClientTcpAuthTest.java
+++ b/modules/clients/src/test/java/org/apache/ignite/internal/client/integration/ClientTcpAuthTest.java
@@ -1,0 +1,101 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.client.integration;
+
+import org.apache.ignite.configuration.IgniteConfiguration;
+import org.apache.ignite.internal.client.GridClient;
+import org.apache.ignite.internal.client.GridClientCompute;
+import org.apache.ignite.internal.client.GridClientConfiguration;
+import org.apache.ignite.internal.client.GridClientData;
+import org.apache.ignite.internal.client.GridClientException;
+import org.apache.ignite.internal.client.GridClientNode;
+import org.apache.ignite.internal.client.GridClientPredicate;
+import org.apache.ignite.internal.client.GridClientProtocol;
+import org.apache.ignite.internal.client.ssl.GridSslContextFactory;
+import org.apache.ignite.internal.processors.security.impl.TestSecurityPluginConfiguration;
+import org.apache.ignite.plugin.security.SecurityCredentials;
+import org.apache.ignite.plugin.security.SecurityCredentialsBasicProvider;
+import org.apache.ignite.spi.discovery.tcp.TestAuthProcessor;
+import org.junit.Test;
+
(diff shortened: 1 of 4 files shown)


modules/clients/src/test/java/org/apache/ignite/internal/client/integration/ClientTcpAuthTest.java
```diff
@@ -0,0 +1,101 @@
...
+import org.junit.Test;
+
+import java.util.Collections;
+
+/**
+ * Tests the Authorization in client-server communication.
+ */
+public class ClientTcpAuthTest extends ClientAbstractSelfTest {
+
+    @Override
```
ibessonov: I see no javadocs here, "codestyle" suite will most likely fail if you run it.

modules/clients/src/test/java/org/apache/ignite/internal/client/integration/ClientTcpAuthTest.java
```diff
@@ -0,0 +1,101 @@
...
+    @Override protected boolean useSsl() {
+        return false;
+    }
+
+    /** {@inheritDoc} */
+    @Override protected GridSslContextFactory sslContextFactory() {
+        return null;
+    }
+
+}
```
ibessonov: Please add empty line at the end of this file.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -1085,7 +1085,7 @@ else if (spiState == LOOPBACK_PROBLEM) {
      */
     private void localAuthentication(SecurityCredentials locCred){
         assert spi.nodeAuth != null;
-        assert locCred != null;
+//        assert locCred != null;
```
ibessonov: Are you sure that this won't break anything? This variable is passed into "authenticateNode" after this line and I believe that this method may not expect null value. Please clarify this moment and either remove comment or return this assert back.

modules/clients/src/test/java/org/apache/ignite/internal/client/integration/ClientTcpAuthTest.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.ignite.plugin.security.SecurityCredentialsBasicProvider;
+import org.apache.ignite.spi.discovery.tcp.TestAuthProcessor;
+import org.junit.Test;
+
+import java.util.Collections;
+
+/**
+ * Tests the Authorization in client-server communication.
+ */
+public class ClientTcpAuthTest extends ClientAbstractSelfTest {
```
ibessonov: Please add this test to some suite, otherwise it won't run on TC.

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TestAuthProcessor.java
```diff
@@ -0,0 +1,272 @@
...
+        return new AllowAll();
+    }
+
+    /** {@inheritDoc} */
+    @Override public boolean isGlobalNodeAuthentication() {
+        return false;
+    }
+
+    /** {@inheritDoc} */
+    @Override public SecurityContext authenticate( AuthenticationContext authCtx) throws IgniteCheckedException {
```
ibessonov: Please reformat this method according to current codestyle.

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TestAuthProcessor.java
```diff
@@ -0,0 +1,272 @@
...
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public boolean systemOperationAllowed( SecurityPermission perm ) {
+            return true;
+        }
+    }
+
+    private static class AllowSome implements SecurityContext, Serializable {
```
ibessonov: There are many missing javadocs in this file as well, please add them.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
Maxoid: That's the point of this issue! We enable authentication, but only for external access. We do not need it for the internal node communications, so there could not be credentials, but the authentication is still presented in the cluster. Only the internal auth-node logic can make a deсision about such case, so we can't insist on this level, that credentials must be present in any way.

modules/clients/src/test/java/org/apache/ignite/internal/client/integration/ClientTcpAuthTest.java
Maxoid: done

PR #6546, opened 2019-05-16
IGNITE-11848: New Monitoring. Phase1

nizhikov:
* Metrics(Gauges, Counters) API
* MonitoringExporterSPI(JMX, SQL View implementations)
* GridMonitoringManager
* MonitoringRegistry


modules/opencensus/src/main/java/org/apache/ignite/internal/processors/monitoring/opencensus/OpenCensusMetricExporterSpi.java
```diff
@@ -0,0 +1,279 @@
...
+    private List<TagKey> tags = new ArrayList<>();
+
+    /**
+     * Opencensus measures.
+     * Values obtained from Ignite recorded to them.
+     */
+    private Map<String, Measure> measures = new HashMap<>();
+
+    /** */
+    private static Function<Metric, Measure> CREATE_LONG = m ->
```
daradurvs: CREATE_LONG -> createLongF

modules/opencensus/src/main/java/org/apache/ignite/internal/processors/monitoring/opencensus/OpenCensusMetricExporterSpi.java
```diff
@@ -0,0 +1,279 @@
...
+        MeasureLong.create(m.name(), m.description(), "");
+
+    /** */
+    private static Function<Metric, Measure> CREATE_DOUBLE = m ->
```
daradurvs: CREATE_DOUBLE -> createDoubleF

modules/opencensus/src/main/java/org/apache/ignite/internal/processors/monitoring/opencensus/OpenCensusMetricExporterSpi.java
```diff
@@ -0,0 +1,279 @@
...
+        MeasureDouble.create(m.name(), m.description(), "");
+
+    /** {@inheritDoc} */
+    @Override public void export() {
+        StatsRecorder recorder = Stats.getStatsRecorder();
+
+        try (Scope globalScope = tagScope()) {
```
daradurvs: 'globalScope' is never used

modules/opencensus/src/main/java/org/apache/ignite/internal/processors/monitoring/opencensus/OpenCensusMetricExporterSpi.java
```diff
@@ -0,0 +1,279 @@
...
+     * @param sendNodeId Flag value.
+     */
+    public void setSendNodeId(boolean sendNodeId) {
+        this.sendNodeId = sendNodeId;
+    }
+
+    /**
+     * If {@code true} then {@link #CONSISTENT_ID_TAG} will be added to each exported measure.
+     *
+     * @param consistentId Flag value.
```
daradurvs: consistentId  -> sendConsistentId

modules/indexing/src/main/java/org/apache/ignite/spi/metric/sql/SqlViewExporterSpi.java
```diff
@@ -0,0 +1,126 @@
...
+import org.apache.ignite.internal.processors.query.h2.SchemaManager;
+import org.apache.ignite.internal.processors.query.h2.sys.view.SqlAbstractLocalSystemView;
+import org.apache.ignite.internal.processors.query.h2.sys.view.SqlSystemViewColumnCondition;
+import org.apache.ignite.internal.util.typedef.F;
+import org.apache.ignite.spi.IgniteSpiAdapter;
+import org.apache.ignite.spi.IgniteSpiContext;
+import org.apache.ignite.spi.IgniteSpiException;
+import org.apache.ignite.spi.metric.Metric;
+import org.apache.ignite.spi.metric.MetricExporterSpi;
+import org.apache.ignite.internal.processors.metric.MetricRegistry;
```
daradurvs: Unused import

modules/indexing/src/main/java/org/apache/ignite/spi/metric/sql/SqlViewExporterSpi.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /** {@inheritDoc} */
+    @Override public void setExportFilter(Predicate<Metric> filter) {
+        this.filter = filter;
+    }
+
+    /** */
+    public class MetricSetLocalSystemView extends SqlAbstractLocalSystemView {
+        /**
+         * @param set Metric set.
```
daradurvs: Obsolete parameter

modules/opencensus/src/main/java/org/apache/ignite/internal/processors/monitoring/opencensus/OpenCensusMetricExporterSpi.java
nizhikov: Why is that?

modules/opencensus/src/main/java/org/apache/ignite/internal/processors/monitoring/opencensus/OpenCensusMetricExporterSpi.java
nizhikov: Why is that?

modules/opencensus/src/main/java/org/apache/ignite/internal/processors/monitoring/opencensus/OpenCensusMetricExporterSpi.java
nizhikov: It's OK. 

Scope used by the undelying structures in `MeasureMap#put` calls.

modules/core/src/main/java/org/apache/ignite/spi/metric/ReadOnlyMetricRegistry.java
```diff
@@ -0,0 +1,59 @@
...
+     *
+     * @param prefixes prefixes for all metrics.
+     * @return Proxy implementation that will search and create only metrics with specified prefixes.
+     */
+    public <T extends ReadOnlyMetricRegistry> T withPrefix(String... prefixes);
+
+    /**
+     * @return Metrics stored in this group.
+     */
+    public Collection<Metric> getMetrics();
```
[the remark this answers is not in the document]
nizhikov: 1. It's OK if someone calls `reset` for metric from the collection. Because `reset` is a member of `Metric` interface.

2. `add` can't be executed for `getMetrics` results, because, for now, we return `ConcurrentHashMap#ValuesView` collection.

I'd rather keep it as is.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricNameUtils.java
```diff
@@ -0,0 +1,140 @@
...
+    /**
+     * Parsed metric name parts.
+     *
+     * Example - metric name - "io.statistics.PRIMARY_KEY_IDX.pagesCount".
+     * root = io - JMX tree root.
+     * subName = statistics.PRIMARY_KEY_IDX - bean name.
+     * msetName = io.statistics.PRIMARY_KEY_IDX - prefix to search metrics for a bean.
+     * mname = pagesCount - metric name.
+     */
+    public static class MetricName {
```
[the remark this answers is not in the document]
nizhikov: I don't think we should separate this class from `MetricNameUtils`.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricRegistryImpl.java
```diff
@@ -0,0 +1,211 @@
...
+     * @param lsnrs Listeners.
+     * @param <T> Type of consumed object.
+     */
+    private <T> void notifyListeners(T t, List<Consumer<T>> lsnrs) {
+        for (Consumer<T> lsnr : lsnrs) {
+            try {
+                lsnr.accept(t);
+            }
+            catch (Exception e) {
+                log.warning("Metric listener error", e);
```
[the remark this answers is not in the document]
nizhikov: Yes, this is expected.
But default constructor is for the test usage only.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricRegistry.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    /**
+     * Register {@link BooleanMetric} that value will be queried from specified supplier.
+     *
+     * @param name Name.
+     * @param supplier Supplier.
+     * @param description Description.
+     */
+    public void register(String name, BooleanSupplier supplier, @Nullable String description);
```
ascherbakoff: What's the point in having register* set of methods ?
It can be replaced by obtaining writeable metric and setting value provided by closure on it.
Please provide more detailed documentation for use cases of this functionality.

modules/core/src/main/java/org/apache/ignite/internal/metric/IoStatisticsHolderCache.java
```diff
@@ -48,12 +50,23 @@
...
+     * @param mreg Metric registry.
      */
-    public IoStatisticsHolderCache(String cacheName, int grpId) {
+    public IoStatisticsHolderCache(String cacheName, int grpId, MetricRegistry mreg) {
         assert cacheName != null;
 
         this.cacheName = cacheName;
         this.grpId = grpId;
+
+        MetricRegistry mset = mreg.withPrefix(CACHE_GROUP.metricGroupName(), cacheName);
```
ascherbakoff: Variable names mset and mreg doesn't follow code style.
Use names registry and locRegistry instead.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -0,0 +1,293 @@
...
+            mset.objectMetric("ThreadFactoryClass", String.class, THRD_FACTORY_DESC).value("");
+        }
+    }
+
+    /**
+     * Creates a MetricSet for an stripped executor.
+     *
+     * @param svc Executor.
+     */
+    private void monitorStrippedPool(StripedExecutor svc) {
```
ascherbakoff: Fix typo monitorStrippedPool -> monitorStripedPool

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricRegistry.java
```diff
@@ -0,0 +1,196 @@
...
+ */
+
+package org.apache.ignite.internal.processors.metric;
+
+import java.util.function.BooleanSupplier;
+import java.util.function.DoubleSupplier;
+import java.util.function.IntSupplier;
+import java.util.function.LongSupplier;
+import java.util.function.Supplier;
+import org.apache.ignite.internal.processors.metric.impl.DoubleMetricImpl;
```
ascherbakoff: What's the point in having Impl suffix ?
org.apache.ignite.internal.processors.metric.impl.DoubleMetric looks cleaner.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricRegistryImpl.java
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * @param log Logger.
+     */
+    public MetricRegistryImpl(IgniteLogger log) {
+        this.log = log;
+    }
+
+    /** {@inheritDoc} */
+    @Override public MetricRegistry withPrefix(String prefix) {
```
ascherbakoff: Calling this method multiple times will produce unnecessary garbage.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricRegistryImpl.java
```diff
@@ -0,0 +1,212 @@
...
+
+import static org.apache.ignite.internal.processors.metric.impl.MetricUtils.metricName;
+import static org.apache.ignite.internal.util.lang.GridFunc.nonThrowableSupplier;
+
+/**
+ * Simple implementation.
+ */
+public class MetricRegistryImpl implements MetricRegistry {
+    /** Registered metrics. */
+    private final ConcurrentHashMap<String, Metric> metrics = new ConcurrentHashMap<>();
```
ascherbakoff: Each key here represents a full hierarchy string and likely to consume a lot of heap space.
Please provide a test measuring heap occupation with large number of registered metrics.
You could look at GridAffinityAssignmentJolBenchmark for reference.
Probably better to use some tree-like strucure here (trie)

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricRegistry.java
nizhikov: 1. It is useful for monitoring some third-party entities.
`ExecutorService` is one of the examples.
Please, take a look into `GridMetricManager#monitorExecutor`.

2. Several existing metrics calculated at the moment they queried. You can find examples in `DataRegionMetricsImpl`: "TotalUsedPages", "TotalAllocatedSize". I think we can rewrite most of it by the way you mentioned, but this is the scope of the following tickets.

modules/core/src/main/java/org/apache/ignite/internal/metric/IoStatisticsHolderCache.java
nizhikov: Why is that? Can you provide a rule I violate with these names?

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
nizhikov: Fixed. Thanks.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricRegistry.java
nizhikov: `DoubleMetric` is the name of the interface. You can find it [here](https://github.com/apache/ignite/pull/6546/files#diff-2f22f348ff33397a422d8b9d1a766c13). I think we should avoid classes with the same names but different packages.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricRegistryImpl.java
nizhikov: What do you mean by "unnecessary"?

`withPrefix` method required to make cleaner code for metrics creation.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricRegistryImpl.java
nizhikov: 1. `GridMetricsJolBenchmark` added.

2. I assume we should optimize the internal implementation of `MetricRegistry` for frequently reading. Seems, it is opposite to lower memory consumption. What do you think?

3. Anyway, I will create a ticket to spend some extra time to optimize the internal implementation of `MetricsRegistry` after all current Ignite metrics will be migrated to it.

nizhikov: [IGNITE-11912](https://issues.apache.org/jira/browse/IGNITE-11912) created.

modules/core/src/main/java/org/apache/ignite/internal/metric/IoStatisticsHolderCache.java
ascherbakoff: I think mreg, mset is not very descriptive, check RULE2 of abbreviation rules.
reg or registry are more descriptive variants for MetricRegistry instance.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricRegistryImpl.java
ascherbakoff: Each call of this method produce new proxy object. I think it's better to change implementation to get rid of new object creation. However, this method is not expected to be called often, so it's up to you to fix it.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricRegistryImpl.java
ascherbakoff: Metrics are not expected to be read by key oftenly from my view. 
Which scenarios do you have in mind? 
Find for me to address this in separate ticket.

modules/core/src/main/java/org/apache/ignite/internal/metric/IoStatisticsHolderCache.java
ascherbakoff: mset doesn't look very descriptive, see RULE2 of abbreviation rules.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricRegistryImpl.java
nizhikov: > Metrics are not expected to be read by key

Each exporter iterate through the set(or subset) of metrics and essentially write in some specific format following information:

1. Metric name.
2. Metric value.
3. Optionally, metric description.

If this scenario is basic then we should keep metric name ready to be written(or returned) somewhere.

Anyway, I agree with you that we should optimize internal implementation. I will investigate this issue in IGNITE-11912.

Thanks for pointing this out.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricRegistryImpl.java
nizhikov: >  get rid of new object creation

I understand your concern.
I agree we will win some memory here.
But we violate code readability a lot, from my point of view.
I expect that we will construct this object only on metric initialization or other rare operations.

So I will keep it as is.

PR #6550, opened 2019-05-17
IGNITE-11671 Thin client: Client may hang when connected to a starting server

alex-plekhanov:


modules/core/src/test/java/org/apache/ignite/client/ConnectToStartingNodeTest.java
```diff
@@ -0,0 +1,90 @@
...
+                catch (Exception ignore) {
+                    // No-op.
+                }
+            }
+        });
+    }
+
+    /**
+     * Test that client can't connect to server before discovery SPI start.
+     */
```
NSAmelchev: Missed javadoc for Exception

PR #6552, opened 2019-05-17
IGNITE-11715 Allow extra options to be passed in to ignite.sh from Docker

sdarlington:
Ticket was specifically about `-nojmx` flag but I don't think there's a good reason to limit it to that.


dmagda:
Stephen, thanks for the contribution!

PR #6553, opened 2019-05-18
IGNITE-11854 Fix performance problem when creating arrays

dmelnichuk:


kulinskyvs:
I've tried the patched and it looks like something is still wrong. 
Below is a simple test case that I use to test the patch (motivated by https://issues.apache.org/jira/browse/IGNITE-11854):

```
for i in range(9):
        arr_len = 3_000_000 * (i + 1)
        content = bytearray(arr_len)

        for i in range(arr_len):
            content[i] = i % 256

        start_time = time.time()
        my_cache.put("key_bin", content, value_hint=ByteArrayObject)
        put_elapsed = time.time() - start_time

        start_time = time.time()
        my_cache.get("key_bin")
        get_elapsed = time.time() - start_time
        print("size: {}, put time:{} secs, get time: {} secs".format(len(content), put_elapsed, get_elapsed))
```
And below the results I've got:

> size: 3000000, put time:1.118593692779541 secs, get time: 0.5311076641082764 secs
size: 6000000, put time:2.2962043285369873 secs, get time: 1.0077407360076904 secs
size: 9000000, put time:3.3157756328582764 secs, get time: 1.4834346771240234 secs
size: 12000000, put time:4.277689695358276 secs, get time: 2.100264549255371 secs
size: 15000000, put time:5.646029949188232 secs, get time: 2.890953540802002 secs
size: 18000000, put time:6.578389883041382 secs, get time: 2.971928596496582 secs
size: 21000000, put time:8.114448547363281 secs, get time: 3.5916993618011475 secs
size: 24000000, put time:8.961003541946411 secs, get time: 4.091071605682373 secs
size: 27000000, put time:10.465842247009277 secs, get time: 5.260945081710815 secs
size: 30000000, put time:10.91893482208252 secs, get time: 5.647061586380005 secs

So, it takes more than 10 secs to put 30MB into Ignite and more than 5 secs to load this data from it. 
And it looks like the most time is spent just to iterate though the given byte array in PrimitiveArray#from_python (see below)

```
buffer = bytearray(header)

for x in value:
     buffer += cls.primitive_type.from_python(x)
return bytes(buffer)
```

Is it the expected behavior? In case yes, what can be solution to put big binary data (more than 100MB ) into Ignite using thin python client?

Also, I've noticed that instead of retrieving bytearray I'm getting a list of ints when calling `cache.get()` and some of the values are negative? Is it a defect?

dmelnichuk:
> I've tried the patch and it looks like something is still wrong.
> Is it the expected behavior?

Well, there is always a room for improvement, but to be able to deliver `pyignite` in reasonable time, I started on the following premises:

- all Ignite data types, that represent signed integers (`Byte`, `Short`, `Integer`, and `Long`) must be represented with Python boundless integer (`int`) type,
- implementation of sequential Ignite data types (arrays, `Map`, `Sequence`) must be based on the corresponding singular types' representations. For example, `ByteArrayObject` uses internal methods of `ByteObject`.

Both measures have simplified the code greatly. So, yes, it is the expected behavior: to create an array, client really should create each of its element iteratively, as if building it from abstract `Iterable[int]`.

>In case yes, what can be solution to put big binary data (more than 100MB ) into Ignite using thin python client?

I must admit, I don't have a simple solution to this problem. I think it may be possible to provide a fast method of creating `ByteArray` from `bytearray` as an in situ optimization, still maintaining its default `Iterable[int]` mapping. Retrieving `ByteArray` from Ignite as `bytearray` will definitely break the user API though.

But I can't help but wonder: is storing/retrieving large amounts of unstructured data into/from Ignite is really a case? Especially with Python client? Python is not particularly famous for its byte stream processing speed.

> Also, I've noticed that instead of retrieving bytearray I'm getting a list of ints when calling `cache.get()` and some of the values are negative? Is it a defect?

Not at all, it is an expected behavior. As I said earlier, Java's `Byte` is a signed integer (-128…127), while Python's `bytearray` consists of unsigned elements (0…255). You are not losing data; it is just a matter of its representation.

kulinskyvs:
Thank you for your extended feedback.

> But I can't help but wonder: is storing/retrieving large amounts of unstructured data into/from Ignite is really a case? Especially with Python client? Python is not particularly famous for its byte stream processing speed.

We were planning to use Ignite as a distributed cache to store already trained and ready for use ML models > the data is actually a pickled bytearray  of real scikitlearn model objects. We expect to have some models around 150-200 MB. 
Ignite is used mainly since this is a selected and approved cache solution in our organization to share the the user session data between java-based microservices > we didn't want to introduce new technology stack.

Keeping in mind your responses may be we should look for another solution in our case.

dmelnichuk:
> We were planning to use Ignite as a distributed cache to store already trained and ready for use ML models > the data is actually a pickled bytearray of real scikitlearn model objects. We expect to have some models around 150-200 MB.

@kulinskyvs Looks like a valid case to me. Please hang on for a moment, I will ask around what can be done about it.

dmelnichuk:
@kulinskyvs I have created a ticket based on our discussion: [IGNITE-11897](https://issues.apache.org/jira/browse/IGNITE-11897). Feel free to watch and comment.

kulinskyvs:
> @kulinskyvs I have created a ticket based on our discussion: [IGNITE-11897](https://issues.apache.org/jira/browse/IGNITE-11897). Feel free to watch and comment.

Thank you a lot for your cooperation!

kyri-ales:
Hello @dmelnichuk,
Do we have an ETA on this optimization?

Thank you.

dspavlov:
This PR is associated with a resolved ticket. Closing the PR, please reopen it if you still need it.

PR #6560, opened 2019-05-21
IGNITE-11757

vldpyatkov:
Missed partitions during rebalancing when new blank node joins


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridCachePartitionedSupplyEventsSelfTest.java
```diff
@@ -82,8 +82,8 @@
 
         lsnrs.put(new IgnitePredicate<CacheRebalancingEvent>() {
                 @Override public boolean apply(CacheRebalancingEvent evt) {
-                    //fail("Should not miss any partitions!");
-                    log.warning("Missed partition " + evt.partition() + " from node " + evt.node().consistentId());
+                    fail("Should not miss any partitions!");
+//                    log.warning("Missed partition " + evt.partition() + " from node " + evt.node().consistentId());
```
alamar: I think you can remove that line.

PR #6563, opened 2019-05-23
IGNITE-11670 Java thin client: Queries are inconsistent in case of fa…

alex-plekhanov:
…ilover


modules/core/src/test/java/org/apache/ignite/client/ReliabilityTest.java
```diff
@@ -168,62 +153,92 @@ public void testSingleServerFailover() throws Exception {
...
+                for (Iterator<Cache.Entry<Integer, String>> it = cur.iterator(); it.hasNext(); it.next()) {
+                    cnt++;
+
+                    if (cnt == 1)
+                        dropAllThinClientConnections(Ignition.allGrids().get(0));
+                }
+
+                assertEquals(2, cnt);
+            }
+            catch (ClientReconnectedException expected) {
```
NSAmelchev: Why you don't check that the expected exception will be thrown?

PR #6595, opened 2019-06-06
IGNITE-11685 Java thin client: Handle multiple async requests in parallel

alex-plekhanov:


modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientChannel.java
```diff
@@ -109,6 +131,10 @@
 
             out = sock.getOutputStream();
             in = sock.getInputStream();
+
+            GridUnsafeDataInput dis = new GridUnsafeDataInput();
```
pavlukhin: What are benefits over standard `DataInputStream`?

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientChannel.java
```diff
@@ -141,52 +174,108 @@
...
-        if (resId != reqId)
-            throw new ClientProtocolError(String.format("Unexpected response ID [%s], [%s] was expected", resId, reqId));
+        BinaryInputStream resIn;
 
-        int status = resIn.readInt();
+        status = readInt();
 
-        if (status != 0) {
-            resIn = new BinaryHeapInputStream(read(resSize - MIN_RES_SIZE));
+        int hdrSize = (int)(totalBytesRead - bytesReadOnStartReq);
```
pavlukhin: An approach with counting `totalBytesRead` seems fragile to me because counting methods should be used everywhere. I think counting a header size could be encapsulated into a method or an object counting just bytes while reading header only.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
```diff
@@ -118,72 +116,49 @@
      */
     public <T> T service(
         ClientOperation op,
-        Consumer<BinaryOutputStream> payloadWriter,
-        Function<BinaryInputStream, T> payloadReader
+        Consumer<PayloadOutputStream> payloadWriter,
```
pavlukhin: I do not fully understand an introduction of _payload_ streams. An abstraction of `PayloadOutputStream` seem only enriches `BinaryOutputStream` with a reference to a channel which is only used as an _attachment_ without adding any related behavior. I can imagine alternative approaches:
1. Use a composition, e.g. `ClientResponse` containing `BinaryInputStream` and `ClientChannel`.
2. Use `BiConsumer` and `BiFuctiontion` in `ReliableChannel.service` receiving a channel (or some complex _context_ object) as a second argument.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
```diff
@@ -216,19 +191,30 @@ public ProtocolVersion serverVersion() {
...
 
-        try {
-            ch.close();
-        }
-        catch (Exception ignored) {
-        }
+    /** */
+    private synchronized void changeServer(ClientChannel oldCh) {
+        if (oldCh == ch && ch != null) {
+            if (!backups.isEmpty()) {
```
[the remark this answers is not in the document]
alex-plekhanov: I don't fully understand what kind of improvement do you mean?
I haven't change naming, there was such naming before PR. I think naming will be changed as far as "Best effort affinity" for java thin client will be implemented.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
```diff
@@ -80,32 +76,34 @@
...
-                this.backups.add(a);
+                backups.add(a);
         }
 
         ClientConnectionException lastEx = null;
 
         for (int i = 0; i < addrs.size(); i++) {
             try {
-                ch = chFactory.apply(new ClientChannelConfiguration(clientCfg).setAddress(primary)).get();
+                ch = chFactory.apply(new ClientChannelConfiguration(clientCfg).setAddress(addrs.get(i))).get();
```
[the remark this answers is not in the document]
alex-plekhanov: In my opinion eager connection makes sense. Problems will be detected at an earlier stage. For example, in case of some wrong server address configuration with eager connection exception will be thrown at some kind of initialization code, but with lazy initialization, we will get an exception only when business logic is running.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientChannel.java
alex-plekhanov: `DataInputStream` is a big-endian input stream, while the protocol implies little-endian.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientChannel.java
alex-plekhanov: In current protocol version header size is fixed, but in next versions, header size depends on different conditions (status, topology change, etc), so to calculate header size we need either to increment the counter each time readLong/readInt/... is called or increment the counter inside readLong/readInt/... There are much more parameters to read from the header in next protocol versions, so incrementing counter inline is more error-prone and more verbose.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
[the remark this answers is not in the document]
pavlukhin: Yes, such reasoning definitely makes sense. Lazy initialization can also have benefits, e.g. reducing time taken by _initialization code_. Well, seems like a trade off. I trust your own decision.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientChannel.java
pavlukhin: Did not know it, thank you!

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
alex-plekhanov: Yes, it was introduced only for a reference to a channel. I've tried to minimize changes to current calls of `ReliableChannel#service`, `ReliableChannel#request`.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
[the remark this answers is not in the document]
pavlukhin: I thought about something like:
```
private synchronized void changeServer(ClientChannel oldCh) {
  if (singleServer) {
    ch = null;
    return;
  }
  ...
}
```
But actually I do not see much win here.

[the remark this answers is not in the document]
pavlukhin: Will stay tuned for name changes with "Best effort affinity".

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
pavlukhin: Do you consider a separate refactoring ticket for improving used abstraction?

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientChannel.java
pavlukhin: Good to know it. But it still be error-prone as long as we have a input streams at hand which allow us to read data without counting.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
alex-plekhanov: No, I will fix it in this ticket.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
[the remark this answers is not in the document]
isapego: Does Java thin client throw an exception if it can not establish connection to one of nodes?

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientChannel.java
alex-plekhanov: Perhaps we can extract input stream and `read...` methods to a new class, but it seems overcomplicated for this task to me.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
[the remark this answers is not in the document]
alex-plekhanov: It throws an exception if it can't establish a connection to all nodes (exception is not thrown if connected to at least one node from the address list).
Actually, I've reverted fix for this line, since `primary` is changed in `changeServer` method and there was no error in the original implementation (I misunderstood this behavior).

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
```diff
@@ -114,76 +118,51 @@
...
+            }
+            catch (ClientConnectionException e) {
+                if (failure == null)
+                    failure = e;
+                else
+                    failure.addSuppressed(e);
 
-                    changeServer();
-                }
+                changeServer(ch);
```
pavlukhin: I can imagine 2 concurrent threads having `ch == null` and `this.ch == null` as well. If so a server list will be rolled twice. In certain configurations (e.g. 2 servers) it can lead to situation when the same (currently down) address will be chosen again. Currently I have only one thought (not very pleasant) how it can be resolved using the same `channel() ... changeServer()` structure. It is using special `BrokenChannel` objects instead of `null` to indicate that channel was invalidated.

Correct me if I was totally wrong.

alex-plekhanov: Yes, such case is possible. I've made another fix to handle this.

PR #6599, opened 2019-06-07
IGNITE-9409 Use current version, modern URL for Ignite download.

alamar:


modules/yarn/src/main/resources/ignite.properties
```diff
@@ -0,0 +1,23 @@
+#
```
shroman: Do we need this file? It's copied from `core`.

alamar: Yes, we need this file since ignite-yarn does not depend on ignire-core: It's just a deployment mechanism.

modules/yarn/src/main/java/org/apache/ignite/yarn/IgniteProvider.java
```diff
@@ -21,33 +21,29 @@
...
     /** */
-    public static final String DOWNLOAD_LINK = "http://tiny.cc/updater/download_community.php";
+    public static final Logger log = Logger.getLogger(IgniteProvider.class.getSimpleName());
 
     /** */
-    private ClusterProperties props;
-
-    /** */
-    private String latestVersion = null;
+    public static final String DOWNLOAD_LINK = "https://www.apache.org/dyn/mirrors/mirrors.cgi?action=download&filename=/ignite/";
```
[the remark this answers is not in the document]
alamar: We're going to add version/file to it so it will look like:
ignite/2.7.0/apache-ignite-2.7.0-bin.zip
in the end

modules/yarn/src/main/resources/ignite.properties
shroman: ok, got it.

PR #6614, opened 2019-06-17
GG-19530  Resend auth failed message to client from router node

akalash:


dspavlov:

PR #6627, opened 2019-06-21
IGNITE-11925: QueryMetrics migration.

nizhikov:


modules/core/src/main/java/org/apache/ignite/internal/processors/metric/impl/MetricUtils.java
```diff
@@ -83,6 +83,32 @@ public static boolean compareAndSet(LongMetricImpl m, long expect, long update)
...
+    /**
+     * Update metrics value only if current value if greater then {@code update}.
+     *
+     * @param m Metric to update.
+     * @param update New value.
+     */
+    public static void setIfGreater(LongMetricImpl m, long update) {
+        long v = m.value();
+
+        while (v < update && !LongMetricImpl.updater.compareAndSet(m, v, update))
```
Mmuzaf: I think it's better to use Java 8 approach here with UnarnaryOperator - 

`oldVal.updateAndGet(x -> x < newVal ? newVal : x) == newVal;`

nizhikov: Proposed implementation is less efficient then current because `updateAndGet` implementation executes at least one CAS operation(see `AtomicLongFieldUpdater#updateAndGet` sources) event value is not change. And my implementation returns immediately in the case we don't want to update the value.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryMetricsAdapter.java
```diff
@@ -126,41 +125,106 @@ public void update(long duration, boolean fail) {
...
+        fails.reset();
     }
 
     /** {@inheritDoc} */
     @Override public String toString() {
         return S.toString(GridCacheQueryMetricsAdapter.class, this);
     }
+
+    /** Query metrics snapshot. */
+    public static class QueryMetricsSnapshot implements QueryMetrics, Externalizable {
```
Mmuzaf: I think this class can be private. Isn't it?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryMetricsAdapter.java
```diff
@@ -126,41 +125,106 @@ public void update(long duration, boolean fail) {
...
+
+        /** */
+        private int fails;
+
+        /** */
+        public QueryMetricsSnapshot() {
+        }
+
+        /** */
+        public QueryMetricsSnapshot(long minTime, long maxTime, double avgTime, int execs, int fails) {
```
Mmuzaf: Can we improve javadoc here? I know that it is an internal class and it is not really necessary, but such a thing will be helpful for further code style analysis.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryMetricsAdapter.java
```diff
@@ -126,41 +125,106 @@ public void update(long duration, boolean fail) {
      *
```
Mmuzaf: It seems to me that javadoc became obsolete and must be fixed. Isn't it?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryManager.java
```diff
@@ -185,7 +185,7 @@
     private int maxIterCnt;
 
     /** */
-    private volatile GridCacheQueryMetricsAdapter metrics = new GridCacheQueryMetricsAdapter();
+    private volatile GridCacheQueryMetricsAdapter metrics;
```
Mmuzaf: We can change here to `QueryMetrics`, isn't it?

Mmuzaf:
I see that both of `GridCacheQueryDetailMetricsAdapter` and `GridCacheQueryMetricsAdapter` used simultaneously for tracking query executions. 

Currently, you've migrated only `GridCacheQueryMetricsAdapter` and it seems that our new query framework starts to track such metrics, but what about detailed metrics? Should we migrate only them (they have a broader scope and it seems more helpful for user) or migrate both of them?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryMetricsAdapter.java
```diff
@@ -126,41 +125,106 @@ public void update(long duration, boolean fail) {
...
+        /** */
+        private double avgTime;
+
+        /** */
+        private int execs;
+
+        /** */
+        private int fails;
+
+        /** */
```
Mmuzaf: I know it's obvious, but such things look much better. Please, see example below:

```
    /**
     * Required by {@link Externalizable}.
     */
    public GridCacheQueryDetailMetricsAdapter() {
        // No-op.
    }
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryMetricsAdapter.java
```diff
@@ -126,41 +125,106 @@ public void update(long duration, boolean fail) {
...
-        out.writeLong(minTime.get());
-        out.writeLong(maxTime.get());
-        out.writeDouble(averageTime());
-        out.writeInt(execs.intValue());
-        out.writeInt(fails.intValue());
+    public QueryMetrics snapshot() {
+        long minTimeVal = minTime.longValue();
+
+        return new QueryMetricsSnapshot(
+            minTimeVal == Long.MAX_VALUE ? 0 : minTimeVal,
```
Mmuzaf: This is not related to your changes directly. But I think it is a time when we can fix it (it affects public metrics value).
`GridCacheQueryDetailMetricsAdapter` sets `minTime` to `-1` and it seems to me that it is more natural to have a negative default value for this parameter rather than `Long.MAX_VALUE`.

Just mark this comment `resolved` if you don't want to.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryMetricsAdapter.java
nizhikov: fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryMetricsAdapter.java
nizhikov: fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryMetricsAdapter.java
nizhikov: I don't think we can confuse the user with `minTime=Long.MAX_VALUE` on the first measurement, so I rather keep things as is.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryMetricsAdapter.java
nizhikov: fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryMetricsAdapter.java
nizhikov: I don't want to make it private for the case if some third-party tools want to aggregate `QueryMetrics`(and creates own instance of `QueryMetricsSnapshot`

nizhikov:
1. This ticket for `QueryMetrics`.
2. 
`QueryMetrics` - aggregated query metrics per cache.
`GridCacheQueryDetailMetricsAdapter` per query metrics, so it's has to be "list", not "metrics".

PR #6638, opened 2019-06-25
 IGNITE-11941: VM metrics moved to GridMetricManager.

nizhikov:


modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -290,4 +393,170 @@ private void monitorStripedPool(StripedExecutor svc) {
...
+        /** Updates metric to the provided values. */
+        public void update(MemoryUsage usage) {
+            init.value(usage.getInit());
+            used.value(usage.getUsed());
+            committed.value(usage.getCommitted());
+            max.value(usage.getMax());
+        }
+    }
 }
+
```
alex-plekhanov: Redundant NL

modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/ClusterMetricsImpl.java
```diff
@@ -270,27 +409,27 @@ public ClusterMetricsImpl(GridKernalContext ctx, GridLocalMetrics vmMetrics, lon
...
     /** {@inheritDoc} */
     @Override public long getUpTime() {
-        return vmMetrics.getUptime();
+        return upTime.value();
     }
 
     /** {@inheritDoc} */
     @Override public long getStartTime() {
-        return vmMetrics.getStartTime();
+        return rt.getStartTime();
```
alex-plekhanov: Why different approaches are used for `getUpTime()` and `getStartTime()`? Both from the same MXBean, both used only once in `ClusterMetricsImpl`. Wouldn't it be better to get start time from the metrics registry too?

nizhikov: `startTime` is a constant that can be derived from `upTime` so we don't need it in `MetricRegistry`.

PR #6646, opened 2019-06-27
IGNITE-11874 Fix mismatch between idle_verify results with and without -dump option.

denis-chudov:


modules/core/src/test/java/org/apache/ignite/util/GridCommandHandlerAbstractTest.java
```diff
@@ -200,4 +209,48 @@ protected void addExtraArguments(List<String> args) {
...
+                        && ((GridNearTxPrepareFutureAdapter) fut).tx().system())
+                        continue;
+
+                    if (fut instanceof GridDhtTxPrepareFuture
+                        && ((GridDhtTxPrepareFuture) fut).tx().system())
+                        continue;
+
+                    log.error("Expecting no active future [node=" + ig.localNode().id() + ", fut=" + fut + ']');
+
+                    hasFutures = true;
```
vmalin: It is possible to move {fail} here and remove hasFutures varaible

modules/core/src/test/java/org/apache/ignite/util/GridCommandHandlerAbstractTest.java
```diff
@@ -200,4 +209,48 @@ protected void addExtraArguments(List<String> args) {
...
     }
+
+    /**
+     * Checks if all non-system txs and non-system mvcc futures are finished.
+     */
+    protected void checkUserFutures() {
+        for (Ignite ignite : G.allGrids()) {
+            IgniteEx ig = (IgniteEx)ignite;
+
+            final Collection<GridCacheFuture<?>> futs = ig.context().cache().context().mvcc().activeFutures();
```
vmalin: Please add more descriptive variable name. mvccFuts of something like this

denis-chudov: fixed

dspavlov:
IGNITE-11874 | Resolved
-- | --

PR #6661, opened 2019-07-02
IGNITE-11954

vmalin:


modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CacheValidateIndexes.java
```diff
@@ -186,20 +189,20 @@ public UUID nodeId() {
...
                         for (IndexValidationIssue is : res.issues())
-                            logger.logWithIndent(is, 2);
+                            logger.info(DOUBLE_INDENT + is);
                     }
                 }
             }
 
             if (!errors)
-                logger.log("no issues found.");
+                logger.severe("no issues found.");
```
denis-chudov: seems that should be 'log' instead of 'severe'

modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -421,40 +526,40 @@ private String requestDataFromConsole(String msg) {
...
-        logger.log("Exit codes:");
-        logger.logWithIndent(EXIT_CODE_OK + " - successful execution.", 2);
-        logger.logWithIndent(EXIT_CODE_INVALID_ARGUMENTS + " - invalid arguments.", 2);
-        logger.logWithIndent(EXIT_CODE_CONNECTION_FAILED + " - connection failed.", 2);
-        logger.logWithIndent(ERR_AUTHENTICATION_FAILED + " - authentication failed.", 2);
-        logger.logWithIndent(EXIT_CODE_UNEXPECTED_ERROR + " - unexpected error.", 2);
+        logger.info("By default commands affecting the cluster require interactive confirmation.");
+        logger.info("Use " + CMD_AUTO_CONFIRMATION + " option to disable it.");
+        logger.info("");
+
```
denis-chudov: empty lines here can be removed

modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -182,30 +277,40 @@ public int execute(List<String> rawArgs) {
                 }
             }
 
+            logger.info("Command [" + commandName + "] finished with code: " + EXIT_CODE_OK);
             return EXIT_CODE_OK;
         }
         catch (IllegalArgumentException e) {
-            logger.error("Check arguments.", e);
+            logger.severe("Check arguments. " + CommandLogger.errorMessage(e));
+            logger.info("Command [" + commandName + "] finished with code: " + EXIT_CODE_INVALID_ARGUMENTS);
```
denis-chudov: I think you should either add empty line or replace 'info' with 'severe' here and 3 places below in this method

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CacheValidateIndexes.java
vmalin: fixed

modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
vmalin: fixed

modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
vmalin: I think empty lines here increase readability. I would prefer to keep them.

PR #6664, opened 2019-07-03
IGNITE-11958 JDBC connection validation should use it's own task instead of cache validation task

denis-chudov:


modules/core/src/main/java/org/apache/ignite/internal/jdbc/JdbcConnectionValidationTask.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.ignite.IgniteException;
+import org.apache.ignite.compute.ComputeJob;
+import org.apache.ignite.compute.ComputeJobAdapter;
+import org.apache.ignite.compute.ComputeJobResult;
+import org.apache.ignite.compute.ComputeTaskSplitAdapter;
+import org.apache.ignite.internal.util.typedef.F;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * This task is used for JDBCConnection validation.
```
vmalin: I think it's better to use {@link JCDCConnection}. Especially when we have to version of this class and one is deprecated.

modules/core/src/main/java/org/apache/ignite/internal/jdbc/JdbcConnectionValidationTask.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @deprecated Using Ignite client node based JDBC driver is preferable.
```
vmalin: Looks like copy-paste from JDBCConnection. Should it be here?

denis-chudov: I think yes, as JdbcConnection and related things are deprecated and will be removed in ignite 3.

modules/core/src/main/java/org/apache/ignite/internal/jdbc/JdbcConnectionValidationTask.java
denis-chudov: fixed

PR #6665, opened 2019-07-03
IGNITE-11959 NullPointerException If transaction failed and failure h…

tkalkirill:
…andler doesn't configured explicitly.


dspavlov:
IGNITE-11959 | Resolved
-- | --

PR #6672, opened 2019-07-08
IGNITE-711 `Need to complete implementation of Java 8 examples`

1vanan:


modules/core/src/main/java/org/apache/ignite/internal/visor/node/VisorNodeDataCollectorJob.java
```diff
@@ -156,7 +156,6 @@ protected void memoryMetrics(VisorNodeDataCollectorJobResult res) {
         try {
             List<VisorMemoryMetrics> memoryMetrics = res.getMemoryMetrics();
 
-            // TODO: Should be really fixed in IGNITE-7111.
```
pavlukhin: Was this line removed intentionally? It refers to another ticket.

examples/src/test/java/org/apache/ignite/examples/IgfsExamplesSelfTest.java
```diff
@@ -20,13 +20,14 @@
 //import org.apache.ignite.examples.igfs.*;
 //import org.apache.ignite.internal.util.typedef.internal.*;
```
pavlukhin: Commented imports can be removed.

PR #6687, opened 2019-07-12
IGNITE-11979 Add ability to set default parallelism of rebuild indexes in ignite system properties

denis-chudov:


modules/core/src/main/java/org/apache/ignite/internal/processors/query/schema/SchemaIndexCacheVisitorImpl.java
```diff
@@ -68,6 +69,15 @@
     /** Whether to stop the process. */
     private volatile boolean stop;
 
+    static {
+        int parallelism = IgniteSystemProperties.getInteger(INDEX_REBUILDING_PARALLELISM, 0);
+
+        if (parallelism > 0)
+            DFLT_PARALLELISM = Math.min(parallelism, Runtime.getRuntime().availableProcessors());
```
AMashenkov: Why we override user defined value here? 
There is a good practice to give user some feedback about wrong value or recommend the optimal one.

denis-chudov: There is a description in javadoc for INDEX_REBUILDING_PARALLELISM, and overriding is mentioned there. Do you suggest to add logging for this overriding?

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1230,6 +1230,14 @@
      */
     public static final String IGNITE_LOG_CLASSPATH_CONTENT_ON_STARTUP = "IGNITE_LOG_CLASSPATH_CONTENT_ON_STARTUP";
 
+    /**
+     * Index rebuilding parallelism level. If specified, sets the count of threads that are used for index rebuilding
+     * and can only be greater than <code>0</code>, otherwise default value will be used. Maximum count of threads
+     * can't be greater than total available processors count.
+     * Default value is minimum of <code>4</code> and processors count / 4, but always greater than <code>0</code>.
```
AMashenkov: Why different markup is used for "4"  and for "processors count / 4" ?
Also, it is not clear what are values allowed.

Can it be like: 
"Index rebuilding parallelism level. 
It sets a number of threads will be used for index rebuilding. Zero value means default should be used.
Default value is calculated as "CPU count / 4"

Note: Number of threads is bounded within the range from "1" up to "CPU count".

What do you think?

denis-chudov: Yes, this version is better, I will fix it.

PR #6688, opened 2019-07-15
IGNITE-3195 Rebalancing: IgniteConfiguration.rebalanceThreadPoolSize is wrongly treated

anton-vinogradov:


modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
```diff
@@ -1572,6 +1572,9 @@ public GridSpringResourceContext springContext() {
         /** Query executor service. */
         private ThreadPoolExecutor schemaExecSvc;
 
+        /** Query executor service. */
```
Mmuzaf: Copy-paste misprint, I think we should fix it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemandMessage.java
```diff
@@ -179,7 +179,9 @@ Object topic() {
 
     /**
      * @param topic Topic.
+     * @deprecated Useless.
```
Mmuzaf: I think we should describe here, why it is not used anymore.
Since it's only internal API, should we just remove it without deprecation?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -459,79 +449,59 @@ private void requestPartitions(final RebalanceFuture fut, GridDhtPreloaderAssign
...
+                    try {
+                        if (log.isInfoEnabled())
+                            log.info("Starting rebalance routine [" + grp.cacheOrGroupName() +
+                                ", topVer=" + fut.topologyVersion() +
+                                ", supplier=" + node.id() + ", topic=" + topicId +
+                                ", fullPartitions=" + S.compact(parts.fullSet()) +
+                                ", histPartitions=" + S.compact(parts.historicalSet()) + "]");
+
+                        ctx.io().sendOrderedMessage(node, rebalanceTopic,
+                            demandMsg.convertIfNeeded(node.version()), grp.ioPolicy(), demandMsg.timeout());
```
Mmuzaf: I think the newly created `GridIoPolicy.REBALANCE_POOL` should be used here, not the group one. WDYT?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPreloader.java
```diff
@@ -351,25 +342,26 @@ private IgniteCheckedException stopError() {
     }
 
     /** {@inheritDoc} */
-    @Override public void handleSupplyMessage(int idx, UUID id, final GridDhtPartitionSupplyMessage s) {
-        if (!enterBusy())
-            return;
+    @Override public void handleSupplyMessage(UUID id, final GridDhtPartitionSupplyMessage s) {
+        if (demander.prepareSupplyMessage(s)) {
+            ctx.kernalContext().closure().runLocalSafe(() -> {
```
Mmuzaf: We don't need `runLocalSafe` here, we should use `GridIoPolicy.REBALANCE_POOL` at the moment of message sending (for demander and supplier). 
Such a technique doesn't free the `PUBLIC_POOL` threads and when it will be a really busy rebalance messages will stuck on it.

Mmuzaf:
I see, that we are still adding a batch of `cctx.io().addOrderedCacheGroupHandler` size `getRebalanceThreadPoolSize`.  Why we still need different handlers for the Rebalance messages since we are not using splitting by `topic` technique?


See: `org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java:438`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -111,8 +112,8 @@
     /** Last exchange future. */
     private volatile GridDhtPartitionsExchangeFuture lastExchangeFut;
 
-    /** Cached rebalance topics. */
-    private final Map<Integer, Object> rebalanceTopics;
+    /** Cache rebalance topic. */
+    private final Object rebalanceTopic;
```
Mmuzaf: I think this can be completely removed. It used only as a `stub` here. Isn't it?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -459,79 +449,59 @@ private void requestPartitions(final RebalanceFuture fut, GridDhtPreloaderAssign
...
                     + ", mode=" + cfg.getRebalanceMode() + ", supplier=" + node.id() + ", partitionsCount=" + parts.size()
-                    + ", topVer=" + fut.topologyVersion() + ", localParallelism=" + locStripes
-                    + ", rmtParallelism=" + rmtStripes + ", parallelism=" + rmtTotalStripes + "]");
+                    + ", topVer=" + fut.topologyVersion() + "]");
             }
 
-            final List<IgniteDhtDemandedPartitionsMap> stripePartitions = new ArrayList<>(stripes);
-            for (int i = 0; i < stripes; i++)
-                stripePartitions.add(new IgniteDhtDemandedPartitionsMap());
+            if (!parts.isEmpty()) {
```
Mmuzaf: I'd recommend here to use such way:

```
if (parts.isEmpty())
   return;
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -459,79 +449,59 @@ private void requestPartitions(final RebalanceFuture fut, GridDhtPreloaderAssign
...
+                // Create copy of demand message with new striped partitions map.
+                final GridDhtPartitionDemandMessage demandMsg = d.withNewPartitionsMap(parts);
```
Mmuzaf: We don't need to do a copy message here. Isn't it?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemandMessage.java
anton-vinogradov: Removal will cause incompatibility with older nodes.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
anton-vinogradov: Kept because of compatibility reasons.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
anton-vinogradov: This will cause code refactoring and merge issues.
Since goal it to perform merge to older versions, that's not a good idea to refactor anything.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
anton-vinogradov: Seems, here we have some sync trick. 
This should be refactored and checked at another issue.

anton-vinogradov:
> I see, that we are still adding a batch of `cctx.io().addOrderedCacheGroupHandler` size `getRebalanceThreadPoolSize`. Why we still need different handlers for the Rebalance messages since we are not using splitting by `topic` technique?
> 
> See: `org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java:438`

We have to support request from older nodes.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemandMessage.java
Mmuzaf: I think we should place exactly your words for deprecated annotation description.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
Mmuzaf: Why we should keep such a thing in the master branch only for `not having problems with cherry-pick`? I think we can do it here.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
anton-vinogradov: fixed

anton-vinogradov: fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemandMessage.java
anton-vinogradov: Let's leave this to future refactoring.

PR #6696, opened 2019-07-17
IGNITE-11987: Metrics configuration.

nizhikov:


nizhikov:
@agura Thanks, for the review!
I will resolve your comments after we came to the consensus in dev-list discussion.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -4532,6 +4534,41 @@ else if (log.isInfoEnabled())
...
 
+    /** {@inheritDoc} */
+    @Override public void configureMetric(String registry, String name, String config) {
+        assert registry != null;
+        assert name != null;
+        assert config != null;
+
+        MetricRegistry mreg = ctx.metric().registry(registry);
+
+        if (mreg == null) {
```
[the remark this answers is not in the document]
nizhikov: Good catch. Thank you.

modules/core/src/test/java/org/apache/ignite/internal/metric/MetricsConfigurationTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+        g0.cluster().active(true);
+
+        assertArrayEquals(bounds, g0.context().metric().registry(TX_METRICS)
+            .<HistogramMetric>findMetric(METRIC_SYSTEM_TIME_HISTOGRAM).bounds());
+
+        assertArrayEquals(bounds, g1.context().metric().registry(TX_METRICS)
+            .<HistogramMetric>findMetric(METRIC_SYSTEM_TIME_HISTOGRAM).bounds());
+    }
+
```
NSAmelchev: unnecessary line break

modules/core/src/test/java/org/apache/ignite/internal/metric/MetricsConfigurationTest.java
```diff
@@ -0,0 +1,174 @@
...
+                .<HistogramMetric>findMetric(METRIC_SYSTEM_TIME_HISTOGRAM).bounds());
+
+            assertArrayEquals(bounds, g1.context().metric().registry(TX_METRICS)
+                .<HistogramMetric>findMetric(METRIC_SYSTEM_TIME_HISTOGRAM).bounds());
+        }
+    }
+
+    /** Tests metric configuration applied on all nodes. */
+    @Test
+    public void testNodeRestart() throws Exception {
```
NSAmelchev: I suggest adding a client node to the test (because some histograms like the pme duration presents at clients).
Also, I think we should throw a user-friendly message when trying to configure metrics on a disconnected client.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -249,6 +271,42 @@ public GridMetricManager(GridKernalContext ctx) {
...
+                            name.substring(HISTOGRAM_CFG_PREFIX.length() + 1), (long[]) val));
+
+                        metastorage.listen(n -> n.startsWith(HITRATE_CFG_PREFIX),
+                            (name, oldVal, newVal) -> onHitRateConfigChanged(
+                                name.substring(HITRATE_CFG_PREFIX.length() + 1), (Long) newVal));
+
+                        metastorage.listen(n -> n.startsWith(HISTOGRAM_CFG_PREFIX),
+                            (name, oldVal, newVal) -> onHistgoramConfigChanged(
+                                name.substring(HISTOGRAM_CFG_PREFIX.length() + 1), (long[]) newVal));
+                    } catch (IgniteCheckedException e) {
```
agura: `catch` must be on the next line.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -267,14 +325,45 @@ public GridMetricManager(GridKernalContext ctx) {
      */
     public MetricRegistry registry(String name) {
         return registries.computeIfAbsent(name, n -> {
-            MetricRegistry mreg = new MetricRegistry(name, log);
+            MetricRegistry mreg = new MetricRegistry(name,
```
agura: That is why I suggested separating `MetricRegistry` from some helper class for metrics creation. It's not registry responsibility for getting config from metastorage. But ok. May be split this later.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -267,14 +325,45 @@ public GridMetricManager(GridKernalContext ctx) {
...
+        metaLock.readLock().lock();
+
+        try {
+            if (metastorage == null)
+                return null;
+
+            try {
+                return metastorage.read(key);
+            }
+            catch (IgniteCheckedException e) {
```
agura: Actually it means that data files are corrupted. Failure process should be invoked in such case. I think it should be done automatically if we pass exception on the next level (but I'm not sure).

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -267,14 +325,45 @@ public GridMetricManager(GridKernalContext ctx) {
...
 
+    /**
+     * Reads value from {@link #metastorage}.
+     *
+     * @param key Key.
+     * @param <T> Key type.
+     * @return Value or {@code null} if not found.
+     */
+    private <T extends Serializable> T readFromMetastorage(String key) {
+        metaLock.readLock().lock();
```
agura: See my question about this lock above.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -302,6 +391,109 @@ public void remove(String regName) {
...
+
+        if (m == null) {
+            if (log.isInfoEnabled())
+                log.info("Metric not found[registry=" + splited[0] + ", metricName=" + splited[1] + ']');
+
+            return null;
+        }
+
+        if (!m.getClass().isAssignableFrom(type)) {
+            log.error("Metric '" + name + "' should be a HitRate[type=" + m.getClass().getSimpleName() + ']');
```
agura: `HitRate` should be removed from log message.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -249,6 +271,42 @@ public GridMetricManager(GridKernalContext ctx) {
             spi.setMetricRegistry(this);
 
         startSpi();
+
+        ctx.internalSubscriptionProcessor().registerDistributedMetastorageListener(
```
agura: About metrics in metastorage: at the moment metastorage size will only grow because we never remove metrics related keys from metastorage. For example if user destroy cache than all key should be removed from metastore for this cache. The same for indexes for example and may be something else.

agura: One more case for hitrate and histogram configuration: during debug I want configure some metric only for one particular node. In this case we should keep this configuration in local metastorage. As result we also should change API and functionality because we should override local config during execution of distributed update. Also local key should override global key on node start. I think it could be done as separate issue.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -4611,6 +4611,24 @@ else if (log.isInfoEnabled())
             log.info("\"" + registry + "\" not found.");
     }
 
+    /** {@inheritDoc} */
+    @Override public void configureHitRateMetric(String name, long rateTimeInterval) {
```
agura: I believe that we shouldn't configure HitRate metrics at all because there is no valid cases for it. See corresponding discussion on dev-list. So only histograms should be configurable.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -4611,6 +4611,24 @@ else if (log.isInfoEnabled())
...
+        try {
+            ctx.metric().configureHitRate(name, rateTimeInterval);
+        } catch (IgniteCheckedException e) {
+            throw new IgniteException(e);
+        }
+    }
+
+    /** {@inheritDoc} */
+    @Override public void configureHistogramMetric(String name, long[] bounds) {
```
agura: I really don't like that `IgniteKernal` interface contains this method. It isn't "kernal" functionality at all.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/impl/MetricUtils.java
```diff
@@ -33,7 +33,10 @@
  */
 public class MetricUtils {
     /** Metric name part separator. */
-    public static final String SEPARATOR = ".";
+    public static final char SEPARATOR_CHAR = '.';
+
+    /** Metric name part separator. */
+    public static final String SEPARATOR = "" + SEPARATOR_CHAR;
```
agura: JS style detected %) You don't use `SEPARATOR_CHAR` anywhere except of `SEPARATOR` value. 

It is enough:

public static final String SEPARATOR = ".";

Or use `SEPARATOR_CHAR` in `indexOf` and `lastIndexOf`. 

Any way `"."` looks better than `"" + SEPARATOR_CHAR`, IMHO

modules/core/src/main/java/org/apache/ignite/mxbean/IgniteMXBean.java
```diff
@@ -506,6 +510,32 @@ void runIoTest(
...
+     * Call of this method will change metric configuration across all cluster nodes.
+     *
+     * @param name Metric name.
+     * @param bounds New bounds.
+     * @throws IgniteException If some error occured.
+     */
+    @MXBeanDescription("Configure histogram metric by name.")
+    @MXBeanParametersNames({"name", "cfg"})
+    @MXBeanParametersDescriptions({"Metric name.", "New bounds."})
+    public void configureHistogramMetric(String name, long[] bounds) throws IgniteException;
```
agura: I think we could add `configure` method to the beans that expose metrics from registries for each histogram metric on the fly. What do you think?

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
nizhikov: OK, I changed it to throw `IgniteException` to a higher level.
Anyway, It's a very confusing decision to make "metastore files corrupted" exception checked.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
nizhikov: Why?

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
nizhikov: > During debug I want configure some metric only for one particular node

I thought we don't want to have local configuration? That's why the previous version of this PR was totally reworked.

agura: No, previous version of PR was reworked because it didn't provide distributed way for configuration. But local configuration still could be very helpful for debug when I can increase dimension of histogram for some node(s).

nizhikov: OK. https://issues.apache.org/jira/browse/IGNITE-12426 created.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
nizhikov: Agree. Let's create separate JMX-bean for metrics configuration purposes?

modules/core/src/main/java/org/apache/ignite/mxbean/IgniteMXBean.java
nizhikov: `JMXExporter` should be explicitly configured.
I think we should create separate JMX bean for metrics configuration purposes.
It will be created independently with `JMXExporter` so metrics configuration will be always accessible.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
agura: I'll introduce bean in IGNITE-11927 for metrics enabling/disabling. So it could be good place for it. I also offer to consider possibility introduce such method for every bean and for every histogram that will be created for registry.

modules/core/src/main/java/org/apache/ignite/mxbean/IgniteMXBean.java
agura: Sounds good. But needs some additional thinking. I think on this iteration separate been will be good. Something like `MetricsMXBean` which will contains management methods.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -249,6 +271,42 @@ public GridMetricManager(GridKernalContext ctx) {
...
+                            (name, oldVal, newVal) -> onHistgoramConfigChanged(
+                                name.substring(HISTOGRAM_CFG_PREFIX.length() + 1), (long[]) newVal));
+                    } catch (IgniteCheckedException e) {
+                        throw new IgniteException(e);
+                    }
+                }
+
+                /** {@inheritDoc} */
+                @Override public void onReadyForWrite(DistributedMetaStorage metastorage) {
+                    metaLock.writeLock().lock();
```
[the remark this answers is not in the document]
nizhikov: Here, we don't need it.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
nizhikov: I want to avoid the following scenario:

0. Metastore not available, yet.
1. T1 creates a metric on node start
2. T1 read config from metastore and got null.
3. T2 execute `onReadyForRead`
4. T2 reads try to configure metric, but it doesn't exists, yet(t1 not finished `addMetric` method execution).
5. T1 finishes.

We metric that not configured with the metastore configuration data.

I fixed it for now, please, see my last commit.

modules/core/src/main/java/org/apache/ignite/mxbean/IgniteMXBean.java
nizhikov: `MetricsMXBean` introduced.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
nizhikov: I moved metrics management methods to the `MetricsMxBean`

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
nizhikov: I think we should keep the ability to configure hitrate metric.
Let's take a look at the `WalLoggingRate`.
The user may want to change that interval to measure the rate more clear or for a longer period of time: 

```
        walLoggingRate = mreg.hitRateMetric("WalLoggingRate",
            "Average number of WAL records per second written during the last time interval.",
            rateTimeInterval,
            subInts);
```

nizhikov:
@agura Seems, I resolve all your comments.
Please, review.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
agura: Proper way to measure the rate for a longer period time is using external tools that can aggregate metric values for any period. It is very simple arithmetic operation. So there is no need in additional functionality in Ignite. Moreover, changing of time interval could lead to complication of calculation of valid hit rate value for long period of time.

nizhikov: We have this configuration for now.
And it's not deprecated or something.
All I do - implementing a universal approach for hitrate configuration.

I think we should keep this.
It doesn't complicate final solution in any way.

```
    /**
     * Sets time interval for {@link DataRegionMetrics#getAllocationRate()}
     * and {@link DataRegionMetrics#getEvictionRate()} monitoring purposes.
     * <p>
     * For instance, after setting the interval to 60 seconds,
     * subsequent calls to {@link DataRegionMetrics#getAllocationRate()}
     * will return average allocation rate (pages per second) for the last minute.
     *
     * @param metricsRateTimeInterval Time interval used for allocation and eviction rates calculations.
     */
    public DataRegionConfiguration setMetricsRateTimeInterval(long metricsRateTimeInterval) {
        this.metricsRateTimeInterval = metricsRateTimeInterval;

        return this;
    }
```

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
agura: Ok. At least now.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
agura: Because you pass metric type as `type` variable.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricsMxBeanImpl.java
```diff
@@ -0,0 +1,73 @@
...
+            mreg.reset();
+        else if (log.isInfoEnabled())
+            log.info("\"" + registry + "\" not found.");
+    }
+
+    /** {@inheritDoc} */
+    @Override public void configureHitRateMetric(String name, long rateTimeInterval) {
+        try {
+            mmgr.configureHitRate(name, rateTimeInterval);
+        } catch (IgniteCheckedException e) {
```
agura: `catch` should be on he next line.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/MetricsMxBeanImpl.java
```diff
@@ -0,0 +1,73 @@
...
+            throw new IgniteException(e);
+        }
+    }
+
+    /** {@inheritDoc} */
+    @Override public void configureHistogramMetric(String name, long[] bounds) {
+        try {
+            mmgr.configureHistogram(name, bounds);
+        } catch (IgniteCheckedException e) {
```
agura: `catch` should be on he next line.

nizhikov:
@agura

I've done following changes according to your comments:

* Metrics management methods moved to new `MetricMXBean`
* configuration remove on metric registry remove
* tests to check configuration remove added.

Please, review.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -300,6 +392,136 @@ public void remove(String regName) {
 
         if (mreg != null)
             notifyListeners(mreg, metricRegRemoveLsnrs, log);
+
+        metaLock.writeLock().lock();
+
+        try {
+            for (Metric m : mreg) {
+                if (m instanceof HitRateMetric)
+                    metastorage.remove(metricName(HITRATE_CFG_PREFIX, m.name()));
```
ibessonov: I suggest adding "removeAsync" to the API and wait only for the last future here, this will speed up this method.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -300,6 +392,136 @@ public void remove(String regName) {
 
         if (mreg != null)
             notifyListeners(mreg, metricRegRemoveLsnrs, log);
+
+        metaLock.writeLock().lock();
```
ibessonov: I don't quite get the intention behind the lock. What is it for?
DistributedMetaStorage is already internally synchronized. Can you please explain? Thank you.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -300,6 +392,136 @@ public void remove(String regName) {
...
+        m.reset(rateTimeInterval);
+    }
+
+    /**
+     * Change {@link HistogramMetric} instance configuration.
+     *
+     * @param name Metric name.
+     * @param bounds New bounds.
+     */
+    private void onHistgoramConfigChanged(String name, @Nullable long[] bounds) {
```
ibessonov: Typo in the name

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -300,6 +392,136 @@ public void remove(String regName) {
...
+
+    /**
+     * @param name Metric name.
+     * @param type Metric type.
+     * @return Metric.
+     */
+    private <T extends Metric> T find(String name, Class<T> type) {
+        A.notNull(name, "name");
+
+        String[] splited = fromFullName(name);
```
ibessonov: It's either "splitted" or "split", I would rename it.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/impl/MetricUtils.java
```diff
@@ -73,6 +73,19 @@ public static String metricName(String... names) {
         return String.join(SEPARATOR, names);
     }
 
+    /**
+     * Splits full metric name to registry name and metric name.
+     *
+     * @param name Full metric name.
+     * @return Array consist of registry name and metric name.
+     */
+    public static String[] fromFullName(String name) {
```
ibessonov: Why not "IgniteBiTuple<String, String>"? Array return type is confusing.
I'm not necessarily asking for the change here, just curious.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
nizhikov: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
nizhikov: Renamed to "splitted".

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/impl/MetricUtils.java
nizhikov: Refactored to `IgniteBiTuple`.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
nizhikov: I've added `removeAsync` as you suggested.
But, I think, we should wait for all remove operations completion before exiting the method.
I think it should be implemented as follows:

```
            GridCompoundFuture opsFut = new GridCompoundFuture<>();

            for (Metric m : mreg) {
                if (m instanceof HitRateMetric)
                    opsFut.add(metastorage.removeAsync(metricName(HITRATE_CFG_PREFIX, m.name())));
                else if (m instanceof HistogramMetric)
                    opsFut.add(metastorage.removeAsync(metricName(HISTOGRAM_CFG_PREFIX, m.name())));
            }

            opsFut.markInitialized();
            opsFut.get();
```

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
nizhikov: It doesn't necessary for now, indeed.
Removed.

nizhikov:
@ibessonov Thanks, for the review. I resolved all your comments.
Please, take a look.

PR #6707, opened 2019-07-19
IGNITE-11410

dgarus:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
```diff
@@ -33,8 +34,8 @@
 public class CacheOperationContext implements Serializable {
     /** */
     //TODO IGNITE-8801 remove this and set default as `false`.
-    public static final boolean DFLT_ALLOW_ATOMIC_OPS_IN_TX =
-        IgniteSystemProperties.getBoolean(IGNITE_ALLOW_ATOMIC_OPS_IN_TX, true);
+    public static final boolean DFLT_ALLOW_ATOMIC_OPS_IN_TX = SecurityUtils.doPrivileged(
```
alamar: Can we just move this into a different class where it is guaranteed to execute outside of sandbox? It's static final anyway.

dgarus: Yes, I think we can. 
But IGNITE-8801 is out of the current issue scope.

alamar: I think we should either read one system property in sandbox, or none of them, but not move a single property to sandbox. This change is unrelated to the feature you are working on, can affect other code.

dgarus: In that place, we don't put property reading to the sandbox.
We allow reading the property with permissions that granted to ProtecionDoman of CacheOperationContext class.
This change is important and related to the feature, IgniteOperationsInsideSandboxTest shows it.

alamar: I think the code can, and should, be rewritten to avoid this.

PR #6734, opened 2019-07-30
IGNITE-9410 Thin client transactions support

alex-plekhanov:


modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientConnectionContext.java
```diff
@@ -85,6 +99,15 @@
     /** Cursor counter. */
     private final AtomicLong curCnt = new AtomicLong();
 
+    /** Tx id. */
+    private final AtomicInteger txIdSeq = new AtomicInteger();
```
ptupitsyn: Let's make it `AtomicLong` as above

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1199,6 +1199,9 @@
      */
     public static final String IGNITE_TX_OWNER_DUMP_REQUESTS_ALLOWED = "IGNITE_TX_OWNER_DUMP_REQUESTS_ALLOWED";
 
+    /** Maximum count of active transactions per connection for thin clients. */
+    public static final String IGNITE_THIN_MAX_ACTIVE_TX_PER_CONNECTION = "IGNITE_THIN_MAX_ACTIVE_TX_PER_CONNECTION";
```
ptupitsyn: I think this should go to `ClientConnectorConfiguration` instead where we already have things like `maxOpenCursorsPerConn`.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientConnectionContext.java
```diff
@@ -216,4 +241,61 @@ public ClientAffinityTopologyVersion checkAffinityTopologyVersion() {
...
+     *
+     * @param txCtx Tx context.
+     */
+    public void addTxContext(ClientTxContext txCtx) {
+        if (txsCnt.incrementAndGet() > ACTIVE_TX_LIMIT) {
+            txsCnt.decrementAndGet();
+
+            throw new IgniteClientException(ClientStatus.TX_LIMIT_EXCEEDED, "Active transactions per connection limit " +
+                "(" + ACTIVE_TX_LIMIT + ") exceeded. To start a new transaction you need to wait for some of currently " +
+                "active transactions complete. To change the limit set up " + IGNITE_THIN_MAX_ACTIVE_TX_PER_CONNECTION +
```
ptupitsyn: See above, let's not use system properties for this.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientRequestHandler.java
```diff
@@ -53,11 +60,40 @@
...
         this.authCtx = authCtx;
         this.ver = ver;
+        log = ctx.kernalContext().log(getClass());
     }
 
     /** {@inheritDoc} */
     @Override public ClientListenerResponse handle(ClientListenerRequest req) {
         try {
+            if (req instanceof ClientTxAwareRequest) {
+                if (((ClientTxAwareRequest)req).isTransactional()) {
```
ptupitsyn: Please introduce a variable to reduce casts.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/tx/ClientTxContext.java
```diff
@@ -0,0 +1,101 @@
...
+
+    /** Lock. */
+    private final Lock lock = new ReentrantLock();
+
+    /**
+     * Constructor.
+     */
+    public ClientTxContext(int txId, GridNearTxLocal tx) {
+        this.txId = txId;
+        this.tx = tx;
```
ptupitsyn: arg checks?

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/tx/ClientTxContext.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Aquire context to work with transaction in the current thread.
+     */
+    @SuppressWarnings("LockAcquiredButNotSafelyReleased")
+    public void aquire() throws IgniteCheckedException {
```
ptupitsyn: `aquire` -> `acquire`

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/tx/ClientTxEndRequest.java
```diff
@@ -0,0 +1,86 @@
...
+            throw new IgniteClientException(ClientStatus.TX_NOT_FOUND, "Transaction with id " + txId + " not found.");
+
+        try (GridNearTxLocal tx = txCtx.tx()) {
+            if (committed)
+                tx.commit();
+            else
+                tx.rollback();
+        }
+        catch (IgniteCheckedException e) {
+            throw new IgniteClientException(ClientStatus.FAILED, e.getMessage(), e);
```
ptupitsyn: Do we need this catch block at all? I think exceptions are handled up the stack anyway.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientConnectionContext.java
alex-plekhanov: But `txId` is int by the protocol, why do we need `AtomicLong` here?

ptupitsyn: My bad, you are right.

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
alex-plekhanov: I was thinking about that, but `maxActiveTxPerConnection` is only reliable to thin clients, while this `ClientConnectorConfiguration` is used to configure JDBC and ODBC connections as well. Property `maxOpenCursorsPerConn` is reliable to all connection types, but property `maxActiveTxPerConnection` is useless for JDBC and ODBC and may be confusing.

alex-plekhanov: I think we better should discuss it on dev-list. I will start a conversation.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/tx/ClientTxEndRequest.java
alex-plekhanov: The `process` method shouldn't throw any checked exception, so this block is required. But up the stack, in `ClientRequstHandler` class it's really don't needed to check `IgniteClientException` as a cause of `IgniteCheckedException` since `IgniteCheckedException` can only be thrown by `acquire` method, which doesn't use `IgniteClientException` at all. I've changed the `ClientRequestHandler#handle` method.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientTransactions.java
```diff
@@ -0,0 +1,216 @@
...
+    private String lb;
+
+    /** Channel. */
+    private final ReliableChannel ch;
+
+    /** Marshaller. */
+    private final ClientBinaryMarshaller marsh;
+
+    /** Current thread transaction. */
+    private final ThreadLocal<TcpClientTransaction> tx = new ThreadLocal<>();
```
pavlukhin: It looks like we have a possible memory leak here. `TcpClientTransaction` has a reference to `TcpClientTransactions` which has a reference to `ThreadLocal<TcpClientTransaction> tx`. If some `TcpClientTransactions` are not in use anymore but left a value in thread-local that value and `TcpClientTransactions` instance will never be garbage-collected. Making `TcpClientTransaction` class static should protect from it.

alex-plekhanov: We can't make `TcpClientTransaction` static, because we need `TcpClientTransactions` context here. 
I don't quite understand how we can get possible leak here. If nobody holds the reference to `TcpClientTransactions` instance and nobody holds the reference to `TcpClientTransaction` instances, which were created by this `TcpClientTransactions`, the only references left outside of garbage objects it's a `WeakReference` to `ThreadLocal` from threads which have used this `TcpClientTransactions` instance. These weak references don't prevent to collect garbage objects. 
Did I miss something?

alex-plekhanov: Sorry, I got it. There is a weak reference only to `ThreadLocal` instance, but regular reference to `TcpClientTransaction` from the thread-local table, so memory leak is possible.
I will try to fix it tomorrow.

alex-plekhanov: I've fixed the implementation. Also, I've found and fixed server-side transactions leak (transaction has left in active set after an attempt to end it when it's impossible to resume the transaction).

pavlukhin: 👍

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cache/ClientCachePutAllRequest.java
```diff
@@ -17,17 +17,17 @@
...
 package org.apache.ignite.internal.processors.platform.client.cache;
 
+import java.util.LinkedHashMap;
+import java.util.Map;
 import org.apache.ignite.internal.binary.BinaryRawReaderEx;
 import org.apache.ignite.internal.processors.platform.client.ClientConnectionContext;
 import org.apache.ignite.internal.processors.platform.client.ClientResponse;
-
-import java.util.LinkedHashMap;
-import java.util.Map;
```
isapego: Here and in other places, imports should not be re-ordered. See this page for details: https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines#CodingGuidelines-PackageImporting

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientConnectionContext.java
isapego: By the way, why do we use `int` and not `long` in protocol?

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cache/ClientCachePutAllRequest.java
alex-plekhanov: It was wrong formated before this patch: `Import should be ordered alphabetically, but packages starting with 'java.' should be declared prior to others.` 
Now it exactly matches coding guidelines.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientConnectionContext.java
alex-plekhanov: I think `int` is more than enough here. Transaction ids can be reused and It's practically impossible that one transaction is still active and after 4*10^9 transactions started by the same connection new one with the same id started.

PR #6738, opened 2019-07-31
IGNITE-12027 NPE on try to read the MinimumNumberOfPartitionCopies metric.

NSAmelchev:
Fixed several bugs on metrics read.


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupMetricsImpl.java
```diff
@@ -126,13 +133,22 @@ public CacheGroupMetricsImpl(CacheGroupContext ctx) {
             this::getTotalAllocatedSize,
             "Total size of memory allocated for group, in bytes.");
 
-        mreg.register("StorageSize",
-            this::getStorageSize,
-            "Storage space allocated for group, in bytes.");
+        if (ctx.shared().database() instanceof GridCacheDatabaseSharedManager) {
```
nizhikov: Let's create `isPDSEnabled` method instead of direct usage `instanceof` operator

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupMetricsImpl.java
```diff
@@ -126,13 +133,22 @@ public CacheGroupMetricsImpl(CacheGroupContext ctx) {
...
-            "Storage space allocated for group adjusted for possible sparsity, in bytes.");
+            mreg.register("SparseStorageSize",
+                () -> database().forGroupPageStores(ctx, PageStore::getSparseSize),
+                "Storage space allocated for group adjusted for possible sparsity, in bytes.");
+
+            storageSize = mreg.findMetric("StorageSize");
+            sparseStorageSize = mreg.findMetric("SparseStorageSize");
+        }
+        else {
+            storageSize = new AtomicLongMetric("NO_OP", null);
```
nizhikov: Why do we need this `NO_OP` metrics?

NSAmelchev: @nizhikov This metric need to provide zero value when PDS disabled (we do not register those metrics in monitoring but it should be available from java API). For example, it can be read from deprecated CacheGroupMetricsMXBeanImpl.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupMetricsImpl.java
NSAmelchev: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupMetricsImpl.java
nizhikov: I think it can be reworked as `storageSize == null ? 0 : storageSize.longValue()`.
Let's do it?

NSAmelchev: @nizhikov Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupMetricsImpl.java
```diff
@@ -59,6 +60,12 @@
     /** */
     private final LongAdderMetric groupPageAllocationTracker;
 
+    /** */
+    private volatile LongMetric storageSize;
```
nizhikov: Why this variable is volatile?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupMetricsImpl.java
```diff
@@ -59,6 +60,12 @@
...
+
+    /** */
+    private volatile LongMetric sparseStorageSize;
```
nizhikov: Why this variable is volatile?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -428,6 +428,9 @@ protected DataRegionMetricsProvider dataRegionMetricsProvider(final DataRegionCo
             private CacheFreeList freeList;
 
             private CacheFreeList getFreeList() {
+                if (freeListMap == null)
```
nizhikov: This change will break something.

PR #6748, opened 2019-08-05
IGNITE-9562

EdShangGG:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ClusterCachesInfo.java
```diff
@@ -1788,6 +1787,53 @@ else if (joiningNodeData instanceof CacheJoinNodeDiscoveryData)
...
+
+                        if (conflictErr != null) {
+                            U.warn(log, "Ignore cache received from joining node. " + conflictErr);
+
+                            continue;
+                        }
+
+                        long flags = cacheInfo.getFlags();
+
+                        if (flags == 1L) {
```
dspavlov: Magic number

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheJoinNodeDiscoveryData.java
```diff
@@ -162,6 +162,13 @@ public boolean isStaticallyConfigured() {
             return staticallyConfigured;
         }
 
+        /**
+         * @return Long which bits represent some flags.
```
dspavlov: How to understand these flags meaning?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridLocalConfigManager.java
```diff
@@ -0,0 +1,333 @@
...
+    ) {
+        if (!caches.containsKey(cacheName)) {
+            if (!cacheType.userCache())
+                stopSeq.addLast(cacheName);
+            else
+                stopSeq.addFirst(cacheName);
+        }
+
+        caches.put(cacheName, new CacheJoinNodeDiscoveryData.CacheInfo(cacheData, cacheType, cacheData.sql(),
+            persistedBefore ? 1 : 0, isStaticallyConfigured));
```
dspavlov: @glukos please avoid accepting magic code into the project. This constant meaning should be documented in code, e.g. introduced public static final fields with the particular bit meaning. Keeping it as-is become error-prone code

PR #6760, opened 2019-08-08
IGNITE-12051: update javadoc for IgniteKernal class

Mmuzaf:


modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -293,52 +294,79 @@
...
- * See <a href="http://en.wikipedia.org/wiki/Kernal">http://en.wikipedia.org/wiki/Kernal</a> for information on the
- * misspelling.
+ * This class represents an implementation of the extended main Ignite API (called {@link IgniteEx}). It also controls
+ * the Ignite life cycle and allows listening for Ignite events.
+ * <p>
+ * Please, refer to the wiki <a href="http://en.wikipedia.org/wiki/Kernal">http://en.wikipedia.org/wiki/Kernal</a>
+ * for the information on the misspelling.
+ * <p>
+ * <h3>Start point</h3>
+ * The main entry point for all the Ignite instances creation is the <tt>org.apache.ignite.internal.IgniteKernal#start()</tt>
```
pavlukhin: Should not we use a link here?

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -293,52 +294,79 @@
...
+ * method. It starts internal Ignites components (see {@link GridComponent}), for instance:
+ * <ul>
+ * <li>{@link GridManager} - a layer of indirection between kernal and SPI modules.</li>
+ * <li>{@link GridProcessor} - an objects responsible for particular internal process implementation.</li>
+ * <li>{@link IgnitePlugin} - an Ignite addition of user-provided functionality.</li>
+ * </ul>
+ * The <tt>start()</tt> method also perfoms additional validation of the provided {@link IgniteConfiguration} and
```
pavlukhin: Why not `@link` or `@code`?

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -293,52 +294,79 @@
...
+ * prints some suggestions such as:
+ * <ul>
+ * <li>Ignites configuration optimizations (e.g. disabling {@link EventType} events).</li>
+ * <li>{@link JvmConfigurationSuggestions} optimizations</li>
```
pavlukhin: `.` in the end.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -293,52 +294,79 @@
...
+ * <li>{@link OsConfigurationSuggestions} optimizations</li>
```
pavlukhin: `.` in the end.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
pavlukhin: I suppose that shutdown procedure via `stop` might be mentioned here as well.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -293,52 +294,79 @@
...
+    /** Currently used instance of JVM pause detector thread. See {@link LongJVMPauseDetector} for details. */
     private LongJVMPauseDetector longJVMPauseDetector;
 
-    /** */
+    /** The main kernal context which holds all the {@link GridComponent}s. */
     @GridToStringExclude
     private GridKernalContextImpl ctx;
 
-    /** Helper that registers MBeans */
+    /** Helper which is register and unregister MBeans. */
```
pavlukhin: `which registers and unregisters`?

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -293,52 +294,79 @@
...
+    /** Ignite runtime configuration instance. */
     private IgniteConfiguration cfg;
 
-    /** */
+    /** Ignite logger instance which enriches log messages with the node instance name and the node id. */
     @GridToStringExclude
     private GridLoggerProxy log;
 
-    /** */
+    /** Instance name of Ignite node. */
```
pavlukhin: `Name of Ignite node`? Does not _instance_ and _node_ have the same meaning here?

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -347,35 +375,44 @@
     /** Spring context, potentially {@code null}. */
     private GridSpringResourceContext rsrcCtx;
 
-    /** */
+    /**
+     * The instance of scheduled thread pool starvation checker. {@code null} if starvation checks have been
+     * disabled by the value {@link IgniteSystemProperties#IGNITE_STARVATION_CHECK_INTERVAL} system property.
```
pavlukhin: -> by the value *of* ...

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -347,35 +375,44 @@
...
+     */
     @GridToStringExclude
     private GridTimeoutProcessor.CancelableTask starveTask;
 
-    /** */
+    /**
+     * The instance of scheduled metrics logger. {@code null} means that the metrics logges have been disabled
```
pavlukhin: logges -> logging

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -347,35 +375,44 @@
...
-    /** */
+    /**
+     * The instance of scheduled long operation checker. {@code null} means that the operations checker is disabled
+     * by the value of {@link IgniteSystemProperties#IGNITE_LONG_OPERATIONS_DUMP_TIMEOUT} system property.
+     */
     @GridToStringExclude
     private GridTimeoutProcessor.CancelableTask longOpDumpTask;
 
-    /** Indicate error on grid stop. */
+    /** {@code true} if an error occur at Ignite instance stop. */
```
pavlukhin: occur -> occurs

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -347,35 +375,44 @@
...
     @GridToStringExclude
     private final AtomicReference<GridKernalGateway> gw = new AtomicReference<>();
 
-    /** Stop guard. */
+    /** Flag indicates that the ignite instance is scheduled to be stopped. */
     @GridToStringExclude
     private final AtomicBoolean stopGuard = new AtomicBoolean();
 
-    /** */
+    /** The state object is used to handle if reconnection occurred. See {@link IgniteKernal#onReconnected(boolean)}. */
```
pavlukhin: Suggestions to rearrange:
* The state object is used _to handle reconnection_.
* The state object is used _when reconnection occurs_.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -4110,7 +4157,7 @@ private void checkNearCacheStarted(IgniteCacheProxy<?, ?> cache) throws IgniteCh
     }
 
     /**
-     * <tt>ctx.gateway().readLock()</tt>
+     * The <tt>ctx.gateway().readLock()</tt> is used underneath.
```
pavlukhin: Can we use `@code` instead of `<tt>`? Here and in a couple of places.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -293,52 +294,79 @@
...
     @GridToStringExclude
     private GridKernalContextImpl ctx;
 
-    /** Helper that registers MBeans */
+    /** Helper which is register and unregister MBeans. */
     @GridToStringExclude
     private IgniteMBeansManager mBeansMgr;
 
-    /** Configuration. */
+    /** Ignite runtime configuration instance. */
```
pavlukhin: A qulifier _runtime_ for an Ignite configuration sounds a bit misleading for me. _Runtime_ is usually related to something that can be changed in a _running time_, i.e. dynamically. But our configuration is static. I suggest to simply use a term _Ignite configuration_. Here and in a couple of other places in this class.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -293,52 +294,79 @@
 import static org.apache.ignite.lifecycle.LifecycleEventType.BEFORE_NODE_START;
 
 /**
- * Ignite kernal.
- * <p/>
- * See <a href="http://en.wikipedia.org/wiki/Kernal">http://en.wikipedia.org/wiki/Kernal</a> for information on the
- * misspelling.
+ * This class represents an implementation of the extended main Ignite API (called {@link IgniteEx}). It also controls
```
pavlukhin: I think it worth to have a link to `Ignite` interface here as the current class is the real implementation of it. `IgniteEx` could be mentioned as well with a note that it adds several methods to `Ignite` for internal purposes. What do you think?

pavlukhin: Should we mention printing metrics, detected long operations and a starvation?

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: Agree, fixed.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: I've added `Endpoint` block to the end of class javadoc. Please, check.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: Agree, fixed.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: Fixed all mentions.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: Agree, I've fixed it.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: Sound good. I've updated the class javadoc. Can you check?

Mmuzaf: Added.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: Agree, fixed.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -293,52 +296,95 @@
...
+ * thread pools state for starvation, detects long JVM pauses and prints out the local node metrics.
+ *
+ * <p>
+ * Please, refer to the wiki <a href="http://en.wikipedia.org/wiki/Kernal">http://en.wikipedia.org/wiki/Kernal</a>
+ * for the information on the misspelling.
+ * <p>
+ * <h3>Startpoint</h3>
+ * The main entry point for all the Ignite instances creation is the method:
+ * <p>
+ * {@link #start(IgniteConfiguration, ExecutorService, ExecutorService, ExecutorService, ExecutorService,StripedExecutor, ExecutorService, ExecutorService, ExecutorService, StripedExecutor, ExecutorService, ExecutorService, ExecutorService, IgniteStripedThreadPoolExecutor, ExecutorService, ExecutorService, Map, GridAbsClosure, WorkersRegistry, Thread.UncaughtExceptionHandler, TimeBag)}
```
pavlukhin: It is possible to add a link name to shorten a rendered representation like `{@link #start(...) start}` (last word _start_ is a link name).

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -293,52 +296,95 @@
...
+ * <p>
+ * It starts internal Ignites components (see {@link GridComponent}), for instance:
+ * <ul>
+ * <li>{@link GridManager} - a layer of indirection between kernal and SPI modules.</li>
+ * <li>{@link GridProcessor} - an objects responsible for particular internal process implementation.</li>
+ * <li>{@link IgnitePlugin} - an Ignite addition of user-provided functionality.</li>
+ * </ul>
+ * The {@code start} method also perfoms additional validation of the provided {@link IgniteConfiguration} and
```
pavlukhin: perfoms -> performs

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -293,52 +296,95 @@
...
- * misspelling.
+ * This class represents an implementation of the main Ignite API {@link Ignite} which is expanded by additional
+ * methods of {@link IgniteEx} for the internal Ignite needs. It also controls the Ignite life cycle, checks
+ * thread pools state for starvation, detects long JVM pauses and prints out the local node metrics.
+ *
+ * <p>
+ * Please, refer to the wiki <a href="http://en.wikipedia.org/wiki/Kernal">http://en.wikipedia.org/wiki/Kernal</a>
+ * for the information on the misspelling.
+ * <p>
+ * <h3>Startpoint</h3>
```
pavlukhin: Startpoint -> Starting (see a comment for _Endpoint_ below).

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -293,52 +296,95 @@
...
+ * </ul>
+ * The {@code start} method also perfoms additional validation of the provided {@link IgniteConfiguration} and
+ * prints some suggestions such as:
+ * <ul>
+ * <li>Ignites configuration optimizations (e.g. disabling {@link EventType} events).</li>
+ * <li>{@link JvmConfigurationSuggestions} optimizations.</li>
+ * <li>{@link OsConfigurationSuggestions} optimizations.</li>
+ * </ul>
+ *
+ * <h3>Endpoint</h3>
```
pavlukhin: _Endpoint_ often means some kind of an exposed interface (e.g. listening connection/port) but not shutdown/termination. I suggest to use _Stopping_ here. And _Starting_ instead of _Startpoint_ for a symmetry.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -1931,7 +1994,7 @@ private String onOff(boolean b) {
     }
 
     /**
-     * @return Whether or not REST is enabled.
+     * @return {@code true} if the rest processor is enabled, {@code false} the otherwise.
```
pavlukhin: *REST* in capital case seems more common (in a javadoc).

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -2418,7 +2481,7 @@ else if (s.contains("clojure")) {
     }
 
     /**
-     * Stops grid instance.
+     * Stops ignite instance.
```
pavlukhin: It seem that _Ignite_ with first capital letter is used elsewhere in this class docs.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: fixed.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: fixed

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: fixed

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: fixed

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: fixed

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
Mmuzaf: fixed.

PR #6783, opened 2019-08-16
ignite-12078 implement loadCache functionality in CacheJdbcBlobStore

luecu:


modules/core/src/main/java/org/apache/ignite/cache/store/jdbc/CacheJdbcBlobStore.java
```diff
@@ -206,6 +213,38 @@
...
+            log.debug("Store load cache");
+
+        Connection conn = null;
+
+        PreparedStatement stmt = null;
+
+        try {
+            conn = connection(tx);
+
+            stmt = conn.prepareStatement(loadCacheQry);
```
qzinwtf: if large resultSet will be loaded I suggest to add fetchSize parameter: 10000 for example

PR #6785, opened 2019-08-16
IGNITE-12080 Add extended logging for rebalance.

tkalkirill:


modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1246,6 +1246,13 @@
      */
     public static final String INDEX_REBUILDING_PARALLELISM = "INDEX_REBUILDING_PARALLELISM";
 
+    /** Enable write rebalnce statistics into log. Default: false */
+    public static final String IGNITE_WRITE_REBALANCE_STATISTICS = "IGNITE_WRITE_REBALANCE_STATISTICS";
```
Mmuzaf: Should we rename to `IGNITE_PRINT_REBALANCE_STATISTIC` ?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/RebalanceStatisticsUtils.java
```diff
@@ -0,0 +1,686 @@
...
+import static java.util.stream.Collectors.toMap;
+import static org.apache.ignite.IgniteSystemProperties.IGNITE_QUIET;
+import static org.apache.ignite.IgniteSystemProperties.IGNITE_WRITE_REBALANCE_PARTITION_STATISTICS;
+import static org.apache.ignite.IgniteSystemProperties.IGNITE_WRITE_REBALANCE_STATISTICS;
+import static org.apache.ignite.IgniteSystemProperties.getBoolean;
+
+/**
+ * Utility class for rebalance statistics.
+ */
+class RebalanceStatisticsUtils {
```
Mmuzaf: Why we are creating a standalone util class? 
Can it be reflected as a part of, for instance, `DiagnosticProcessor`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -351,17 +364,18 @@ Runnable addAssignments(
...
-                        try {
-                            if (f.get()) // Not cancelled.
-                                next.run(); // Starts next cache rebalancing (according to the order).
-                        }
-                        catch (IgniteCheckedException e) {
-                            if (log.isDebugEnabled())
-                                log.debug(e.getMessage());
-                        }
-                    });
+                fut.listen(f -> {
```
Mmuzaf: What the reason of removing `if (next != null)`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/RebalanceStatisticsUtils.java
```diff
@@ -0,0 +1,686 @@
...
+        ) {
+            assert nonNull(topicId);
+            assert nonNull(supplierNode);
+            assert nonNull(supplyMsg);
+
+            if (!printRebalanceStatistics)
+                return;
+
+            List<PartitionStatistics> partStats = supplyMsg.infos().entrySet().stream()
+                .map(entry -> new PartitionStatistics(entry.getKey(), entry.getValue().infos().size()))
```
Mmuzaf: Entry value can be `null` in case of historical rebalance used. Will we get `NullPointerException` here?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/RebalanceStatisticsUtils.java
```diff
@@ -0,0 +1,686 @@
...
+     * #SUCCESSFUL_REBALANCE_TEXT} into header.
+     *
+     * @param rebFutrs Participating in successful and not rebalances, require not null.
+     * @param nodeAliases For print nodeId=1 instead long string, require not null.
+     * @param joiner For write statistics, require not null.
+     * @param finish Is finish rebalance.
+     */
+    private static void writeCacheGroupsRebalanceStatistics(
+        final Map<CacheGroupContext, Collection<RebalanceFuture>> rebFutrs,
+        final Map<ClusterNode, Integer> nodeAliases,
```
Mmuzaf: Do we need `nodeAliases`? Why UUID or node name can't be used here?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1514,4 +1536,74 @@ private void sendRebalanceFinishedEvent() {
...
+     * <p/>
+     * If the whole rebalance is over, print statistics for all cache groups.
+     * The end of the rebalance is determined by the successful done all
+     * {@code RebalanceFuture}'s.
+     *
+     * @throws IgniteCheckedException when get result {@code RebalanceFuture}
+     * @see RebalanceFuture RebalanceFuture
+     */
+    private void printRebalanceStatistics() throws IgniteCheckedException {
+        if (!RebalanceStatisticsUtils.printRebalanceStatistics())
```
Mmuzaf: Is it correct to use `static` methods for such things?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1514,4 +1536,74 @@ private void sendRebalanceFinishedEvent() {
...
+            return;
+
+        RebalanceFuture currRebFut = rebalanceFut;
+        assert currRebFut.isDone() : "RebalanceFuture should be done.";
+
+        currRebFut.stat.endTime(currentTimeMillis());
```
Mmuzaf: Is it correct to set the time at the moment of this method called but not at the time then the rebalance future actually ends?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1514,4 +1536,74 @@ private void sendRebalanceFinishedEvent() {
...
+            return;
+
+        for (GridCacheContext gridCacheContext : ctx.cacheContexts()) {
+            IgniteInternalFuture<Boolean> rebalanceFuture = gridCacheContext.preloader().rebalanceFuture();
+
+            if (!rebalanceFuture.isDone() || !rebalanceFuture.get()) //Rebalance not done or not success
+                return;
+        }
+
+        List<GridDhtPartitionDemander> demanders = demanders();
```
Mmuzaf: Why single-instance `demander` which is responsible for the particular cache group should know something about the other `demanders`? What if the others are not initialized yet?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPreloader.java
```diff
@@ -610,4 +610,13 @@ public void onPartitionEvicted(GridDhtLocalPartition part, boolean updateSeq) {
     @Override public void dumpDebugInfo() {
         // No-op
     }
+
+    /**
+     * Return demander.
+     *
+     * @return Demander.
+     * */
+    public GridDhtPartitionDemander demander() {
```
Mmuzaf: Why we need to make it `public`?

PR #6796, opened 2019-08-20
IGNITE-12049 Add user attributes to thin clients

SomeFire:


modules/core/src/main/java/org/apache/ignite/internal/processors/rest/protocols/tcp/GridTcpRestNioListener.java
```diff
@@ -145,6 +149,9 @@
     /** Handler for all Redis requests. */
     private GridRedisNioListener redisLsnr;
 
+    /** Client SSL enabled. */
+    private final boolean sslEnabled;
```
andrey-kuznetsov: isSslClientAuth could be a better name.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -3386,6 +3390,30 @@ public IgniteConfiguration setAuthenticationEnabled(boolean authEnabled) {
         return this;
     }
 
+    /**
+     * Returns {@code true} if SSL certificates transmission is enabled for cluster. Otherwise returns {@code false}.
```
andrey-kuznetsov: Please add more details: it's unclear what kind of transmission is mentioned.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/protocols/tcp/GridTcpRestNioListener.java
SomeFire: I renamed field to `sslClientAuth`, because `is...` is a getter name, not field.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
SomeFire: Done.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -3386,6 +3395,33 @@ public IgniteConfiguration setAuthenticationEnabled(boolean authEnabled) {
...
+     * Default value is false; transmission is disabled.
+     *
+     * If enabled, then node attributes of node join messages will contain SSL certificates,
+     * with which new node trying to connect to the cluster.
+     *
+     * @return {@code true} if SSL certificates transmission is enabled for cluster. Otherwise returns {@code false}.
+     */
+    public boolean isSslCertsTransmissionEnabled() {
```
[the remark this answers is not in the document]
SomeFire: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/util/nio/ssl/GridNioSslFilter.java
```diff
@@ -330,6 +330,11 @@ public ByteBuffer encrypt(GridNioSession ses, ByteBuffer input) throws SSLExcept
 
         GridNioSslHandler hnd = sslHandler(ses);
 
+        GridSslMeta meta = ses.meta(SSL_META.ordinal());
+
+        if (meta.sslEngine() == null)
```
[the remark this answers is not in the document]
SomeFire: Fixed.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -6775,6 +6776,16 @@ else if (msg instanceof TcpDiscoveryJoinRequestMessage) {
                             TcpDiscoveryJoinRequestMessage req = (TcpDiscoveryJoinRequestMessage)msg;
 
                             if (!req.responded()) {
+                                if (sock instanceof SSLSocket &&
+                                    spi.ignite().configuration().isSslCertsTransmissionEnabled()) {
+                                    HashMap<String, Object> attrs = new HashMap<>(req.node().getAttributes());
+
+                                    attrs.put(ATTR_SECURITY_CERTIFICATES, ((SSLSocket) sock).getSession()
+                                        .getPeerCertificateChain());
```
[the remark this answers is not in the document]
SomeFire: Default implementation (`X509V1CertImpl`) implements `Serializable`. We can add check here, and in case certificates aren't serializable - don't put them.

modules/core/src/test/java/org/apache/ignite/internal/processors/security/SslAbstractNodeAttributesFactory.java
```diff
@@ -0,0 +1,73 @@
...
+
+            try (InputStream ksInputStream = new FileInputStream(file)) {
+                clientKeyStore.load(ksInputStream, pass);
+
+                Certificate[] certs = clientKeyStore.getCertificateChain(alias);
+                JdkMarshaller marshaller = MarshallerUtils.jdkMarshaller(null);
+
+                HashMap<String, Object> map = new HashMap<>();
+
+                map.put(ATTR_SECURITY_CERTIFICATES, U.marshal(marshaller, certs));
```
ptupitsyn: I understand that this is test code, but this is the only demonstration of the new mechanism, so I'll raise my concern here:

The approach is Java-specific. Imagine trying to authenticate like this from Python or Node.js - how would you do it?

SomeFire: We can send a byte array and create a certificate from the bytes. Even Java KeyStore can be created from the byte array.

ptupitsyn: Byte array makes sense

PR #6811, opened 2019-08-26
GG-23282: Several open iterator by lazy SQL query can lead to assertion error

ygerzhedovich:
GG-23282: Several open iterator by lazy SQL query can lead to assertion error


pavlukhin:
Closing as it is not related to any Ignite ticket.

PR #6817, opened 2019-08-27
IGNITE-12112

denis-chudov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/BPlusTree.java
```diff
@@ -5920,12 +5923,39 @@ protected IoStatisticsHolder statisticsHolder() {
...
+    }
+
+    /**
+     * Creates a formatted message even if "toString" of optioanl parameters failed.
+     *
+     * @param msg Detailed error message.
+     * @param rows Optional parameters.
+     * @return New instance of {@link CorruptedTreeException}.
+     */
+    private String formatMsg(String msg, Object... rows) {
```
AMashenkov: Looks like this method is never used.

denis-chudov: Removed it.

PR #6819, opened 2019-08-28
IGNITE-12116 Support array as key for caches.

ingvard:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheDefaultAffinityKeyMapper.java
```diff
@@ -77,6 +78,9 @@
     @Override public Object affinityKey(Object key) {
         GridArgumentCheck.notNull(key, "key");
 
+        if(key.getClass().isArray())
+            return IgniteUtils.hashCode(key);
```
pavlukhin: @ingvard It is puzzling for me why an _affinity key_ for an array is calculated as a _hash_ for an array content? Could you please elaborate?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -5359,6 +5359,42 @@ public static void writeMap(ObjectOutput out, Map<?, ?> map) throws IOException
...
+                return Arrays.hashCode((double[])obj);
+            if (obj instanceof char[])
+                return Arrays.hashCode((char[])obj);
+            if (obj instanceof boolean[])
+                return Arrays.hashCode((boolean[])obj);
+
+            int result = 1;
+
+            for (Object element : (Object[])obj)
+                result = 31 * result + hashCode(element);
```
xtern: Is NPE possible here if one of the elements is null (for example for an array of strings)?
I know that Ignite doesn't allow nulls as key and/or value, but I think "arrays-as-key" restrictions should be documented somehow.

xtern: When we put null-key we get NPE with th problem description  
"java.lang.NullPointerException: Ouch! Argument cannot be null: key"

But when we put "new String[] {"c", null, "a"}" we get just NPE, and  I think we should at least explain the problem in the same way if this is the expected behavior.

PR #6823, opened 2019-08-29
IGNITE-1666 Added test on serial node shutdown

Diana-Iakovleva:


dspavlov:
IGNITE-1666 | Resolved
-- | --

PR #6837, opened 2019-09-03
IGNITE-10223: affinity list methods are added, tests are changed

Diana-Iakovleva:
Deprecated methods:
Affinity::mapPartitionToPrimaryAndBackups
Affinity::mapKeyToPrimaryAndBackups
New methods are introduced:
Affinity::mapPartitionToPrimaryAndBackupsList
Affinity::mapKeyToPrimaryAndBackupsList

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/cache/affinity/Affinity.java
+++ b/modules/core/src/main/java/org/apache/ignite/cache/affinity/Affinity.java
@@ -17,11 +17,13 @@
 
 package org.apache.ignite.cache.affinity;
 
-import java.util.Collection;
-import java.util.Map;
 import org.apache.ignite.IgniteException;
 import org.apache.ignite.cluster.ClusterNode;
 import org.jetbrains.annotations.Nullable;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
 
 /**
  * Provides affinity information to detect which node is primary and which nodes are
@@ -180,6 +182,9 @@ public interface Affinity<K> {
     @Nullable public ClusterNode mapKeyToNode(K key);
 
     /**
+     *
+     * @deprecated Use Affinity#mapKeyToPrimaryAndBackups(java.lang.Object) instead
+     *
      * Gets primary and backup nodes for the key. Note that primary node is always
      * first in the returned collection.
      *
@@ -188,7 +193,19 @@ public interface Affinity<K> {
      *      always first.
      * @throws IgniteException If there are no alive nodes for this cache.
      */
+    @Deprecated
     public Collection<ClusterNode> mapKeyToPrimaryAndBackups(K key);
+
+    /**
+     * Gets primary and backup nodes for the key. Primary node is always
+     * first in the returned list.
+     *
+     * @param key Key to get affinity nodes for.
+     * @return List of primary and backup nodes for the key with primary node
+     *      always first.
+     * @throws IgniteException If there are no alive nodes for this cache.
+     */
+    public List<ClusterNode> mapKeyToPrimaryAndBackupsList(K key);
 
     /**
      * Gets primary node for the given partition.
@@ -215,6 +232,7 @@ public interface Affinity<K> {
     public Map<Integer, ClusterNode> mapPartitionsToNodes(Collection<Integer> parts);
 
     /**
+     * @deprecated Use Affinity#mapPartitionToPrimaryAndBackupsList(int) instead
      * Gets primary and backup nodes for partition. Note that primary node is always
      * first in the returned collection.
      *
@@ -223,5 +241,17 @@ public interface Affinity<K> {
      *      always first.
      * @throws IgniteException If there are no alive nodes for this cache.
      */
+    @Deprecated
     public Collection<ClusterNode> mapPartitionToPrimaryAndBackups(int part);
+
+    /**
+     * Gets primary and backup nodes for partition. Note that primary node is always
+     * first in the returned list.
+     *
+     * @param part Partition to get affinity nodes for.
+     * @return List of primary and backup nodes for partition with primary node
+     *      always first.
+     * @throws IgniteException If there are no alive nodes for this cache.
+     */
+    public List<ClusterNode> mapPartitionToPrimaryAndBackupsList(int part);
 }
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/affinity/GridAffinityProcessor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/affinity/GridAffinityProcessor.java
@@ -17,16 +17,6 @@
 
 package org.apache.ignite.internal.processors.affinity;
 
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.concurrent.ConcurrentSkipListMap;
 import org.apache.ignite.IgniteCheckedException;
 import org.apache.ignite.IgniteException;
 import org.apache.ignite.IgniteLogger;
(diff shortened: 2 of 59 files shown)


modules/core/src/main/java/org/apache/ignite/cache/affinity/Affinity.java
```diff
@@ -188,8 +192,20 @@
...
+    /**
+     * Gets primary and backup nodes for the key. Primary node is always
+     * first in the returned list.
+     *
+     * @param key Key to get affinity nodes for.
+     * @return List of primary and backup nodes for the key with primary node
+     *      always first.
+     * @throws IgniteException If there are no alive nodes for this cache.
+     */
+    public List<ClusterNode> mapKeyToPrimaryAndBackupsList(K key);
```
nizhikov: We should have default implementation of this method to preserve backward compatibility.
Something like:
```
Collection<ClusterNode> res = mapKeyToPrimaryAndBackups(key);
if (res instanceof List)
    return (List<ClusterNode>)res;
else
    return new ArrrayList(res);
```

modules/core/src/main/java/org/apache/ignite/cache/affinity/Affinity.java
```diff
@@ -215,6 +231,7 @@
     public Map<Integer, ClusterNode> mapPartitionsToNodes(Collection<Integer> parts);
 
     /**
+     * @deprecated Use Affinity#mapPartitionToPrimaryAndBackupsList(int) instead
```
nizhikov: This should be after @throws, at the end of the Javadoc.
Please, add a dot at the end of the statement.

modules/core/src/main/java/org/apache/ignite/internal/processors/affinity/GridAffinityProcessor.java
```diff
@@ -1050,7 +1050,8 @@ public CacheAffinityProxy(String cacheName) {
             }
         }
 
-        /** {@inheritDoc} */
+        /** @deprecated Use mapKeyToPrimaryAndBackupsList instead */
```
nizhikov: Please, use `/** {@inheritDoc} */` for overriding method.
The only reason to write a new JavaDoc comment if the implementation has another method contract compared to the parent one.

modules/core/src/main/java/org/apache/ignite/internal/processors/affinity/GridAffinityProcessor.java
```diff
@@ -1067,6 +1068,24 @@ public CacheAffinityProxy(String cacheName) {
             }
         }
 
+        /** {@inheritDoc} */
+        @Override
```
nizhikov: Please, write it in one line: `@Override public`

modules/core/src/main/java/org/apache/ignite/internal/processors/affinity/GridAffinityProcessor.java
```diff
@@ -1067,6 +1068,24 @@ public CacheAffinityProxy(String cacheName) {
...
+        public List<ClusterNode> mapKeyToPrimaryAndBackupsList(K key) {
+            ctx.gateway().readLock();
```
nizhikov: Any reason to copy-paste implementation from `mapKeyToPrimaryAndBackups`?

modules/core/src/main/java/org/apache/ignite/internal/processors/affinity/GridAffinityProcessor.java
```diff
@@ -1106,7 +1125,8 @@ public CacheAffinityProxy(String cacheName) {
             }
         }
 
-        /** {@inheritDoc} */
+        /** @deprecated use mapPartitionToPrimaryAndBackupsList instead */
```
nizhikov: Please, use `/** {@inheritDoc} */` for overriding method.
The only reason to write a new JavaDoc comment if the implementation has another method contract compared to the parent one.

modules/core/src/main/java/org/apache/ignite/internal/processors/affinity/GridAffinityProcessor.java
```diff
@@ -1121,6 +1141,22 @@ public CacheAffinityProxy(String cacheName) {
             }
         }
 
+        /** {@inheritDoc} */
+        @Override
```
nizhikov: Please, write it in one line: `@Override public`

modules/core/src/main/java/org/apache/ignite/internal/processors/affinity/GridAffinityProcessor.java
```diff
@@ -1121,6 +1141,22 @@ public CacheAffinityProxy(String cacheName) {
...
+        public List<ClusterNode> mapPartitionToPrimaryAndBackupsList(int part) {
```
nizhikov: Any reason to copy-paste implementation from `mapPartitionToPrimaryAndBackups`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/affinity/GridCacheAffinityImpl.java
```diff
@@ -218,20 +219,40 @@ public GridCacheAffinityImpl(GridCacheContext<K, V> cctx) {
         return res;
     }
 
-    /** {@inheritDoc} */
+    /** @deprecated Use mapKeyToPrimaryAndBackupsList(K) instead */
```
nizhikov: Please, use `/** {@inheritDoc} */` for overriding method.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/affinity/GridCacheAffinityImpl.java
```diff
@@ -218,20 +219,40 @@ public GridCacheAffinityImpl(GridCacheContext<K, V> cctx) {
...
+    @Deprecated
     @Override public Collection<ClusterNode> mapKeyToPrimaryAndBackups(K key) {
         A.notNull(key, "key");
 
         return cctx.affinity().nodesByPartition(partition(key), topologyVersion());
     }
 
     /** {@inheritDoc} */
+    @Override
+    public List<ClusterNode> mapKeyToPrimaryAndBackupsList(K key) {
```
nizhikov: Please, write it in one line: `@Override public`
Any reason to copy-paste implementation from `mapKeyToPrimaryAndBackupsList`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/affinity/GridCacheAffinityImpl.java
```diff
@@ -218,20 +219,40 @@ public GridCacheAffinityImpl(GridCacheContext<K, V> cctx) {
...
+        A.notNull(key, "key");
+
+        return cctx.affinity().nodesByPartition(partition(key), topologyVersion());
+    }
+
+    @Deprecated
```
nizhikov: Annotations should be written after  JavaDocs.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/affinity/GridCacheAffinityImpl.java
```diff
@@ -218,20 +219,40 @@ public GridCacheAffinityImpl(GridCacheContext<K, V> cctx) {
...
+     * @deprecated Use GridCacheAffinityImpl#mapPartitionToPrimaryAndBackupsList(int) instead
+     */
     @Override public Collection<ClusterNode> mapPartitionToPrimaryAndBackups(int part) {
         A.ensure(part >= 0 && part < partitions(), "part >= 0 && part < total partitions");
 
         return cctx.affinity().nodesByPartition(part, topologyVersion());
     }
 
+    /** {@inheritDoc} */
+    @Override
```
nizhikov: Please, write it in one line: `@Override public`

modules/core/src/main/java/org/apache/ignite/cache/affinity/Affinity.java
Diana-Iakovleva: Can you please explain why do we need it? Clients with old version would use old method which hadn't been changed. All interface implementations are cited in the same module as I see

nizhikov: Classes located in `org/apache/ignite/internal` package are Internal API.
`org/apache/ignite/cache/affinity/Affinity.java` package are Public API.
So, we should keep Public API backwards compatible between minor releases.
It means, we can't add new methods to the public API without default implementation, because, there are implementations of this interface out of Ignite codebase.

modules/core/src/main/java/org/apache/ignite/internal/processors/affinity/GridAffinityProcessor.java
Diana-Iakovleva: no one!

PR #6839, opened 2019-09-04
IGNITE-11815: replace GridTestUtils.retryAssert with GridTestUtils.waitForCondition

Diana-Iakovleva:


modules/core/src/test/java/org/apache/ignite/igfs/IgfsFragmentizerSelfTest.java
```diff
@@ -248,19 +247,16 @@ public void testDeleteFragmentizing() throws Exception {
...
-                        g.igfsx("igfs").configuration().getDataCacheConfiguration().getName());
+            GridCacheAdapter<Object, Object> cache = ((IgniteKernal)g).internalCache(
+                    g.igfsx("igfs").configuration().getDataCacheConfiguration().getName());
 
-                    assertTrue("Data cache is not empty [keys=" + cache.keySet() +
-                        ", node=" + g.localNode().id() + ']', cache.isEmpty());
-                }
-            }
-        });
+            boolean isCacheEmpty = GridTestUtils.waitForCondition(() -> cache.isEmpty(), 5000);
```
AMashenkov: In worst case we will wait for NODE_CNT * 5 sec,
but before this change w'd wait for 5 sec in total.

modules/core/src/test/java/org/apache/ignite/internal/managers/checkpoint/GridCheckpointManagerAbstractSelfTest.java
```diff
@@ -751,13 +711,13 @@ private boolean checkCheckpointManager(Ignite g) throws Exception {
     }
 
     /**
-     * Wrapper around {@link GridTestUtils#retryAssert(org.apache.ignite.IgniteLogger, int, long, GridAbsClosure)}.
+     * Wrapper around {@link GridTestUtils#waitForCondition(org.apache.ignite.internal.util.lang.GridAbsPredicate, long).
      * For the given closure provides count of retries, configured by {@link #retries} attribute.
      * @param assertion Closure with assertion inside.
```
AMashenkov: Method semantic has been changed.
Now "assertion" param is a closure with condition and no assertion is expected to be thrown from the closure.

AMashenkov: Let's get rid of this method either and rewrite places where it is used with waitForCondition.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridCacheDhtPreloadDelayedSelfTest.java
```diff
@@ -461,22 +462,28 @@ private final void checkMaps(final boolean strict, final GridDhtCacheAdapter<Str
...
-                            assert state != GridDhtPartitionState.EVICTED;
-                            assert !strict || state == GridDhtPartitionState.OWNING : "Invalid partition state: " + state;
-                            assert state == nodeEntry.getValue();
+                            boolean isOk =
+                                state != null
+                                && state != GridDhtPartitionState.EVICTED
+                                && (!strict || state == GridDhtPartitionState.OWNING)
+                                && state == nodeEntry.getValue();
+                            if (!isOk) {
+                                info("Entry state is invalid, rerun");
```
AMashenkov: Typo. Seems, you mean "Partition state", not an entry state.

modules/core/src/test/java/org/apache/ignite/internal/processors/igfs/IgfsStreamsSelfTest.java
```diff
@@ -282,12 +282,10 @@ public void testCreateFileFragmented() throws Exception {
...
 
-        GridTestUtils.retryAssert(log, ASSERT_RETRIES, ASSERT_RETRY_INTERVAL, new CAX() {
-            @Override public void applyx() {
-                for (int i = 0; i < NODES_CNT; i++)
-                    assertTrue(grid(i).cachex(dataCacheName).isEmpty());
-            }
-        });
+        for (int i = 0; i < NODES_CNT; i++) {
+            IgniteEx grid = grid(i);
+            assertTrue(GridTestUtils.waitForCondition(() -> grid.cachex(dataCacheName).isEmpty(), ASSERT_RETRIES*ASSERT_RETRY_INTERVAL));
```
AMashenkov: Wait time increased up to Nodes_Cnt*retries*interval

modules/core/src/test/java/org/apache/ignite/internal/processors/service/GridServiceProcessorAbstractSelfTest.java
```diff
@@ -293,20 +292,15 @@ public void testGetServicesByName() throws Exception {
...
-                        cnt += svcs.size();
-                }
-
-                assertEquals(nodeCount() * 2, cnt);
-            }
-        });
+            if (svcs != null)
+                cnt.addAndGet(svcs.size());
+        }
+        assertTrue(GridTestUtils.waitForCondition(() -> cnt.get() == nodeCount() * 2, 10000));
```
AMashenkov: Semantic is broken here. 'cnt' is calculated only once and will never be changed. 
So, if condition doesn't met there is no sense to wait neither 10 sec nor more nor less.

modules/core/src/test/java/org/apache/ignite/testframework/GridTestUtils.java
```diff
@@ -1701,6 +1701,9 @@ private static boolean areCompatible(Object[] objs, Class[] types) {
      * {@code assertion} closure will be executed {@code retries} + 1 times and
      * thread will spend approximately {@code retries} * {@code retryInterval} sleeping.
      *
+     * Please consider GridTestUtils#waitForCondition(GridAbsPredicate, long)
+     * to make logs clearer and to avoid possible false positive results.
+     *
```
AMashenkov: Let's drop this method at all.

Diana-Iakovleva: Done. I've made another method, which in general do the same but it's exception-free to make tests clearer, is it acceptable?

modules/core/src/test/java/org/apache/ignite/testframework/GridTestUtils.java
```diff
@@ -1798,6 +1766,35 @@ public static boolean waitForCondition(GridAbsPredicate cond, long timeout) thro
...
+                U.sleep(DFLT_BUSYWAIT_SLEEP_INTERVAL);
+
+                curTime = U.currentTimeMillis();
+            }
+        }
+        catch (Exception ex) {
+            return false;
+        }
+
+        return false;
```
AMashenkov: This can lead to false-negative result as condition wasn't check after last sleep.
Should be "return con.get()"

modules/core/src/test/java/org/apache/ignite/testframework/GridTestUtils.java
```diff
@@ -1798,6 +1766,35 @@ public static boolean waitForCondition(GridAbsPredicate cond, long timeout) thro
...
+     * @return {@code true} if condition was met, {@code false} otherwise or in case of timeout/exception.
+     */
+    public static boolean waitUntil(Supplier<Boolean> cond, long timeout) {
+        assert timeout > 0;
+        long curTime = U.currentTimeMillis();
+        long endTime = curTime + timeout;
+
+        try {
+            while (curTime < endTime) {
+                if (cond.get()) {
```
AMashenkov: Code style convention does't allow braces for one-liner.

PR #6845, opened 2019-09-06
IGNITE-12145: Monitoring list engine.

nizhikov:
This PR contains:

* Monitoring list engine.
  * Support for SQL, JMX, Log exporters implemented.
  * Internal MonitoringList API added.
  * Enable/Disable of lists implemented.
* Following list implemented:
  * Cache list
  * Cache group list
  * Compute task list
  * Service list.

Engine details:
* `MonitoringList` added to store list data.
* Base interface `MonitoringRow` for list data created.
* Corresponding method added to `MetricExporterSpi`
* `JmxMetricExporterSpi`, `SqlViewExporterSpi`, `LogExporterSpi` updated to support list export.
* JMX, SQL and other column-oriented SPI uses `MonitoringRowAttributeWalker` to quickly traverse all list row attributes.
* Implementation of `MonitoringRowAttributeWalker` for specific `MonitoringRow` can be generated with `MonitoringRowAttributeWalkerGenerator`


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ClusterCachesInfo.java
```diff
@@ -145,6 +173,14 @@
...
         this.ctx = ctx;
 
+        ctx.metric().list(CACHES_MON_LIST, CACHES_MON_LIST_DESC, CacheView.class,
+            l -> cachesMonList = l,
+            l -> cachesMonList = null);
+
+        ctx.metric().list(CACHE_GRPS_MON_LIST, CACHE_GRPS_MON_LIST_DESC, CacheGroupView.class,
+            l -> cachesGrpMonList = l,
+            l -> cachesGrpMonList = null);
+
```
alex-plekhanov: List will be inconsistent when someone will disable it in runtime, create new caches and enable again.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ClusterCachesInfo.java
```diff
@@ -1397,7 +1438,10 @@ else if (!GridFunc.eqNotOrdered(desc.schema().entities(), localQueryEntities))
 
             desc.receivedOnDiscovery(true);
 
-            registeredCaches.put(cacheData.cacheConfiguration().getName(), desc);
+            String cacheName = cacheData.cacheConfiguration().getName();
```
alex-plekhanov: Redundant change, `cacheName` has only one usage

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ClusterCachesInfo.java
```diff
@@ -2514,7 +2572,7 @@ public void removeRestartingCaches() {
                 if (o1.cacheType().userCache() ^ o2.cacheType().userCache())
                     return o2.cacheType().userCache() ? -1 : 1;
 
-                return o1.cacheId().compareTo(o2.cacheId());
+                return o1.cacheId() - o2.cacheId();
```
alex-plekhanov: Undesirable change. Has almost the same logic as before, but produce errors on some corner cases. Not related to this ticket.

nizhikov: It's better to use primitives in `MonitoringRow` for the sake of speed. 
Replaced with `Integer.compare(o1.cacheId(), o2.cacheId())`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ClusterCachesInfo.java
nizhikov: It'is expected behavior, from my point of view.
It's not easy to build a consistent list if it enabled in runtime.

Transactions list, locks list, can be taken as an example.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/list/walker/CacheGroupViewWalker.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Generated by {@code org.apache.ignite.codegen.MonitoringRowAttributeWalkerGenerator}.
+ * {@link CacheGroupView} attributes walker.
+ * 
+ * @see CacheGroupView
+ */
+public class CacheGroupViewWalker implements MonitoringRowAttributeWalker<CacheGroupView> {
+    /** {@inheritDoc} */
+    @Override public void visitAll(AttributeVisitor v) {
+        v.accept(0, "cacheGroupName", String.class);
```
daradurvs: It may irritate in future to support such walkers because of matching id -> name in several methods.
Maybe it makes sense describing fields names in private enum, possibly it makes code cleaner and #count() will return #values().length

```    
    private enum CacheFields {
        CACHE_GROUP_NAME(0, "cacheGroupName"),
        CACHE_COUNT(1, "cacheCount"),
        // etc.;

        private final int idx;
        private final String name;

        CacheFields(int idx, String name) {
            this.idx = idx;
            this.name = name;
        }
    }
 ```

File: modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java [hunk 5]
```diff
@@ -124,6 +133,14 @@
      */
     private final ConcurrentMap<IgniteUuid, ServiceInfo> registeredServices = new ConcurrentHashMap<>();
 
+    /**
+     * Services monitoring list.
+     *
+     * @see ReadOnlyMonitoringListRegistry
+     * @see GridMetricManager
+     */
+    @Nullable private volatile MonitoringListImpl<IgniteUuid, ServiceView> svcsMonList;
```
daradurvs: Use interface `MonitoringList` instead of an implementation class

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -265,45 +377,132 @@ public MetricRegistry registry(String name) {
...
+     *
+     * @param name List name.
+     * @param description List description.
+     * @param row Row class
+     * @param storer Storer callback.
+     * @param cleaner Cleaner callback.
+     * @param <Id> Type of the id of the monitoring row.
+     * @param <R> Type of the monitoring row.
+     */
+    public <Id, R extends MonitoringRow<Id>> void list(String name, String description,
```
daradurvs: I believe this method should return MonitoringList.
You always use it to register metrics and initialize fields MonitoringList in a functional style.
If the method returns `<Id, R extends MonitoringRow<Id>> MonitoringList<Id, R>` initialize blocks in IgniteServiceProcessor, GridTaskProcessor, ClusterCachesInfo will be clearer.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ClusterCachesInfo.java
nizhikov: All lists made consistent after enable/disable.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -959,29 +959,29 @@ public void initQueryStructuresForNotStartedCache(DynamicCacheDescriptor cacheDe
      */
     @SuppressWarnings({"unchecked"})
     private void stopCache(GridCacheAdapter<?, ?> cache, boolean cancel, boolean destroy) {
-        GridCacheContext ctx = cache.context();
+        GridCacheContext cctx = cache.context();
```
daradurvs: Can't see a reason for this change.

modules/core/src/main/java/org/apache/ignite/spi/metric/list/MonitoringList.java
```diff
@@ -0,0 +1,41 @@
...
+
+package org.apache.ignite.spi.metric.list;
+
+/**
+ * Implementation provides data for some internal Ignite objects.
+ *
+ * @param <Id> Type of the row identificator.
+ * @param <R> Type of the row.
+ */
+public interface MonitoringList<Id, R extends MonitoringRow<Id>> extends Iterable<R> {
```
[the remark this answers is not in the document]
nizhikov: @alex-plekhanov 
@agura 
Guys, what do you think about naming?

modules/core/src/main/java/org/apache/ignite/spi/metric/list/MonitoringRow.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Monitoring list row base interface.
+ * Each row idenitified by the instance of {@code Id}.
+ *
+ * @see MonitoringList
+ * @see GridMetricManager
+ * @see ReadOnlyMonitoringListRegistry
+ */
+public interface MonitoringRow<Id> {
```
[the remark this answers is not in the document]
nizhikov: Hmm, it's a good point. Thanks!

I think the answer is no. 
We are not going to provide general serializability.
We can track some `Thread`s or other not serializable data later.

What do you think?

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/list/walker/CacheGroupViewWalker.java
nizhikov: Actually, this code was generated by `MonitoringRowAttributeWalkerGenerator`.
You can find its source code in this PR.
So, when you change some view class, you simply run it and got up to date walker implementation.

What do you think?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
nizhikov: Reverted. Thanks.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/list/walker/CacheGroupViewWalker.java
daradurvs: Thanks for the explanation, no problem then.

modules/core/src/main/java/org/apache/ignite/spi/metric/list/MonitoringRow.java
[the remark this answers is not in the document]
daradurvs: If Id's type becomes a part of public API then it may be hard to change it in future, otherwise, we can address it later when it really needs.

modules/codegen/src/main/java/org/apache/ignite/codegen/MonitoringRowAttributeWalkerGenerator.java
```diff
@@ -0,0 +1,289 @@
...
+import org.apache.ignite.spi.metric.list.MonitoringList;
+import org.apache.ignite.spi.metric.list.MonitoringRow;
+import org.apache.ignite.spi.metric.list.MonitoringRowAttributeWalker;
+import org.apache.ignite.spi.metric.list.view.CacheGroupView;
+import org.apache.ignite.spi.metric.list.view.CacheView;
+import org.apache.ignite.spi.metric.list.view.ComputeTaskView;
+import org.apache.ignite.spi.metric.list.view.ServiceView;
+import org.apache.ignite.spi.metric.sql.MonitoringListLocalSystemView;
+
+import static org.apache.ignite.codegen.MessageCodeGenerator.DFLT_SRC_DIR;
```
alex-plekhanov: Sorry, I misplaced my previous comment about static import from `MessageCodeGenerator`. I mean DFLT_SRC_DIR need to be independent for these classes. But this comment replated to `TAB` constant too. If you will fix this, there is no need to change `MessageCodeGenerator` at all.

modules/codegen/src/main/java/org/apache/ignite/codegen/MonitoringRowAttributeWalkerGenerator.java
```diff
@@ -0,0 +1,289 @@
...
+
+        final int[] cnt = {0};
+        forEachMethod(clazz, (m, i) -> cnt[0]++);
+
+        code.add(TAB + "/** {@inheritDoc} */");
+        code.add(TAB + "@Override public int count() {");
+        code.add(TAB + TAB + "return " + cnt[0] + ';');
+        code.add(TAB + "}");
+        code.add("}");
+        code.add("");
```
alex-plekhanov: Redundant NL generation

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/list/AbstractMonitoringList.java
```diff
@@ -0,0 +1,79 @@
...
+    /** Description of the list. */
+    private final String description;
+
+    /** Class of the row */
+    private final Class<R> rowClass;
+
+    /**
+     * Row attribute walker.
+     *
+     * @see org.apache.ignite.codegen.MonitoringRowAttributeWalkerGenerator
```
alex-plekhanov: IDEA highlighted this line as an error, since MonitoringRowAttributeWalkerGenerator from module `ignite-codegen` which is not in dependencies.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/list/AbstractMonitoringList.java
```diff
@@ -0,0 +1,79 @@
...
+     */
+    private final MonitoringRowAttributeWalker<R> walker;
+
+    /**
+     * @param name Name of the list.
+     * @param description Description of the list.
+     * @param rowClass Class of the row.
+     * @param walker Row attribute walker.
+     */
+    AbstractMonitoringList(String name, String description, Class<R> rowClass, MonitoringRowAttributeWalker<R> walker) {
```
alex-plekhanov: It's desirable to use abbreviations for "description" and "class"

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/list/AbstractMonitoringList.java
nizhikov: So it's the issue in the IDEA, isn't it?)

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -177,9 +224,60 @@
...
+    /** Lists registry. */
+    private final ReadOnlyMonitoringListRegistry listRegistry = new ReadOnlyMonitoringListRegistry() {
+        /** {@inheritDoc} */
+        @Override public void addListCreationListener(Consumer<MonitoringList<?>> lsnr) {
+            listCreationLsnrs.add(lsnr);
+        }
+
+        /** {@inheritDoc} */
+        @Override public void addListRemoveListener(Consumer<MonitoringList<?>> lsnr) {
+            listRemoveLsnrs.add(lsnr::accept);
```
alex-plekhanov: Nested consumer creation, `lsnr::accept` should be replaced with `lsnr`

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -265,45 +378,116 @@ public MetricRegistry registry(String name) {
...
+            notifyListeners(mreg, metricRegCreationLsnrs, log);
 
             return mreg;
         });
     }
 
-    /** {@inheritDoc} */
-    @NotNull @Override public Iterator<MetricRegistry> iterator() {
-        return registries.values().iterator();
+    public <R extends MonitoringRow, D> void list(String name, String description,
```
alex-plekhanov: Missed JavaDoc

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -265,45 +378,116 @@ public MetricRegistry registry(String name) {
...
+        Class<R> rowClazz, Supplier<ConcurrentMap<?, D>> data, Function<D, R> rowFunc, Consumer<D> clearer) {
+
```
alex-plekhanov: Redundand NL

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -265,45 +378,116 @@ public MetricRegistry registry(String name) {
...
+            description,
+            rowClazz,
+            (MonitoringRowAttributeWalker<R>)walkers.get(rowClazz),
+            data.get(),
+            rowFunc));
+
+        //Create new instance of the list.
+        listCreator.get();
+
+        ctx.metric().addEnableListListener(listenOnlyEqual(name, identity(), n -> listCreator.get()));
```
alex-plekhanov: A very complicated logic here. It's up to 5-6 nested consumers/suppliers used for event listener. Perhaps it can be simplified somehow (but I'm not sure). In current implementation is hard to understand such a code.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -265,45 +378,116 @@ public MetricRegistry registry(String name) {
...
+        Class<R> rowClazz, Supplier<ConcurrentMap<?, D>> data, Function<D, R> rowFunc, Consumer<D> clearer) {
+
+        Supplier<MonitoringList<R>> listCreator = () -> list(name, () -> new MonitoringListAdapter<>(name,
+            description,
+            rowClazz,
+            (MonitoringRowAttributeWalker<R>)walkers.get(rowClazz),
+            data.get(),
+            rowFunc));
+
+        //Create new instance of the list.
```
alex-plekhanov: Space missed

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
nizhikov: 1. We create a list which exports map content.
2. On list removal event(via `IgniteMBean#disableList`) `rowClearer` will clear data needed only for list(if applicable).
3. On list enable event(via `IgniteMBean#enableList`) it will be recreated(fresh instance of the `data` map will be obtained.

I think it pretty simple and easy to use approach.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
alex-plekhanov: Perhaps `createList` is a better name since you already have overloaded method with the same name, which only returns the existing list.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -265,45 +378,116 @@ public MetricRegistry registry(String name) {
...
+
+    /**
+     * Gets or creates {@link MonitoringList}.
+     *
+     * @param name Name of the list.
+     * @param listSupplier List supplier.
+     * @param <R> Type of the row.
+     * @return Monitoring list.
+     */
+    private <R extends MonitoringRow> MonitoringList<R> list(String name,
```
alex-plekhanov: Only one usage. Can be inlined into `list` (`createList`) and simplified

nizhikov: It will be more usages in the follow-up PRs.
Let's keep it?

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
alex-plekhanov: And actually it's not a listener. It's something like "list enabler", it's lack of abstraction here since listener should only react to this event, but not performing event itself.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -265,45 +378,116 @@ public MetricRegistry registry(String name) {
...
 
             return mreg;
         });
     }
 
-    /** {@inheritDoc} */
-    @NotNull @Override public Iterator<MetricRegistry> iterator() {
-        return registries.values().iterator();
+    public <R extends MonitoringRow, D> void list(String name, String description,
+        Class<R> rowClazz, Supplier<ConcurrentMap<?, D>> data, Function<D, R> rowFunc, Consumer<D> clearer) {
```
alex-plekhanov: Do we really need `Supplier` for `data` parameter here? Why can't we use data without supplier?

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/list/ListUtils.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package org.apache.ignite.internal.processors.metric.list;
+
+import java.util.Objects;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import org.apache.ignite.spi.metric.list.MonitoringRow;
+import org.jetbrains.annotations.Nullable;
```
alex-plekhanov: Unused imports

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/list/ListUtils.java
```diff
@@ -0,0 +1,47 @@
...
+
+/** Utils methods for monitoring list feature. */
+public final class ListUtils {
```
alex-plekhanov: I think this class don't need at all after simplification of `GridMetricManager#list` method

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/list/MonitoringListAdapter.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.ignite.spi.metric.list.MonitoringRow;
+import org.apache.ignite.spi.metric.list.MonitoringRowAttributeWalker;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Monitoring list backed by {@code data} {@link ConcurrentMap}.
+ */
+public class MonitoringListAdapter<R extends MonitoringRow, D> extends AbstractMonitoringList<R> {
+    /** Data backed by this list. */
+    private final ConcurrentMap<?, D> data;
```
alex-plekhanov: Why do we need `ConcurrentMap` here? It brings some restrictions to data source for lists. We only use `values()` and `size()` methods of `data` field, why can't we use `Collection`?

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/list/AbstractMonitoringList.java
alex-plekhanov: Not sure it's IDEA issue, perhaps we shouldn't link to unreachable classes even in JavaDoc.

alex-plekhanov: Or at least you can use `@see "text"` syntax, but not `@see reference`

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
nizhikov: A map may be recreated in the class who owns data.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/list/MonitoringListAdapter.java
nizhikov: Because Map is not a Collection, I think :)

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
nizhikov: I doesn't understand your proposal, actually. 
What is the different between "enable list event" and "list enabler"?

modules/codegen/src/main/java/org/apache/ignite/codegen/MonitoringRowAttributeWalkerGenerator.java
alex-plekhanov: Because classes not related to each other. Why I should change `MessageCodeGenerator` if I only want to change the path for `MonitoringRowAttributeWalkerGenerator`?

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
alex-plekhanov: I don't say that I didn't understand the code, I say that code is hard to understand because it uses 5-6 nested lambdas. It's certainly not a pretty simple.  

I don't propose to rename it to "list enabler", I say that "listener" is a bad abstraction for this case.

Perhaps code can be simplified in this way:
1. Create maps for list creators and removers (name -> closure) instead of lists for listeners 
2. On enabling/disabling just call closure from this map by name.
3. Get rid of as most nested lambdas as possible. Now it looks redundant. Ideally, for list creation case there should be only two labdas: one for list creation closure, and one for `computeIfAbsent` mapping function.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/list/MonitoringListAdapter.java
alex-plekhanov: You use only `values()` collection from this map. I see no reason to pass map here.

And `Map` say nothing about thread safety. Thread safety provides by the implementation, not by the interface.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
alex-plekhanov: Let's don't introduce any functional prematurely, follow-up PRs also can have some fixes after review, so it's not a fact that this method will be needed.

And here is naming problem again. You have overloaded `registerList` methods, one of it register some listeners, another one creates the list after enabling.

modules/codegen/src/main/java/org/apache/ignite/codegen/MonitoringRowAttributeWalkerGenerator.java
nizhikov: It's not a path for `MonitoringRowAttributeWalkerGenerator`, it's a path of source in the core module. You should use another constant if you want to generate code in another module.

But if the path for the core source changes, I think we want to change only one constant.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
nizhikov: OK, fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
alex-plekhanov: In this case, you will have the same problem with the current implementation too (the only difference is you can reenable the list to work correctly). 
But I think it's better to re-register the list if class who owns the data recreates the map. In this case, the list will always be up to date and no reenabling is needed.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/list/MonitoringListAdapter.java
nizhikov: > And Map say nothing about thread safety.

Thats why I use `ConcurrentMap` which declares:

```
 * A {@link java.util.Map} providing thread safety and atomicity
 * guarantees.
```

If we use `Collection` we can't force "user" of the list to pass thread-safety implementation of Collection.

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
nizhikov: OK, fixed.

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxMetricExporterSpiTest.java
```diff
@@ -144,4 +210,125 @@ public void testFilterAndExport() throws Exception {
...
+
+    /** */
+    @Test
+    public void testCachesList() throws Exception {
+        Set<String> cacheNames = new HashSet<>(Arrays.asList("cache-1", "cache-2"));
+
+        for (String name : cacheNames)
+            ignite.createCache(name);
+
+        TabularDataSupport data = monitoringList("caches");
```
alex-plekhanov: "caches" -> `GridMetricManager#CACHES_MON_LIST`

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxMetricExporterSpiTest.java
```diff
@@ -144,4 +210,125 @@ public void testFilterAndExport() throws Exception {
...
+
+        assertEquals(3, data.size());
```
alex-plekhanov: I think something like `ignite.context().cache().cacheDescriptors().size()` will be better then just a constant.

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxMetricExporterSpiTest.java
```diff
@@ -144,4 +210,125 @@ public void testFilterAndExport() throws Exception {
...
+
+        for(int i=0; i<data.size(); i++) {
```
alex-plekhanov: Spaces missed (after "for", around "<")

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxMetricExporterSpiTest.java
```diff
@@ -144,4 +210,125 @@ public void testFilterAndExport() throws Exception {
...
+
+    /** */
+    @Test
+    public void testCacheGroupsList() throws Exception {
+        Set<String> grpNames = new HashSet<>(Arrays.asList("grp-1", "grp-2"));
+
+        for (String grpName : grpNames)
+            ignite.createCache(new CacheConfiguration<>("cache-" + grpName).setGroupName(grpName));
+
+        TabularDataSupport grps = monitoringList("cacheGroups");
```
alex-plekhanov: `GridMetricManager#CACHE_GRPS_MON_LIST`

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxMetricExporterSpiTest.java
```diff
@@ -144,4 +210,125 @@ public void testFilterAndExport() throws Exception {
...
+
+        assertEquals("ignite-sys, grp-1, grp-2", 3, grps.size());
```
alex-plekhanov: Strange error message. "3" can be replaced with `ignite.context().cache().cacheGroupDescriptors().size()`

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxMetricExporterSpiTest.java
```diff
@@ -144,4 +210,125 @@ public void testFilterAndExport() throws Exception {
...
+    public void testServices() throws Exception {
+        ServiceConfiguration srvcCfg = new ServiceConfiguration();
+
+        srvcCfg.setName("service");
+        srvcCfg.setMaxPerNodeCount(1);
+        srvcCfg.setService(new DummyService());
+
+        ignite.services().deploy(srvcCfg);
+
+        TabularDataSupport srvs = monitoringList("services");
```
alex-plekhanov: `SVCS_MON_LIST`

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxMetricExporterSpiTest.java
```diff
@@ -144,4 +210,125 @@ public void testFilterAndExport() throws Exception {
...
+                try {
+                    Thread.sleep(3_000L);
+                }
+                catch (InterruptedException e) {
+                    throw new RuntimeException(e);
+                }
+            });
+        }
+
+        TabularDataSupport tasks = monitoringList("tasks");
```
alex-plekhanov: `TASKS_MON_LIST`

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxMetricExporterSpiTest.java
```diff
@@ -144,4 +210,125 @@ public void testFilterAndExport() throws Exception {
...
+        assertEquals(DummyService.class.getName(), sview.get("serviceClass"));
+    }
+
+    /** */
+    @Test
+    public void testComputeBroadcast() throws Exception {
+        for (int i = 0; i < 5; i++) {
+            ignite.compute().broadcastAsync(() -> {
+                try {
+                    Thread.sleep(3_000L);
```
alex-plekhanov: Please use some synchronizer, but not `sleep()`.
Also, you need synchronization before checking the list. I think CyclicBarrier will be helpful here, which can be used twice.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/SqlSystemViewsSelfTest.java
```diff
@@ -1293,7 +1302,7 @@ public void testCachesViews() throws Exception {
         assertEquals(0L, execSql("SELECT COUNT(*) FROM " + systemSchemaName() + ".CACHES WHERE CACHE_NAME = 'no_such_cache'").get(0)
             .get(0));
 
-        assertEquals(0L, execSql("SELECT COUNT(*) FROM " + systemSchemaName() + ".CACHES WHERE CACHE_NAME = 1").get(0).get(0));
+        //assertEquals(0L, execSql("SELECT COUNT(*) FROM " + systemSchemaName() + ".CACHES WHERE CACHE_NAME = 1").get(0).get(0));
```
alex-plekhanov: Why this line is commented?

nizhikov: Actually, this is error in the existing test. `CACHE_NAME` is a string and 1 is number.

Fixed, for now. Thanks!

alex-plekhanov: It's not an error, it was made on purpose to cover case that comparison with wrong data type doesn't break the view, SQL should return an empty result set instead.

modules/core/src/main/java/org/apache/ignite/spi/metric/list/MonitoringList.java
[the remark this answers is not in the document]
agura: I've post my proposal to the div-list

modules/core/src/main/java/org/apache/ignite/spi/metric/list/MonitoringRow.java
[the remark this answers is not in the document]
agura: Serialization is not needed, IMO.

modules/core/src/test/java/org/apache/ignite/internal/metric/SystemViewSelfTest.java
```diff
@@ -0,0 +1,144 @@
...
+            assertEquals(DummyService.class, sview.serviceClass());
+        }
+    }
+
+    /** */
+    @Test
+    public void testComputeBroadcast() throws Exception {
+        latch = new CountDownLatch(1);
+
+        try(IgniteEx g1 = startGrid(0)) {
```
alex-plekhanov: Space missed

modules/core/src/test/java/org/apache/ignite/internal/metric/SystemViewSelfTest.java
```diff
@@ -0,0 +1,144 @@
...
+            SystemView<ComputeTaskView> tasks = g1.context().metric().view(TASKS_VIEW);
+
+            for (int i=0; i<5; i++)
```
alex-plekhanov: Spaces missed around "<"

modules/core/src/test/java/org/apache/ignite/internal/metric/SystemViewSelfTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+            assertEquals(srvcCfg.getName(), sview.name());
+            assertEquals(srvcCfg.getMaxPerNodeCount(), sview.maxPerNodeCount());
+            assertEquals(DummyService.class, sview.serviceClass());
+        }
+    }
+
+    /** */
+    @Test
+    public void testComputeBroadcast() throws Exception {
```
alex-plekhanov: I think we should cover more cases for tasks (for example run tasks by name/by class/by closure, run tasks from another node), to see different values for different rows in one column of the list. The same for services.

nizhikov: Fixed!!!

modules/core/src/main/java/org/apache/ignite/spi/metric/jmx/JmxMetricExporterSpi.java
```diff
@@ -31,41 +32,99 @@
...
 
+    /** Syste view registry. */
+    private ReadOnlySystemViewRegistry mlreg;
+
     /** Metric filter. */
-    private @Nullable Predicate<MetricRegistry> filter;
+    private @Nullable Predicate<MetricRegistry> metricFilter;
+
+    /** System view filter. */
+    private @Nullable Predicate<SystemView<?>> systeViewFilter;
```
alex-plekhanov: syste**m**ViewFilter

modules/core/src/main/java/org/apache/ignite/internal/cluster/ClusterGroupAdapter.java
```diff
@@ -805,7 +805,7 @@ private CachesFilter(String cacheName, boolean affNodes, boolean nearNodes, bool
 
     /**
      */
-    private static class AttributeFilter implements IgnitePredicate<ClusterNode> {
+    public static class AttributeFilter implements IgnitePredicate<ClusterNode> {
```
alex-plekhanov: I think we can (optionally) set custom node filter in a test and left this class private.

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxMetricExporterSpiTest.java
```diff
@@ -144,4 +196,129 @@ public void testFilterAndExport() throws Exception {
...
+                }
+                catch (InterruptedException e) {
+                    throw new RuntimeException(e);
+                }
+            });
+        }
+
+        TabularDataSupport tasks = systemView(TASKS_VIEW);
+
+        assertEquals(5, tasks.size());
```
alex-plekhanov: Perhaps we can get flaky failure here, I think it's better to use `CyclicBarrier` instead of the latch (before checks and after checks).

modules/clients/src/test/java/org/apache/ignite/internal/jdbc2/JdbcMetadataSelfTest.java
```diff
@@ -320,10 +319,8 @@ public void testGetTableTypes() throws Exception {
      */
     @Test
     public void testGetAllView() throws Exception {
-        List<String> expViews = Arrays.asList(
+        Set<String> expViews = new HashSet<>(Arrays.asList(
             "BASELINE_NODES",
-            "CACHES",
-            "CACHE_GROUPS",
```
alex-plekhanov: Wouldn't it be better to have enabled system view by default? To enable views admin must restart the node, sometimes it's not a good issue.

modules/core/src/main/java/org/apache/ignite/spi/metric/ReadOnlySystemViewRegistry.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Read only system view registry.
+ *
+ * @see GridMetricManager
+ * @see SystemView
+ * @see MetricExporterSpi
+ */
+public interface ReadOnlySystemViewRegistry extends Iterable<SystemView<?>> {
+    /**
+     * Adds listener of list creation events.
```
alex-plekhanov: list -> view

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -268,53 +338,69 @@ public MetricRegistry registry(String name) {
...
-    }
+    /**
+     * Registers view which exports {@link Collection} content.
+     *
+     * @param name Name.
+     * @param desc Description.
+     * @param rowCls Row class.
+     * @param data Data.
+     * @param rowFunc value to row function.
+     * @param <R> List row type.
```
alex-plekhanov: List -> view

modules/core/src/main/java/org/apache/ignite/spi/metric/view/SystemViewRow.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.apache.ignite.spi.metric.view;
+
+import org.apache.ignite.internal.processors.metric.GridMetricManager;
+import org.apache.ignite.spi.metric.ReadOnlySystemViewRegistry;
+
+/**
+ * System view row base interface.
+ * All public non-static methods, except system one, will be exporter as a list columns.
```
alex-plekhanov: list -> view

modules/clients/src/test/java/org/apache/ignite/internal/jdbc2/JdbcMetadataSelfTest.java
nizhikov: I see your point.
Let's discuss it on the dev-list?

Seems, it's up to community consensus.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -78,6 +78,8 @@
 import org.apache.ignite.spi.loadbalancing.LoadBalancingSpi;
 import org.apache.ignite.spi.loadbalancing.roundrobin.RoundRobinLoadBalancingSpi;
 import org.apache.ignite.spi.metric.MetricExporterSpi;
+import org.apache.ignite.spi.systemview.SystemViewExporterSpi;
+import org.apache.ignite.spi.systemview.jmx.JmxSystemViewExporterSpi;
```
alex-plekhanov: Unused import

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -2347,14 +2353,36 @@ public IgniteConfiguration setMetricExporterSpi(MetricExporterSpi... metricExpor
...
      * @return Metric exporter SPI implementations.
      */
     public MetricExporterSpi[] getMetricExporterSpi() {
         return metricExporterSpi;
     }
 
+    /**
+     * Gets fully configured system view SPI implementations.
+     *
+     * @return Metric exporter SPI implementations.
```
alex-plekhanov: "System view exporter SPI ..."

modules/core/src/main/java/org/apache/ignite/spi/systemview/view/ServiceView.java
```diff
@@ -0,0 +1,98 @@
...
+    public String cacheName() {
+        return serviceInfo.cacheName();
+    }
+
+    /** @return Affininty key value. */
+    public String affinityKey() {
+        Object affKey = serviceInfo.configuration().getAffinityKey();
+
+        return affKey == null ? null : affKey.toString();
+
```
alex-plekhanov: Redundant NL

modules/core/src/main/java/org/apache/ignite/spi/systemview/view/ServiceView.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.ignite.internal.managers.systemview.walker.Order;
+import org.apache.ignite.internal.processors.service.ServiceInfo;
+import org.apache.ignite.lang.IgnitePredicate;
+import org.apache.ignite.lang.IgniteUuid;
+
+/**
+ * Service representation for a {@link SystemView}.
+ */
+public class ServiceView {
+    private final ServiceInfo serviceInfo;
```
alex-plekhanov: JavaDoc

modules/core/src/main/java/org/apache/ignite/spi/systemview/view/ServiceView.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public ServiceView(ServiceInfo serviceInfo) {
```
alex-plekhanov: JavaDoc

modules/core/src/test/java/org/apache/ignite/internal/metric/SystemViewSelfTest.java
```diff
@@ -0,0 +1,400 @@
...
+            IgniteCache<Integer, Integer> cache = g1.createCache("test-cache");
+
+            cache.put(1, 1);
+
+            g1.compute().executeAsync(new ComputeTask<Object, Object>() {
+                @Override public @NotNull Map<? extends ComputeJob, ClusterNode> map(List<ClusterNode> subgrid,
+                    @Nullable Object arg) throws IgniteException {
+                    return Collections.singletonMap(new ComputeJob() {
+                        @Override public void cancel() {
+                            System.out.println("SystemViewSelfTest.cancel");
```
alex-plekhanov: Some debug output, perhaps can be removed

modules/core/src/test/java/org/apache/ignite/internal/metric/SystemViewSelfTest.java
```diff
@@ -0,0 +1,400 @@
...
+                            // No-op.
+                        }
+
+                        @Override public Object execute() throws IgniteException {
+                            System.out.println("SystemViewSelfTest.execute");
```
alex-plekhanov: Some debug output, perhaps can be removed

modules/core/src/test/java/org/apache/ignite/internal/metric/SystemViewSelfTest.java
nizhikov: Sorry for that. Fixed.

PR #6847, opened 2019-09-06
IGNITE-12147 Thin client: Fixed partitions response for caches with n…

alex-plekhanov:
…ot applicable affinity awareness

### DIFF
--- a/modules/platforms/dotnet/Apache.Ignite.Core.Tests/Client/Cache/AffinityAwarenessTest.cs
+++ b/modules/platforms/dotnet/Apache.Ignite.Core.Tests/Client/Cache/AffinityAwarenessTest.cs
@@ -272,6 +272,7 @@ namespace Apache.Ignite.Core.Tests.Client.Cache
             // Test cache for which affinity awareness is not applicable.
             var cfg = new CacheClientConfiguration("replicated_cache") {CacheMode = CacheMode.Replicated};
             var cache = Client.CreateCache<int, int>(cfg);
+            cache.Put(1, 1);
             ClearLoggers();
 
             cache.Get(1);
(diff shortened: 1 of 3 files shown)


modules/platforms/dotnet/Apache.Ignite.Core.Tests/Client/Cache/AffinityAwarenessTest.cs
```diff
@@ -268,6 +268,32 @@ public void CacheGet_RepeatedCall_DoesNotRequestAffinityMapping()
             };
 
             Assert.AreEqual(expectedRequests, requests);
+
+            // Test cache for which affinity awareness is not applicable.
```
ptupitsyn: Please move this to a separate test method.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Client/Cache/AffinityAwarenessTest.cs
```diff
@@ -268,6 +268,32 @@ public void CacheGet_RepeatedCall_DoesNotRequestAffinityMapping()
...
+
+            cache.Put(1, 1);
+            Client.CreateCache<int, int>("repeat-call-test-replicated");
+            ClearLoggers();
+
+            cache.Get(1);
+            cache.Get(1);
+            cache.Get(1);
+
+            requests = null;
```
ptupitsyn: So we need to verify that for Replicated cache all requests go to a single (default) node, because affinity awareness does not matter. Let's state this in a comment for clarity.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Client/Cache/AffinityAwarenessTest.cs
```diff
@@ -268,6 +268,32 @@ public void CacheGet_RepeatedCall_DoesNotRequestAffinityMapping()
...
+
+            foreach (var logger in _loggers)
+            {
+                var cacheRequestNames = GetCacheRequestNames(logger);
+
+                if (cacheRequestNames.Any())
+                {
+                    requests = cacheRequestNames.ToArray();
```
ptupitsyn: 1. `cacheRequestNames` is a lazy `IEnumerable`, calling `Any` then `ToArray` causes multiple enumeration, which is an anti-pattern.

2. We overwrite `requests` here, possibly missing some data, which can be confusing.

I propose to replace the loop with the following:
```
            var reqs = _loggers
                .Select(l => new {Logger = l, Requests = GetCacheRequestNames(l).ToArray()})
                .Where(r => r.Requests.Length > 0)
                .ToArray();
            
            Assert.AreEqual(1, reqs.Length);
            Assert.AreEqual(expectedRequests, reqs[0].Requests);
```

I think the intent is more obvious this way.

PR #6849, opened 2019-09-06
IGNITE-12148: [ML] Recommendation Engine

zaleslaw:


zaleslaw:
Add updatable Recommendation Engine + support for ModelStorage in Windows

zaleslaw:
@avplatonov Please make review

modules/ml/pom.xml
```diff
@@ -102,7 +102,6 @@
             <groupId>org.springframework</groupId>
             <artifactId>spring-context</artifactId>
             <version>${spring.version}</version>
-            <scope>test</scope>
```
YuriBabak: We really need this change?

modules/ml/src/main/java/org/apache/ignite/ml/recommendation/RecommendationModel.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.ignite.ml.IgniteModel;
+import org.apache.ignite.ml.math.primitives.vector.Vector;
+
+/**
+ * Recommendation model that predicts rating for {@link ObjectSubjectPair}.
+ *
+ * @param <O> Type of an object of recommendation.
+ * @param <S> Type of a subject of recommendation.
+ */
+// TODO: GG-22915 Investigate ability to distribute recommendation model
```
YuriBabak: Please, cleanup this

modules/ml/pom.xml
zaleslaw: yep, we use spring in the resolving resource path

modules/ml/src/main/java/org/apache/ignite/ml/recommendation/RecommendationModel.java
zaleslaw: Agree

examples/src/main/java/org/apache/ignite/examples/ml/recommendation/MovieLensExample.java
```diff
@@ -0,0 +1,149 @@
...
+                    .withLearningRate(10)
+                    .withLearningEnvironmentBuilder(envBuilder)
+                    .withTrainerEnvironment(envBuilder.buildForTrainer());
+
+                RecommendationModel<Integer, Integer> mdl = trainer.fit(
+                    new CacheBasedDatasetBuilder<>(ignite, movielensCache)
+                );
+
+                double mean = 0;
+                try (QueryCursor<Cache.Entry<Integer, RatingPoint>> cursor = movielensCache.query(new ScanQuery<>())) {
```
avplatonov: We can compute RSS using just one query, there is no need to do two-pass process of estimation. Please, have a look at this page: https://en.wikipedia.org/wiki/Variance . Variance could be computed using two means - mean value of original data and mean value of squared data. We can compute these two values to determine variance that used in RSS formula.

PR #6853, opened 2019-09-08
IGNITE-12024: [ML] PeerClassloading for ml related lambdas

avplatonov:
@zaleslaw, Please review this PR


modules/ml/src/main/java/org/apache/ignite/ml/composition/ModelsComposition.java
```diff
@@ -23,13 +23,15 @@
...
 import org.apache.ignite.ml.composition.predictionsaggregator.PredictionsAggregator;
+import org.apache.ignite.ml.environment.deploy.DeployableObject;
 import org.apache.ignite.ml.math.primitives.vector.Vector;
 import org.apache.ignite.ml.util.ModelTrace;
 
 /**
  * Model consisting of several models and prediction aggregation strategy.
  */
-public class ModelsComposition implements IgniteModel<Vector, Double>, Exportable<ModelsCompositionFormat> {
+public class ModelsComposition implements IgniteModel<Vector, Double>, Exportable<ModelsCompositionFormat>,
```
zaleslaw: Maybe we should make this class final too?

modules/ml/src/main/java/org/apache/ignite/ml/clustering/kmeans/KMeansModel.java
```diff
@@ -125,4 +128,9 @@ public DistanceMeasure distanceMeasure() {
             .addField("centroids", centersList)
             .toString();
     }
+
+    /** {@inheritDoc} */
+    @Override public List<Object> getDependencies() {
```
zaleslaw: We need to start the something like "Template Trainer Design Document" based on parent class to describe the purpose of all these methods and hidden limitations of implementation like this method, what do you think, @avplatonov ?

modules/ml/src/main/java/org/apache/ignite/ml/composition/ModelsComposition.java
avplatonov: I'm not sure, there is a child of this class (GDBModel) and I don't want to change an inheritance schema in this PR. Maybe we should create another ticker in Jira with a review of hierarchies of models (of course without changing APIs)

modules/ml/src/main/java/org/apache/ignite/ml/clustering/kmeans/KMeansModel.java
avplatonov: What do you mean? Is this a kind of example class with a lot of comments or is it a document in wiki?

avplatonov:
@zaleslaw I fixed conflicts. Please, have a look.

modules/ml/src/main/java/org/apache/ignite/ml/clustering/kmeans/KMeansModel.java
zaleslaw: I suggest the page for Wiki in AI Wiki (you could start from the draft in the Google Doc and after review merge it into Wiki)

avplatonov: Ok! I'll do it. Unfortunately, I won't have free time the next week but I'll try to prepare some draft.

PR #6876, opened 2019-09-18
IGNITE-12182 ExecutorService defaults to only server nodes

sdarlington:
Makes the executor service consistent with compute and service grids, only executing on server nodes by default.


dmagda:
Stephen,

I'm afraid we can't merge this change until Ignite 3.0 because it breaks backward compatibility. There are already might be many applications that use this API and execute the service on all the nodes purposefully. Let's wait until Ignite 3.0 is put on rails.

PR #6879, opened 2019-09-18
IGNITE-12197 Fixed method of getting value of persistent enabled in CacheGroupMetricsImpl

agura:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupMetricsImpl.java
```diff
@@ -85,22 +85,13 @@ public CacheGroupMetricsImpl(CacheGroupContext ctx) {
 
         mreg.register("Caches", this::getCaches, List.class, null);
 
-        if (isPDSEnabled()) {
-            mreg.register("StorageSize",
-                () -> database().forGroupPageStores(ctx, PageStore::size),
-                "Storage space allocated for group, in bytes.");
+        storageSize = mreg.register("StorageSize",
+            () -> database().forGroupPageStores(ctx, PageStore::size),
```
nizhikov: We would have `ClassCastException` here if using `IgniteCacheDatabaseSharedManager` instead of `GridCacheDatabaseSharedManager`

```
    private GridCacheDatabaseSharedManager database() {
        return (GridCacheDatabaseSharedManager)ctx.shared().database();
    }
```

agura:
Merged to master branch.

PR #6881, opened 2019-09-19
IGNITE-12199 Flush WAL records on backups for transactional caches

agura:


agura:
Merged to master branch.

PR #6882, opened 2019-09-19
IGNITE-12044: Export histogram values as separate JMX attributes.

nizhikov:


modules/core/src/main/java/org/apache/ignite/internal/processors/metric/impl/MetricUtils.java
```diff
@@ -133,6 +138,43 @@ private static boolean ensureAllNamesNotEmpty(String... names) {
...
 
+    /**
+     * Gets histogram bucket names.
+     *
+     * Example of metric names if bounds are 10,100:
+     *  histogram_0_10 (less than 10)
+     *  histogram_10_100 (between 10 and 100)
+     *  histogram_100_inf (more than 100)
+     *
+     * @param metric Histogram metric.
```
NSAmelchev: missed parameter "cache"

modules/core/src/main/java/org/apache/ignite/spi/metric/jmx/MetricRegistryMBean.java
```diff
@@ -146,4 +176,55 @@ else if (metric instanceof ObjectMetric)
...
+        if (!sc.hasNextLong()) {
+            if (sc.hasNext() && INF.equals(sc.next()) && bounds[bounds.length-1] == lowBound)
+                return values[values.length-1];
+
+            return null;
+        }
+
+        long highBound = sc.nextLong();
+
+        for (int i=0; i<bounds.length; i++) {
```
NSAmelchev: You missed spaces

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxMetricExporterSpiTest.java
```diff
@@ -131,17 +134,70 @@ public void testFilterAndExport() throws Exception {
...
 
-        DynamicMBean bean2 = metricSet(ignite.name(), "other", "prefix2");
+        DynamicMBean bean2 = metricRegistry(ignite.name(), "other", "prefix2");
 
         assertEquals(44L, bean2.getAttribute("test3"));
     }
+
+    /** */
+    @Test
+    public void testHistogramSearchByName() throws Exception {
```
NSAmelchev: I think we should add the case where histogram intervals change.

modules/core/src/main/java/org/apache/ignite/spi/metric/jmx/MetricRegistryMBean.java
```diff
@@ -146,4 +176,55 @@ else if (metric instanceof ObjectMetric)
...
+            if (bounds[i] == highBound) {
```
NSAmelchev: Bounds are sorted so we can use the binary search here instead of a full scan.

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxMetricExporterSpiTest.java
nizhikov: For now, we don't support interval changes.
So, I don't think we should test it.

Seems, we should re-register MBean if some metric change its properties.

modules/core/src/main/java/org/apache/ignite/spi/metric/jmx/MetricRegistryMBean.java
nizhikov: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/metric/impl/MetricUtils.java
nizhikov: Fixed

modules/core/src/main/java/org/apache/ignite/spi/metric/jmx/MetricRegistryMBean.java
nizhikov: Fixed

PR #6886, opened 2019-09-19
IGNITE-12200 More informative assertion message at constructor of Cac…

x-kreator:
…hedDeploymentInfo (GridCacheDeploymentManager class)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheDeploymentManager.java
```diff
@@ -921,7 +921,8 @@ private boolean isLocallyExcluded(String name) {
          */
         private CachedDeploymentInfo(UUID sndId, IgniteUuid ldrId, String userVer, DeploymentMode depMode,
             Map<UUID, IgniteUuid> participants) {
-            assert sndId.equals(ldrId.globalId()) || participants != null;
+            assert sndId.equals(ldrId.globalId()) || participants != null
```
Mmuzaf: Let's log all the input method params. WDYT?

x-kreator: Assertion error can be fired only if participants=null, I met the IDE warning about that when wrote that message construction expression earlier. But if you strongly recommend to do that, I'll do.

PR #6894, opened 2019-09-20
IGNITE-11868 GridClient#data() should be deprecated/removed.

kcmvp:
Removed method GridClient#data()

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/client/GridClient.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/client/GridClient.java
@@ -33,7 +33,6 @@ import java.util.UUID;
  * <p>
  * Use following methods to get access to remote cache functionality:
  * <ul>
- * <li>{@link #data()}</li>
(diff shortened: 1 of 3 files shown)


kcmvp:
TeamCity is green : https://ci.ignite.apache.org/viewLog.html?buildId=4621223&buildTypeId=IgniteTests24Java8_RunCache

modules/core/src/main/java/org/apache/ignite/internal/client/GridClient.java
```diff
@@ -33,7 +33,6 @@
  * <p>
  * Use following methods to get access to remote cache functionality:
```
sk0x50: For now, there is only one method which can be used to get access to cache functionality. Let's change the JavaDoc as follows: `Use the following method to get access to remote cache functionality`

kcmvp: ok

PR #6899, opened 2019-09-24
IGNITE-12208 Fix silent ignitevisorcmd.sh fail when JAVA_HOME is empty

liyuj:
IGNITE-12208 Fix silent ignitevisorcmd.sh fail when JAVA_HOME is empty


liyuj:
fixed

PR #6900, opened 2019-09-24
IGNITE-11312: Fix JDBC Driver properties in UI tools like DBeaver

lev555:
Issue: 
JDBC driver properties in UI tools like DBeaver doesn't work.

Root cause: 
JDBC driver connect method requires prefixed property names, like _ignite.jdbc.enforceJoinOrder_, but  _getPropertyInfo()_ method (which is used by UI tools for getting available properties) returns property names without prefix, like _enforceJoinOrder_.

Solution:
Add prefix _ignite.jdbc_ to property names, returned by _getPropertyInfo()_ method.

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/jdbc/thin/ConnectionPropertiesTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/jdbc/thin/ConnectionPropertiesTest.java
@@ -1,3 +1,20 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


modules/core/src/test/java/org/apache/ignite/internal/jdbc/thin/ConnectionPropertiesTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * {@link ConnectionPropertiesImpl} unit tests.
+ */
+public class ConnectionPropertiesTest {
+
```
pavlukhin: Redundant empty line according to our code-style (automated checks on TeamCity will report that problem).

modules/core/src/test/java/org/apache/ignite/internal/jdbc/thin/ConnectionPropertiesTest.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Properties;
+import org.junit.Test;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * {@link ConnectionPropertiesImpl} unit tests.
+ */
+public class ConnectionPropertiesTest {
```
pavlukhin: Please add this test to `IgniteJdbcDriverTestSuite` test suite.

PR #6911, opened 2019-09-25
IGNITE-12232 NPE while node join processing

SomeFire:


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -4937,6 +4941,17 @@ else if (spiState == CONNECTING)
                 sendMessageAcrossRing(msg);
         }
 
+        /**
+         * @param secCtx Sec context.
```
andrey-kuznetsov: Please use full word in parameter description.

SomeFire: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/security/InvalidServerTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.ignite.plugin.security.SecurityCredentials;
+import org.apache.ignite.spi.discovery.tcp.TcpDiscoverySpi;
+import org.apache.ignite.spi.discovery.tcp.messages.TcpDiscoveryAbstractMessage;
+import org.apache.ignite.spi.discovery.tcp.messages.TcpDiscoveryNodeAddedMessage;
+import org.junit.Test;
+
+import static org.apache.ignite.testframework.GridTestUtils.assertThrowsWithCause;
+
+/**
+ *
```
andrey-kuznetsov: What kind of invalidness does it have? This should be described in class name or in comment.

modules/core/src/test/java/org/apache/ignite/internal/processors/security/AbstractSecurityTest.java
```diff
@@ -35,6 +35,9 @@
     /** Empty array of permissions. */
     protected static final SecurityPermission[] EMPTY_PERMS = new SecurityPermission[0];
 
+    /** Global authentication. */
```
andrey-kuznetsov: ... sign (or flag).

modules/core/src/test/java/org/apache/ignite/internal/processors/security/impl/TestSecurityProcessor.java
```diff
@@ -46,29 +46,36 @@
  */
 public class TestSecurityProcessor extends GridProcessorAdapter implements GridSecurityProcessor {
     /** Permissions. */
-    private static final Map<SecurityCredentials, SecurityPermissionSet> PERMS = new ConcurrentHashMap<>();
+    public static final Map<SecurityCredentials, SecurityPermissionSet> PERMS = new ConcurrentHashMap<>();
```
andrey-kuznetsov: `public` can be reduced to default visibility.

modules/core/src/test/java/org/apache/ignite/internal/processors/security/AbstractSecurityTest.java
SomeFire: Done.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -2044,7 +2044,11 @@ private boolean recordable(TcpDiscoveryAbstractMessage msg) {
      * @param rmtPerms The second set of permissions.
      * @return {@code True} if given parameters contain the same permissions, {@code False} otherwise.
      */
-    private boolean permissionsEqual(SecurityPermissionSet locPerms, SecurityPermissionSet rmtPerms) {
+    private boolean permissionsEqual(@Nullable SecurityPermissionSet locPerms,
+        @Nullable SecurityPermissionSet rmtPerms) {
+        if (locPerms == null || rmtPerms == null)
```
alex-plekhanov: What if both `locPerms` and `rmtPerms` are `null`?

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -4737,7 +4741,7 @@ else if (!locNodeId.equals(node.id()) && ring.node(node.id()) != null) {
                                 spi.marshaller(), U.resolveClassLoader(spi.ignite().configuration()), node
                             );
 
-                            if (!permissionsEqual(coordSubj.subject().permissions(), subj.subject().permissions())) {
+                            if (!permissionsEqual(getPermissions(coordSubj), getPermissions(subj))) {
```
alex-plekhanov: Why we only fix one usage of this method? Are `null`'s possible in another branch?

SomeFire: No, other method usages contain null check.

alex-plekhanov: Where is that check? I see only usage of the same method `nodeSecurityContext` to get `coordSubj` and no null check after it.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
SomeFire: Both nulls means something wrong with configuration because it shouldn't be possible, because Ignite checks configuration consistency.

`null` `rmtPerms` means authenticator on current node doesn't authenticated node.
`null` `locPerms` means authenticator on joining node doesn't authenticated node. In this case node shouldn't start because of `null` in `localAuthentication`.

PR #6917, opened 2019-09-26
IGNITE-12189 Implementation correct limit for TextQuery

shuliga:


modules/core/src/main/java/org/apache/ignite/cache/query/TextQuery.java
```diff
@@ -79,6 +82,19 @@ public TextQuery(String type, String txt) {
         setText(txt);
     }
 
+    /**
+     * Constructs query for the given search string.
+     *
+     * @param type Type.
+     * @param txt Search string.
+     * @param limit Limit.
```
AMashenkov: This is a public API class.
Let's make javadoc more verbose for 'limit' param at least.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
```diff
@@ -327,9 +334,14 @@ protected void onNodeLeft(UUID evtNodeId) {
     protected void enqueue(Collection<?> col) {
         assert Thread.holdsLock(this);
 
-        queue.add((Collection<R>)col);
-
-        cnt.addAndGet(col.size());
+        if(limit <= 0 || total.addAndGet(col.size()) <= limit) {
```
AMashenkov: Is it possible to rewrite this with no adding 'total' field, but using CAS-loop or cnt.getAndUpdate() ?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryManager.java
```diff
@@ -1951,7 +1951,9 @@ public String cacheName() {
      */
     @GridInternal
     private static class MetadataJob implements IgniteCallable<Collection<CacheSqlMetadata>> {
-        /** */
+        /**
+         *
```
AMashenkov: Let's keep one-lined empty javadoc.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryManager.java
```diff
@@ -2751,19 +2754,19 @@ public int size() {
      * @return Created query.
      */
     public CacheQuery<Map.Entry<K, V>> createFullTextQuery(String clsName,
-        String search, boolean keepBinary) {
+        String search, int limit, boolean keepBinary) {
```
AMashenkov: Missed javadoc for 'limit' param

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheProxyImpl.java
```diff
@@ -520,13 +520,13 @@ public CountDownLatch getInitLatch(){
     }
 
     /**
-     * @param filter Filter.
+     * @param query Filter.
```
AMashenkov: Javadoc is incorrect.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryRequest.java
```diff
@@ -257,6 +260,7 @@ public GridCacheQueryRequest(
         GridCacheQueryType type,
         boolean fields,
         String clause,
+        int limit,
```
AMashenkov: Missed javadoc.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/cache/PlatformCache.java
```diff
@@ -1420,9 +1420,10 @@ private Query readTextQuery(BinaryRawReader reader) {
         boolean loc = reader.readBoolean();
         String txt = reader.readString();
         String typ = reader.readString();
+        final int limit = reader.readInt();
         final int pageSize = reader.readInt();
 
-        return new TextQuery(typ, txt).setPageSize(pageSize).setLocal(loc);
+        return new TextQuery(typ, txt, limit).setPageSize(pageSize).setLocal(loc);
```
AMashenkov: Platforms do not support 'limit' param yet. 
So, attempt to read it can lead an error. 

Let's set a default value here and create a separate ticket for the issue.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryIndexing.java
```diff
@@ -138,7 +138,7 @@ public long streamUpdateQuery(String schemaName, String qry, @Nullable Object[]
      * @throws IgniteCheckedException If failed.
      */
     public <K, V> GridCloseableIterator<IgniteBiTuple<K, V>> queryLocalText(String schemaName, String cacheName,
-        String qry, String typeName, IndexingQueryFilter filter) throws IgniteCheckedException;
+        String qry, String typeName, IndexingQueryFilter filter, int limit) throws IgniteCheckedException;
```
AMashenkov: Missed javadoc.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
```diff
@@ -2712,7 +2712,7 @@ public void remove(GridCacheContext cctx, CacheDataRow row)
      * @throws IgniteCheckedException If failed.
      */
     public <K, V> GridCloseableIterator<IgniteBiTuple<K, V>> queryText(final String cacheName, final String clause,
-        final String resType, final IndexingQueryFilter filters) throws IgniteCheckedException {
+        final String resType, final IndexingQueryFilter filters, int limit) throws IgniteCheckedException {
```
AMashenkov: Missed javadoc.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/opt/GridLuceneIndex.java
```diff
@@ -241,7 +241,7 @@ public void remove(CacheObject key) throws IgniteCheckedException {
      * @throws IgniteCheckedException If failed.
      */
     public <K, V> GridCloseableIterator<IgniteBiTuple<K, V>> query(String qry,
-        IndexingQueryFilter filters) throws IgniteCheckedException {
+        IndexingQueryFilter filters, int limit) throws IgniteCheckedException {
```
AMashenkov: Missed javadoc.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQuerySelfTest.java
```diff
@@ -212,86 +255,84 @@ private static void clearCache(IgniteEx ignite) {
...
 
-                checkForMissedKeys(ignite, exp0, all);
+                assertResult(ignite, qry, testPair);
             }
+        }
+    }
 
-            try (QueryCursor<Cache.Entry<Integer, Person>> cursor = cache.query(qry)) {
-                Set<Integer> exp0 = new HashSet<>(exp);
+    private static void assertResult(IgniteEx ignite, TextQuery qry,
```
AMashenkov: Missed javadoc

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQuerySelfTest.java
```diff
@@ -212,86 +255,84 @@ private static void clearCache(IgniteEx ignite) {
...
+        TestPair testPair) throws IgniteCheckedException {
+        if (qry.getLimit() > 0){
+            assertTrue(testPair.all.size() <= QUERY_LIMIT);
+        } else {
+            checkForMissedKeys(ignite, testPair.expected, testPair.all);
+        }
+    }
 
-                List<Cache.Entry<Integer, ?>> all = new ArrayList<>();
+    @NotNull private static GridCacheFullTextQuerySelfTest.TestPair processExpectedWithBinary(Set<Integer> exp,
```
AMashenkov: Missed javadoc.
Method modificators order.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQuerySelfTest.java
```diff
@@ -212,86 +255,84 @@ private static void clearCache(IgniteEx ignite) {
...
-                    exp0.remove(entry.getKey());
-                }
+            testPair.expected.remove(entry.getKey());
+        }
+        return testPair;
+    }
 
-                checkForMissedKeys(ignite, exp0, all);
-            }
+    @NotNull private static GridCacheFullTextQuerySelfTest.TestPair processExpected(Set<Integer> exp,
```
AMashenkov: Missed javadoc.
Method modificators order.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQuerySelfTest.java
```diff
@@ -212,86 +255,84 @@ private static void clearCache(IgniteEx ignite) {
...
             }
+        }
+    }
 
-            try (QueryCursor<Cache.Entry<Integer, Person>> cursor = cache.query(qry)) {
-                Set<Integer> exp0 = new HashSet<>(exp);
+    private static void assertResult(IgniteEx ignite, TextQuery qry,
+        TestPair testPair) throws IgniteCheckedException {
+        if (qry.getLimit() > 0){
+            assertTrue(testPair.all.size() <= QUERY_LIMIT);
```
AMashenkov: Braces for one-liners are redundant.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
shuliga: 'total' counter was added because 'cnt' used here as queue size indicator and is incremented on inserts and decremented when item is consumed. So we cannot rely on it as on total counter.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/cache/PlatformCache.java
```diff
@@ -1422,6 +1422,11 @@ private Query readTextQuery(BinaryRawReader reader) {
         String typ = reader.readString();
         final int pageSize = reader.readInt();
 
+        //TODO uncomment when limit parameter is added to Platforms
```
AMashenkov: Let's either create a ticket for this or remove commented code.

https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines#CodingGuidelines-TODOs

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQuerySelfTest.java
```diff
@@ -59,6 +59,19 @@
...
+
+    private static final class TestPair{
+
+        public final Set<Integer> expected;
+        public final List<Cache.Entry<Integer, ?>> all = new ArrayList<>();
+
+        public TestPair(Set<Integer> exp) {
+            this.expected = new HashSet<>(exp);
+        }
+    }
```
AMashenkov: This class also should have minimal javadoc.

https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines#CodingGuidelines-JavadocComments

shuliga:
@AMashenkov in order to access Apache Ignite Teamcity Bot, TC credentials are required; and i do not have them yet.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQuerySelfTest.java
AMashenkov: Should be:
```
if (qry.getLimit() > 0)
    assertTrue(testPair.all.size() <= QUERY_LIMIT);
else 
    checkForMissedKeys(ignite, testPair.expected, testPair.all);
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQuerySelfTest.java
```diff
@@ -59,6 +59,22 @@
     /** Cache name */
     private static final String PERSON_CACHE = "Person";
 
+    /** Limitation to query response size */
+    private static final int QUERY_LIMIT = 5;
+
+    /**
+     * Container for expected values and all available entries
+     */
+    private static final class TestPair{
```
AMashenkov: Missed space before '{'

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
```diff
@@ -327,9 +331,15 @@ protected void onNodeLeft(UUID evtNodeId) {
...
 
-        queue.add((Collection<R>)col);
-
-        cnt.addAndGet(col.size());
+        if(limitCnt <= 0 || limitCnt >= col.size()) {
+            queue.add((Collection<R>)col);
+            cnt.addAndGet(col.size());
+            limitCnt -= col.size();
+        } else {
+            int toAdd = limitCnt;
```
AMashenkov: limitCnt should be decreased in 'else' case as well

shuliga: Formally - yes, but this case will be the last pass under such condition and whole function will not be called any more.

AMashenkov: This is true for local query, but isn't for distributed
.
enqueue() method is called on every page received from remote node.

Assume we receive 2 pages from 2 nodes and try to enqueue them concurrently.
Both pages will be trimmed and added to queue, but the only first is expected to be added.

shuliga: I this case let us set 'limitCnt'=0 as soon as we have first node's portion that fulfills our limit. But arrived data from 2nd node still makes enqueue() to be called one more time. For 'limitCnt' =0 the first condition will fire as if it is no-limit case (limit<=0 is treated as no limit) what is not good specially here. So we need to add kind a 'limitReached' flag, set it in 'else' block and check at the method entry.

AMashenkov: I've made a PR with the only one additional field
https://github.com/apache/ignite/pull/6997/

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
```diff
@@ -327,9 +331,15 @@ protected void onNodeLeft(UUID evtNodeId) {
...
+            queue.add(new ArrayList(col).subList(0, toAdd));
```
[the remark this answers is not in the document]
shuliga: I used `new ArrayList()` because array size here will always be smaller then `pageSize`. Usually it's around 1000 (please correct me if no). This is small array and there should not be performance impact. Working with streams seems to be more resource consuming in this case.
Please consider what should match here, I will update the code.

modules/core/src/main/java/org/apache/ignite/cache/query/TextQuery.java
```diff
@@ -79,6 +82,19 @@ public TextQuery(String type, String txt) {
...
     }
 
+    /**
+     * Constructs query for the given search string.
+     *
+     * @param type Type.
+     * @param txt Search string.
+     * @param limit Limits response records count. If 0 or less, the limit considered to be Integer.MAX_VALUE, that is virtually no limit.
+     */
+    public TextQuery(String type, String txt, int limit) {
```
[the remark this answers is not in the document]
shuliga: 'Limit' parameter is important part of Queries in Lucene text search. This is because responses usually contain many records and only certain relevant top of them interests. That's why limits are used in most cases when working with Lucene. This differs TextQuery from SQL ones.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryAdapter.java
```diff
@@ -234,6 +234,7 @@ public GridCacheQueryAdapter(
      * @param part Partition.
      * @param clsName Class name.
      * @param clause Clause.
+     * @param limit Response limit. Set to -1 for no limits.
```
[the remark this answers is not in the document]
shuliga: Correct, we a going with 0 as no limit.
@AMashenkov , please, cofirm.  We have to end-up this part of review at last.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
```diff
@@ -330,15 +333,18 @@ protected void onNodeLeft(UUID evtNodeId) {
...
-        if(limitCnt <= 0 || limitCnt >= col.size()) {
-            queue.add((Collection<R>)col);
-            cnt.addAndGet(col.size());
-            limitCnt -= col.size();
-        } else {
-            int toAdd = limitCnt;
-            queue.add(new ArrayList(col).subList(0, toAdd));
-            cnt.addAndGet(toAdd);
+        if(!limitReached){
+            if(capacity <= 0 || capacity >= col.size()) {
```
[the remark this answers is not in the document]
shuliga: A really good finding!
Applying your fix snippet in the code.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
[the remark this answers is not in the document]
shuliga: Don't think that we will deal with huge arrays here ever. So why to add code complexity?)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
shuliga: Concurrent test has been added

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
```diff
@@ -60,37 +59,65 @@
     /** Logger. */
     protected static IgniteLogger log;
 
-    /** */
+    /**
```
pavlukhin: We permit _single-line_ javadoc comments for fields. Please use them (in all changed places in this PR).

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
```diff
@@ -325,17 +354,34 @@ protected void onNodeLeft(UUID evtNodeId) {
      * @param col Collection.
      */
     protected void enqueue(Collection<?> col) {
+
```
pavlukhin: Redundant empty line. Can lead to automatic style check failure.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
```diff
@@ -2821,11 +2821,12 @@ public void remove(GridCacheContext cctx, CacheDataRow row)
      * @param filters Key and value filters.
      * @param <K> Key type.
      * @param <V> Value type.
+     * @param limit Limits response records count. If 0 or less, the limit considered to be Integer.MAX_VALUE, that is virtually no limit.
```
pavlukhin: Clarification is needed why it is no more than `Integer.MAX_VALUE`? Cannot we have more with `GridCloseableIterator`?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQueryMultithreadedSelfTest.java
```diff
@@ -76,6 +76,7 @@ public void testH2Text() throws Exception {
         final int keyCnt = 5000;
         final int logFreq = 50;
         final String txt = "Value";
+        final int limit = 1000;
```
pavlukhin: Not sure that we should change the behavior in old test. What are the arguments to do it?

modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
pavlukhin: Ah, it seems that we use Lucene in that way. How does Lucene interpret `Integer.MAX_VALUE`? Can it return more than `Integer.MAX_VALUE`?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQuerySelfTest.java
```diff
@@ -59,6 +63,26 @@
...
+    private static final int QUERY_LIMIT = 5;
+
+    /** Concurrent threads number */
+    private static final int N_THREADS = 20;
+
+    /**
+     * Container for expected values and all available entries
+     */
+    private static final class TestPair {
+
```
pavlukhin: Please remove this empty line and add javadocs to class members (automated checks require it).

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQuerySelfTest.java
```diff
@@ -120,28 +160,87 @@ public void testTextQueryWithKeepBinary() throws Exception {
...
+    /**
+     * Creates Runnable for TextQuery
+     *
+     * @param ignite Ignite insance.
+     * @param clause Query clause.
+     * @param exp Expected results for validation.
+     * @return TextQuery and validation wrapped into Runnable functional interface.
+     */
+
```
pavlukhin: Extra empty line.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQuerySelfTest.java
```diff
@@ -120,28 +160,87 @@ public void testTextQueryWithKeepBinary() throws Exception {
...
+    @NotNull private Runnable textQueryTask(IgniteEx ignite, String clause, Set<Integer> exp) {
+        return () -> {
+            try {
+                TextQuery qry = new TextQuery<>(Person.class, clause).setLocal(false);
```
pavlukhin: It seems no need to call `setLocal(false)` as it is a default value.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQuerySelfTest.java
```diff
@@ -212,86 +311,109 @@ private static void clearCache(IgniteEx ignite) {
...
+     *
+     * @param ignite Ignite context.
+     * @param qry Initial text query.
+     * @param testPair pair containing expected and all entries.
+     * @throws IgniteCheckedException if key check failed.
+     */
+    private static void assertResult(IgniteEx ignite, TextQuery qry,
+        TestPair testPair) throws IgniteCheckedException {
+        if (qry.getLimit() > 0) {
+            assertTrue(testPair.all.size() <= QUERY_LIMIT);
```
pavlukhin: Should not we `checkForMissedKeys` here as well? Checking that we got lesser than limit seems insufficient. By the way are there really more matching items than specified limit?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQuerySelfTest.java
```diff
@@ -212,86 +311,109 @@ private static void clearCache(IgniteEx ignite) {
...
+        }
+        else {
```
pavlukhin: According to [code style](https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines#CodingGuidelines-BracesandIdentation) braces should not be used for single-line statements after _if_ and _else_.

pavlukhin:
Regarding `List.sublist`. Generally we should do a benchmark to check a performance impact. On the other hand we can extract a method `firstElements` doing a needed trick. Does it sound as over-complication?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQueryMultithreadedSelfTest.java
shuliga: Yes, there's a reason not to set limits here, as we have created special tests for this.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryAdapter.java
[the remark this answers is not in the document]
shuliga: No comments for a long time. Setting to 0 as no-limit value.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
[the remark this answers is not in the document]
pavlukhin: Unfortunately it is a common practice in Ignite. We do not have cheap performance measurement feedback loop, consequently we are trying to use faster approaches but possibly sacrificing code simplicity. There were a lot of cases from a real practice where a single "innocent" code line (not lock and not system call) caused visible performance drop.

Also I can thing about another option. Apparently a collection coming to `enqueue` method is always a `List`. We can try to do a refactoring and change an argument type to `List`. Then `List.sublist` method could be used directly. Possibly, it deserves a separate ticket. What do you think?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheFullTextQuerySelfTest.java
```diff
@@ -120,28 +165,86 @@ public void testTextQueryWithKeepBinary() throws Exception {
...
+        final IgniteEx ignite = grid(0);
+
+        final String clause = "1*";
+
+        // 1. Populate cache with data, calculating expected count in parallel.
+        Set<Integer> exp = populateCache(ignite, false, MAX_ITEM_COUNT, (IgnitePredicate<Integer>)x -> String.valueOf(x).startsWith("1"));
+
+        ExecutorService executor = Executors.newFixedThreadPool(N_THREADS);
+
+        IntStream.range(0, N_THREADS).forEach((i) -> executor.submit(textQueryTask(ignite, clause, exp)));
```
pavlukhin: It seems errors from submitted tasks do not propagate to main test thread. Consequently failures in tasks will not fail the test. I can suggest here to use `org.apache.ignite.testframework.GridTestUtils#runMultiThreaded(java.lang.Runnable, int, java.lang.String)` or someting similar to `java.util.concurrent.ExecutorService#invokeAll(java.util.Collection<? extends java.util.concurrent.Callable<T>>)` (and waiting on futures).

pavlukhin:
Once again about `List.sublist`.

Unfortunately it is a common practice in Ignite. We do not have cheap performance measurement feedback loop, consequently we are trying to use faster approaches but possibly sacrificing code simplicity. There were a lot of cases from a real practice where a single "innocent" code line (not lock and not system call) caused visible performance drop.

Also I can thing about another option. Apparently a collection coming to enqueue method is always a List. We can try to do a refactoring and change an argument type to List. Then List.sublist method could be used directly. Possibly, it deserves a separate ticket. What do you think?

shuliga:
Regarding `List.subList` in `enqueue()` method.
Your point seems reasonable; but conversion of  `Collection` into `List` touches methods `onPage()` and `onPageReady()`; there is a lot of reworking here. It should be definitely a separate ticket.
Also please keep in mind that `subList()` is taken only for last page of data and its size is always less then `pageSize`. This might never be a performance issue point.

PR #6927, opened 2019-09-30
IGNITE-12212: Cluster nodes system view.

nizhikov:


modules/core/src/test/java/org/apache/ignite/internal/metric/SystemViewSelfTest.java
```diff
@@ -461,6 +464,63 @@ public void testClientsConnections() throws Exception {
         }
     }
 
+    /** */
+    @Test
+    public void testNodes() throws Exception {
+        try(IgniteEx g1 = startGrid(0)) {
+            SystemView<ClusterNodeView> nodes = g1.context().systemView().view(NODES_SYS_VIEW);
```
dgarus: rename nodes -> views

modules/core/src/test/java/org/apache/ignite/internal/metric/SystemViewSelfTest.java
```diff
@@ -461,6 +464,63 @@ public void testClientsConnections() throws Exception {
...
+                assertEquals(2, nodes2.size());
+
+                for (ClusterNodeView node : nodes2) {
+                    if(node.nodeId().equals(g2.localNode().id()))
+                        checkNodeView(node, g2.localNode(), true);
+                    else
+                        checkNodeView(node, g1.localNode(), false);
+                }
+            }
+        }
```
dgarus: we should check that views size is 0

nizhikov: View size can't be 0, because we have alive node.
`assertEquals(1, views.size());` added.

PR #6928, opened 2019-10-01
IGNITE-12250 Adds rest requests processing logging.

ololo3000:


modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
```diff
@@ -150,7 +150,11 @@
         }
 
         @Override public IgniteInternalFuture<GridRestResponse> handleAsync(GridRestRequest req) {
-            return handleAsync0(req);
+            IgniteInternalFuture<GridRestResponse> res = handleAsync0(req);
+
+            res.listen(fut -> logRestRequestResult(req, fut));
```
nizhikov: Can we use closure here?

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
```diff
@@ -1039,6 +1043,30 @@ private void startProtocol(GridRestProtocol proto) throws IgniteCheckedException
...
+        IgniteCheckedException reqProcErr = null;
+
+        try {
+            resp = fut.get();
+        }
+        catch (IgniteCheckedException e) {
+            reqProcErr = e;
+        }
+
+        log.info(String.format("REST request handled [req=%s, resp=%s, err=%s].", req, resp, reqProcErr));
```
nizhikov: Why do we need `info` level here? I think every parameter of every request should be on `debug` or even `trace` level.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
ololo3000: No, we can't. We need GridRestRequest to be logged and IgniteInternalFuture result doesn't provide it.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
ololo3000: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
nizhikov:
```
res.listen(fuct -> {
});
```

Why this wouldn't work?

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
```diff
@@ -1039,6 +1043,30 @@ private void startProtocol(GridRestProtocol proto) throws IgniteCheckedException
...
     }
 
+    /**
+     * Logs REST request processing result.
+     *
+     * @param req Received REST request.
+     * @param fut Result of REST request processing.
+     */
+    private void logRestRequestResult(GridRestRequest req, IgniteInternalFuture<GridRestResponse> fut) {
+        if (log == null || !log.isDebugEnabled())
```
nizhikov: When log may be null?

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
```diff
@@ -1039,6 +1043,30 @@ private void startProtocol(GridRestProtocol proto) throws IgniteCheckedException
...
+            return;
+
+        GridRestResponse resp = null;
+
+        IgniteCheckedException reqProcErr = null;
+
+        try {
+            resp = fut.get();
```
nizhikov: Can we use `fut.error() == null` here or `fut.isCancelled()` and not use try catch logic?

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
nizhikov: Seems, we can do not register listener in case debug disabled.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
ololo3000: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
ololo3000: My mistake. Ignite log4j logger is used if no specified. Done.

PR #6929, opened 2019-10-01
IGNITE-12181 Fixed assertion for non-persisted group in PDS enabled cluster.

xtern:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/IgnitePdsAtomicCacheHistoricalRebalancingTest.java
```diff
@@ -61,4 +63,11 @@
         if (!walRebalanceInvoked)
             throw new AssertionError("WAL rebalance hasn't been invoked.");
     }
+
+    /** {@inheritDoc} */
+    @Test
+    @Ignore
+    @Override public void testTopologyChangesWithConstantLoadOnInMemoryCache() {
+        // No-op.
+    }
```
Mmuzaf: I think it is not good for adding new ignored tests. Why we should do so?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/IgnitePdsTxHistoricalRebalancingTest.java
```diff
@@ -61,4 +63,11 @@
         if (!walRebalanceInvoked)
             throw new AssertionError("WAL rebalance hasn't been invoked.");
     }
+
+    /** {@inheritDoc} */
+    @Test
+    @Ignore
+    @Override public void testTopologyChangesWithConstantLoadOnInMemoryCache() {
+        // No-op.
+    }
```
Mmuzaf: Why we should do so?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/IgnitePdsAtomicCacheHistoricalRebalancingTest.java
xtern: Because testTopologyChangesWithConstantLoad is the best that I found to check complex rebalancing with indexing, but as you may know, historical (WAL) rebalancing doesn't support for in-memory caches, so test case with in-memory cache should be ignored somehow.

xtern: I've updated PR, added a new test for mixed configurations of the data regions.

PR #6934, opened 2019-10-04
IGNITE-12254 IO errors during write header of WAL files in FSYNC mode…

alex-plekhanov:
… should be handled by failure handler


nizhikov:
LGTM

PR #6936, opened 2019-10-04
IGNITE-11230 Fix Javadoc for DataStorageConfiguration

shishkovilja:


modules/core/src/main/java/org/apache/ignite/configuration/DataStorageConfiguration.java
```diff
@@ -473,7 +482,9 @@ public DataStorageConfiguration setStoragePath(String persistenceStorePath) {
 
     /**
      * Gets checkpoint frequency.
+     * Default value {@link #DFLT_CHECKPOINT_FREQ} .
```
NSAmelchev: unnecessary space

modules/core/src/main/java/org/apache/ignite/configuration/DataStorageConfiguration.java
```diff
@@ -308,7 +308,9 @@ public DataStorageConfiguration() {
 
     /**
      * Initial size of a data region reserved for system cache.
+     * Default value is {@link #DFLT_SYS_REG_INIT_SIZE}.
      *
+     * @see #setSystemRegionInitialSize(long).
```
NSAmelchev: It seems that it’s not customary to write a dot after the see link

shishkovilja: Hi, @NSAmelchev ! I removed excess dots after JavaDoc see links and unnecessary whitespace in 485th line. Thank you!

PR #6937, opened 2019-10-04
IGNITE-12186 TDE - Phase-2. Master key rotation.

NSAmelchev:
Implementation of master key change. [See design.](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=95652381)

Introduced new public interface `IgniteEncryption` with methods:
- `String getMasterKeyName()`
- `void changeMasterKey(String masterKeyName)`

Changed public API of `EncryptionSpi`. New methods are:
- `String getMasterKeyName()`
- `void setMasterKeyName(String masterKeyName)`
The behavior of the following methods depends on the master key id set:
- `byte[] encryptKey(Serializable key)`
- `Serializable decryptKey(byte[] key)`
- `byte[] masterKeyDigest()`

The process can be started on a server node. Clients and daemons can't get or change the current cluster master key name.

Each master key change creates a logical WAL record (`MasterKeyChangeRecord`). The node must apply these records before starting memory restore to able decrypt encrypted cache WAL records. In PR it happens during matastore recovery.

If a node was unavailable during a master key change process it won't be able to join to cluster with old the master key id. New system property implemented to re-encrypt cache keys on startup and join to cluster with valid master key name.

Cache start and node join will be rejected during the master key change process.


modules/core/src/main/java/org/apache/ignite/spi/encryption/keystore/KeystoreEncryptionSpi.java
```diff
@@ -483,19 +469,54 @@ public void setKeySize(int keySize) {
...
-        this.masterKeyName = masterKeyName;
+        masterKey.set1(masterKeyName);
+    }
+
+    /**
+     * Sets master key.
+     *
+     * @param masterKeyId Master key id.
+     */
+    private void setMasterKey(String masterKeyId) {
```
nizhikov: Let's rename it to the `loadMasterKey`

modules/core/src/main/java/org/apache/ignite/spi/encryption/keystore/KeystoreEncryptionSpi.java
```diff
@@ -113,24 +113,13 @@
...
-     * Master key name.
-     */
-    private String masterKeyName = DEFAULT_MASTER_KEY_NAME;
-
-    /**
-     * Master key.
-     */
-    private KeystoreEncryptionKey masterKey;
+    /** Tuple of the master key id and the master key. */
+    private volatile T2<String, KeystoreEncryptionKey> masterKey = new T2<>(DEFAULT_MASTER_KEY_NAME, null);
```
nizhikov: Not clear to me.

Why do we need volatile guarantees here?

modules/core/src/main/java/org/apache/ignite/spi/encryption/keystore/KeystoreEncryptionSpi.java
```diff
@@ -332,6 +302,22 @@ private void doEncryption(ByteBuffer data, Cipher cipher, Serializable key, Byte
         return BLOCK_SZ;
     }
 
+    /** {@inheritDoc} */
+    @Override public String getMasterKeyId() {
```
nizhikov: Why do we need new method `getMasterKeyId` instead of `getMasterKeyName`?
Let's keep existing method and not deprecate it.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,388 @@
...
+
+            fut.listen(f -> {
+                if (f.error() != null)
+                    p.resFut.onDone(f.error());
+                else
+                    p.resFut.onDone(f.result());
+
+                ClusterNode crdNode = coordinator();
+
+                if (crdNode != null && !ctx.clientNode())
```
nizhikov: `!ctx.clientNode()` - not clear for me.
Why we execute process on client node but don't want to know results?

nizhikov: Why we just ignore the fact that coordinator unknown?
We should print this in log, at least.
But I think we should throw an exception in that case.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,388 @@
...
+        });
+
+        ctx.event().addDiscoveryEventListener((evt, discoCache) -> {
+            UUID leftNodeId = evt.eventNode().id();
+
+            for (Process p : processes.values()) {
+                p.initFut.listen(fut -> {
+                    boolean crdChanged = F.eq(leftNodeId, p.crdId);
+
+                    if (crdChanged) {
```
nizhikov: 1. `crdChanged` variable can be omitted.
2. `F.eq` and `localNodeId().equals` - let's use the same technique to check equality.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,388 @@
...
+                        ClusterNode crd = coordinator();
+
+                        if (crd != null) {
```
nizhikov: We shouldn't ignore the fact that coordinator is null.
Let's rewrite this as

```
if (coordinator == null) {
    onAllServerNodesLeft();
    return;
}

// main code here
```

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,388 @@
...
+        }
+    }
+
+    /**
+     * Sends single node message to coordinator.
+     *
+     * @param p Process.
+     * @param crd Coordinator node to send message.
+     */
+    private void sendSingleSingleMessage(Process p, ClusterNode crd) {
```
nizhikov: nit: word `single` repeats two times

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,388 @@
...
+            (Exception)p.resFut.error());
+
+        if (crd.isLocal())
+            onSingleNodeMessageReceived(singleMsg, crd.id());
+        else {
+            try {
+                ctx.io().sendToGridTopic(crd, GridTopic.TOPIC_DISTRIBUTED_PROCESS, singleMsg, SYSTEM_POOL);
+            }
+            catch (IgniteCheckedException e) {
+                log.warning("Unable to send message.", e);
```
nizhikov: We can't ignore this exception.
Otherwise, the process can never end.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/FullMessage.java
```diff
@@ -0,0 +1,119 @@
...
+    private static final long serialVersionUID = 0L;
+
+    /** Custom message ID. */
+    private final IgniteUuid id = IgniteUuid.randomUuid();
+
+    /** Process id. */
+    private final UUID processId;
+
+    /** Process type. */
+    private final DistributedProcessType type;
```
nizhikov: How this will be serialized?
We should transfer int by the wire.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/InitMessage.java
```diff
@@ -0,0 +1,108 @@
...
+    private static final long serialVersionUID = 0L;
+
+    /** Custom message ID. */
+    private final IgniteUuid id = IgniteUuid.randomUuid();
+
+    /** Process id. */
+    private final UUID processId;
+
+    /** Process type. */
+    private final DistributedProcessType type;
```
nizhikov: How this will be serialized?
We should transfer int by the wire.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/SingleNodeMessage.java
```diff
@@ -0,0 +1,193 @@
...
+    private static final long serialVersionUID = 0L;
+
+    /** Initial channel message type (value is {@code 177}). */
+    public static final short TYPE_CODE = 177;
+
+    /** Process id. */
+    private UUID processId;
+
+    /** Process type. */
+    private DistributedProcessType type;
```
nizhikov: How this will be serialized?
We should transfer int by the wire.

modules/core/src/main/java/org/apache/ignite/spi/encryption/keystore/KeystoreEncryptionSpi.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/spi/encryption/keystore/KeystoreEncryptionSpi.java
NSAmelchev: I have implemented two variables.
But synchronization not needed because `EncryptionManager` guarantee that no one key encryption method will be called during the `setMasterKeyName` executes.

modules/core/src/main/java/org/apache/ignite/spi/encryption/keystore/KeystoreEncryptionSpi.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1284,6 +1284,16 @@
...
     public static final String IGNITE_PAGES_LIST_DISABLE_ONHEAP_CACHING = "IGNITE_PAGES_LIST_DISABLE_ONHEAP_CACHING";
 
+    /**
+     * The master key name that node will use during the recovery procedure.
+     * <p>
+     * If a node was unavailable during a master key change process it won't be able to join to cluster with old the
+     * master key name. Set up this property to re-encrypt cache keys on startup and join to cluster with valid master
+     * key name.
+     */
+    public static final String IGNITE_MASTER_KEY_NAME_TO_CHANGE_ON_STARTUP =
```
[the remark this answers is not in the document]
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -1190,6 +1190,10 @@ public void start(
 
             ctx.add(discoMgr, false);
 
+            // Start the encryption manager after assigning the discovery manager to context, so it will be
```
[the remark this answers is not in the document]
NSAmelchev: We should change the start order because of `EncryptionManager` should listen to custom discovery events. See `GridEncryptionManager.start()` and constructor of `DistributedProcess`.

modules/core/src/main/java/org/apache/ignite/internal/commandline/encryption/EncryptionCommand.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.ignite.internal.visor.encryption.VisorEncryptionTaskResult;
+
+import static org.apache.ignite.internal.commandline.CommandList.ENCRYPTION;
+import static org.apache.ignite.internal.commandline.TaskExecutor.executeTask;
+import static org.apache.ignite.internal.commandline.encryption.EncryptionSubcommand.CHANGE_MASTER_KEY;
+import static org.apache.ignite.internal.commandline.encryption.EncryptionSubcommand.GET_MASTER_KEY;
+import static org.apache.ignite.internal.commandline.encryption.EncryptionSubcommand.of;
+
+/**
+ * Commands assosiated with encryption functionality.
```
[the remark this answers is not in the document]
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/MasterKeyChangeRecord.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.Map;
+
+import static org.apache.ignite.internal.pagemem.wal.record.WALRecord.RecordType.MASTER_KEY_CHANGE_RECORD;
+
+/**
+ * Logical data record indented for master key change action.
+ */
+public class MasterKeyChangeRecord extends WALRecord {
+    /** */
```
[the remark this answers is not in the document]
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/MasterKeyChangeRecord.java
```diff
@@ -0,0 +1,73 @@
...
+    private final String masterKeyName;
+
+    /** */
```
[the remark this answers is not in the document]
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ClusterCachesInfo.java
```diff
@@ -958,7 +961,23 @@ private boolean processStartNewCacheRequest(
...
 
-        CacheConfiguration<?, ?> ccfg = req.startCacheConfiguration();
+        GridEncryptionManager encMgr = ctx.encryption();
+
+        if (ccfg.isEncryptionEnabled() && (encMgr.isMasterKeyChangeInProgress() ||
+            (encMgr.lastChangedMasterKeyDigest() != null &&
+                !Arrays.equals(encMgr.lastChangedMasterKeyDigest(), req.masterKeyDigest())))) {
+            U.warn(log, "Ignore cache start request during the master key change process.");
+
+            IgniteCheckedException error = new IgniteCheckedException("Cache start during the master key change " +
```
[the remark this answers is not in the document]
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ClusterCachesInfo.java
```diff
@@ -930,7 +932,9 @@ private boolean processStartNewCacheRequest(
         DynamicCacheChangeRequest req,
         String cacheName
     ) {
-        String conflictErr = checkCacheConflict(req.startCacheConfiguration());
+        CacheConfiguration<?, ?> ccfg = req.startCacheConfiguration();
```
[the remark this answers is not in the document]
NSAmelchev: Reverted.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -783,7 +786,10 @@ private void readMetastore() throws IgniteCheckedException {
 
                 metaStorage = createMetastorage(true);
 
-                applyLogicalUpdates(status, onlyMetastorageGroup(), onlyMetastorageRecords(), false);
+                boolean encryptionDisabled = cctx.gridConfig().getEncryptionSpi() instanceof NoopEncryptionSpi;
```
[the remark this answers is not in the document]
NSAmelchev: I have implemented without it.

modules/core/src/main/java/org/apache/ignite/mxbean/EncryptionMXBean.java
```diff
@@ -0,0 +1,46 @@
...
+    public String getMasterKeyName();
+
+    /**
+     * Starts master key change process.
+     *
+     * @param masterKeyName Master key name.
+     *
+     * @see IgniteEncryption#changeMasterKey(String)
+     */
+    @MXBeanDescription("Change master key name.")
```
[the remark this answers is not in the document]
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/spi/encryption/keystore/KeystoreEncryptionSpi.java
```diff
@@ -332,6 +302,22 @@ private void doEncryption(ByteBuffer data, Cipher cipher, Serializable key, Byte
...
     }
 
+    /** {@inheritDoc} */
+    @Override public String getMasterKeyName() {
+        return masterKey.get1();
+    }
+
+    /** {@inheritDoc} */
+    @Override public void setMasterKeyName(String masterKeyName) {
+        if (!started()) {
```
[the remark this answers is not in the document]
NSAmelchev: Done.

modules/core/src/test/java/org/apache/ignite/spi/encryption/KeystoreEncryptionSpiSelfTest.java
```diff
@@ -28,25 +29,27 @@
...
+import static org.apache.ignite.internal.encryption.AbstractEncryptionTest.MASTER_KEY_NAME_2;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
 /** */
 public class KeystoreEncryptionSpiSelfTest {
-    /** @throws Exception If failed. */
+    /** */
```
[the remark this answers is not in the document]
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,388 @@
...
+     */
+    public DistributedProcess(GridKernalContext ctx, DistributedProcessType type,
+        Function<I, IgniteInternalFuture<R>> exec,
+        CI3<UUID, Map<UUID, R>, Map<UUID, Exception>> finish) {
+        this.ctx = ctx;
+        this.type = type;
+
+        log = ctx.log(getClass());
+
+        ctx.discovery().setCustomEventListener(InitMessage.class, (topVer, snd, msg) -> {
```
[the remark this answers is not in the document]
NSAmelchev: Events in the `ctx.event().addDiscoveryEventListener` will be processed by `disco-event-worker`. Events in the `ctx.discovery().setCustomEventListener` will be processed by `disco-notifier-worker`.

We should be able to reject a cache start during a master key change to encrypt group keys using the relevant master key. 

For these purposes I have implemented the `isMasterKeyChangeInProgress` flag. A node checks this flag when process a cache start. So it should be updated in the `disco-notifier-worker` as cache start processed there too.

Otherwise, nodes can inconsistency reject starting the cache and it will lead to a hang of the exchange.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,388 @@
...
+            // May be completed in case of double delivering.
+            if (p.initFut.isDone())
+                return;
+
+            ClusterNode crd = coordinator();
+
+            if (crd == null) {
+                p.initFut.onDone();
+
+                onAllServersLeft();
```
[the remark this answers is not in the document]
NSAmelchev: Clients do not participate in the aggregation of single messages. This situation can be only in the client disconnected case. I think it should not throws esception.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,388 @@
...
+     * Initiates process coordinator.
+     *
+     * @param p Process.
+     * @param topVer Topology version.
+     */
+    private void initCoordinator(Process p, AffinityTopologyVersion topVer) {
+        Set<UUID> aliveSrvNodesIds = new HashSet<>();
+
+        for (ClusterNode node : ctx.discovery().serverNodes(topVer)) {
+            if (ctx.discovery().alive(node))
```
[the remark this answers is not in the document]
NSAmelchev: Done. i have add a node fails test.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
NSAmelchev: Clients do not participate in the aggregation of single messages. Coordinator do not wait for client responses. So we should not send it to a coordinator in the current implementation.

NSAmelchev: I left a check on the client node only. Coordinator can not be null on a server node.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
NSAmelchev: Done. I have added failure handler call.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/FullMessage.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/InitMessage.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/SingleNodeMessage.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,388 @@
...
+     */
+    public DistributedProcess(GridKernalContext ctx, DistributedProcessType type,
+        Function<I, IgniteInternalFuture<R>> exec,
+        CI3<UUID, Map<UUID, R>, Map<UUID, Exception>> finish) {
+        this.ctx = ctx;
+        this.type = type;
+
+        log = ctx.log(getClass());
+
+        ctx.discovery().setCustomEventListener(InitMessage.class, (topVer, snd, msg) -> {
```
Mmuzaf: I see that each `DistributedProcess` adds a new event listener but I haven't found when it removes the event listener on its end. Is it correct?

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,388 @@
...
+        ctx.io().addMessageListener(GridTopic.TOPIC_DISTRIBUTED_PROCESS, (nodeId, msg0, plc) -> {
+            if (msg0 instanceof SingleNodeMessage && ((SingleNodeMessage)msg0).type() == type.ordinal()) {
+                SingleNodeMessage<R> msg = (SingleNodeMessage<R>)msg0;
+
+                if (msg.type() == type.ordinal())
+                    onSingleNodeMessageReceived(msg, nodeId);
+            }
+        });
+
+        ctx.event().addDiscoveryEventListener((evt, discoCache) -> {
```
Mmuzaf: Same as `ctx.discovery().setCustomEventListener`. A listener added but not removed. Is it correct?

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,388 @@
...
+     * @param p Process.
+     * @param crd Coordinator node to send message.
+     */
+    private void sendSingleMessage(Process p, ClusterNode crd) {
+        assert p.resFut.isDone();
+
+        SingleNodeMessage<R> singleMsg = new SingleNodeMessage<>(p.id, type.ordinal(), p.resFut.result(),
+            (Exception)p.resFut.error());
+
+        if (crd.isLocal())
```
Mmuzaf: According to the `coordinator()` method `crd` can be `null`. Will we get `NullPointerException` here?

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,388 @@
...
+        }
+    }
+
+    /**
+     * Sends single node message to coordinator.
+     *
+     * @param p Process.
+     * @param crd Coordinator node to send message.
+     */
+    private void sendSingleMessage(Process p, ClusterNode crd) {
```
Mmuzaf: Do we really need `crd` here since the `Process` already has its `crdId`?

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,388 @@
...
+        return U.oldest(ctx.discovery().aliveServerNodes(), null);
+    }
+
+    /** */
+    private class Process {
+        /** Process id. */
+        private final UUID id;
+
+        /** Init coordinator future. */
+        private final GridFutureAdapter<Void> initCrdFut = new GridFutureAdapter<>();
```
Mmuzaf: Should this future be re-inited in case of the process coordinator changed?

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
NSAmelchev: Could you tell API for removal, please?

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
NSAmelchev: This method is private. Еhere are checks for null before each call.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
NSAmelchev: Yes, the parameter is redundant. I'll refactor it soon.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
NSAmelchev: No, each node has its own pre-created crd futures. When the coordinator changes, coordinator node will initialize the init crd future locally.

modules/core/src/main/java/org/apache/ignite/internal/commandline/encryption/EncryptionCommand.java
```diff
@@ -0,0 +1,112 @@
...
+                String masterKeyName = argIter.nextArg("Expected master key name.");
+
+                encryptionArgs = new VisorEncryptionArgs(CHANGE_MASTER_KEY, masterKeyName);
+
+                break;
+
+            default:
+                throw new IllegalArgumentException("Unknown encryption subcommand: " + cmd);
+        }
+
```
nizhikov: nit: redundant empty line

modules/aws/src/test/java/org/apache/ignite/spi/encryption/AmazonEncryptionSpiSelfTest.java
```diff
@@ -0,0 +1,173 @@
...
+        JceMasterKey key2 = getInstance(retrieveEncryptionKey(),
+            "Test provider.", MASTER_KEY_NAME_2, "AES/GCM/NoPadding");
+
+        MasterKeyProvider<JceMasterKey> provider = (MasterKeyProvider<JceMasterKey>)buildMultiProvider(key1, key2);
+
+        encSpi.setProvider(provider);
+
+        encSpi.setMasterKeyName(DEFAULT_MASTER_KEY_NAME);
+
+        GridTestUtils.invoke(encSpi, "onBeforeStart");
```
nizhikov: `onBeforeStart` is public. We can call it directly.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -247,18 +311,22 @@ public GridEncryptionManager(GridKernalContext ctx) {
 
     /** {@inheritDoc} */
     @Override public void onDisconnected(IgniteFuture<?> reconnectFut) {
-        synchronized (genEcnKeyMux) {
+        synchronized (opsMux) {
             assert !disconnected;
 
             disconnected = true;
 
+            pendingMasterKey = null;
```
nizhikov: Why we just nullify variables here?
What will happen if process in progress?
How it would be finished?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -308,23 +376,37 @@ public void onLocalJoin() {
...
         if (res != null)
             return res;
 
-        if (!discoData.hasJoiningNodeData()) {
-            U.quietAndInfo(log, "Joining node doesn't have encryption data [node=" + node.id() + "]");
+        NodeEncryptionKeys nodeEncKeys;
 
-            return null;
+        if (!discoData.hasJoiningNodeData() ||
+            (nodeEncKeys = (NodeEncryptionKeys)discoData.joiningNodeData()) == null) {
```
nizhikov: Please, don't use the assignment operator in if.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -308,23 +376,37 @@ public void onLocalJoin() {
...
+            return new IgniteNodeValidationResult(ctx.localNodeId(),
+                "Joining node doesn't have encryption data [node=" + node.id() + "]",
+                "Joining node doesn't have encryption data.");
         }
 
-        NodeEncryptionKeys nodeEncKeys = (NodeEncryptionKeys)discoData.joiningNodeData();
+        if (!Arrays.equals(getSpi().masterKeyDigest(), nodeEncKeys.masterKeyDigest)) {
```
nizhikov: We should check `masterKeyDigest` even joining node doesn't sent `nodeEncKeys`.
Consider the following case:

0. Node started.
1. `EncryptionSPI` configured. No encrypted caches created.
2. Node joins to the cluster with other encryption configuration.

Please, add explicit test to check this case.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
NSAmelchev: Reconnecting is identical to the node join procedure. If the client node disconnects from the cluster during the key change process, it should reset these variables. The client node will not be able to join/reconnect during the master key change procedure.

modules/aws/src/test/java/org/apache/ignite/spi/encryption/AmazonEncryptionSpiSelfTest.java
NSAmelchev: Done

modules/core/src/main/java/org/apache/ignite/internal/commandline/encryption/EncryptionCommand.java
NSAmelchev: Done

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -156,13 +201,13 @@ public GridEncryptionManager(GridKernalContext ctx) {
     @Override public void start() throws IgniteCheckedException {
         startSpi();
 
-        if (getSpi().masterKeyDigest() != null)
+        if (!ctx.clientNode() && getSpi().masterKeyDigest() != null)
```
nizhikov: We don't need this attribute anymore.
Let's remove it completely.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
nizhikov: Should we complete change master key future?
>     public IgniteFuture<Void> changeMasterKey(String masterKeyName);

NSAmelchev: Yes, it will be completed in `cancelFutures("Client node was disconnected from topology (operation result is unknown).");`

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -187,20 +232,33 @@ public GridEncryptionManager(GridKernalContext ctx) {
...
 
                     assert req.keyCount() != 0;
 
                     List<byte[]> encKeys = new ArrayList<>(req.keyCount());
 
-                    for (int i = 0; i < req.keyCount(); i++)
-                        encKeys.add(getSpi().encryptKey(getSpi().create()));
+                    byte[] masterKeyDigest;
+
+                    masterKeyChangeLock.readLock().lock();
```
nizhikov: Can we made a utility method for executing login under read lock?
```
private void withReadLock(Runnable method) {
        masterKeyChangeLock.readLock().lock();

        try {
            logic.run();
        }
        finally {
            masterKeyChangeLock.readLock().unlock();
        }
}
```

nizhikov:
We need to provide the user proper way to start process from client nodes.
Let's do the following:

1. Allow starting process from a client node
2. Don't encrypt `masterKeyName` in `PendingMasterKey`.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -494,6 +566,63 @@ public void groupKey(int grpId, byte[] encGrpKey) {
...
+    @Override public IgniteFuture<Void> changeMasterKey(String masterKeyName) {
+        if (ctx.clientNode())
+            throw new UnsupportedOperationException("Client and daemon nodes can not perform this operation.");
+
+        if (!ctx.state().clusterState().active())
+            throw new IgniteException("Master key change was rejected. The cluster is inactive.");
+
+        checkMasterKeyChangeSupported();
+
+        if (masterKeyName.equals(getMasterKeyName()))
```
nizhikov: I think we should complete the future with the error here.
Change master key is the maintenance procedure that executes on timeout of key compromise.
If the change was not done user should know about it.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
NSAmelchev: Done. I have used `Callable` instead of `Runnable`. Please, take a look.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ClusterCachesInfo.java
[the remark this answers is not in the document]
nizhikov: > Group keys can be encrypted by the old master key.

Why "can be"? If local digest doesn't equal to the digest from request then key encrypted with the different MK for sure.
Let's remove "can be" from the message.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -762,21 +1206,95 @@ private void cancelFutures(String msg) {
...
-                    crd = node;
-            }
+            return crd != null && F.eq(ctx.localNodeId(), crd.id());
+        }
+    }
+
+    /**
+     * Pending master key.
+     */
+    private static class PendingMasterKey implements Serializable {
```
nizhikov: Seems, we should rename this to the `MasterKeyChangeRequest`?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -853,4 +1375,27 @@ public int keyCount() {
             return S.toString(GenerateEncryptionKeyFuture.class, this);
         }
     }
+
+    /**
+     * Master key change future.
+     */
+    private static class MasterKeyChangeFuture extends GridFutureAdapter<Void> {
+        /** */
+        private final UUID id;
```
nizhikov: Please, add javadoc for this field.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -734,18 +884,309 @@ private void writeToMetaStore(int grpId, byte[] encGrpKey) {
...
+     */
+    private void finishMasterKeyChange(UUID id, Map<UUID, MasterKeyChangeResult> res, Map<UUID, Exception> err) {
+        doneMasterKeyChangeFuture(id, err);
+    }
+
+    /**
+     * @param reqId Request id.
+     * @param err Exception.
+     */
+    private void doneMasterKeyChangeFuture(UUID reqId, Map<UUID, Exception> err) {
```
nizhikov: Let's rename this to the `completeMasterKeyChangeFuture`

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -482,7 +535,7 @@ else if (newKeys != null) {
     public void groupKey(int grpId, byte[] encGrpKey) {
         assert !grpEncKeys.containsKey(grpId);
 
-        Serializable encKey = getSpi().decryptKey(encGrpKey);
+        Serializable encKey = withMasterKeyChangeReadLock(() -> getSpi().decryptKey(encGrpKey));
```
nizhikov: What is the MK will change after this line and before `encKey` will be written to the metastore?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -494,6 +547,56 @@ public void groupKey(int grpId, byte[] encGrpKey) {
         }
     }
 
+    /** {@inheritDoc} */
+    @Override public IgniteFuture<Void> changeMasterKey(String masterKeyName) {
+        if (ctx.clientNode())
+            throw new UnsupportedOperationException("Client and daemon nodes can not perform this operation.");
+
+        if (!ctx.state().clusterState().active())
```
nizhikov: It's not clear for me. Why we can't perform MK change on inactive cluster?
Please, clarify. 

I think it should be reflected in code comment, also.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -734,18 +884,309 @@ private void writeToMetaStore(int grpId, byte[] encGrpKey) {
...
+            buf.get(encKeyName);
+
+            byte[] serKeyName = getSpi().decrypt(encKeyName, getSpi().decryptKey(encKey));
+
+            return U.fromBytes(serKeyName);
+        });
+    }
+
+    /** Checks that the master key change process supported by all nodes in cluster. */
+    public void checkMasterKeyChangeSupported() {
```
nizhikov: This function has only one usage. Let's inline it.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -494,6 +547,56 @@ public void groupKey(int grpId, byte[] encGrpKey) {
...
+        if (ctx.clientNode())
+            throw new UnsupportedOperationException("Client and daemon nodes can not perform this operation.");
+
+        if (!ctx.state().clusterState().active())
+            throw new IgniteException("Master key change was rejected. The cluster is inactive.");
+
+        checkMasterKeyChangeSupported();
+
+        if (masterKeyName.equals(getMasterKeyName()))
+            throw new IgniteException("Master key change was rejected. Master key with provided name is already set.");
```
nizhikov: > Master key change was rejected. Master key with provided name is already set.

Let's reformulate it "Master key change was rejected. New name equal to the current"

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
NSAmelchev: WAL is unavailable for write on the inactive cluster. Master key change will not be logged and group keys can be partially reencrypted in case of node stop without the possibility of recovery.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
NSAmelchev: We do not encrypt group key here. Decrypt is under the read lock because of it can be at the same time with the local new master key check.

nizhikov:
@NSAmelchev Please, add logging actual master key id that is used by Encryption manager. It may be different from the config value.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -734,18 +899,309 @@ private void writeToMetaStore(int grpId, byte[] encGrpKey) {
...
+            ctx.failure().process(new FailureContext(CRITICAL_ERROR,
+                new IgniteException("Unable to change master key locally.", e)));
+        }
+        finally {
+            masterKeyChangeLock.writeLock().unlock();
+        }
+    }
+
+    /** Logs the record with all keys and master key name to WAL. */
+    private void logKeysToWal() throws IgniteCheckedException {
```
nizhikov: Let's rename this to `writeKeysToWal`.

modules/core/src/main/java/org/apache/ignite/IgniteEncryption.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.ignite.configuration.IgniteConfiguration;
+import org.apache.ignite.lang.IgniteFuture;
+import org.apache.ignite.spi.encryption.EncryptionSpi;
+
+/**
+ * Defines encryption features.
+ * <p>
+ * Ignite provides Transparent Data Encryption of cache data on disk. Encryption features are provided by {@link
+ * EncryptionSpi} and can be configured via {@link IgniteConfiguration#setEncryptionSpi(EncryptionSpi)}. Cache
+ * encryption is configured via {@link CacheConfiguration#setEncryptionEnabled(boolean)}.
```
nizhikov: nit: "Cache encryption can be enabled via ..."

modules/core/src/main/java/org/apache/ignite/IgniteEncryption.java
```diff
@@ -0,0 +1,73 @@
...
+ * <p>
+ * Instance of {@code IgniteEncryption} is obtained from {@link Ignite} as follows:
+ * <pre class="brush:java">
+ * Ignite ignite = Ignition.ignite();
+ *
+ * IgniteEncryption encryption = ignite.encryption();
+ * </pre>
+ * Two types of keys are involved in data encryption: group and master keys.
+ * <p>
+ * Group key encrypts data of cache group caches. Each group key is encrypted by a master key. Encrypted group key and
```
nizhikov: nit: ".. by the master key"

modules/core/src/main/java/org/apache/ignite/IgniteEncryption.java
```diff
@@ -0,0 +1,73 @@
...
+ * encrypted data are written to disk.
+ * <p>
+ * Ignite provides the ability to change the master key. Master keys is identified by a master key name (see {@link
```
nizhikov: nit: "Master keys are identified by a name"

modules/core/src/main/java/org/apache/ignite/IgniteEncryption.java
```diff
@@ -0,0 +1,73 @@
...
+ * EncryptionSpi#getMasterKeyName()}). Follow operations are available for master keys:
```
nizhikov: nit: master key~s~

modules/core/src/main/java/org/apache/ignite/IgniteEncryption.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public String getMasterKeyName();
+
+    /**
+     * Starts master key change process.
+     * <p>
+     * Each node will re-encrypt group keys stored on the disk.
+     * <p>
+     * <b>NOTE:</b> The new master key should be available to {@link EncryptionSpi} for each server node. Cache start
+     * and node join during the key change process will be rejected.
```
nizhikov: nit: "Cache start and node join during the key change process is prohibited and will be rejected."

modules/core/src/main/java/org/apache/ignite/IgniteEncryption.java
```diff
@@ -0,0 +1,73 @@
...
+     * <p>
+     * If some node was unavailable during a master key change process it won't be able to join to cluster with old the
```
nizhikov: nit: "with the old master key ~name~"

modules/core/src/main/java/org/apache/ignite/IgniteEncryption.java
```diff
@@ -0,0 +1,73 @@
...
+     * master key name. The node should re-encrypt group keys during the startup and recovery process. Set up a valid
```
nizhikov: nit: "The node should re-encrypt group keys during recovery on startup. The actual master key name should be set via ..."

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1284,6 +1284,16 @@
      */
     public static final String IGNITE_PAGES_LIST_DISABLE_ONHEAP_CACHING = "IGNITE_PAGES_LIST_DISABLE_ONHEAP_CACHING";
 
+    /**
+     * The master key name that the node will use during the recovery procedure.
```
nizhikov: nit: "recovery ~procedure~"

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1284,6 +1284,16 @@
...
+     * <p>
+     * If a node was unavailable during a master key change process it won't be able to join to cluster with old the
```
nizhikov: nit: "with the old master key ~name~"

modules/core/src/main/java/org/apache/ignite/internal/GridTopic.java
```diff
@@ -142,7 +142,10 @@
     TOPIC_SERVICES,
 
     /** */
-    TOPIC_DEADLOCK_DETECTION;
+    TOPIC_DEADLOCK_DETECTION,
+
+    /** */
+    TOPIC_DISTRIBUTED_PROCESS;
```
nizhikov: Please, add JavaDoc here.

modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -76,7 +76,10 @@
     LRT_SYSTEM_USER_TIME_DUMP_SETTINGS(18),
 
     /** Partition Map Exchange-free switch on baseline node left at fully rebalanced cluster.  */
-    PME_FREE_SWITCH(19);
+    PME_FREE_SWITCH(19),
+
+    /** Master key change. */
```
nizhikov: Please, add `@see GridEncryptionManager#changeMasterKey(String)` here.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GenerateEncryptionKeyResponse.java
```diff
@@ -68,6 +73,13 @@ public IgniteUuid requestId() {
         return encKeys;
     }
 
+    /**
+     * @return Master key digest that encrypted group encryption keys.
```
nizhikov: Let's write this comment in one line.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/MasterKeyChangeRecord.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+package org.apache.ignite.internal.pagemem.wal.record;
+
+import java.util.Map;
+
+import static org.apache.ignite.internal.pagemem.wal.record.WALRecord.RecordType.MASTER_KEY_CHANGE_RECORD;
+
+/**
+ * Logical data record indented for master key change action.
```
nizhikov: nit: "Logical record that stores encryption keys. Written to the WAL on the master key change."

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/MasterKeyChangeRecord.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+public class MasterKeyChangeRecord extends WALRecord {
+    /** Master key name. */
+    private final String masterKeyName;
+
+    /** Group keys. */
+    private final Map<Integer, byte[]> grpKeys;
```
nizhikov: nite: "Group keys encryption with the master key."

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/MasterKeyChangeRecord.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * @param masterKeyName Master key name.
+     * @param grpKeys Encrypted cache keys.
```
nizhikov: nit: "Group keys".

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/MasterKeyChangeRecord.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * @return Master key name.
+     */
+    public String getMasterKeyName() {
+        return masterKeyName;
+    }
+
+    /**
+     * @return Encrypted cache keys.
```
nizhikov: nit: "group keys".

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/MasterKeyChangeRecord.java
```diff
@@ -0,0 +1,73 @@
...
+     * @param masterKeyName Master key name.
+     * @param grpKeys Encrypted cache keys.
+     */
+    public MasterKeyChangeRecord(String masterKeyName, Map<Integer, byte[]> grpKeys) {
+        this.masterKeyName = masterKeyName;
+        this.grpKeys = grpKeys;
+    }
+
+    /**
+     * @return Master key name.
```
nizhikov: Let's write comments in this class in a single line.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/MasterKeyChangeRecord.java
```diff
@@ -0,0 +1,73 @@
...
+        return grpKeys;
+    }
+
+    /** {@inheritDoc} */
+    @Override public RecordType type() {
+        return MASTER_KEY_CHANGE_RECORD;
+    }
+
+    /**
+     * @return Record's data size.
```
nizhikov: nit: "Record~'s~ data size".

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ClusterCachesInfo.java
```diff
@@ -962,6 +964,33 @@ private boolean processStartNewCacheRequest(
 
         CacheConfiguration<?, ?> ccfg = req.startCacheConfiguration();
 
+        GridEncryptionManager encMgr = ctx.encryption();
+
+        if (ccfg.isEncryptionEnabled()) {
+            IgniteCheckedException error = null;
+
+            if (encMgr.isMasterKeyChangeInProgress()) {
+                error = new IgniteCheckedException("Cache start during the master key change " +
```
nizhikov: nit: "Cache start failed. Master key change in progress".

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ClusterCachesInfo.java
```diff
@@ -962,6 +964,33 @@ private boolean processStartNewCacheRequest(
...
+                    "process is not supported.");
+            }
+            else if (encMgr.masterKeyDigest() != null &&
+                !Arrays.equals(encMgr.masterKeyDigest(), req.masterKeyDigest())) {
+                error = new IgniteCheckedException("Cache start failed. The request was initiated before " +
+                    "the master key change and can't be processed.");
+            }
+
+            if (error != null) {
+                U.warn(log, "Ignore cache start request during the master key change process.");
```
nizhikov: We should log error text also.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/DynamicCacheChangeRequest.java
```diff
@@ -101,6 +101,9 @@
     /** Encryption key. */
     @Nullable private byte[] encKey;
 
+    /** Master key digest that encrypted the group encryption key. */
```
nizhikov: nit: "Master key digest".

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/DynamicCacheChangeRequest.java
```diff
@@ -467,6 +470,20 @@ public void encryptionKey(@Nullable byte[] encKey) {
         return encKey;
     }
 
+    /**
+     * @param masterKeyDigest Master key digest that encrypted the group encryption key.
```
nizhikov: nit: "Master key digest".
Let's write this comment in single line.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/DynamicCacheChangeRequest.java
```diff
@@ -467,6 +470,20 @@ public void encryptionKey(@Nullable byte[] encKey) {
...
+     */
+    public void masterKeyDigest(@Nullable byte[] masterKeyDigest) {
+        this.masterKeyDigest = masterKeyDigest;
+    }
+
+    /**
+     * @return Master key name that encrypted the group encryption key.
```
nizhikov: nit: "Master key digest".
Let's write this comment in single line.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,385 @@
...
+import static org.apache.ignite.events.EventType.EVT_NODE_LEFT;
+import static org.apache.ignite.failure.FailureType.CRITICAL_ERROR;
+import static org.apache.ignite.internal.managers.communication.GridIoPolicy.SYSTEM_POOL;
+
+/**
+ * Distributed process is a cluster-wide process that accumulates single nodes results to finish itself.
+ * <p>
+ * The process consists of the following phases:
+ * <ol>
+ *  <li>The initial request starts the process. (Sent via discovery)</li>
```
nizhikov: nit: braces are useless. Please, remove it.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,385 @@
...
+ *  <li>Each server node processes an initial request and sends the single node result to the coordinator. (Sent via
+ *  communication)</li>
+ *  <li>The coordinator accumulate all single nodes results and finish process. (Sent via discovery)</li>
```
nizhikov: Let's add {@link } to the FinishMessage here.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,385 @@
...
+ * </ol>
+ * <p>
+ * Several processes of one type can be started at the same time.
+ *
+ * @param <I> Request type.
+ * @param <R> Result type.
```
nizhikov: Please, add @see InitMessage, @see FinishMessage here.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,385 @@
...
+        try {
+            ctx.discovery().sendCustomEvent(msg);
+        }
+        catch (IgniteCheckedException e) {
+            log.warning("Unable to send action message.", e);
+        }
+    }
+
+    /**
+     * Handles case when all server nodes have left the grid.
```
nizhikov: nit: Let's write it in the single line.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,385 @@
...
+        processes.clear();
+    }
+
+    /** @return Cluster coordinator, {@code null} if failed to determine. */
+    private @Nullable ClusterNode coordinator() {
+        return U.oldest(ctx.discovery().aliveServerNodes(), null);
+    }
+
+    /** */
+    private class Process {
```
nizhikov: Please, add javadoc here.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -0,0 +1,385 @@
...
+        private final ConcurrentHashMap<UUID, SingleNodeMessage<R>> singleMsgs = new ConcurrentHashMap<>();
+
+        /** @param id Process id. */
+        private Process(UUID id) {
+            this.id = id;
+        }
+    }
+
+    /**
+     * Defines distributed processes.
```
nizhikov: nit: Let's write it in the single line.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/FullMessage.java
```diff
@@ -0,0 +1,118 @@
...
+import org.apache.ignite.internal.managers.discovery.GridDiscoveryManager;
+import org.apache.ignite.internal.processors.affinity.AffinityTopologyVersion;
+import org.apache.ignite.internal.util.typedef.internal.S;
+import org.apache.ignite.lang.IgniteUuid;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Full process message. Contains single nodes results.
+ *
+ * @param <R> Result type.
```
nizhikov: Let's add here @see DistributedProcess.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/FullMessage.java
```diff
@@ -0,0 +1,118 @@
...
+    private final IgniteUuid id = IgniteUuid.randomUuid();
+
+    /** Process id. */
+    private final UUID processId;
+
+    /** Process type. */
+    private final int type;
+
+    /** Result. */
+    private Map<UUID, R> res;
```
nizhikov: nit: 'Result*s*'

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/FullMessage.java
```diff
@@ -0,0 +1,118 @@
...
+
+    /** Error. */
```
nizhikov: nit: 'Error*s*'

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/FullMessage.java
```diff
@@ -0,0 +1,118 @@
...
+    private Map<UUID, Exception> err;
+
+    /**
+     * @param processId Process id.
+     * @param type Process type.
+     * @param res Results.
+     * @param err Errors
+     */
+    public FullMessage(UUID processId, int type, Map<UUID, R> res, Map<UUID, Exception> err) {
```
nizhikov: It seems we should change `type` from int to the `DistributedProcessType` and use type.ordinal() if required.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/InitMessage.java
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.ignite.internal.managers.discovery.GridDiscoveryManager;
+import org.apache.ignite.internal.processors.affinity.AffinityTopologyVersion;
+import org.apache.ignite.internal.util.typedef.internal.S;
+import org.apache.ignite.lang.IgniteUuid;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Initiate message.
+ *
+ * @param <I> Request type.
```
nizhikov: Let's add here @see DistributedProcess.
@see FullMessage
@see SingleMessage

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/InitMessage.java
```diff
@@ -0,0 +1,107 @@
...
+
+    /** Request. */
+    private final I req;
+
+    /**
+     * @param processId Process id.
+     * @param type Process type.
+     * @param req Request.
+     */
+    public InitMessage(UUID processId, int type, I req) {
```
nizhikov: It seems we should change type from int to the DistributedProcessType and use type.ordinal() if required.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/SingleNodeMessage.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.UUID;
+import org.apache.ignite.internal.util.typedef.internal.U;
+import org.apache.ignite.plugin.extensions.communication.Message;
+import org.apache.ignite.plugin.extensions.communication.MessageReader;
+import org.apache.ignite.plugin.extensions.communication.MessageWriter;
+
+/**
+ * Single node result message.
+ *
+ * @param <R> Result type.
```
nizhikov: Let's add here @see DistributedProcess.
@see FullMessage
@see InitMessage

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/SingleNodeMessage.java
```diff
@@ -0,0 +1,192 @@
...
+    public SingleNodeMessage() {
+    }
+
+    /**
+     * @param processId Process id.
+     * @param type Process type.
+     * @param resp Single node response.
+     * @param err Error.
+     */
+    public SingleNodeMessage(UUID processId, int type, R resp, Exception err) {
```
nizhikov: It seems we should change type from int to the DistributedProcessType and use type.ordinal() if required.

modules/core/src/main/java/org/apache/ignite/mxbean/EncryptionMXBean.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Encryption features MBean.
+ */
+@MXBeanDescription("MBean that provides access to encryption features.")
+public interface EncryptionMXBean {
+    /**
+     * Gets the current master key name.
+     *
+     * @return Master key name.
+     *
```
nizhikov: nit: please, remove redundant empty line.

modules/core/src/main/java/org/apache/ignite/mxbean/EncryptionMXBean.java
```diff
@@ -0,0 +1,48 @@
...
+     * @see IgniteEncryption#getMasterKeyName()
+     */
+    @MXBeanDescription("Current master key name.")
+    public String getMasterKeyName();
+
+    /**
+     * Starts master key change process.
+     *
+     * @param masterKeyName Master key name.
+     *
```
nizhikov: nit: please, remove redundant empty line.

modules/core/src/main/java/org/apache/ignite/spi/encryption/EncryptionSpi.java
```diff
@@ -110,4 +111,22 @@
      * @return Encrypted data block size.
      */
     int blockSize();
+
+    /**
+     * Gets the current master key name.
+     *
+     * @return Master key name.
+     *
```
nizhikov: nit: please, remove redundant empty line.

modules/core/src/test/java/org/apache/ignite/internal/encryption/EncryptionMXBeanTest.java
```diff
@@ -0,0 +1,146 @@
...
+import org.apache.ignite.internal.managers.encryption.EncryptionMXBeanImpl;
+import org.apache.ignite.internal.util.typedef.internal.U;
+import org.apache.ignite.mxbean.EncryptionMXBean;
+import org.junit.Test;
+
+import static org.apache.ignite.spi.encryption.keystore.KeystoreEncryptionSpi.DEFAULT_MASTER_KEY_NAME;
+import static org.apache.ignite.testframework.GridTestUtils.assertThrowsWithCause;
+
+/**
+ * Tests {@link EncryptionMXBean}.
```
nizhikov: nit: Let's write this in one line.

PR #6941, opened 2019-10-07
FOR TEST POC evict index partitions

zkillingjar:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/GridDhtLocalPartition.java
```diff
@@ -164,6 +164,9 @@
      * reservation is released. */
     private volatile boolean delayedRenting;
 
+    /** Set if partition is requested for exclusive evict. */
+    private volatile boolean exclClear;
```
nizhikov: Let's rename it to the `exclPurge`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/GridDhtLocalPartition.java
```diff
@@ -1469,6 +1477,44 @@ public void beforeApplyBatch(boolean last) {
         // No-op.
     }
 
+    /**
+     * Initialize this partition to be a part of a exclusive clear over a set of partitions.
+     *
+     * @param fut Future to listen for completion of a larger clear procedure.
+     * @return {@code true} if successfully initialized, otherwise a retry is required.
+     */
+    public boolean exclusiveClearAsync(IgniteInternalFuture<Void> fut) {
```
nizhikov: Who protect this method in case of concurrent invocation?

nizhikov:
@zkillingjar 

We have`GridDhtLocalPartition#exclClear` flag and `GridDhtLocalPartition#clearFut` which seems to be the same.

Seems, we should encapsulate `exclClear` flag to the `clearFut`.

nizhikov:
We have to ensure consistent naming in PR.
I think `purge` or `partClear` is OK.
sweep, clear and other prefixes that are used through PR should be renamed.

PR #6942, opened 2019-10-07
IGNITE-9913

anton-vinogradov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ExchangeContext.java
```diff
@@ -77,6 +95,28 @@ public ExchangeContext(boolean crd, GridDhtPartitionsExchangeFuture fut) {
...
+
+            GridAffinityAssignmentCache aff = grp.affinity();
+
+            Set<Integer> failedPrimaries = aff.primaryPartitions(fut.exchangeId().eventNode().id(), aff.lastVersion());
+            Set<Integer> locBackups = aff.backupPartitions(fut.sharedContext().localNodeId(), aff.lastVersion());
+
+            for (int part : failedPrimaries) {
+                if (locBackups.contains(part))
+                    return true;
+            }
```
agoncharuk: This code does not look correct. A node may OWN a partition, but it will not be reported in aff.backupPartitions because the method returns current assignment, not the local partitions state. GridDhtPartitionTopology should be read for this information. A corresponding test need to be added.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ExchangeContext.java
```diff
@@ -58,7 +66,17 @@
     public ExchangeContext(boolean crd, GridDhtPartitionsExchangeFuture fut) {
         int protocolVer = exchangeProtocolVersion(fut.firstEventCache().minimumNodeVersion());
 
-        if (compatibilityNode || (crd && fut.localJoinExchange())) {
+        GridDiscoveryManager disco = fut.sharedContext().discovery();
+
+        if (protocolVer > 2 && fut.exchangeId().isLeft() && fut.isEventNodeInBaseline() &&
```
Jokser: This protocol works only if cluster is in-memory and baseline-autoadjust is enabled.
How it works when persistence is enabled and with/without baseline autoadjust is enabled?

anton-vinogradov: What protocol do you mean? 
protocolVer == 3 was introduced by me.

Jokser: Yes, I mean the protocol introduced by yourself.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -1021,6 +1025,15 @@ private boolean isLocalNodeInBaseline() {
...
     }
 
+    /**
+     * @return {@code true} if Event node is in baseline and {@code false} otherwise.
+     */
+    public boolean isEventNodeInBaseline() {
+        BaselineTopology top = cctx.discovery().discoCache().state().baselineTopology();
+
+        return top != null && top.consistentIds().contains(firstDiscoEvt.eventNode().consistentId());
+    }
```
agoncharuk: Looks like there is a race here. discovery().discoCache() will return a current snapshot, which is updated in discovery thread, while the method is executed in exchange thread. Returned disco cache may have topology version different from the exchange future topology version.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -1628,6 +1660,100 @@ private void changeWalModeIfNeeded() {
         return null;
     }
 
+    /**
+     * @throws IgniteCheckedException If failed.
+     */
+    private void waitRecovery() throws IgniteCheckedException {
```
agoncharuk: The code in this method is very much similar to the waitPartitionRelease(). Can we avoid the duplication?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
agoncharuk: BTW, what happens when exchanges are merged? Two failed nodes -> we need to check both? Failed and joined -> We should not allow fast-path exchange.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ExchangeContext.java
Jokser: `disco.baselineChanged(fut.sharedContext().exchange().readyAffinityVersion(), fut.initialVersion())`
This code will return `true` only if baseline has changed by `NODE_LEFT` event. Baseline is changed automatically in discovery cache only if it's in-memory cluster and auto-adjust is enabled. In other cases baseline is changed manually (if auto-adjust disabled) or by separate event (ChangeGlobalState) right after topology event if auto-adjust is enabled.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheAffinitySharedManager.java
```diff
@@ -1982,7 +1982,37 @@ public boolean onCentralizedAffinityChange(final GridDhtPartitionsExchangeFuture
...
+     * @return {@code True} if affinity should be assigned by coordinator.
+     */
+    public boolean onBaselineNodeLeft(final GridDhtPartitionsExchangeFuture fut, boolean crd) {
+        assert (fut.events().hasServerLeft() && !fut.firstEvent().eventNode().isClient()) : fut.firstEvent();
+
+        forAllRegisteredCacheGroups(new IgniteInClosureX<CacheGroupDescriptor>() {
+            @Override public void applyx(CacheGroupDescriptor desc) throws IgniteCheckedException {
+                CacheGroupHolder cache = getOrCreateGroupHolder(fut.initialVersion(), desc);
+
+                cache.aff.reinitializeWithoutOfflineNodes(fut.initialVersion());
```
Jokser: This is not correct affinity initialization.
You don't take into account that cluster may have `MOVING` partitions. In this case, a primary node can be chosen from `MOVING` backup and affinity is not recalculated after PME end. You can see correct affinity recalculation at `org.apache.ignite.internal.processors.cache.CacheAffinitySharedManager#onServerLeftWithExchangeMergeProtocol`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ExchangeContext.java
anton-vinogradov: It still looks like code works, but disco.baselineChanged(...) can be used on in case of nodeFail, correct?
Persistence were checked by TxPartitionCounterStateConsistencyTest#testPartitionConsistencyWithPrimaryRestart. 
As you told, persistence case (auto-adjust in [true, false]) on nodeLeft always leads to recovery instead of regular PME.

I don't like the idea to have a check like persistence.enabled() here, because we should check the result (was baseline changed or not) instead of checking reasons leads to this (which, definitely, will be changed in future, and can be (already) overcomplicated).
Could you help with writing a simple and proper "if" statement?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ExchangeContext.java
anton-vinogradov: Copying answer from issue.

anton-vinogradov: Copying answer from issue. 
Do you mean we have to serve also moving and renting partitions? 

So, Set<Integer> failedPrimaries = aff.primaryPartitions(fut.exchangeId().eventNode().id(), aff.lastVersion()); is a correct calculation, but
Set<Integer> locBackups = aff.backupPartitions(fut.sharedContext().localNodeId(), aff.lastVersion()); should be replaced with dht.localPartitions() usage?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
anton-vinogradov: 1) Correct fix is to use versioned disco? 
`cctx.discovery().discoCache(topVer)...` correct?

2) We newer merge left with anything 
`if (baselineNodeLeft) baselineNodeLeft = true;
merge = false;`
so, even cascade failure will be processed one-by-one.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheAffinitySharedManager.java
anton-vinogradov: Since I'm using assignment (not idealAssignment), which took moving into account (correct?) this should be еру correct initialization?

Jokser: There is no guarantee that the next backup will have a partition in OWNING status.
Case 1:
Suppose that you have a partition with 2 backups. Primary is OWNER other nodes are MOVING:
OMM
The second backup completed rebalancing and now have OWNING partition state:
OMO
Now the primary node is left and you have:
MO in affinity distribution. MOVING partition becomes primary which is incorrect behavior.
Case 2:
You lost your last OWNING partition in affinity. In this case, detectLostPartitions should be triggered.

You also nullify WaitRebalanceInfo. You shouldn't do this because in this case if you have currently running rebalancing late affinity assignment will not be triggered after rebalancing completion.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ExchangeContext.java
Jokser: I think, there should be no relying on the changed baseline or not.
Why do you check it here?

anton-vinogradov: Because "recovery only PME" possible only in case affinity was not changed, and only primary-backup switches are expected.

Jokser: The primary-backup switch is only possible when baseline is not changed :)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheAffinitySharedManager.java
anton-vinogradov: 1) Let's clarify, aff.assignment contains only owning partitions (correct?)
I'm just shrinking previous assignment, without recalculation.
I see no MOVING issues here, am I missed something?

2) Rebalance will be restarted in case of moving partitions.
See no problem here :(

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ExchangeContext.java
anton-vinogradov: That's exactly the case covered by this fix :)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheAffinitySharedManager.java
Jokser: 1. Affinity assignment contains OWNING and MOVING partitions.
2. Rebalance is started but late affinity assignment because of nullified WaitRebalanceInfo - not.
3. Detect lost partitions should be triggered after PME because if you don't have OWNING partitions you should mark others as LOST due to loss policy.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ExchangeContext.java
anton-vinogradov: My bad :) 

public boolean baselineChanged(AffinityTopologyVersion topVer1, AffinityTopologyVersion topVer2) {
...
return Objects.equals(baseline1, baseline2); 
}

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
NSAmelchev: 1. Seems you can use `firstEvtDiscoCache`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ExchangeContext.java
anton-vinogradov: Issue solved?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -3826,10 +3952,14 @@ private void assignPartitionsStates() {
...
-     * failed to send update counter deltas to backup.
+     * Removes gaps in the local update counters. Affects specified node's primary partitions in case node specified and
+     * all local partitins otherwise.
+     *
+     * Gaps in update counters are possible on backup node when primary failed to send update counter deltas to backup.
+     *
+     * @param nodeId Failed node id, {@code null} in case of merged failures.
      */
-    private void finalizePartitionCounters() {
+    private void finalizePartitionCounters(UUID nodeId) {
```
ascherbakoff: @Nullable is missing

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ExchangeContext.java
```diff
@@ -58,7 +66,17 @@
     public ExchangeContext(boolean crd, GridDhtPartitionsExchangeFuture fut) {
         int protocolVer = exchangeProtocolVersion(fut.firstEventCache().minimumNodeVersion());
 
-        if (compatibilityNode || (crd && fut.localJoinExchange())) {
+        GridDiscoveryManager disco = fut.sharedContext().discovery();
+
+        if (protocolVer > 2 && fut.exchangeId().isLeft() && fut.isFirstEventNodeInBaseline() &&
```
ascherbakoff: Protocol v3 should be disabled if previous assignment is not ideal. Otherwise it means rebalancing is not finished and some nodes can see different local partition states at the moment of assignment calculation.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ExchangeContext.java
```diff
@@ -77,6 +95,28 @@ public ExchangeContext(boolean crd, GridDhtPartitionsExchangeFuture fut) {
...
+     */
+    private boolean localRecoveryNeeded(GridDhtPartitionsExchangeFuture fut) {
+        for (CacheGroupContext grp : fut.sharedContext().cache().cacheGroups()) {
+            if (grp.isLocal())
+                continue;
+
+            GridAffinityAssignmentCache aff = grp.affinity();
+
+            Set<Integer> failedPrimaries = aff.primaryPartitions(fut.exchangeId().eventNode().id(), aff.lastVersion());
+            Set<Integer> loc = grp.topology().localPartitionMap().keySet();
```
ascherbakoff: You should only take into account owning partitions here (avoid renting partitions, moving are not possible because previous assignment is ideal).

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -1357,7 +1370,9 @@ private ExchangeType onServerNodeEvent(boolean crd) throws IgniteCheckedExceptio
 
             exchCtx.events().warnNoAffinityNodes(cctx);
 
-            centralizedAff = cctx.affinity().onCentralizedAffinityChange(this, crd);
+            centralizedAff = exchCtx.baselineNodeLeft() ?
```
ascherbakoff: centralizedAff  is a part of old protocol. It's better not to mix old and new protocols, some refactoring is needed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -2225,7 +2351,7 @@ private String exchangeTimingsLogMessage(String header, List<String> timings) {
 
                 boolean locNodeNotCrd = crd == null || !crd.isLocal();
 
-                if (locNodeNotCrd && (serverNodeDiscoveryEvent() || localJoinExchange()))
+                if ((locNodeNotCrd && (serverNodeDiscoveryEvent() || localJoinExchange())) || exchCtx.baselineNodeLeft())
```
ascherbakoff: Make sure partition loss policy works well with new protocol. I suggest removing nodes one by one under load until some subset of partitions will have no owners more.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -3839,7 +3969,14 @@ private void finalizePartitionCounters() {
                 cctx.kernalContext().getSystemExecutorService(),
                 nonLocalCacheGroups(),
                 grp -> {
-                    grp.topology().finalizeUpdateCounters();
+                    AffinityTopologyVersion topVer = sharedContext().exchange().readyAffinityVersion();
+
+                    // Failed node's primary partitions or just all local backups in case of possible exchange merge.
+                    Set<Integer> parts = nodeId != null ?
```
ascherbakoff: It this intended as an optimization ?
This is not really needed, because here all possible updates are received and all remaining gaps could be safely removed.

modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/GridDiscoveryManager.java
```diff
@@ -1798,6 +1799,24 @@ public int size() {
...
+
+        DiscoCache disco1 = discoCache(topVer1);
+        DiscoCache disco2 = discoCache(topVer2);
+
+        BaselineTopology baseline1 = disco1 != null ? disco1.state().baselineTopology() : null;
+        BaselineTopology baseline2 = disco2 != null ? disco2.state().baselineTopology() : null;
+
+        return !Objects.equals(baseline1, baseline2);
+    }
+
```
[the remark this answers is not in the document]
anton-vinogradov: Looks like it's possible to just use `firstEvtDiscoCache.state().localBaselineAutoAdjustment()`?
Updated the PR.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
anton-vinogradov: What's the reason for using this annotation? 
Everything in nullable until checked to be not null.

ascherbakoff: To emphasize special meaning of null argument. We have lots of use of @Nullable annotation in the codebase.

anton-vinogradov: Not sure I understand who checks them... and how?
What contract we have here, what difference with absent annotation?
Seems, that's a holywar issue :)
Will you mind if I keep this as is?

ascherbakoff: I wouldn't mind )

anton-vinogradov:
continued at https://github.com/apache/ignite/pull/7069

PR #6947, opened 2019-10-08
IGNITE-12261 - Issue with adding nested index dynamically

kotari4u:
http://apache-ignite-users.70518.x6.nabble.com/Issue-with-adding-nested-index-dynamically-tt29571.html

As discussed in the user group ( please check the link ), I have added fix in this class, please let me know if it can be merged and new version can be released. Please let me know for any issue with this commit. Thank you

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
@@ -835,7 +835,7 @@ public class QueryUtils {
 
             fullName.append(prop);
 
-            String alias = aliases.get(fullName.toString());
+            String alias = aliases.get(fullName.toString()) == null ? fullName.toString() : aliases.get(fullName.toString());
 
             // The key flag that we've found out is valid for the whole path.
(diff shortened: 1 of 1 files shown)


kotari4u:
Can you please check this pull request. Thanks

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
```diff
@@ -835,7 +835,7 @@ public static QueryBinaryProperty buildBinaryProperty(GridKernalContext ctx, Str
 
             fullName.append(prop);
 
-            String alias = aliases.get(fullName.toString());
+            String alias = aliases.get(fullName.toString()) == null ? fullName.toString() : aliases.get(fullName.toString());
```
skarpushin: Why do `aliases.get(fullName.toString());` twice?
Suggestion:
```
String alias = aliases.get(fullName.toString());
if (alias == null) {
            alias = fullName.toString();
}
```

kotari4u:
Creatdd another pull request with comment
Closing this so that we can look at other
https://github.com/apache/ignite/pull/6949

PR #6948, opened 2019-10-08
IGNITE-12268 Adds ability to set up custom REST processor.

ololo3000:


modules/core/src/test/java/org/apache/ignite/internal/processors/rest/RestProcessorInitializationTest.java
```diff
@@ -0,0 +1,224 @@
...
+            return new IgnitePlugin() {
+                // No-op.
+            };
+        }
+    }
+
+    /**
+     * Test no-op implementation of {@link IgniteRestProcessor}.
+     */
+    private static class TestGridRestProcessorImpl implements IgniteRestProcessor {
```
dgarus: should use GridProcessorAdapter to avoid a boilerplate code

ololo3000: Thanks! Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/IgniteRestProcessor.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.processors.rest;
+
+import org.apache.ignite.internal.processors.GridProcessor;
+
+/**
+ * REST processor marker interface.
```
nizhikov: Let's add `@see GridRestProcessor` here.

nizhikov:
We need to change the return type of `StandaloneGridKernalContext#rest`, also

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/IgniteRestProcessor.java
ololo3000: Thanks. Done.

PR #6960, opened 2019-10-09
IGNITE-12193 Add cacheGroup metrics.

zzzadruga:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupMetricsImpl.java
```diff
@@ -188,6 +224,98 @@ public void decrementIndexBuildCountPartitionsLeft() {
...
+    public long getRebalancingStartTime() {
+        return rebalancingStartTime.value();
+    }
+
+    /**
+     * Set the current time to rebalancing start time.
+     *
+     * @param delay Delay in milliseconds.
+     */
+    public void startRebalance(long delay) {
```
Mmuzaf: I think we should leave this logic in `demander`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -277,6 +277,9 @@ Runnable addAssignments(
         long delay = grp.config().getRebalanceDelay();
 
         if ((delay == 0 || force) && assignments != null) {
+            grp.metrics().clearRebalanceCounters();
+            grp.metrics().startRebalance(0);
```
Mmuzaf: It is two different stages: rebalance future inited and rebalance future started.

I think we should set `rebalanceStartTime` for a cache group when the first request message sends and right after the cache group eviction happened. See the `requestPartitions` method.

WDYT?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -687,6 +697,8 @@ public void handleSupplyMessage(
 
             final GridDhtPartitionTopology top = grp.topology();
 
+            grp.metrics().onRebalanceBatchReceived(supplyMsg.messageSize());
```
Mmuzaf: Let's rename `onRebalanceBatchReceived` with a `addRebalancingReceivedBytes`. It's more clear to read such callbacks.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1133,6 +1145,8 @@ private void updateGroupMetrics() {
         // TODO: IGNITE-11330: Update metrics for touched cache only.
         // Due to historical rebalancing "EstimatedRebalancingKeys" metric is currently calculated for the whole cache
         // group (by partition counters), so "RebalancedKeys" and "RebalancingKeysRate" is calculated in the same way.
+        grp.metrics().onRebalanceKeyReceived();
```
Mmuzaf: I don't think that this is the right place for incrementing of receiving keys. Let's call this outside the `updateGroupMetrics` since this is a method-candidate to remove.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupMetricsImpl.java
```diff
@@ -188,6 +224,98 @@ public void decrementIndexBuildCountPartitionsLeft() {
...
+     * Rebalance partition callback.
+     */
+    public void onPartitionRebalanced() {
+        rebalancingPartitionsLeft.decrement();
+    }
+
+    /**
+     * Rebalance entry store callback.
+     */
+    public void onRebalanceKeyReceived() {
```
Mmuzaf: Sometimes you use `rebalancing` word and sometimes `rebalance`. Let's leave only `rebalancing`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupMetricsImpl.java
Mmuzaf: I think this method don't need at all, since we set the rebalance time only when it actually starts.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1213,6 +1227,8 @@ private void updateGroupMetrics() {
 
                 remaining.put(k.id(), v.partitions());
 
+                grp.metrics().addRebalancingPartitions(v.partitions().size());
```
Mmuzaf: Let's rename to `rebalancingPartitionsLeft(value)`.
I think this should be set when the rebalance actually starts.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1318,7 +1334,10 @@ private void cancel(UUID nodeId) {
 
                 cleanupRemoteContexts(nodeId);
 
-                remaining.remove(nodeId);
+                IgniteDhtDemandedPartitionsMap rmvd = remaining.remove(nodeId);
+
+                if (rmvd != null)
+                    grp.metrics().cancelRebalancingPartitions(rmvd.size());
```
Mmuzaf: Do we need this method? Can the `reset` method be used? Are we need such callback here at all?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupMetricsImpl.java
```diff
@@ -188,6 +224,98 @@ public void decrementIndexBuildCountPartitionsLeft() {
...
+     * Rebalanced keys count.
+     */
+    public long getRebalancingReceivedKeys() {
+        return rebalancingReceivedKeys.value();
+    }
+
+    /**
+     * Rebalanced bytes count.
+     */
+    public long getRebalancingReceivedBytes() {
```
Mmuzaf: Do we need a `getters` at all? It seems all the metrics can be obtained from the metric registry.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupMetricsImpl.java
```diff
@@ -57,6 +59,21 @@
     /** Number of partitions need processed for finished indexes create or rebuilding. */
     private final AtomicLongMetric idxBuildCntPartitionsLeft;
 
+    /** Number of rebalanced partitions left. */
```
Mmuzaf: I think we should track all the metrics in `Demander` class and do not place it here.
The metric registry instance can be obtained on the `Demander` startup.
The `register(String name, LongSupplier supplier, @Nullable String desc) ` method with a lambda expression can be used to track a state of the `RebalanceFuture`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -141,6 +145,27 @@ public GridDhtPartitionDemander(CacheGroupContext grp) {
         Map<Integer, Object> tops = new HashMap<>();
 
         rebalanceTopic = GridCachePartitionExchangeManager.rebalanceTopic(0);
+
+        String metricGroupName = metricName(CACHE_GROUP_METRICS_PREFIX, grp.cacheOrGroupName());
+
+        MetricRegistry mreg = grp.shared().kernalContext().metric().registry(metricGroupName);
+
+        mreg.register("RebalancingPartitionsLeft", () -> rebalanceFut.remaining.values().stream()
```
Mmuzaf: I think it's better to pre-calculate all partition on future creation time and decrement them on demand. Seeking thought map for metrics can be too expensive.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -141,6 +145,27 @@ public GridDhtPartitionDemander(CacheGroupContext grp) {
...
+                        .sum(), "Number of rebalanced partitions left.");
+
+        mreg.register("RebalancingReceivedKeys",
+            () -> rebalanceFut.rebalancingReceivedKeys.get(), "Number of already rebalanced keys.");
+
+        mreg.register("RebalancingReceivedBytes",
+            () -> rebalanceFut.rebalancingReceivedBytes.get(), "Number of already rebalanced bytes.");
+
+        mreg.register("RebalancingStartTime",
+            () -> rebalanceFut.rebalancingStartTime, "Rebalancing start time.");
```
Mmuzaf: `Cache group actual rebalancing start time. Cache groups within a cluster rebalances one by one.`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -141,6 +145,27 @@ public GridDhtPartitionDemander(CacheGroupContext grp) {
...
+        mreg.register("RebalancingPartitionsLeft", () -> rebalanceFut.remaining.values().stream()
+                        .map(IgniteDhtDemandedPartitionsMap::fullSet)
+                        .mapToInt(Collection::size)
+                        .sum(), "Number of rebalanced partitions left.");
+
+        mreg.register("RebalancingReceivedKeys",
+            () -> rebalanceFut.rebalancingReceivedKeys.get(), "Number of already rebalanced keys.");
+
+        mreg.register("RebalancingReceivedBytes",
+            () -> rebalanceFut.rebalancingReceivedBytes.get(), "Number of already rebalanced bytes.");
```
Mmuzaf: `The number of currently rebalanced bytes of this cache group`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -141,6 +145,27 @@ public GridDhtPartitionDemander(CacheGroupContext grp) {
...
+
+        MetricRegistry mreg = grp.shared().kernalContext().metric().registry(metricGroupName);
+
+        mreg.register("RebalancingPartitionsLeft", () -> rebalanceFut.remaining.values().stream()
+                        .map(IgniteDhtDemandedPartitionsMap::fullSet)
+                        .mapToInt(Collection::size)
+                        .sum(), "Number of rebalanced partitions left.");
+
+        mreg.register("RebalancingReceivedKeys",
+            () -> rebalanceFut.rebalancingReceivedKeys.get(), "Number of already rebalanced keys.");
```
Mmuzaf: `The number of currently rebalanced keys for the whole cache group`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -141,6 +145,27 @@ public GridDhtPartitionDemander(CacheGroupContext grp) {
...
         rebalanceTopic = GridCachePartitionExchangeManager.rebalanceTopic(0);
+
+        String metricGroupName = metricName(CACHE_GROUP_METRICS_PREFIX, grp.cacheOrGroupName());
+
+        MetricRegistry mreg = grp.shared().kernalContext().metric().registry(metricGroupName);
+
+        mreg.register("RebalancingPartitionsLeft", () -> rebalanceFut.remaining.values().stream()
+                        .map(IgniteDhtDemandedPartitionsMap::fullSet)
+                        .mapToInt(Collection::size)
+                        .sum(), "Number of rebalanced partitions left.");
```
Mmuzaf: `The number of cache group partitions left to be rebalanced`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -446,6 +471,9 @@ private void requestPartitions(final RebalanceFuture fut, GridDhtPreloaderAssign
                 if (fut.isDone())
                     break;
 
+                if (rebalanceFut.rebalancingStartTime == -1)
```
Mmuzaf: Do we need such check?

zzzadruga: Without this check, the rebalancing start time can be set several times, because it is in the “for each” cycle

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1305,6 +1365,23 @@ public boolean isInitial() {
...
+            assert !cancel;
+
+            long time = System.currentTimeMillis();
+
+            if (startTime != -1)
+                endTime.compareAndSet(-1, time);
+
+            boolean byThisCall = super.onDone(res, err, cancel);
+
+            if (byThisCall && (res == false || isFailed()))
```
Mmuzaf: Let's change to `res == Boolean.FALSE` to avoid unboxing.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1305,6 +1365,23 @@ public boolean isInitial() {
...
         }
 
+        /** {@inheritDoc} */
+        @Override public boolean onDone(@Nullable Boolean res, @Nullable Throwable err, boolean cancel) {
+            assert !cancel;
+
+            long time = System.currentTimeMillis();
+
+            if (startTime != -1)
+                endTime.compareAndSet(-1, time);
```
Mmuzaf: The endTime should be set only on a successful method call, right?

PR #6962, opened 2019-10-10
IGNITE-12167

denis-chudov:


modules/core/src/test/java/org/apache/ignite/testframework/junits/GridAbstractTest.java
```diff
@@ -245,6 +247,12 @@ public String getName() {
     /** Lazily initialized current test method. */
     private volatile Method currTestMtd;
 
+    /**
+     * Page handler wrapper for {@link BPlusTree}, it can be saved here and overrided for test purposes,
+     * then it must be restored using value of this field.
+     */
+    private transient PageHandlerWrapper<BPlusTree.Result> regularPageHndWrapper;
```
AMashenkov: Why 'transient' is needed here?

modules/core/src/test/java/org/apache/ignite/testframework/junits/GridAbstractTest.java
```diff
@@ -334,7 +342,8 @@ protected void beforeTestsStarted() throws Exception {
      * @throws Exception If failed.
      */
     protected void afterTestsStopped() throws Exception {
-        // No-op.
+        //restoring page handler wrapper
+        BPlusTree.pageHndWrapper = regularPageHndWrapper == null ? ((tree, hnd) -> hnd) : regularPageHndWrapper;
```
AMashenkov: Seems, regularPageHndWrapper is never null here.
If BPlusTree.pageHndWrapper was null in beforeTestsStarted, then null value will not be restored.

AMashenkov: Do we really need change pageHndWrapper for all tests (in GridAbstractTest)?

PR #6976, opened 2019-10-14
IGNITE-6804 Warning if unordered map used in locking cache operation.

alamar:


modules/core/src/main/java/org/apache/ignite/IgniteCache.java
```diff
@@ -639,6 +639,10 @@
      * <p>
      * Please refer to documentation for {@link CacheAtomicityMode#ATOMIC} for information on
      * system behavior in crash scenarios for atomic caches.
+     * <p>
```
dmagda: Ilya, how about this version? I would use it here and in similar places below.

_Note, always pass the keys in ordered maps like TreeMap to avoid distributed deadlocks. Ignite always locks/accesses the keys in a way they ordered within a data structure. Don't use unordered maps like HashMap to avoid distributed deadlocks that happen if this or similar method is called by multiple threads/applications and keys are passed in an unordered map._

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -5175,6 +5197,58 @@ protected final void validateCacheKeys(Iterable<?> keys) {
...
+        if (m instanceof SortedMap || m instanceof LinkedHashMap || m instanceof GridSerializableMap)
+            return;
+
+        Transaction tx = ctx.kernalContext().cache().transactions().tx();
+        if (tx != null && !tx.implicit() && tx.concurrency() == OPTIMISTIC)
+            return;
+
+        LT.warn(log, "Unordered map of type " + m.getClass().getSimpleName() +
+            " was passed to " + op + " operation on cache: " + name() + ". " +
+            "Locking order of keys cannot be guaranteed - this will lead to deadlock! " +
```
dmagda: Please rephrase this sentence a bit - "Unordered map {type} is used for {name} operation. This can lead to a distributed deadlock. Switch to a sorted map like TreeMap instead"

Let's use it in similar places below.

modules/core/src/main/java/org/apache/ignite/IgniteCache.java
alamar: I think this is a good passage for dedicated docs page, but not for copying over and over across javadoc. I will check my message to see if it can be improved.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -5176,6 +5196,62 @@ protected final void validateCacheKeys(Iterable<?> keys) {
...
+     *
+     * Issues developer warning otherwise.
+     *
+     * @param coll Collection to examine.
+     */
+    protected void checkKeysOrdered(Collection coll, String op) {
+        if (coll == null || coll.size() <= 1)
+            return;
+
+        if (coll instanceof SortedSet || coll instanceof GridCacheAdapter.KeySet)
```
agura: Why `GridCacheAdapter.KeySet` is allowed?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -5176,6 +5196,62 @@ protected final void validateCacheKeys(Iterable<?> keys) {
...
+        if (curTxDeadlockDetecting())
+            return;
+
+        LT.warn(log, "Unordered collection " + coll.getClass().getSimpleName() +
+            " is used for " + op + " operation on cache " + name() + ". " +
+            "This can lead to a distributed deadlock. Switch to a sorted set like TreeSet instead.");
+    }
+
+    /** */
+    private boolean curTxDeadlockDetecting() {
```
agura: Will not work for `tx.concurrency() == PESSIMISTIC`. Deadlock detection works for both optimistic and pessimistic transactions.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/local/atomic/GridLocalAtomicCache.java
```diff
@@ -650,6 +652,8 @@ public GridLocalAtomicCache(GridCacheContext<K, V> ctx) {
         if (keyCheck)
             validateCacheKeys(keys);
 
+        checkKeysOrdered(keys, "Invoke All");
```
agura: Here should be `"Invoke All Sync"` operation.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/local/atomic/GridLocalAtomicCache.java
```diff
@@ -717,6 +723,8 @@ public GridLocalAtomicCache(GridCacheContext<K, V> ctx) {
         if (keyCheck)
             validateCacheKeys(map.keySet());
 
+        checkKeysOrdered(map, "Invoke All");
```
agura: Here should be `"Invoke All Sync"` operation.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/local/atomic/GridLocalAtomicCache.java
```diff
@@ -858,9 +866,6 @@ private Object updateAllInternal(GridCacheOperation op,
         boolean readThrough,
         boolean keepBinary
     ) throws IgniteCheckedException {
-        if (keyCheck)
```
agura: Unsafe change. This method could be invoked form `puAll0()` for example. The same for `GridDhtAtomicCache`.

alamar: I have checked absolutely all callers of this method, and they all have key check upstream. putAll0 has this check in upstream putAll.

I think we never need to repeat it here. I can add javadoc stating that keys need to already be checked.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/local/atomic/GridLocalAtomicCache.java
alamar: Do you mean "Async"?? Others also

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
alamar: Invoked by our own code, no warning is needed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -5176,6 +5196,62 @@ protected final void validateCacheKeys(Iterable<?> keys) {
...
+     *
+     * Issues developer warning otherwise.
+     *
+     * @param m Map to examine.
+     */
+    protected void checkKeysOrdered(Map m, String op) {
+        if (m == null || m.size() <= 1)
+            return;
+
+        if (m instanceof SortedMap || m instanceof GridSerializableMap)
```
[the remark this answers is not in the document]
alamar: Invoked by our own code, no warning is needed.

modules/core/src/main/java/org/apache/ignite/IgniteCache.java
```diff
@@ -898,6 +902,11 @@
 
     /**
      * {@inheritDoc}
+     * <p>
+     * Keys are locked in the order in which they appear in map. It is caller's responsibility to
+     * make sure keys always follow same order, such as by using {@link TreeMap}. Using unordered map,
+     * such as {@link HashMap}, while calling this method in parallel <b>will lead to deadlock</b>.
```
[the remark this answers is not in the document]
alamar: Will Do

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/local/atomic/GridLocalAtomicCache.java
agura: Yes, I do.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/local/atomic/GridLocalAtomicCache.java
agura: Okay.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -5176,6 +5198,91 @@ protected final void validateCacheKeys(Iterable<?> keys) {
...
+
+        if (op == BulkOperation.GET && tx == null)
+            return true;
+
+        return false;
+    }
+
+    /** */
+    protected enum BulkOperation {
+        GET("Get All"),
```
agura: It seems that `title` is redundant.  We know that operation is bulk and we know operation type, it is enough.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -5176,6 +5198,91 @@ protected final void validateCacheKeys(Iterable<?> keys) {
...
+        LT.warn(log, "Unordered collection " + coll.getClass().getSimpleName() +
+            " is used for " + op.title() + " operation on cache " + name() + ". " +
+            "This can lead to a distributed deadlock. Switch to a sorted set like TreeSet instead.");
+    }
+
+    /** */
+    private boolean curTxDeadlockDetecting(BulkOperation op) {
+        Transaction tx = ctx.kernalContext().cache().transactions().tx();
+
+        if (tx != null && !tx.implicit() &&
```
agura: What is `!tx.implicit()` check's purpose? Deadlock is still possible for implicit transaction.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -5176,6 +5198,91 @@ protected final void validateCacheKeys(Iterable<?> keys) {
...
+            ((ctx.tm().deadlockDetectionEnabled() && tx.timeout() > 0L) ||
+            (tx.concurrency() == OPTIMISTIC && tx.isolation() == SERIALIZABLE) ||
```
agura: Deadlock is impossible for `tx.concurrency() == OPTIMISTIC && tx.isolation() == SERIALIZABLE`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -5176,6 +5198,91 @@ protected final void validateCacheKeys(Iterable<?> keys) {
...
+            (tx.concurrency() == PESSIMISTIC && tx.isolation() == READ_COMMITTED)))
```
agura: Deadlock is possible for any pessimistic transaction. Why do you ignore read-committed?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/local/atomic/GridLocalAtomicCache.java
alamar: I have dropped "Async" indication altogether, it will be misreported often.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -5176,6 +5198,91 @@ protected final void validateCacheKeys(Iterable<?> keys) {
...
+     */
+    protected void checkKeysOrdered(Collection coll, BulkOperation op) {
+        if (coll == null || coll.size() <= 1)
+            return;
+
+        if (coll instanceof SortedSet || coll instanceof GridCacheAdapter.KeySet)
+            return;
+
+        // To avoid false positives, once removeAll() is called, cache will never issue Remove All warnings.
+        if (ctx.lastRemoveAllJobFut().get() != null && op == BulkOperation.REMOVE)
```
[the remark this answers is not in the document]
alamar: Yes, this is a possible case where we can't detect problem - once cache is cleared even once we never report removeAll(). I could not figure out a way to mitigate it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
alamar: I will improve title so it is not redundant.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -5176,6 +5198,91 @@ protected final void validateCacheKeys(Iterable<?> keys) {
...
+        if (m == null || m.size() <= 1)
+            return;
+
+        if (m instanceof SortedMap || m instanceof GridSerializableMap)
+            return;
+
+        if (curTxDeadlockDetecting(op))
+            return;
+
+        LT.warn(log, "Unordered map " + m.getClass().getSimpleName() +
```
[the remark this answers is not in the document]
alamar: will do

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
alamar: If tx is implicit we will actually report tx as not detecting, i.e. warning will be issued.

I took this condition from @agoncharuk I guess

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
alamar: This is what we are checking for.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
alamar: I could not reproduce read-committed deadlocks, and you have confirmed that it should not happen. This is for getAll especially. I can try to reproduce it.

alamar: I can get them for putAll, but not getAll. I will update code to reflect this finding.

PR #6978, opened 2019-10-15
IGNITE-12259 Create new module for support spring-5.2.X and spring-da…

SurkovAleksandr:


modules/spring-data-2.2/src/test/java/org/apache/ignite/testsuites/IgniteSpringData2TestSuite.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Ignite Spring Data 2.2 test suite.
+ */
+@RunWith(Suite.class)
+@Suite.SuiteClasses({
+    IgniteSpringDataCrudSelfTest.class,
+    IgniteSpringDataQueriesSelfTest.class
+})
+public class IgniteSpringData2TestSuite {
```
alamar: Can you please rename this suite to IgniteSpringData22TestSuite or something like that? Per our TC we can't run suites if they have same class name.

Please also rebase on top of fresh master if possible.

SurkovAleksandr: As I understand you already merged PR to master. Should I make this change?

alamar: I think it is no longer relevant :) took me three commits to figure it out, though.

PR #6986, opened 2019-10-17
IGNITE-12302 Test ZookeeperDiscoveryTopologyChangeAndReconnectTest.testDuplicatedNodeId is broken.

NSAmelchev:
Fixed test by disabling registering system views.
Optimized test speed by decreasing server join tries counts.
Added correct SPI exception check.

### DIFF
--- a/modules/zookeeper/src/test/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryTopologyChangeAndReconnectTest.java
+++ b/modules/zookeeper/src/test/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryTopologyChangeAndReconnectTest.java
@@ -75,15 +75,8 @@ public class ZookeeperDiscoveryTopologyChangeAndReconnectTest extends ZookeeperD
 
         cfg.setIncludeEventTypes(EventType.EVTS_ALL);
 
-        if (indexingDisabled) {
+        if (indexingDisabled)
             GridQueryProcessor.idxCls = DummyQueryIndexing.class;
-
-            cfg.setSystemViewExporterSpi(new JmxSystemViewExporterSpi() {
-                @Override protected void register(SystemView<?> sysView) {
-                    // No-op.
-                }
-            });
-        }
 
         return cfg;
     }
@@ -93,6 +86,8 @@ public class ZookeeperDiscoveryTopologyChangeAndReconnectTest extends ZookeeperD
         super.afterTest();
 
         indexingDisabled = false;
+
+        GridQueryProcessor.idxCls = null;
     }
 
     /**
--- a/modules/zookeeper/src/test/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryTopologyChangeAndReconnectTest.java
+++ b/modules/zookeeper/src/test/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryTopologyChangeAndReconnectTest.java
@@ -31,7 +31,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import org.apache.ignite.Ignite;
 import org.apache.ignite.IgniteCache;
-import org.apache.ignite.IgniteCheckedException;
 import org.apache.ignite.IgniteLogger;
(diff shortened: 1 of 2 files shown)


modules/zookeeper/src/test/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryTopologyChangeAndReconnectTest.java
```diff
@@ -562,32 +564,23 @@ public void testForcibleClientFail() throws Exception {
      * @throws Exception If failed.
      */
     @Test
+    @WithSystemProperty(key = IGNITE_SQL_DISABLE_SYSTEM_VIEWS, value = "true")
```
alex-plekhanov: `IgniteH2Indexing` connects to H2 instance by URL based on nodeId, so there can be more errors in indexing here in the future even if system views are disabled.
Perhaps it's better to disable indexing at all for this test (something like `GridQueryProcessor.idxCls = DummyQueryIndexing.class;` before each grid start).

NSAmelchev: @alex-plekhanov, thanks for the suggestion. I have fixed the test. Could you take a look? [TC test suite](https://ci.ignite.apache.org/viewLog.html?buildId=4713749&buildTypeId=IgniteTests24Java8_ZooKeeperDiscovery1) looks fine.

modules/zookeeper/src/test/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryTopologyChangeAndReconnectTest.java
```diff
@@ -62,11 +66,35 @@
...
+
+        return cfg;
     }
-    
+
+    /** {@inheritDoc} */
+    @Override protected void afterTest() throws Exception {
+        super.afterTest();
+
+        indexingDisabled = false;
```
alex-plekhanov: You should also return the value of `GridQueryProcessor.idxCls` to `null` (in case startGrid fails)

modules/zookeeper/src/test/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryTopologyChangeAndReconnectTest.java
```diff
@@ -62,11 +66,35 @@
...
     @Override protected IgniteConfiguration getConfiguration(String igniteInstanceName) throws Exception {
-        return super.getConfiguration(igniteInstanceName).setIncludeEventTypes(EventType.EVTS_ALL);
+        IgniteConfiguration cfg = super.getConfiguration(igniteInstanceName);
+
+        cfg.setIncludeEventTypes(EventType.EVTS_ALL);
+
+        if (indexingDisabled) {
+            GridQueryProcessor.idxCls = DummyQueryIndexing.class;
+
+            cfg.setSystemViewExporterSpi(new JmxSystemViewExporterSpi() {
```
alex-plekhanov: Let's fix `SqlViewExporterSpi` instead of excluding it (looks like a bug in its implementation)

modules/zookeeper/src/test/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryTopologyChangeAndReconnectTest.java
NSAmelchev: Done.

modules/zookeeper/src/test/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryTopologyChangeAndReconnectTest.java
NSAmelchev: @alex-plekhanov, I have fixed it. Take a look, please. [Tests OK.](https://ci.ignite.apache.org/viewLog.html?buildId=4714492&buildTypeId=IgniteTests24Java8_ZooKeeperDiscovery1)

PR #6991, opened 2019-10-21
IGNITE-12301 Free-lists system view

alex-plekhanov:


nizhikov:
Hello, @alex-plekhanov can you, please, move refactoring to the other PR?

nizhikov:
Can we provide information about fields that supports quick filtering in the separate method of the `SystemViewRowAttributeWalker`. 
Looks like this will simplify both final solution and changes in current implementation.

```
public interface SystemViewRowAttributeWalker {
    default Collection<String> quickFilterFields() { 
        return Collections.EMPTY_LIST();       
    }
}
```

nizhikov:
I think we should introduce separate annotations instead of renaming `Order`.
Let's add `QuickFilter`?

modules/indexing/src/main/java/org/apache/ignite/spi/systemview/SystemViewLocal.java
```diff
@@ -56,19 +62,45 @@
...
+            }
+        });
     }
 
     /** {@inheritDoc} */
     @Override public Iterator<Row> getRows(Session ses, SearchRow first, SearchRow last) {
-        Iterator<R> rows = sysView.iterator();
+        Iterator<R> rows;
+
+        if (sysView instanceof FiltrableSystemView) {
```
nizhikov: We should store the result of this check inside some boolean variable.

modules/indexing/src/main/java/org/apache/ignite/spi/systemview/SystemViewLocal.java
```diff
@@ -56,19 +62,45 @@
...
      * @param ctx Kernal context.
      * @param sysView View to export.
      */
     public SystemViewLocal(GridKernalContext ctx, SystemView<R> sysView) {
-        super(sqlName(sysView.name()), sysView.description(), ctx, columnsList(sysView));
+        super(sqlName(sysView.name()), sysView.description(), ctx, indexes(sysView), columns(sysView));
 
         this.sysView = sysView;
+
+        attributeNames = new String[sysView.walker().count()];
```
nizhikov: We should initialize `attributeNames` only if `(sysView instanceof FiltrableSystemView) == true`

modules/core/src/main/java/org/apache/ignite/internal/managers/systemview/walker/DataRegionPagesListViewWalker.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.apache.ignite.internal.processors.cache.persistence.freelist.PagesListView;
+import org.apache.ignite.spi.systemview.view.SystemViewRowAttributeWalker;
+
+/**
+ * {@link PagesListView} attributes walker for data regions.
+ * 
+ * @see PagesListView
+ */
+public class DataRegionPagesListViewWalker implements SystemViewRowAttributeWalker<PagesListView> {
```
nizhikov: We should avoid several walkers for a same row class. Let's create two separate class for new system views of use single walker.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesListView.java
```diff
@@ -0,0 +1,102 @@
...
+package org.apache.ignite.internal.processors.cache.persistence.freelist;
+
+import org.apache.ignite.internal.managers.systemview.walker.ViewAttribute;
+import org.apache.ignite.internal.pagemem.PageIdUtils;
+import org.apache.ignite.spi.systemview.view.SystemView;
+
+/**
+ * Pages-list representation for a {@link SystemView}.
+ */
+public class PagesListView {
```
nizhikov: Let's place this class with the other system view row class in package `org.apache.ignite.spi.systemview.view`

modules/core/src/main/java/org/apache/ignite/internal/managers/systemview/walker/DataRegionPagesListViewWalker.java
alex-plekhanov: Why not? We avoid code duplication when use 2 walkers for 1 view class. 
In case of two different views we also should expose two different methods from free-lists or expose one method with some parameters class and do some unclear class casting.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesListView.java
alex-plekhanov: Most of the fields accessed from PagesListView are protected or package private. I think it's better to put class in the same package than increase access level for these fields.

nizhikov: You can provide getters for these values.

modules/core/src/main/java/org/apache/ignite/internal/managers/systemview/walker/DataRegionPagesListViewWalker.java
nizhikov: Because it's how system view designed -  the walker should be auto-generated.
I can't see a reason to break it.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sys/view/SqlAbstractSystemView.java
```diff
@@ -131,4 +136,9 @@ protected static Column newColumn(String name, int type) {
 
         return sql.toString();
     }
+
+    /** {@inheritDoc} */
+    @Override public String getDropSQL() {
```
nizhikov: I think we shouldn't reregister system view.
We should return empty data when applicable instead of dropping it.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/SchemaManager.java
```diff
@@ -129,7 +128,7 @@
     private final ConcurrentMap<QueryTable, GridH2Table> dataTables = new ConcurrentHashMap<>();
 
     /** System VIEW collection. */
-    private final Set<SqlSystemView> systemViews = new GridConcurrentHashSet<>();
+    private final Map<String, SqlSystemView> sysViews = new ConcurrentHashMap<>();
```
nizhikov: As far as I understand these changes related to support of dropping system view.
We shouldn't support dropping the system view.

Let's revert it and return empty data in case system view temporary can't return a real one.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -5322,6 +5334,58 @@ public CacheConfigurationEnricher enricher() {
         return enricher;
     }
 
+    /**
+     * Pages list view supplier.
+     *
+     * @param filter Filter.
+     */
+    private Iterable<CachePagesListView> pagesListViewSupplier(Map<String, Object> filter) {
+        Object cacheGrpId = filter == null ? null : filter.get(CachePagesListViewWalker.CACHE_GROUP_ID_FILTER);
```
nizhikov: We shouldn't check type of the attributes on the system view level.
It's exporter's responsibility to check parameters type.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -5322,6 +5334,58 @@ public CacheConfigurationEnricher enricher() {
...
+        else
+            cacheGrps = this.cacheGrps.values();
+
+        Object partFilter = filter == null ? null : filter.get(CachePagesListViewWalker.PARTITION_ID_FILTER);
+        Integer partId = partFilter instanceof Integer ? (Integer)partFilter : null;
+
+        Object bucketFilter = filter == null ? null : filter.get(CachePagesListViewWalker.BUCKET_NUMBER_FILTER);
+        Integer bucketNum = bucketFilter instanceof Integer ? (Integer)bucketFilter : null;
+
+        return F.flat(
```
nizhikov: Can we extract `dataStores` variable to make this code more readable:

```
        Iterable<IgniteCacheOffheapManager.CacheDataStore> dataStores =
            F.flat(F.iterator(cacheGrps, grp -> grp.offheap().cacheDataStores(), true));

        return F.flat(F.iterator(dataStores, dataStore -> {
            RowStore rowStore = dataStore.rowStore();

            if (rowStore == null || !(dataStore instanceof GridCacheOffheapManager.GridCacheDataStore))
                return Collections.emptySet();

            PagesList pagesList = (PagesList)rowStore.freeList();

            if (bucketNum != null) {
                return bucketNum >= 0 && bucketNum < pagesList.bucketsCount() ?
                    Collections.singleton(new CachePagesListView(pagesList, bucketNum, dataStore.partId())) :
                    Collections.emptyList();
            }

            return IntStream.range(0, pagesList.bucketsCount())
                .mapToObj(bucket -> new CachePagesListView(pagesList, bucket, dataStore.partId()))
                .collect(Collectors.toList());
        }, true, cacheDataStore -> partId == null || cacheDataStore.partId() == partId));
```

modules/core/src/main/java/org/apache/ignite/internal/managers/systemview/FiltrableSystemViewAdapter.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @NotNull @Override public Iterator<R> iterator(Map<String, Object> filter) {
+        if (filter == null)
+            filter = Collections.emptyMap();
+
+        Iterator<D> dataIter = dataSupplier.apply(filter).iterator();
+
+        return new Iterator<R>() {
```
nizhikov: Let's rewrite it with the regular F.iterator `F.iterator(dataSupplier.apply(filter), rowFunc::apply, true);`?

nizhikov:
LGTM

PR #6997, opened 2019-10-22
IGNITE-12189: Implement correct limit for TextQuery

AMashenkov:
…tureAdapter with int primitive.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
@@ -57,6 +57,9 @@ public abstract class GridCacheQueryFutureAdapter<K, V, R> extends GridFutureAda
     /** Logger reference. */
     private static final AtomicReference<IgniteLogger> logRef = new AtomicReference<>();
 
+    /** */
+    private static final int NO_LIMIT = -1;
+
     /** Logger. */
     protected static IgniteLogger log;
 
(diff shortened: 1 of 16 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
```diff
@@ -57,6 +57,9 @@
     /** Logger reference. */
     private static final AtomicReference<IgniteLogger> logRef = new AtomicReference<>();
 
+    /** */
+    private static final int NO_LIMIT = -1;
```
shuliga: This constant is not corresponding to original description in ticket https://issues.apache.org/jira/browse/IGNITE-12189
>>> Special value: : 0 -> No limit (Integer.MAX_VALUE);

We can proceed with this, then the dscription and all the correspondiing JavaDocs are needed to be updated.

AMashenkov: Actually this magic constant belong to Lucene code and can be changed in any next Lucene version (e.g. to Long.MAX_VALUE). From architectural point Ignite code shouldn't depends on this constant (and any other Lucene abstracts as well) because this may lead us to unwanted changes and issues in future if Lucene guys will change their contracts.

I'd remove any mention of special value Integer.MAX_VALUE from javadocs and bound magic constant usage to method with Lucene call only. In Ignite code we can use 'zero' for 'unlimit'.

What about 'GridCacheQueryFutureAdapter', NO_LIMIT=-1 is a poison value here for private usage only. This can be safely changed to some flag that will indicate 'unlimited' e.g. if negative values will make sense at once. Thanks to incapsulation principle.

shuliga: I have updated the original [PR](https://github.com/apache/ignite/pull/6917) with special flag `limitReached` and hav corrected the JavaDocs. Please review.

PR #7004, opened 2019-10-24
IGNITE-1606 NPE during node stop due to nullified logger in TcpCommunicationSpi

x-kreator:


modules/core/src/main/java/org/apache/ignite/internal/processors/resource/GridResourceIoc.java
```diff
@@ -574,6 +574,16 @@ public boolean inject(Object target,
             ResourceAnnotation.IGNITE_INSTANCE,
             ResourceAnnotation.LOGGER,
             ResourceAnnotation.SERVICE
+        ),
+
+        /**
+         * Same as {@link #GENERIC} excluding {@link ResourceAnnotation#LOGGER}.
+         */
+        GENERIC_CLEANABLE(
```
nizhikov: Can we just exclude `LOGGER` from `GENERIC` set instead of the creation of a new set?
New set is error-prone.
Some contributor can forget to add new `ReourceAnnotation` to it.
This will lead to hard to find issues.

x-kreator: Is it right that you suggest to add a new AnnotationSet.LOGGER with ResourceAnnotation.LOGGER only and use it at the points where AnnotationSet.GENERIC is injecting but don't use one at cleanup?

nizhikov: I think we can just use existing LOGGER annotation but don't clean up it.

x-kreator: Thanks for review, code was updated according your recommendation.

PR #7008, opened 2019-10-25
IGNITE-12261 Issue with adding nested index dynamically

kotari4u:
This will ensure we are using fullName instead of lastAttribute in nested attribute
Ex: Address.street
With out this fix, field name will be street
with this fix, field name will be Address.street. It will help when we query from other node


kotari4u:
https://github.com/apache/ignite/pull/7016
Created another PR with all the changes at one place. Hence closing this

kotari4u:
https://github.com/apache/ignite/pull/7016
Created another PR with all the changes at one place. Hence closing this

PR #7009, opened 2019-10-25
IGNITE-12261 Issue with adding nested index dynamically

kotari4u:
Instead of using default constructor to build binary property, use utils method which will build binary property properly with proper aliases


kotari4u:
https://github.com/apache/ignite/pull/7016
Created another PR with all the changes at one place. Hence closing this

PR #7045, opened 2019-11-11
IGNITE-12366: Cancel file transmission on a node-receiver

Mmuzaf:


modules/core/src/main/java/org/apache/ignite/internal/managers/communication/TransmissionCancelException.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.apache.ignite.internal.managers.communication;
+
+/**
+ * Exception is used to cancel a file transmission operation on the receiver. This exception may be thrown
+ * at anytime during session handling by a {@link TransmissionHandler} to gracefully interrupt the transmission
+ * session on a node-sender.
+ */
+public class TransmissionCancelException extends RuntimeException {
```
xtern: I think that TransmissionCancelledException sounds better.

xtern: Also, In my opinion, it's better to inherit this exception from the IgniteException.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/GridIoManager.java
```diff
@@ -3289,8 +3289,15 @@ public void send(
...
 
-                throw new IgniteCheckedException("Exception while sending file [rmtId=" + rmtId +
-                    ", file=" + file.getName() + ", sesKey=" + sesKey + ", retries=" + retries + ']', e);
+                if (X.hasCause(e, TransmissionCancelException.class)) {
+                    throw new TransmissionCancelException("File transmission has been cancelled on the remote node " +
+                        "[rmtId=" + rmtId + ", file=" + file.getName() + ", sesKey=" + sesKey +
+                        ", retries=" + retries + ", cause='" + e.getCause().getMessage() + "']");
+                }
+                else {
+                    throw new IgniteCheckedException("Exception while sending file [rmtId=" + rmtId +
```
xtern: Why not just re-throw `e` here?

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/GridIoManager.java
```diff
@@ -3289,8 +3289,15 @@ public void send(
             catch (IgniteCheckedException e) {
                 closeChannelQuiet();
 
-                throw new IgniteCheckedException("Exception while sending file [rmtId=" + rmtId +
-                    ", file=" + file.getName() + ", sesKey=" + sesKey + ", retries=" + retries + ']', e);
+                if (X.hasCause(e, TransmissionCancelException.class)) {
+                    throw new TransmissionCancelException("File transmission has been cancelled on the remote node " +
+                        "[rmtId=" + rmtId + ", file=" + file.getName() + ", sesKey=" + sesKey +
+                        ", retries=" + retries + ", cause='" + e.getCause().getMessage() + "']");
```
xtern: It seems that `e.getCause()`may not always contain expected exception (TransmissionCancelException).

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/TransmissionCancelException.java
Mmuzaf: Fixed.

Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/GridIoManager.java
Mmuzaf: Fixed.

PR #7047, opened 2019-11-12
IGNITE-10760 Fix confusing message about system worker blocking

alievmirza:


modules/core/src/main/java/org/apache/ignite/internal/worker/WorkersRegistry.java
```diff
@@ -225,8 +225,9 @@ public void setSystemWorkerBlockedTimeout(long val) {
 
                         if (worker0 != null && worker0 == worker) {
                             log.error("Blocked system-critical thread has been detected. " +
-                                "This can lead to cluster-wide undefined behaviour " +
-                                "[threadName=" + worker.name() + ", blockedFor=" + heartbeatDelay / 1000 + "s]");
+                                    "This can lead to cluster-wide undefined behaviour " +
+                                    "[workerName=" + worker.name() + ", threadName=" + worker.runner().getName() +
```
agura: `worker.runner()` can return `null` in some cases, so it makes sense to add additional null checking.

alievmirza: As far as I can see from the code, the check has already been done at https://github.com/apache/ignite/blob/c1dba9c80cb48ead9a3d6b1e015d54c81503bdec/modules/core/src/main/java/org/apache/ignite/internal/worker/WorkersRegistry.java#L208
So it seems to me that we don't need to do additional check. 

I can change `worker.runner()` to `runner` to make code cleaner, if it needed.

agura:
It's classic check-then-act error. In your case NullPointerException is still possible. But using `runner` instead of `worker.runner()` solves the problem.

modules/core/src/main/java/org/apache/ignite/internal/worker/WorkersRegistry.java
agura: It's classic check-then-act error. In your case NullPointerException is still possible. But using runner instead of worker.runner() solves the problem.

PR #7051, opened 2019-11-14
IGNITE-12373: skip recovery descriptor reservation for a channel connection

Mmuzaf:
…ection


modules/core/src/test/java/org/apache/ignite/internal/managers/communication/GridIoManagerFileTransmissionSelfTest.java
```diff
@@ -840,6 +840,88 @@ public void testChunkHandlerCancelTransmission() throws Exception {
         }
     }
 
+    /**
+     * @throws Exception If fails.
+     */
+    @Test
+    public void testFileHandlerCrossConnections() throws Exception {
```
NSAmelchev: This test does not need configured persistence region. This will significantly reduce the test execution time from ~5-10 seconds to 200 milliseconds.

Mmuzaf: Fixed

PR #7059, opened 2019-11-18
IGNITE-12224: SQL query & SQL query history system views.

nizhikov:


modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryHistoryTracker.java
```diff
@@ -114,37 +112,37 @@ else if (removeLink(node)) {
...
-    private boolean removeLink(ConcurrentLinkedDeque8.Node<QueryHistoryMetrics> node) {
+    private boolean removeLink(ConcurrentLinkedDeque8.Node<QueryHistory> node) {
         return evictionQueue.unlinkx(node);
     }
 
     /**
-     * Gets SQL query history. Size of history could be configured via {@link
-     * IgniteConfiguration#setSqlQueryHistorySize(int)}
+     * Gets SQL query history. Size of history could be configured via
+     * {@link IgniteConfiguration#setSqlQueryHistorySize(int)}.
```
NSAmelchev: The link doesn't refer correctly without importing class.: `import org.apache.ignite.configuration.IgniteConfiguration;`

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryHistory.java
```diff
@@ -162,13 +162,13 @@ public long lastStartTime() {
      * @param updatedLink New link which should be set.
      * @return {@code true} If link has been updated.
      */
-    public boolean replaceLink(ConcurrentLinkedDeque8.Node<QueryHistoryMetrics> expLink,
-        ConcurrentLinkedDeque8.Node<QueryHistoryMetrics> updatedLink) {
+    public boolean replaceLink(ConcurrentLinkedDeque8.Node<QueryHistory> expLink,
+        ConcurrentLinkedDeque8.Node<QueryHistory> updatedLink) {
```
NSAmelchev: Seems  the only `Node` class can be imported:  `import org.jsr166.ConcurrentLinkedDeque8.Node;`

PR #7069, opened 2019-11-24
IGNITE-9913 PR #2

anton-vinogradov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -1653,7 +1709,9 @@ private void waitPartitionRelease(boolean distributed, boolean doRollback) throw
             if (distributed)
                 releaseLatch = cctx.exchange().latch().getOrCreate(DISTRIBUTED_LATCH_ID, initialVersion());
 
-            partReleaseFut = cctx.partitionReleaseFuture(initialVersion());
+            partReleaseFut = context().exchangeFreeSwitch() ?
```
NSAmelchev: I think we should update javadoc because the method will not wait for all ongoing updates in case of `exchangeFreeSwitch`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ExchangeContext.java
```diff
@@ -100,6 +106,13 @@ public boolean fetchAffinityOnJoin() {
         return fetchAffOnJoin;
     }
 
+    /**
+     * @return {@code True} if it's safe to perform PME-free switch.
+     */
+    public boolean exchangeFreeSwitch() {
```
NSAmelchev: Should we consider this PME non-blocking cache operations? If so, we need to fix the metric (see `GridDhtPartitionsExchangeFuture#currentPMEDuration`).

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/GridDhtLocalPartition.java
```diff
@@ -500,7 +500,9 @@ private void release0(int sizeChange) {
             if (this.state.compareAndSet(state, newState)) {
                 // If no more reservations try to continue delayed renting.
                 if (reservations == 0) {
-                    if (delayedRenting)
+                    if (delayedRentingTopVer != 0 &&
+                        // Prevents delayed renting on topology which expects owninership.
```
NSAmelchev: Typo in `owninership`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ExchangeContext.java
anton-vinogradov: Since exchange-free switch blocks upcoming operations during the tx recovery, seems, it's correct to consider it's blocking during this period.
In other words exchange-free does not means non-blocking, but period is short
Blocking period can be shortened in future by recovery procedure refactoring.
Also we, theoretically, can block only affected nodes (we have some cornercases here, so it should be additional fix)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheAffinitySharedManager.java
```diff
@@ -1180,36 +1175,19 @@ public void onChangeAffinityMessage(
...
-                            ']';
-
-                        assignment.set(part, nodes);
-                    }
+                if (!aff.partitionPrimariesDifferentToIdeal(affTopVer).isEmpty())
+                    aff.initialize(topVer, aff.idealAssignmentRaw());
+                else {
+                    assert aff.assignments(aff.lastVersion()).equals(aff.idealAssignmentRaw()) :
+                        "Not an ideal distribution duplication attempt on LAA [grp=" + aff.cacheOrGroupName() +
+                            ", lastAffinity=" + aff.lastVersion() + ", cacheAffinity=" + aff.cachedVersions() + "]";
```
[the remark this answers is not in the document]
anton-vinogradov: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheAffinitySharedManager.java
```diff
@@ -1180,36 +1175,19 @@ public void onChangeAffinityMessage(
...
 
-                    aff.initialize(topVer, assignment);
-                }
-                else
                     aff.clientEventTopologyChange(exchFut.firstEvent(), topVer);
+                }
+
+                assert aff.assignments(topVer).equals(aff.idealAssignmentRaw()) :
+                    "Not an ideal final distribution on LAA [grp=" + aff.cacheOrGroupName() + ", lastAffinity=" +
+                        aff.lastVersion() + "]";
```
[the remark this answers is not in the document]
anton-vinogradov: Agree.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -364,6 +364,9 @@
     /** Partitions scheduled for clearing before rebalance for this topology version. */
     private Map<Integer, Set<Integer>> clearingPartitions;
 
+    /** This future finished with 'cluster is fully rebalanced' state. */
+    private volatile boolean rebalanced;
```
agoncharuk: I have a concern about this flag and surrounding logic:
Consider the following scenario
1) A fully rebalanced cluster, latest future has this flag set to true
2) A node leaves, optimization is possible, the flag is set to true. The node which left the cluster only has backup partitions assigned to it
3) The node returns to the cluster. Rebalance is needed, but no primary should be assigned to this node according to the ideal assignment. Q1: will waitRebalanceInfo be non-null in this case?
4) Rebalance is finished. Given that no primary node change should happen, Q2: will late affinity assignment exchange be triggered?

If the answer is yes for both questions, looks like all is good, even though the late affinity assignment could have been avoided.
If the answers are yes, no - the optimization will be lost for the next node leave event (the rebalanced flag is false, but an exchange that flips the flag is never triggered)
If the answers are no, no - the flag is named improperly because the rebalance for backups is happening, but the flag is true
(the combination no, yes does not make sense)

anton-vinogradov: @agoncharuk 
Both is yes. 
Explicit test added, please check.
See https://github.com/apache/ignite/pull/7069/commits/265b35a0963310a4de018dfb2f7ea699d8956e8c for details.

agoncharuk: Got it, looks good then

PR #7079, opened 2019-11-26
IGNITE-12188  Metric CacheGroupMetrics.IndexBuildCountPartitionsLeft doesn't work correctly if there is more then one cache in cache group

NSAmelchev:
Fix `IndexBuildCountPartitionsLeft` metric


nizhikov:
@NSAmelchev LGTM.

@alex-plekhanov Can you, please, take a look too.

PR #7080, opened 2019-11-27
IGNITE-12053 Total time threads parked if checkpoint throttling occurred

NSAmelchev:
Add the `totalThrottleDuration` metric.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataRegionMetricsImpl.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataRegionMetricsImpl.java
@@ -97,6 +97,9 @@ public class DataRegionMetricsImpl implements DataRegionMetrics {
     /** */
     private final HitRateMetric pageReplaceAge;
 
+    /** Total threads throttle duration in milliseconds. */
+    private final LongAdderMetric totalThrottleDuration;
+
     /** */
     private final DataRegionConfiguration memPlcCfg;
 
@@ -180,6 +183,10 @@ public class DataRegionMetricsImpl implements DataRegionMetrics {
             dataRegionMetricsProvider::emptyDataPages,
             "Calculates empty data pages count for region. It counts only totally free pages that can be reused " +
                 "(e. g. pages that are contained in reuse bucket of free list).");
+
+        totalThrottleDuration = mreg.longAdderMetric("totalThrottleDuration",
+            "Total threads throttle duration in milliseconds. The Ignite throttles threads that generate " +
+                "dirty pages during the ongoing checkpoint.");
     }
 
     /** {@inheritDoc} */
@@ -559,6 +566,12 @@ public class DataRegionMetricsImpl implements DataRegionMetrics {
         evictRate.reset();
         pageReplaceRate.reset();
         pageReplaceAge.reset();
+    }
+
+    /** @param duration Duration to add to {@code totalThrottleDuration} metric. */
+    public void addThrottleDuration(long duration) {
+        if (metricsEnabled)
+            totalThrottleDuration.add(duration);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteSpeedBasedThrottle.java
```diff
@@ -225,6 +226,11 @@ public PagesWriteSpeedBasedThrottle(
             doPark(throttleParkTimeNs);
         }
 
+        DataRegionMetricsImpl metrics = pageMemory.metrics();
+
+        if (metrics != null)
```
Mmuzaf: Can the metric instance be `null`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataRegionMetricsImpl.java
```diff
@@ -561,6 +568,12 @@ public void clear() {
         pageReplaceAge.reset();
     }
 
+    /** @param time Time to add to {@code totalThrottleTime} metric. */
```
Mmuzaf: I think we should add to javadoc a measurement metric unit.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataRegionMetricsImpl.java
```diff
@@ -180,6 +183,10 @@ public DataRegionMetricsImpl(DataRegionConfiguration memPlcCfg,
             dataRegionMetricsProvider::emptyDataPages,
             "Calculates empty data pages count for region. It counts only totally free pages that can be reused " +
                 "(e. g. pages that are contained in reuse bucket of free list).");
+
+        totalThrottleDuration = mreg.longAdderMetric("totalThrottleDuration",
```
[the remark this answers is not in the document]
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataRegionMetricsImpl.java
```diff
@@ -180,6 +183,10 @@ public DataRegionMetricsImpl(DataRegionConfiguration memPlcCfg,
...
+            "Total threads throttle duration in milliseconds. The Ignite throttles threads that generate " +
```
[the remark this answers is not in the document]
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataRegionMetricsImpl.java
NSAmelchev: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteSpeedBasedThrottle.java
NSAmelchev: I have implemented NO-OP metrics for some tests. Now it can't be null.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PageMemoryImpl.java
```diff
@@ -283,7 +283,7 @@
     private long[] sizes;
 
     /** Memory metrics to track dirty pages count and page replace rate. */
-    private DataRegionMetricsImpl memMetrics;
+    private final DataRegionMetricsImpl memMetrics;
```
Mmuzaf: We need to add an assertion on `memMetrics` since it cannot be `null` value

NSAmelchev: I think this will be an unnecessary check. Because it cannot be null in prod files. If it will be null, we will get NPE much earlier. This is only possible in tests, e.g if you create a fake class through the mock.

PR #7090, opened 2019-11-29
IGNITE-12376 Change level of "TCP client created ..." log messages to…

SurkovAleksandr:


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -3128,7 +3128,7 @@ else if (log.isDebugEnabled())
 
             if (time > CONNECTION_ESTABLISH_THRESHOLD_MS) {
                 if (log.isInfoEnabled())
```
alex-plekhanov: We shouldn't check "INFO" level if we now log to "WARNING" level

nizhikov: I propose following message text
`Slow TCP client created[client=YYY,threshold=XXX,duration=ZZZms]`

What do you think?

SurkovAleksandr: You're right.

PR #7091, opened 2019-11-29
IGNITE-12406 Refactor GridCommonAbstractTest#findKeys to use enum

SurkovAleksandr:


modules/core/src/test/java/org/apache/ignite/testframework/junits/common/KeyType.java
```diff
@@ -0,0 +1,20 @@
+package org.apache.ignite.testframework.junits.common;
+
+/**
+ * Defined type of key
```
nizhikov: Please, format java doc properly.
Sentence should end with the dot.

modules/core/src/test/java/org/apache/ignite/testframework/junits/common/KeyType.java
```diff
@@ -0,0 +1,20 @@
...
+ * */
+public enum KeyType {
+    PRIMARY(0),
```
nizhikov: Do we really need `getKeyIndex`?
We can use `Enum#ordinal` instead of `getKeyIndex`.

modules/core/src/test/java/org/apache/ignite/testframework/junits/common/KeyType.java
```diff
@@ -0,0 +1,20 @@
+package org.apache.ignite.testframework.junits.common;
+
+/**
+ * Defined type of key
+ * */
+public enum KeyType {
```
nizhikov: Let's move this enum inside `GridCommonAbstractTest`

modules/core/src/test/java/org/apache/ignite/testframework/junits/common/KeyType.java
```diff
@@ -0,0 +1,20 @@
...
+    PRIMARY(0),
+    BACKUP(1),
+    NEAR(2);
```
nizhikov: Let's rename `NEAR` to `ANY`

modules/core/src/test/java/org/apache/ignite/testframework/junits/common/GridCommonAbstractTest.java
```diff
@@ -1140,16 +1142,19 @@ public static ClusterNode localNode(IgniteCache<?, ?> cache) {
...
+                                ok = aff.isPrimary(node0, key);
+                                break;
+                            case BACKUP:
+                                ok = aff.isBackup(node0, key);
+                                break;
+                            case NEAR:
+                                ok = !aff.isPrimaryOrBackup(node0, key);
+                                break;
+                            default:
+                                fail();
```
anton-vinogradov: newline here (please fix similar too).
see explanation here: https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines#CodingGuidelines-SemanticUnits

modules/core/src/test/java/org/apache/ignite/testframework/junits/common/KeyType.java
SurkovAleksandr: Ok.

modules/core/src/test/java/org/apache/ignite/testframework/junits/common/KeyType.java
SurkovAleksandr: It’s not a good practice to be tied to the order of enums.

modules/core/src/test/java/org/apache/ignite/testframework/junits/common/KeyType.java
SurkovAleksandr: The class `GridCommonAbstractTest `is big enough and class `KeyType `is used in other packages.
What do you say?

modules/core/src/test/java/org/apache/ignite/testframework/junits/common/GridCommonAbstractTest.java
SurkovAleksandr: Ok.

modules/core/src/test/java/org/apache/ignite/testframework/junits/common/KeyType.java
anton-vinogradov: We should provide solutions as small/simple as possible.
In case you're not required this now, code should not contain this.
For every code fragment at PR, you should be able to provide proper explanation"why it's here", "good practice" is not acceptable explanation.

BTW, `keyType.getKeyIndex() * 100_00` looks odd to me, can we refactor this to?

modules/core/src/test/java/org/apache/ignite/testframework/junits/common/GridCommonAbstractTest.java
```diff
@@ -1140,16 +1142,26 @@ public static ClusterNode localNode(IgniteCache<?, ?> cache) {
...
+                                ok = aff.isPrimary(node0, key);
 
-                            return false;
+                                break;
+
+                            case BACKUP:
+                                ok = aff.isBackup(node0, key);
+
+                                break;
+
```
anton-vinogradov: no need to have newline here :)

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteOptimisticTxSuspendResumeTest.java
```diff
@@ -252,17 +253,17 @@ private void doCheckSuspendTxAndResume(Ignite node, Map<String, List<List<Intege
...
                 if (cfg.getCacheMode() == LOCAL)
                     continue;
 
-                if (type == 1 && cfg.getCacheMode() == PARTITIONED && cfg.getBackups() == 0)
+                if (keyType == KeyType.BACKUP && cfg.getCacheMode() == PARTITIONED && cfg.getBackups() == 0)
                     continue;
 
-                if (type == 2 && cfg.getCacheMode() == REPLICATED)
+                if (keyType == KeyType.NEAR && cfg.getCacheMode() == REPLICATED)
                     continue;
```
anton-vinogradov: Looks like this block should be simplified to single if?

modules/core/src/test/java/org/apache/ignite/testframework/junits/common/KeyType.java
anton-vinogradov: We should avoid names like `KeyType`. 
Each class, even internal, should be called properly according to it's semantic.

modules/core/src/test/java/org/apache/ignite/testframework/junits/common/KeyType.java
```diff
@@ -0,0 +1,18 @@
+package org.apache.ignite.testframework.junits.common;
+
+/** Defined type of key. */
```
anton-vinogradov: Comments should be useful. 
Please provide description of class features.

PR #7098, opened 2019-12-02
IGNITE-12412 Fixed ABA issue for PagePool#releaseFreePage

agoncharuk:


modules/core/src/main/java/org/apache/ignite/internal/util/GridUnsafe.java
```diff
@@ -1361,6 +1361,42 @@ public static boolean compareAndSwapLong(Object obj, long off, long exp, long up
...
     }
 
+    /**
+     * Atomically increments value stored in an integer pointed by {@code ptr}.
+     *
+     * @param ptr Pointer to an integer.
+     * @return Updated value.
+     */
+    public static int incrementAndGetInt(long ptr) {
+        while (true) {
```
agura: Method body could be replaced by `Unsafe.getAndAddInt(null, ptr, 1) + 1`.

modules/core/src/main/java/org/apache/ignite/internal/util/GridUnsafe.java
```diff
@@ -1361,6 +1361,42 @@ public static boolean compareAndSwapLong(Object obj, long off, long exp, long up
...
+    }
+
+    /**
+     * Atomically increments value stored in an integer pointed by {@code ptr}.
+     *
+     * @param ptr Pointer to an integer.
+     * @return Updated value.
+     */
+    public static int decrementAndGetInt(long ptr) {
+        while (true) {
```
agura: Method body could be replaced by `Unsafe.getAndAddInt(null, ptr, -1) - 1`.

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -8720,6 +8720,31 @@ public static String fl(String msg, Object... args) {
         return sb.toString();
     }
 
+    /**
+     * Round up the argument to the next highest power of 2;
+     *
+     * @param v Value to round up.
+     * @return Next closest power of 2.
+     */
+    public static int nextPowerOf2(int v) {
```
agura: `Integer.highestOneBit(value) << 1` ?

agoncharuk: Not quite because it does not check if the integer is already is a power of 2.
Replaced with 1 << (32 - Integer.numberOfLeadingZeros(v - 1))

modules/core/src/main/java/org/apache/ignite/internal/util/GridUnsafe.java
agoncharuk: Done.

modules/core/src/main/java/org/apache/ignite/internal/util/GridUnsafe.java
agoncharuk: Done.

dspavlov:
IGNITE-12412 | Resolved
-- | --

PR #7103, opened 2019-12-05
IGNITE-12419 Fixed JCache TCK

sk0x50:


modules/core/pom.xml
```diff
@@ -402,6 +402,26 @@
             </properties>
 
             <dependencies>
+                <dependency>
+                    <groupId>org.mockito</groupId>
+                    <artifactId>mockito-core</artifactId>
```
Mmuzaf: We already have `mockito-all` dependency which seems to cover all needs, do we need another one?

parent/pom.xml
```diff
@@ -82,7 +82,7 @@
         <jaxb.impl.version>2.1.14</jaxb.impl.version>
         <javassist.version>3.20.0-GA</javassist.version>
         <javax.cache.bundle.version>1.0.0_1</javax.cache.bundle.version>
-        <javax.cache.tck.version>1.0.1</javax.cache.tck.version>
+        <javax.cache.tck.version>1.1.0</javax.cache.tck.version>
```
Mmuzaf: Why the version has changed? Why not `1.1.1` if it's necessary?

modules/core/pom.xml
sk0x50: The main reason is that `mockito-all` includes some dependencies, one of them is `harmcrest`. Unfortunately, that version conflicts with the version which is used by AI. Moreover, “mockito-all” distribution has been discontinued in Mockito 2.*.

parent/pom.xml
sk0x50: This version is already used by TeamCity (the value is specified via an environment variable) in accordance with https://issues.apache.org/jira/browse/IGNITE-8548. I just aligned the version with the value which is actually used.

modules/core/pom.xml
Mmuzaf: I've checked `>mvn dependency:tree -Dverbose` command for the `ignite-core` module.
It seems we already have `hamcrest` dependency from `junit` module.
{code}
[INFO] +- junit:junit:jar:4.11:test
[INFO] |  \- org.hamcrest:hamcrest-core:jar:1.3:test
{code}

Also,  it seems `mockito-core` is not required here. I've removed these lines from my local branch an it seems it works fine.
Am I missing something?

modules/core/src/test/java/org/apache/ignite/internal/processors/security/cache/EntryProcessorPermissionCheckTest.java
```diff
@@ -79,7 +79,7 @@ private void runOperation(Ignite verifierNode, BiConsumer<String, T2<String, Int
 
         c.accept(CACHE_NAME, entry);
 
-        assertThat(verifierNode.cache(CACHE_NAME).get(entry.getKey()), is(entry.getValue()));
+        assertThat((Integer)(verifierNode.cache(CACHE_NAME).get(entry.getKey())), is(entry.getValue()));
```
Mmuzaf: I think we should not cast here, but parametrize cache directly:
`verifierNode.<String, Integer>cache(CACHE_NAME).get(entry.getKey())`

modules/core/pom.xml
sk0x50: Well, the problem was the version of hamcrest which is bundled with mockito-all and mockito-core (it is 1.1 if I am not mistaken),
and that version conflicted with the version included into jcache-tests (1.3 actually). So, I just excluded the former.

I just removed dependency from mockito-core and re-checked tck tests with `mvn clean test -P-release,jcache-tck -pl :ignite-core -am` and it works. So, it can be updated. Will do.

modules/core/src/test/java/org/apache/ignite/internal/processors/security/cache/EntryProcessorPermissionCheckTest.java
sk0x50: ok

PR #7107, opened 2019-12-06
IGNITE-11923 add IgniteMXBean attributes to metrics

zkillingjar:


modules/core/src/main/java/org/apache/ignite/mxbean/IgniteMXBean.java
```diff
@@ -27,36 +27,152 @@
...
+    public String ATTR_DESC_READ_ONLY_MODE = "Cluster read-only mode status.";
+
+    /** */
+    public String ATTR_DESC_READ_ONLY_MODE_DURATION = "Duration of read-only mode enabled on cluster.";
+
     /**
      * Gets string presentation of the version.
      *
      * @return String presentation of the version.
+     * @deprecated Use metric API.
```
nizhikov: Let's replace this and all others deprecation comments with the comment that equal all over the code:\

` * @deprecated Use {@link GridMetricManager} instead.`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -424,7 +424,7 @@ private void requestPartitions(final RebalanceFuture fut, GridDhtPreloaderAssign
             return;
         }
 
-        if (!ctx.kernalContext().grid().isRebalanceEnabled()) {
+        if (!ctx.isRebalanceEnabled()) {
```
nizhikov: Why this change?

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -596,14 +596,24 @@ public IgniteKernal(@Nullable GridSpringResourceContext rsrcCtx) {
 
     /** {@inheritDoc} */
     @Override public boolean isNodeInBaseline() {
-        ClusterNode locNode = localNode();
+        ctx.gateway().readLockAnyway();
```
nizhikov: Why this change?

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxExporterSpiTest.java
```diff
@@ -730,6 +735,105 @@ public void testScanQuery() throws Exception {
         }
     }
 
+    /**
+     *
```
nizhikov: nit: Let's write this comment in one line.

modules/core/src/main/java/org/apache/ignite/mxbean/IgniteMXBean.java
```diff
@@ -27,36 +27,152 @@
  */
 @MXBeanDescription("MBean that provides access to Kernal information.")
 public interface IgniteMXBean {
+    /** */
+    public String ATTR_DESC_FULL_VER = "String presentation of the Ignite version.";
```
nizhikov: Can we rename this constant to `FULL_VER_DESC`?
Other constants introduced by this change should be renamed with this pattern, also.

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxExporterSpiTest.java
```diff
@@ -730,6 +735,105 @@ public void testScanQuery() throws Exception {
...
+
+        int cnt = 36;
+
+        for (MBeanAttributeInfo info : mbn.getMBeanInfo().getAttributes()) {
+            cnt--;
+
+            assertFalse(F.isEmpty(info.getDescription()));
+            Object val = mbn.getAttribute(info.getName());
+
+            switch (info.getName()) {
```
nizhikov: Can we rewrite this test with the more dumb fashion? 

```
assertEquals(36, mbn.getMBeanInfo().getAttributes().size());

assertFalse(F.isEmpty(mbn.getAttribute("fullVersion")));
assertFalse(F.isEmpty(mbn.getAttribute("copyright")));
```

And so on...

modules/core/src/main/java/org/apache/ignite/mxbean/IgniteMXBean.java
zkillingjar: done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
zkillingjar: Removed it.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
zkillingjar: `localNode()` takes the gw lock internally and the latter throws an exception if state is not `STARTED`. As a result the test `ReadMetricsOnNodeStartUpTest` was failing.
Rechecked it now and it doesn't reproduce anymore. Most likely due to the fact that I moved the call to `registerMetrics()` past the place where gw state is set to `STARTED`.

So I reverted the change for isNodeBaselne() back to original.

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxExporterSpiTest.java
zkillingjar: done

modules/core/src/main/java/org/apache/ignite/mxbean/IgniteMXBean.java
zkillingjar: done

modules/core/src/test/java/org/apache/ignite/internal/metric/JmxExporterSpiTest.java
zkillingjar: done

PR #7141, opened 2019-12-13
IGNITE-12450 Transaction operations metrics

NSAmelchev:
Add commitTime, rollbackTime metrics


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheAbstractMetricsSelfTest.java
```diff
@@ -1460,6 +1460,48 @@ public void testRemoveTime() {
         assertEquals(1, Arrays.stream(m.value()).filter(v -> v == 1).count());
     }
 
+    /** */
+    @Test
+    public void testCommitTime() {
```
nizhikov: Can we move these tests in `GridCacheTransactionalAbstractMetricsSelfTest`?

NSAmelchev: Done.

PR #7147, opened 2019-12-16
IGNITE-12442: Fix unnecessary sync on PagesCache.

zstan:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesList.java
```diff
@@ -2033,6 +2033,9 @@ public synchronized boolean add(long pageId) {
             int stripeIdx = (int)pageId & (STRIPES_COUNT - 1);
 
             synchronized (stripeLocks[stripeIdx]) {
+                if (size >= MAX_SIZE)
```
alex-plekhanov: It's a redundant check. There already exists a stripe size check under the lock.

zstan: not this, but upper, if so - i agree.

PR #7159, opened 2019-12-18
IGNITE-12283 Access restriction to IgniteKernal

dgarus:


modules/core/src/test/java/org/apache/ignite/internal/processors/security/sandbox/IgnitionComponentProxyTest.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /**
+     * @param s Supplier that returns a collection of Ignite instances.
+     */
+    private void checkCollection(Supplier<Collection<Ignite>> s) {
+        Ignite srv = grid(SRV);
+
+        Ignite clnt = grid(CLNT);
+        //Checks that inside the sandbox we should get a proxied instance of Ignite.
```
alex-plekhanov, line 173: NL before comment and space after `//`

modules/core/src/test/java/org/apache/ignite/internal/processors/security/sandbox/IgnitionComponentProxyTest.java
```diff
@@ -0,0 +1,191 @@
...
+        clnt.compute(clnt.cluster().forNodeId(srv.cluster().localNode().id()))
+            .broadcast(() -> {
+                Collection<Ignite> nodes = s.get();
+
+                for (Ignite node : nodes) {
+                    assertTrue(Proxy.isProxyClass(node.getClass()));
+                    assertFalse(node instanceof IgniteEx);
+                }
+            });
+        //If we run outside of the sandbox, we should get an instance of IgniteEx.
```
alex-plekhanov, line 183: NL before comment and space after `//`

modules/core/src/test/java/org/apache/ignite/testframework/junits/GridAbstractTest.java
```diff
@@ -1439,7 +1439,7 @@ protected void waitForDiscovery(Ignite... ignites) throws IgniteCheckedException
      */
     protected IgniteEx grid(String name) {
         if (!isRemoteJvm(name))
-            return (IgniteEx)G.ignite(name);
+            return (IgniteEx)IgnitionEx.grid(name);
```
alex-plekhanov, line 1442: Why do we need it? I think it can fail only if we cast to `IgniteEx` inside test and can affect only sandbox tests, all other tests should work well. If we cast to `IgniteEx` inside sandbox test we should better rewrite this logic.

modules/core/src/test/java/org/apache/ignite/internal/processors/security/sandbox/IgnitionComponentProxyTest.java
```diff
@@ -0,0 +1,191 @@
...
+        Ignite srv = grid(SRV);
+
+        Ignite clnt = grid(CLNT);
+        //Checks that inside the sandbox we should get a proxied instance of Ignite.
+        clnt.compute(clnt.cluster().forNodeId(srv.cluster().localNode().id()))
+            .broadcast(() -> {
+                Collection<Ignite> nodes = s.get();
+
+                for (Ignite node : nodes) {
+                    assertTrue(Proxy.isProxyClass(node.getClass()));
```
alex-plekhanov, line 179: Proxy it's not a goal itself. We should check that we can perform some operations using these ignite instances which we can't perform without proxy.

modules/core/src/main/java/org/apache/ignite/Ignition.java
```diff
@@ -538,7 +556,43 @@ public static Ignite ignite(@Nullable String name) throws IgniteIllegalStateExce
...
+    private static Ignite sandboxIgniteProxy(Ignite ignite) {
+        return sandboxIgniteProxy(ignite, true);
+    }
+
+    /**
+     * @param ignite Ignite.
+     * @return Ignite component proxy if the Ignite Sandbox is enabled.
+     */
+    private static Ignite sandboxIgniteProxy(Ignite ignite, boolean doInsideSandboxCheck) {
+        if (ignite instanceof IgniteEx && (!doInsideSandboxCheck || insideSandbox())) {
```
alex-plekhanov, line 575: Do we really need `IgniteEx` check? I can't imagine in which cases `IgnitionEx` will return something else.

modules/core/src/main/java/org/apache/ignite/Ignition.java
```diff
@@ -490,7 +498,17 @@ public static Ignite ignite() throws IgniteIllegalStateException {
      * @return List of all grids started so far.
      */
     public static List<Ignite> allGrids() {
-        return IgnitionEx.allGrids();
+        List<Ignite> res = IgnitionEx.allGrids();
+
+        if (res == null || res.isEmpty())
+            return res;
+
+        Ignite ignite = res.get(0);
```
alex-plekhanov, line 506: We should check every node (or none of them, see comment about `IgniteEx`), but not only the first node.

modules/core/src/test/java/org/apache/ignite/internal/processors/security/sandbox/IgnitionComponentProxyTest.java
alex-plekhanov: Space after `//`

modules/core/src/main/java/org/apache/ignite/internal/processors/security/IgniteSecurityProcessor.java
```diff
@@ -63,6 +63,13 @@
     /** Number of started nodes with the sandbox enabled. */
     private static final AtomicInteger SANDBOXED_NODES_COUNTER = new AtomicInteger();
 
+    /**
+     * @return True if there are nodes with the sandbox enabled.
+     */
+    public static boolean hasSandboxedNodes() {
```
alex-plekhanov, line 69: Package private?

modules/core/src/test/java/org/apache/ignite/internal/processors/security/sandbox/IgnitionComponentProxyTest.java
```diff
@@ -0,0 +1,193 @@
...
+
+        Ignite clnt = grid(CLNT);
+
+        //Checks that inside the sandbox we should get a proxied instance of Ignite.
+        clnt.compute(clnt.cluster().forNodeId(srv.cluster().localNode().id()))
+            .broadcast(() -> {
+                Collection<Ignite> nodes = s.get();
+
+                for (Ignite node : nodes)
+                    node.compute().call(() -> true);
```
alex-plekhanov, line 190: It's better to add a comment, why `() -> true` shouldn't work if we get not a proxy instance here.

PR #7162, opened 2019-12-18
Checkstyle expects OS specific line separator for NewlineAtEndOfFile rule.

agura:
Windows users can have  problems with checkstyle plugin.


agura:
Closed. Another PR #7167 created.

PR #7167, opened 2019-12-19
IGNITE-12472 Checkstyle expects OS specific line separator for NewlineAtEndOfFile rule.

agura:


agura:
Merged other soultion #7174

PR #7212, opened 2019-12-27
IGNITE-12471 .NET Thin Client: Fix WithExpiryPolicy crash

ptupitsyn:
Old servers do not support the extra flag, which causes exception and disconnect.

* Refactor client operation handling to be able to check ProtocolVersion of the exact connection that is used for the given operation
* Add `ProtocolVersion` check for cache operations with expiry policy.
* Fix `ProtocolVersion` check for `CacheConfiguration`


modules/platforms/dotnet/Apache.Ignite.Core.Tests/JavaServer/pom.xml
```diff
@@ -35,7 +35,7 @@
         <dependency>
             <groupId>org.apache.ignite</groupId>
             <artifactId>ignite-core</artifactId>
-            <version>2.7.6</version>
+            <version>2.5.9</version>
```
ashapkin: Central maven repo doesn't contain this professional release. It's ok since the version is replaced prior to a server startup, but could be confusing for new readers.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/JavaServer/src/main/java/Runner.java
```diff
@@ -16,25 +16,39 @@
...
-                .setAddresses(Collections.singleton("127.0.0.1:47500..47501"));
+                .setAddresses(Collections.singleton("127.0.0.1:47500"));
 
-        TcpDiscoverySpi discoSpi = new TcpDiscoverySpi().setIpFinder(ipFinder).setSocketTimeout(300);
+        TcpDiscoverySpi discoSpi = new TcpDiscoverySpi()
+            .setIpFinder(ipFinder)
+            .setSocketTimeout(300)
+            .setNetworkTimeout(300);
+
+        CacheConfiguration expiryCacheCfg = new CacheConfiguration("twoSecondCache")
```
ashapkin: Minor: could be a const that makes TestServerSideExpiryPolicyWorksOnAllVersions a little bit cleaner.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/ClientContextBase.cs
```diff
@@ -0,0 +1,79 @@
...
+        private readonly IBinaryStream _stream;
+
+        /** */
+        private readonly Marshaller _marshaller;
+
+        /** */
+        private readonly ClientProtocolVersion _protocolVersion;
+        
+        /// <summary>
+        /// Initializes a new instance of <see cref="ClientRequestContext"/> class.
```
ashapkin: This class initializes both Request and Response after the refactoring, let's change the summary.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/JavaServer/src/main/java/Runner.java
ptupitsyn: But this is Java file, and `TestServerSideExpiryPolicyWorksOnAllVersions` is C#, how should I share a constant between languages?

modules/platforms/dotnet/Apache.Ignite.Core.Tests/JavaServer/pom.xml
ptupitsyn: Fixed, thanks

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/ClientContextBase.cs
ptupitsyn: Fixed, thanks

PR #7215, opened 2019-12-30
IGNITE-12512: Compact log and drop log level down to info for cancelled queries

AMashenkov:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -668,7 +668,17 @@ private void sendError(ClusterNode node, long qryReqId, Throwable err) {
             GridQueryFailResponse msg = new GridQueryFailResponse(qryReqId, err);
 
             if (node.isLocal()) {
-                U.error(log, "Failed to run map query on local node.", err);
+                if (err instanceof QueryCancelledException) {
+                    String errMsg = "Failed to run cancelled map query on local node: [localNodeId="
+                        + node.id() + ", reqId=" + qryReqId + ']';
+
+                    if (log.isDebugEnabled())
+                        U.warn(log, errMsg, err);
```
pavlukhin: Cannot realize why do we need to log at WARN level when debug is enabled? From the first glance looks strange.

AMashenkov: Unfortunately stacktrace can't be printed in info\debug levels.

pavlukhin: Complicated =( But ok.

PR #7219, opened 2019-12-30
IGNITE-12514 Fixed WAL don't flush several last records in LOG-ONLY/FSYNC mode if flush ptr=null

ibessonov:


modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/IgniteWriteAheadLogManager.java
```diff
@@ -90,8 +90,10 @@ public WALPointer log(WALRecord entry, RolloverType rolloverType)
      * @throws IgniteCheckedException If failed to write.
      * @throws StorageException If IO exception occurred during the write. If an exception is thrown from this
      *      method, the WAL will be invalidated and the node will be stopped.
+     * @return Last WAL position which was flushed to WAL segment file. May be greater than or equal to a {@code ptr}.
+     * May be {@code null}, it means nothing has been flushed.
      */
-    public void flush(WALPointer ptr, boolean explicitFsync) throws IgniteCheckedException, StorageException;
+    public WALPointer flush(WALPointer ptr, boolean explicitFsync) throws IgniteCheckedException, StorageException;
```
AMashenkov: Let's add @Nullable annotation to return type.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/IgniteWriteAheadLogManager.java
```diff
@@ -90,8 +90,10 @@ public WALPointer log(WALRecord entry, RolloverType rolloverType)
      * @throws IgniteCheckedException If failed to write.
      * @throws StorageException If IO exception occurred during the write. If an exception is thrown from this
      *      method, the WAL will be invalidated and the node will be stopped.
+     * @return Last WAL position which was flushed to WAL segment file. May be greater than or equal to a {@code ptr}.
```
AMashenkov: @return Last WAL position which was flushed to WAL segment file. Greater than or equal to a {@code ptr} if flush was successful, {@code null} otherwise.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/filehandle/FileHandleManager.java
```diff
@@ -66,6 +66,8 @@
      * @param explicitFsync {@code true} if fsync required.
      * @throws IgniteCheckedException if fail.
      * @throws StorageException if storage was fail.
+     * @return Last WAL position which was flushed to WAL segment file. May be greater than or equal to a {@code ptr}.
```
AMashenkov: Let's avoid "May be" here and describe certain cases properly.

dspavlov:
This PR is associated with a resolved ticket. Closing the PR, please reopen it if you still need it.

PR #7230, opened 2020-01-09
IGNITE-12527 repair Deployment Modes-Un-Deployment and User Versions function can not use

pop443:
When using continuous deployment mode, the custom entryprocessor works correctly the first time it is loaded; however, it does not work properly after that, because the custom entryprocessor is cached by igniteutils concurrentmap < classloader, concurrentmap < string, class > >.
When processing custom entryprocessor, it will directly return to class without adding cache

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
@@ -8841,8 +8841,13 @@ public abstract class IgniteUtils {
                 throw new RuntimeException("Deserialization of class " + clsName + " is disallowed.");
 
             // Avoid class caching inside Class.forName
-            if (ldr instanceof CacheClassLoaderMarker)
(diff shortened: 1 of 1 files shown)


pop443:
user CONTINUOUS DeploymentMode,but user versions function is not correct work when I use custom EntryProcessor .
because class is cache by IgniteUtils ConcurrentMap<ClassLoader, ConcurrentMap<String, Class>> classCache . so when deal custom EntryProcessor , the class is not add to cache

pop443:
other class:

pop443:
other class like:
org.apache.ignite.internal.processors.cache.distributed.dht.preloader.GridDhtPartitionMap
org.apache.ignite.internal.processors.cache.distributed.dht.preloader.GridDhtPartitionFullMap
org.apache.ignite.internal.processors.cache.EntryProcessorResourceInjectorProxy
will be cached in org.apache.ignite.internal.managers.deployment.GridDeploymentLocalStore class

PR #7234, opened 2020-01-09
IGNITE-12523 Added throttling for thread dumps generation on system failure.

agura:


agura:
Merged to master branch.

PR #7245, opened 2020-01-13
IGNITE-12530 Pages list cache limit added to prevent IgniteOOME on checkpoint

alex-plekhanov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheOffheapManager.java
```diff
@@ -118,6 +120,17 @@
...
+    /**
+     * Threshold to calculate limit for pages list on-heap caches.
+     * <p>
+     * Note: When a checkpoint is triggered, we need some amount of page memory to store pages list on-heap cache.
+     * If a checkpoint is triggered by "too many dirty pages" reason and pages list cache is rather big, we can get
+     * {@code IgniteOutOfMemoryException}. To prevent this, we can limit the total amount of cached page list buckets,
+     * assuming that checkpoint will be triggered if no more then 3/4 of pages will be marked as dirty (there will be
+     * at least 1/4 of clean pages) and each cached page list bucket can be stored to up to 2 pages.
+     */
+    private static final double PAGE_LIST_CACHE_LIMIT_THRESHOLD = 1.0 / 4.0 / 2.0;
```
zstan: more exactly this formula is not static, but depends on PagesCache.MAX_SIZE, so if PagesCache.MAX_SIZE > (empty)PagesListNodeIO#getCapacity it can take more than 2 pages.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesList.java
```diff
@@ -1951,14 +1951,19 @@ public int cachedPagesCount(int bucket) {
...
         private int emptyFlushCnt;
 
+        /** Global (per data region) limit of caches for page lists. */
+        private final AtomicLong pagesCacheLimit;
+
         /**
          * Default constructor.
          */
-        public PagesCache() {
+        public PagesCache(AtomicLong pagesCacheLimit) {
```
zstan: append @Nullable here.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheOffheapManager.java
```diff
@@ -118,6 +120,19 @@
...
+     * <p>
+     * Note: When a checkpoint is triggered, we need some amount of page memory to store pages list on-heap cache.
+     * If a checkpoint is triggered by "too many dirty pages" reason and pages list cache is rather big, we can get
+     * {@code IgniteOutOfMemoryException}. To prevent this, we can limit the total amount of cached page list buckets,
+     * assuming that checkpoint will be triggered if no more then 3/4 of pages will be marked as dirty (there will be
+     * at least 1/4 of clean pages) and each cached page list bucket can be stored to up to 2 pages (this value is not
+     * static, but depends on PagesCache.MAX_SIZE, so if PagesCache.MAX_SIZE > PagesListNodeIO#getCapacity it can take
+     * more than 2 pages). Also some amount of page memory needed to store page list metadata.
+     */
+    public static final double PAGE_LIST_CACHE_LIMIT_THRESHOLD = 0.1;
```
zstan: i still have no clue how u define this constant is it really enough to be static (not dynamically) calculated here?

zstan: i still have no clue how u define this constant is it really enough to be static (not dynamically) calculated here?

alex-plekhanov: 1/4 (clean pages) / 2 (at most 2 pages per bucket cache) + some reserve for metadata.
It's hard to calculate it dynamically, it depends on metadata size (which we don't know at runtime)

zstan: it really enough to be static (not dynamically) calculated here?

alex-plekhanov: We can't calculate it dynamically, so yes, reasonable static value is the best we can do.

PR #7260, opened 2020-01-14
IGNITE-12540 Update versions of some dependencies.

alamar:


alamar:
> My current understanding is that just bumping the version of beanutils doesn't fully make the trick. You need to exclude transitive usages, for example from hadoop-common.

I think we will ignore Hadoop for now.

PR #7263, opened 2020-01-15
IGNITE-12545 Introduce listener interface for components to react to …

glukos:
…partition map exchange events


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/PartitionsExchangeAware.java
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
sk0x50: Please change this license to ASF-2.0 :)

PR #7269, opened 2020-01-16
IGNITE-12552: Move ReadOnlyMetricRegistry to public API.

nizhikov:


modules/core/src/main/java/org/apache/ignite/spi/metric/jmx/MetricRegistryMBean.java
```diff
@@ -25,7 +25,6 @@
 import java.util.Scanner;
 import javax.management.MBeanAttributeInfo;
 import javax.management.MBeanInfo;
-import org.apache.ignite.internal.processors.metric.MetricRegistry;
 import org.apache.ignite.internal.processors.metric.impl.HistogramMetric;
 import org.apache.ignite.internal.processors.metric.impl.MetricUtils;
```
agoncharuk: Since exporter is supposed to be implemented by a user, we need to move HistogramMetric and supporting logic to public package as well. An exporter must be implemented using public API solely.

nizhikov: @agoncharuk Fixed.

Please, take a look, one more time.

PR #7289, opened 2020-01-22
IGNITE-12565 Extend test coverage [IGNITE-9279] Support custom default SQL schema name

pavlukhin:


modules/core/src/test/java/org/apache/ignite/testframework/GridTestUtils.java
```diff
@@ -2129,7 +2129,22 @@ public static void addTestIfNeeded(@NotNull final List<Class<?>> suite, @NotNull
...
-        int len = rnd.nextInt(maxLen);
+        return randomString(rnd, 0, maxLen);
+    }
+
+    /**
+     * Generate random alphabetical string.
+     *
+     * @param rnd Random object.
+     * @param maxLen Maximal length of string
+     * @param minLen Minimum length of string
```
korlov42: I suggest swapping the parameters maxLen and minLen so that they match the signature of the method. Also trailing dot is missing.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/AbstractDefaultSchemaTest.java
```diff
@@ -0,0 +1,140 @@
...
+    protected void sql(String qry) {
+        sql(qry, null);
+    }
+
+    /** */
+    protected void sql(String qry, @Nullable Predicate<List<List<?>>> validator) {
+        List<List<?>> res = execSql(qry);
+
+        if (validator != null)
+            validator.test(res);
```
korlov42: Looks like validation result is ignored. Let's wrap it with `Assert.assertTrue()`. Also constants on rows 115 and 118 should be changed to long (0 -> 0L and 1 -> 1L accordingly).

modules/core/src/test/java/org/apache/ignite/testframework/GridTestUtils.java
pavlukhin: fixed

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/AbstractDefaultSchemaTest.java
pavlukhin: fixed

PR #7298, opened 2020-01-23
IGNITE-12557 Fix possible IgniteOOM during cache destroy.

ascherbakoff:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheOffheapManagerImpl.java
```diff
@@ -295,8 +295,9 @@ private void removeCacheData(int cacheId) {
             if (grp.sharedGroup()) {
                 assert cacheId != CU.UNDEFINED_CACHE_ID;
 
-                for (CacheDataStore store : cacheDataStores())
+                for (CacheDataStore store : cacheDataStores()) {
```
alex-plekhanov: Braces should not be used with one line statements

ascherbakoff: Missed that. Will fix it later.

PR #7310, opened 2020-01-24
IGNITE-12576 nodeId replaced with consistentId in TcpCommunicationSpi for usability purposes.

ibessonov:


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
```diff
@@ -121,11 +134,17 @@ public TcpCommunicationMetricsListener(GridMetricManager mmgr) {
...
         sentMsgsCntByNodeIdMetricFactory = nodeId ->
-            mmgr.registry(metricName(COMMUNICATION_METRICS_GROUP_NAME, nodeId.toString()))
-                .findMetric(SENT_MESSAGES_BY_NODE_ID_METRIC_NAME);
+            allSentMsgsMetricsByNodeId.computeIfAbsent(nodeId, k -> new AtomicLong());
 
         rcvdMsgsCntByNodeIdMetricFactory = nodeId ->
-            mmgr.registry(metricName(COMMUNICATION_METRICS_GROUP_NAME, nodeId.toString()))
+            allRcvdMsgsMetricsByNodeId.computeIfAbsent(nodeId, k -> new AtomicLong());
+
+        sentMsgsCntByConsistentIdMetricFactory = consistentId ->
```
Mmuzaf: Metric named `ConsistentIdMetric` but the constant for looking up named as by nodeId `SENT_MESSAGES_BY_NODE_ID_METRIC_NAME` why?

Mmuzaf:
@ibessonov Can you clarify why the `consistentId` has been added to ConnectionKey class? I think this may not be correct keeping this filed only for `metrics logging purposes` and it can easily be obtained via `discovery().node(nodeId);`

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
```diff
@@ -121,11 +134,17 @@ public TcpCommunicationMetricsListener(GridMetricManager mmgr) {
         );
 
         sentMsgsCntByNodeIdMetricFactory = nodeId ->
```
Mmuzaf: Why we should keep this metric if it totally replaced with the new one?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
```diff
@@ -82,10 +83,22 @@
     private final Function<Short, LongAdderMetric> rcvdMsgsCntByTypeMetricFactory;
 
     /** Function to be used in {@link Map#computeIfAbsent(Object, Function)} of {@code sentMsgsMetricsByNodeId}. */
-    private final Function<UUID, LongAdderMetric> sentMsgsCntByNodeIdMetricFactory;
+    private final Function<UUID, AtomicLong> sentMsgsCntByNodeIdMetricFactory;
```
nizhikov: Why this change?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
```diff
@@ -404,36 +423,50 @@ public static String receivedMessagesByTypeMetricName(Short directType) {
...
         /**
          * Collects statistics for message sent by SPI.
-         *
-         * @param msg Sent message.
+         *  @param msg Sent message.
+         * @param consistentId Receiver node consistent id.
          * @param nodeId Receiver node id.
          */
-        private void onMessageSent(Message msg, UUID nodeId) {
+        private void onMessageSent(Message msg, Object consistentId, UUID nodeId) {
```
nizhikov: `nodeId` should be removed as it not used.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
```diff
@@ -404,36 +423,50 @@ public static String receivedMessagesByTypeMetricName(Short directType) {
...
         /**
          * Collects statistics for message received by SPI.
-         *
-         * @param msg Received message.
+         *  @param msg Received message.
+         * @param consistentId Sender node consistent id.
          * @param nodeId Sender node id.
          */
-        private void onMessageReceived(Message msg, UUID nodeId) {
+        private void onMessageReceived(Message msg, Object consistentId, UUID nodeId) {
```
nizhikov: nodeId should be removed as it not used.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
nizhikov: Are we still need this factory?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
```diff
@@ -82,10 +83,22 @@
     private final Function<Short, LongAdderMetric> rcvdMsgsCntByTypeMetricFactory;
 
     /** Function to be used in {@link Map#computeIfAbsent(Object, Function)} of {@code sentMsgsMetricsByNodeId}. */
-    private final Function<UUID, LongAdderMetric> sentMsgsCntByNodeIdMetricFactory;
+    private final Function<UUID, AtomicLong> sentMsgsCntByNodeIdMetricFactory;
 
     /** Function to be used in {@link Map#computeIfAbsent(Object, Function)} of {@code rcvdMsgsMetricsByNodeId}. */
-    private final Function<UUID, LongAdderMetric> rcvdMsgsCntByNodeIdMetricFactory;
+    private final Function<UUID, AtomicLong> rcvdMsgsCntByNodeIdMetricFactory;
```
nizhikov: Are we still need this factory?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
```diff
@@ -82,10 +83,22 @@
...
+
+    /** Function to be used in {@link Map#computeIfAbsent(Object, Function)} of {@code #sentMsgsMetricsByConsistentId}. */
+    private final Function<Object, LongAdderMetric> sentMsgsCntByConsistentIdMetricFactory;
+
+    /** Function to be used in {@link Map#computeIfAbsent(Object, Function)} of {@code #rcvdMsgsMetricsByConsistentId}. */
+    private final Function<Object, LongAdderMetric> rcvdMsgsCntByConsistentIdMetricFactory;
+
+    /** All sent messages count metrics grouped by message node id. */
+    public volatile Map<UUID, AtomicLong> allSentMsgsMetricsByNodeId = new ConcurrentHashMap<>();
```
nizhikov: Are we still need this factory?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
```diff
@@ -82,10 +83,22 @@
...
+
+    /** Received messages metrics count grouped by message node id. */
+    public volatile Map<UUID, AtomicLong> allRcvdMsgsMetricsByNodeId = new HashMap<>();
```
nizhikov: Are we still need this factory?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
ibessonov: These things won't be exported as metrics. We have them for JMX support only.

ibessonov: And yes, factory saves us from extra allocations, we discussed this privately.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
ibessonov: Yes

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
ibessonov: Yes, this collection allows us to use HashMap in ThreadLocal entities without worrying about concurrent access - that was the main optimization point.
Basically I emulate MetricRegistry with this collection.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
ibessonov: Yes, replied above.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
ibessonov: old JMX API

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
ibessonov: I'll rename it.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
ibessonov: It is used in last line of the method.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
ibessonov: Same as above.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
nizhikov: Can we map `consistentId -> nodeId` when the user requests data to reduce the amount of metrics we store internally?

nizhikov:
```
        Map<UUID, Long> msgs = metricsLsnr.receivedMessagesByNode();

        Map<Object, Long> consistentIdMap = new HashMap<>();

        IgniteCluster cluster = ignite().cluster();

        for (Map.Entry<UUID, Long> entry : msgs.entrySet()) 
            consistentIdMap.put(cluster.node(entry.getKey()).consistentId(), entry.getValue());
        
        return consistentIdMap;
```

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
nizhikov: As I see in ticket descriptions we replacing `nodeId` with the `consistentId`, but the changes just add another copy of the same numbers, not replacing.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/ConnectionKey.java
```diff
@@ -25,6 +25,9 @@
  * Connection Key.
  */
 public class ConnectionKey {
+    /** */
+    private final Object consistentId;
```
Mmuzaf: `consistentId` should be a part of session meta (`ses.meta(CONSISTENT_ID_META)`. It has no meaning for the connectionKey class. An thus it will require much less changes in this PR.

ibessonov: I don't think that this would require less changes, pretty much the same amount.
I don't see any issues in storing consistentId near to nodeId, looks logical to me. Why do you think that it has no meaning?

nizhikov:
LGTM

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/ConnectionKey.java
Mmuzaf: I see the following major reasons:
1. ConnectionKey is a public class from **public API**. Do you think changing public API is a good idea for a minor bug?
1. `consistentId` doesn't participate in `equals` and `hashCode` methods. For the person, who will take a look at ConnectionKey class the first time this would look weird because `consistentId` is not the same as `nodeId`.
2. This field is used only for metrics and does not have any meaning for a session connection key itself.
3. You are passing `consistentId` key for each constructor rather obtaining it from session attributes directly, so you should change each constructor and this will lead to unnecessary changes.

ibessonov: 1. No, it's not: org.apache.ignite.spi.communication.tcp.**internal**.ConnectionKey
2. It's absolutely fine, other fields don't participate in equals and hashCode either.

Fine, I'll do it and we'll see if it got simpler. Maybe it is, I don't know for sure.

ibessonov: just for the protocol - I disagree with this change)

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -562,20 +569,22 @@
...
                 UUID sndId;
-
-                ConnectionKey connKey;
+                int connIdx;
+                long connCnt;
 
                 if (msg instanceof NodeIdMessage) {
                     sndId = U.bytesToUuid(((NodeIdMessage)msg).nodeIdBytes(), 0);
-                    connKey = new ConnectionKey(sndId, 0, -1);
+                    connIdx = 0;
```
nizhikov: It seems these changes are not required.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/ConnectionKey.java
Mmuzaf: @ibessonov thank you, please, let me know when you'll be done with changes. I'll take a final look.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationMetricsListener.java
agura: @ibessonov It seems there are no release with metrics by node. Old metric shows total sent/received message count.

PR #7317, opened 2020-01-27
IGNITE-12580: NPE in GridMetricManager.

nizhikov:


modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -386,14 +386,19 @@ public void remove(String regName) {
 
         notifyListeners(mreg, metricRegRemoveLsnrs, log);
 
+        DistributedMetaStorage metastorage0 = metastorage;
+
+        if (metastorage0 == null)
```
agura: Good catch!

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/LocalWalModeChangeDuringRebalancingSelfTest.java
```diff
@@ -234,16 +230,6 @@ public void testWalDisabledDuringRebalancing() throws Exception {
         doTestSimple();
     }
 
-    /**
```
agura: Why so many tests are removed? It isn't related with NPE in `GridMetricManager`.

nizhikov: The changes in the tests doesn't relate to the fix.
I removed it to run test that reproduces error many times on the TC - https://ci.ignite.apache.org/viewLog.html?buildId=4969452&.

To fix the issue only changes in GridMetricManager required.

If you know the way write the reproducer, please, let me know.

agura: I believe it's impossible to write adequate reproducer except of a lot of test repetition. But I think that configuration should not be deleted on cache stop, only on cache destroy.

PR #7320, opened 2020-01-28
IGNITE-12468 Java thin client: deserialization of arrays, collections and maps fixed

alex-plekhanov:


modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientUtils.java
```diff
@@ -522,12 +528,73 @@ void writeObject(BinaryOutputStream out, Object obj) {
...
+    /**
+     * Unwrap binary object.
+     */
+    private Object unwrapBinary(Object obj, BinaryReaderHandles hnds) {
+        if (obj instanceof BinaryObjectImpl) {
+            BinaryObjectImpl obj0 = (BinaryObjectImpl)obj;
+
+            return marsh.deserialize(BinaryHeapInputStream.create(obj0.array(), obj0.start()), hnds);
+        }
+        else if (obj instanceof BinaryObject )
```
pavlukhin: Cosmetic. Extra space between `BinaryObject` and `)`.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientUtils.java
```diff
@@ -522,12 +528,73 @@ void writeObject(BinaryOutputStream out, Object obj) {
...
+    }
+
+    /**
+     * Unwrap array with binary objects.
+     */
+    private Object[] unwrapArray(Object[] arr, BinaryReaderHandles hnds) {
+        if (BinaryUtils.knownArray(arr))
+            return arr;
+
+        Object[] res = new Object[arr.length];
```
pavlukhin: Is not it a problem that we upcast every array to a plain `Object` array? I mean that it can be more an array of more specific type (user) objects. Do we have the same behavior for a thick client?

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientUtils.java
```diff
@@ -522,12 +528,73 @@ void writeObject(BinaryOutputStream out, Object obj) {
...
+        else if (obj instanceof Object[])
+            return unwrapArray((Object[])obj, hnds);
+        else
+            return obj;
+    }
+
+    /**
+     * Unwrap collection with binary objects.
+     */
+    private Collection<Object> unwrapCollection(Collection<Objects> col, BinaryReaderHandles hnds) {
```
pavlukhin: `Collection<Objects> col` seems to be a typo (Object**s**). It seems we can use raw `Collection` type here.

modules/core/src/test/java/org/apache/ignite/client/FunctionalTest.java
```diff
@@ -278,6 +284,157 @@ public void testPutGet() throws Exception {
...
+            checkDataType(thinCache, thickCache, new HashMap<>(F.asMap(1, person)));
+            checkDataType(thinCache, thickCache, new HashMap<>(F.asMap(new HashSet<>(Arrays.asList(1, 2)),
+                Arrays.asList(person, person))));
+        }
+    }
+
+    /**
+     * Check that we get the same value from the cache as we put before.
+     *
+     * @param thinCache Cache.
```
pavlukhin: There is no `@param thickCache`. We might omit a parameters list here.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientUtils.java
alex-plekhanov: Information about user type is already lost when marshalling/unmarshalling client request on server side (IGNITE-12625).
For thick client we have the same behaviour:
```
            ignite.getOrCreateCache(DEFAULT_CACHE_NAME).put(1, new Person[] {new Person(1, "name")});
            Object obj = ignite.getOrCreateCache(DEFAULT_CACHE_NAME).get(1);
            System.out.println(obj.getClass().getName()); // [Ljava.lang.Object
```

pavlukhin: Got it.

PR #7332, opened 2020-01-29
IGNITE-12598: Cache metrics configuration should be removed only on cache destroy

nizhikov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -2060,7 +2063,7 @@ private void stopCacheSafely(GridCacheContext<?, ?> cctx, boolean clearDbObjects
             prepareCacheStop(cctx.name(), false, clearDbObjects);
 
             if (!cctx.group().hasCaches())
-                stopCacheGroup(cctx.group().groupId());
+                stopCacheGroup(cctx.group().groupId(), clearDbObjects);
```
agura: Here `clearDbObjects` is incorrect because `destroy == false`. IMHO valid value for flag is `false`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -1021,7 +1024,7 @@ private void stopCache(GridCacheAdapter<?, ?> cache, boolean cancel, boolean des
 
             sharedCtx.removeCacheContext(ctx);
 
-            cache.stop();
+            cache.stop(destroy);
```
agura: It seems this place (and below at line 1041) are only places where we should remove metric configuration. IF this code will be invoked from this place then we don't need any additional flags for `GridCacheAdapter.stop()` method.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -659,13 +659,15 @@ protected final String startInfo() {
     /**
      * Stops this cache. Child classes should override this method
      * to provide custom stop behavior.
+     *
+     * @param destroy Destroy data flag. Setting to <code>true</code> will remove all cache data.
      */
-    public void stop() {
+    public void stop(boolean destroy) {
```
agura: We can avoid this change. See my comment to `GridCacheProcessor.stopCache()` method.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -659,13 +659,15 @@ protected final String startInfo() {
...
         // Nulling thread local reference to ensure values will be eventually GCed
         // no matter what references these futures are holding.
         lastFut = null;
 
-        if (!ctx.kernalContext().isStopping())
+        if (!ctx.kernalContext().isStopping() && destroy)
```
agura: Could be just removed (condition and statement under it). See my comment to `GridCacheProcessor.stopCache()` method.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -1035,7 +1038,7 @@ private void stopCache(GridCacheAdapter<?, ?> cache, boolean cancel, boolean des
 
                 // Check whether dht cache has been started.
                 if (dht != null) {
-                    dht.stop();
+                    dht.stop(destroy);
```
agura: See comment above (line 1027).

modules/core/src/test/java/org/apache/ignite/internal/metric/CacheMetricsAddRemoveTest.java
```diff
@@ -82,17 +98,58 @@ public void testCacheMetricsAddRemove() throws Exception {
...
+
+        checkMetricsNotEmpty(cachePrefix);
+
+        grid("client").cluster().state(ClusterState.ACTIVE);
+
+        assertEquals(1L, grid("client").cache("other-cache").get(1L));
+
+        checkMetricsNotEmpty(cachePrefix);
+
+        destroyCache();
```
agura: After cache destroying should be asserted that metric configuration is removed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
nizhikov: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
nizhikov: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
nizhikov: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/metric/CacheMetricsAddRemoveTest.java
nizhikov: Fixed.

nizhikov:
@agura Thanks for the review.

I fixed all your comments.
Please, take a look.

PR #7346, opened 2020-01-31
IGNITE-9216 Uncomment tests in IgniteCacheTestSuite

alamar:
Also remove 'multiple update' in GridDhtCacheAdapter.


modules/core/src/test/java/org/apache/ignite/internal/managers/communication/GridIoManagerSelfTest.java
```diff
@@ -89,6 +90,7 @@ public GridIoManagerSelfTest() throws IgniteCheckedException {
      * @throws Exception If failed.
      */
     @Test
+    @Ignore
```
aealeksandrov: Maybe we should add ticket name.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/IgniteCacheObjectPutSelfTest.java
```diff
@@ -38,9 +38,9 @@
     @Override protected IgniteConfiguration getConfiguration(String gridName) throws Exception {
         IgniteConfiguration cfg = super.getConfiguration(gridName);
 
-        CacheConfiguration ccfg = new CacheConfiguration(DEFAULT_CACHE_NAME);
+        CacheConfiguration ccfg = new CacheConfiguration(CACHE_NAME);
 
-        ccfg.setName(CACHE_NAME);
+        ccfg.setOnheapCacheEnabled(true);
```
aealeksandrov: Did you add this for some reason? Should we test heap instead of off-heap cases?

PR #7375, opened 2020-02-06
PoC security context spread

dgarus:


modules/core/src/main/java/org/apache/ignite/internal/processors/security/IgniteSecurityProcessor.java
```diff
@@ -125,10 +120,25 @@ public IgniteSecurityProcessor(GridKernalContext ctx, GridSecurityProcessor secP
...
         throws IgniteCheckedException {
-        return secPrc.authenticateNode(node, cred);
+        SecurityContext res = secPrc.authenticateNode(node, cred);
+
+        if (res != null) {
+            Map<String, Object> attrs = new HashMap<>(node.attributes());
+
+            attrs.put(ATTR_SECURITY_SUBJECT_ID, res.subject().id());
+
+            ((TcpDiscoveryNode)node).setAttributes(attrs);
```
ascherbakoff: The approach looks wrong.
What if we have ZK discovery SPI ?
This will trigger class cast exception.

dgarus: Yes, you are right.
To avoid this exception, we can, for example, add the IgniteClusterNode#setAttributes method (I've updated the PR).
I tried to illustrate that we can do placing the existing logic of spreading security context to GridSecurityProcessor, 
but this PoC doesn't pretend to be a merge-ready approach.

PR #7377, opened 2020-02-06
IGNITE-12602 Calcite integration. JDBC Thin driver support.

gvvinblade:


modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/jdbc/JdbcRequestHandler.java
```diff
@@ -723,6 +745,21 @@ private JdbcResponse executeQuery(JdbcQueryExecuteRequest req) {
...
     }
 
+    /** */
+    private List<FieldsQueryCursor<List<?>>> querySqlFields(SqlFieldsQueryEx qry, GridQueryCancel cancel) {
+        if (experimentalQueryEngine != null) {
+            try {
+                return experimentalQueryEngine.query(QueryContext.of(qry, cancel), qry.getSchema(), qry.getSql(), qry.getArgs());
+            }
+            catch (IgniteSQLException e) {
+                U.warn(log, "Failed to execute SQL query using experimental engine. [qry=" + qry + ']', e);
```
rkondakov: I think we should not write this warning to log for DDL/DML queries.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/message/MessageServiceImpl.java
```diff
@@ -189,32 +187,16 @@ public FailureProcessor failureProcessor() {
...
-    @Override public void send(Collection<UUID> nodeIds, CalciteMessage msg) {
-        for (UUID nodeId : nodeIds)
-            send(nodeId, msg);
-    }
-
-    /** {@inheritDoc} */
-    @Override public void send(UUID nodeId, CalciteMessage msg) {
+    @Override public void send(UUID nodeId, CalciteMessage msg) throws IgniteCheckedException {
         if (localNodeId().equals(nodeId))
             onMessage(nodeId, msg, true);
```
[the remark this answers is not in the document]
gvvinblade: We submit it to another thread (async=true)

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/jdbc/JdbcRequestHandler.java
gvvinblade: Not sure, actually here we highlight an issue - it's been handled but still an issue and a user should be aware on it, because this mean you were wrong setting "useExperimentalEngine" flag to true.

PR #7383, opened 2020-02-07
IGNITE-12638 DMS classes are IgniteDataTransferObjects now.

ibessonov:


modules/core/src/main/java/org/apache/ignite/internal/processors/metastorage/persistence/DistributedMetaStorageHistoryItem.java
```diff
@@ -33,15 +37,19 @@
...
     @GridToStringInclude
-    public final byte[][] valBytesArray;
+    public byte[][] valBytesArray;
 
     /** */
     private transient long longHash;
 
+    /** Default constructor for deserialization. */
+    public DistributedMetaStorageHistoryItem() {
+    }
```
agura: Please, add `// No-op.` comment into the body of constructor.

modules/core/src/main/java/org/apache/ignite/internal/processors/metastorage/persistence/DistributedMetaStorageHistoryItem.java
```diff
@@ -68,6 +78,31 @@ public long estimateSize() {
...
+
+        for (int i = 0; i < keys.length; i++) {
+            U.writeString(out, keys[i]);
+
+            U.writeByteArray(out, valBytesArray[i]);
+        }
+    }
+
+    /** {@inheritDoc} */
+    @Override
```
agura: `@Override` should be on the same line as method signature.

modules/core/src/main/java/org/apache/ignite/internal/processors/metastorage/persistence/DistributedMetaStorageHistoryItem.java
```diff
@@ -68,6 +78,31 @@ public long estimateSize() {
...
+    protected void readExternalData(byte protoVer, ObjectInput in) throws IOException {
+        int len = in.readInt();
+
+        keys = new String[len];
```
agura: Do we really need empty arrays in case if `len == 0`?

modules/core/src/main/java/org/apache/ignite/internal/processors/metastorage/persistence/DistributedMetaStorageVersion.java
```diff
@@ -44,14 +48,18 @@ private static long nextHash(long hash, DistributedMetaStorageHistoryItem update
...
      * Hash of the whole updates list. Hashing algorinthm is almost the same as in {@link List#hashCode()}, but with
      * {@code long} value instead of {@code int}.
      */
     @GridToStringInclude
-    public final long hash;
+    public long hash;
+
+    /** Default constructor for deserialization. */
+    public DistributedMetaStorageVersion() {
+    }
```
agura: Please, add `// No-op.` comment into the body of constructor.

modules/core/src/main/java/org/apache/ignite/internal/processors/metastorage/persistence/DistributedMetaStorageVersion.java
```diff
@@ -133,6 +141,19 @@ public DistributedMetaStorageVersion nextVersion(
...
     }
 
+    /** {@inheritDoc} */
+    @Override protected void writeExternalData(ObjectOutput out) throws IOException {
+        out.writeLong(id);
+        out.writeLong(hash);
+    }
+
+    /** {@inheritDoc} */
+    @Override
```
agura: `@Override` should be on the same line as method signature.

modules/core/src/main/java/org/apache/ignite/internal/processors/metastorage/persistence/DistributedMetaStorageHistoryItem.java
```diff
@@ -50,6 +58,8 @@ public DistributedMetaStorageHistoryItem(String key, byte[] valBytes) {
 
     /** */
     public DistributedMetaStorageHistoryItem(String[] keys, byte[][] valBytesArray) {
+        assert keys.length == valBytesArray.length;
```
agura: Abbreviation should be used for `valBytesArray` parameter and corresponding class field.

modules/core/src/main/java/org/apache/ignite/internal/processors/metastorage/persistence/DistributedMetaStorageHistoryItem.java
ibessonov: len == 0 is impossible, I'll assert it in the constructor.

PR #7417, opened 2020-02-12
IGNITE-12668 Extends test coverage for plugin providers configuration

agura:


modules/core/src/test/java/org/apache/ignite/plugin/PluginConfigurationTest.java
```diff
@@ -0,0 +1,196 @@
...
+        return col.stream().map(PluginProvider::name).collect(Collectors.toList());
+    }
+
+    /** Plugin with own message factory. */
+    private static class TestPlugin implements IgnitePlugin {
+    }
+
+    /** */
+    @SuppressWarnings("RedundantThrows")
+    public static class TestPluginProvider implements PluginProvider<TestPluginConfiguration> {
```
ololo3000: Use AbstractTestPluginProvider to avoid boilerplate code, please.

agura: @ololo3000 Thanks. I'll do it in other change where some set of tests will be changed.

PR #7428, opened 2020-02-14
IGNITE-11147 Re-balance cancellation occur by non-affected event

vldpyatkov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1248,10 +1251,22 @@ private void updateGroupMetrics() {
...
+
         /**
          * @param grp Cache group.
          * @param assignments Assignments.
          * @param log Logger.
+         * @param rebalanceId Rebalance id.
+         * @param next Next rebalance future.
+         * @param lastCancelledTime Cancelled time.
          *
             @param rebalanceId Rebalance id.
```
xtern: duplicate parameter description

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -3484,6 +3462,31 @@ else if (r != null) {
                 }
             }
         }
+
+        /**
+         * Rebalance is not required on client node and always required when exchange future is null.
+         * In other cases the method checks all caches and decides required rebalance or not for specific exchange.
```
xtern: I suggest to correct this description as follows:

Rebalancing is not required on the client node and is always required when the exchange future is null. In other cases,
this method checks all caches and decides whether rebalancing is required or not for a specific exchange.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/WalStateManager.java
```diff
@@ -393,12 +394,19 @@ else if (!F.eq(grpDesc.deploymentId(), curGrpDesc.deploymentId())) {
      * @param topVer Topology version.
      * @param changedBaseline The exchange is caused by Baseline Topology change.
```
xtern: should be replaced with exchFut parameter

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/WalStateManager.java
```diff
@@ -466,78 +467,64 @@ else if (!grp.localWalEnabled())
...
-
-                for (Integer grpId0 : session0.disabledGrps) {
-                    CacheGroupContext grp = cctx.cache().cacheGroup(grpId0);
-
-                    assert grp != null;
-
-                    if (!grp.localWalEnabled())
-                        grp.localWalEnabled(true, false);
-                }
+        grpId = groupsToEnable.stream().findFirst().get();
```
xtern: can be replaced with
`grpId = F.first(groupsToEnable);`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -289,16 +280,19 @@ void onTopologyChanged(GridDhtPartitionsExchangeFuture lastFut) {
      * @param assignments Assignments to process.
      * @param force {@code True} if preload request by {@link ForceRebalanceExchangeTask}.
      * @param rebalanceId Rebalance id generated from exchange thread.
-     * @param next Runnable responsible for cache rebalancing chain.
+     * @param next That is next rebalance future wich will be executed after this rebalance.
```
xtern: I suggest changing the description to something like:

The subsequent rebalancing routine, which will be launched after the completion of the current rebalance.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -310,21 +304,60 @@ Runnable addAssignments(
...
+            }
 
             if (!oldFut.isInitial())
                 oldFut.cancel();
             else
                 fut.listen(f -> oldFut.onDone(f.result()));
 
+            // Make sure partitions sceduled for full rebalancing are first cleared.
+            for (Map.Entry<ClusterNode, GridDhtPartitionDemandMessage> e : assignments.entrySet()) {
+                for (Integer partId: e.getValue().partitions().fullSet()) {
```
xtern: missed space after partId

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePreloader.java
```diff
@@ -87,13 +87,15 @@
      * @param rebalanceId Rebalance id created by exchange thread.
      * @param next Runnable responsible for cache rebalancing chain.
```
xtern: old description

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePreloader.java
```diff
@@ -87,13 +87,15 @@
...
      * @param forcedRebFut External future for forced rebalance.
+     * @param compatibleRebFut Future for waiting for compatible rebalances.
      * @return Rebalancing runnable.
```
ascherbakoff: Actually returns rebalance future, not runnable.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -310,21 +304,60 @@ Runnable addAssignments(
...
                     }
                 });
+            }
 
             if (!oldFut.isInitial())
                 oldFut.cancel();
             else
                 fut.listen(f -> oldFut.onDone(f.result()));
 
+            // Make sure partitions sceduled for full rebalancing are first cleared.
```
ascherbakoff: typo - scheduled

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/GridDhtPartitionTopologyImpl.java
```diff
@@ -2708,23 +2690,30 @@ private void removeNode(UUID nodeId) {
...
+            if (lastAffChangeVer.compareTo(rebFinishedTopVer) > 0) {
+                if (log.isInfoEnabled()) {
+                    log.info("Affinity topology changed, no MOVING partitions will be owned " +
+                        "[rebFinishedTopVer=" + rebFinishedTopVer +
+                        ", lastAffChangeVer=" + lastAffChangeVer + "]");
+                }
 
-        try {
-            for (GridDhtLocalPartition locPart : grp.topology().currentLocalPartitions()) {
+                grp.preloader().forceRebalance();
```
ascherbakoff: This is poor solution because it breaks rebalancing compatibility optimization.
We should always own partitions if topology changes are compatible.
Also we need explicit test for this scenario.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1587,6 +1627,62 @@ private void sendRebalanceFinishedEvent() {
...
+                return false;
+
+            AffinityTopologyVersion previousTopVer =
+                grp.affinity().cachedVersions().stream().skip(grp.affinity().cachedVersions().size() - 2).findFirst().get();
+
+            p0 = Stream.concat(grp.affinity().cachedAffinity(previousTopVer).primaryPartitions(ctx.localNodeId()).stream(),
+                grp.affinity().cachedAffinity(previousTopVer).backupPartitions(ctx.localNodeId()).stream())
+                .collect(Collectors.toSet());
+
+            p1 = Stream.concat(grp.affinity().cachedAffinity(otherAssignments.topologyVersion()).primaryPartitions(ctx.localNodeId()).stream(),
```
NSAmelchev: Too long line

modules/core/src/test/java/org/apache/ignite/platform/PlatformCacheAffinityVersionTask.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.ignite.resources.IgniteInstanceResource;
+
+/**
+ * Task retrieves a affinity topology version for cache.
+ */
+public class PlatformCacheAffinityVersionTask extends ComputeTaskAdapter<String, Object> {
+    /** {@inheritDoc} */
+    @Override public Map<? extends ComputeJob, ClusterNode> map(List<ClusterNode> subgrid,
+        String arg) throws IgniteException {
+        return Collections.singletonMap(new Job(arg), subgrid.stream().filter(ClusterNode::isLocal).findFirst().get());
```
NSAmelchev: `F#find` can be used here

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1276,6 +1295,11 @@ private void updateGroupMetrics() {
 
                 partitionsLeft.addAndGet(v.partitions().size());
 
+                rebalancingParts.put(k, new HashSet<Integer>() {{
```
NSAmelchev: Can initial capacity be set?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/GridDhtPartitionTopologyImpl.java
vldpyatkov: I suggest that will be a temporary solution.
Until we haven't implement the issue
https://ggsystems.atlassian.net/browse/GG-27642

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1309,9 +1335,29 @@ private void updateGroupMetrics() {
...
+         * Init future.
+         */
+        public void init() {
+            requestPartitions(this, assignments);
+        }
+
+        /** {@inheritDoc} */
+        @Override public boolean onDone(@Nullable Boolean res, @Nullable Throwable err) {
+            if (super.onDone(res, err)) {
+                if (next != null)
```
Mmuzaf: Is it true that the `next` will start regardless the current RebalanceFuture finishes with an error or cancelled e.g. on stop() method called? Is it correct?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
vldpyatkov: Done

modules/core/src/test/java/org/apache/ignite/platform/PlatformCacheAffinityVersionTask.java
vldpyatkov: Is will look more complicated:
F.find(subgrid, null, new IgnitePredicate<ClusterNode>() {
            @Override public boolean apply(ClusterNode node) {
                return node.isLocal();
            }
        })

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
vldpyatkov: I think yes. All rebalance's futures are executed independently of other.

modules/core/src/test/java/org/apache/ignite/platform/PlatformCacheAffinityVersionTask.java
vldpyatkov: I found a way to fix it
F.find(subgrid, null, (IgnitePredicate<? super ClusterNode>)ClusterNode::isLocal)

NSAmelchev: It seems there is another option without cast: `F.find(subgrid, null, ClusterNode::isLocal)`

vldpyatkov: Unfortunately not. Because we have a method with another predicate here.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
Mmuzaf: It is OK that rebalance futures are independent, but it's strange when the current future has cancelled, but the next cache group starts nevertheless. We should:
1. cancel the whole chain since we are building one 
2. use an external arbiter (an executor service) for such tasks.

Please also note that parallel rebalancing of cache groups will flood the whole cluster an can lead to the cluster death.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1309,9 +1334,29 @@ private void updateGroupMetrics() {
...
             this.routines = 0;
             this.cancelLock = new ReentrantReadWriteLock();
+            this.next = null;
             this.lastCancelledTime = new AtomicLong();
         }
 
+        /**
+         * Init future.
+         */
+        public void init() {
```
Mmuzaf: Let's rename it to `start`

PR #7445, opened 2020-02-18
IGNITE-6530. Attempt to fix the deadlock

javaronok:
Removed thread-unsafe method for CP write lock checking. It's safer to use honest unlock method.


javaronok:
Removed thread-unsafe method for CP write lock checking. It's safer to use honest unlock method.

PR #7461, opened 2020-02-20
IGNITE-12710 Extension log in rebuild indexes to search problems.

antonovsergey93:


modules/core/src/main/java/org/apache/ignite/internal/processors/query/schema/SchemaIndexCacheVisitorImpl.java
```diff
@@ -157,8 +171,55 @@ public SchemaIndexCacheVisitorImpl(GridCacheContext cctx, SchemaIndexCacheFilter
             throw e;
         }
 
-        if (fut != null)
-            fut.get();
+        if (fut != null && stat != null) {
```
ygerzhedovich: Before the change we were waiting complete compound future, now we can skip it. It looks strange.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/schema/SchemaIndexCachePartitionWorker.java
```diff
@@ -227,7 +237,7 @@ private void processKey(KeyCacheObject key) throws IgniteCheckedException {
                 GridCacheEntryEx entry = cctx.cache().entryEx(key);
 
                 try {
-                    entry.updateIndex(rowFilter, clo);
+                    entry.updateIndex(rowFilter, clo, indexCacheStat);
```
AMashenkov: Looks like all of these 3 params can be squashed into single closure right here.

PR #7473, opened 2020-02-25
IGNITE-12682 IgniteMessageFactoryImpl.registerCustom() method is removed as potentially dangerous

agura:


agura:
Merged to master branch.

PR #7479, opened 2020-02-27
IGNITE-12723: getMXBean for tests introduced.

nizhikov:


Vladsz83:
@nizhikov , 

1) Should we slightly rename `GridAbstractTest#metricRegistry()`?

2) Do we need to keep `CacheMetricsManageTest#mxBean()` `ClusterBaselineNodesMetricsSelfTest#mxBean()`, `GridCacheLongRunningTransactionDiagnosticsTest#txMXBean()`, `SetTxTimeoutOnPartitionMapExchangeTest#txMXBean()`, `TcpCommunicationStatisticsTest#mbean()`? Why not using `GridAbstractTest#getMxBean()` directly?

nizhikov:
@Vladsz83 Thanks for the review

> Should we slightly rename GridAbstractTest#metricRegistry()?

Rename of the existing methods is not the goal of this ticket.
What name is more convenient?

> Do we need to keep

Yes. These methods used multiple times, so I think it OK to keep them.

Vladsz83:
@nizhikov , 
>>> What name is more convenient?

I was confused with 'metric registry' and dynamic mbeans, dynamic mbean creation. The method is about dynamic mbean to show some numbers, but not exactly for a metric registry . Better name might be ‘metricMXBean’

nizhikov:
> Better name might be ‘metricMXBean’

Got it. Thank you.

PR #7484, opened 2020-02-28
IGNITE-12729 LT.warn() doesn't ignore exception

agura:


agura:
Merged to the master branch.

PR #7486, opened 2020-03-02
IGNITE-12733: TransmissionHandler can now notify user listeners when it ends

Mmuzaf:


modules/core/src/main/java/org/apache/ignite/internal/managers/communication/TransmissionHandler.java
```diff
@@ -40,6 +40,15 @@
  * {@link TransmissionPolicy#CHUNK} the <em>#chunkHandler()</em> will be picked up.
  */
 public interface TransmissionHandler {
+    /**
+     * The end of the handled transmission. This means that all resources associated with previously opened
+     * session are freed and can be reused. Generally, it means that transmission topic from now can
```
xtern: "session is freed"

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/GridIoManager.java
```diff
@@ -2835,7 +2835,11 @@ private void receiveFromChannel(
                 boolean exit = in.readBoolean();
 
                 if (exit) {
-                    rcvCtxs.remove(topic);
+                    ReceiverContext rcv0 = rcvCtxs.remove(topic);
```
xtern: Why not name this variable simply "rcv" (instead of rcv0)?

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/TransmissionHandler.java
Mmuzaf: fixed

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/GridIoManager.java
Mmuzaf: fixed

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/TransmissionHandler.java
```diff
@@ -40,6 +40,15 @@
...
  */
 public interface TransmissionHandler {
+    /**
+     * The end of the handled transmission. This means that all resources associated with previously opened
+     * session is freed and can be reused. Generally, it means that transmission topic from now can
+     * safely accept new files.
+     *
+     * @param rmtNodeId Remote node id from which the source request comes from.
+     */
+    public void onEnd(UUID rmtNodeId);
```
xtern: As I can see. this method is not mandatory - maybe it’s worth adding a default implementation?

PR #7489, opened 2020-03-02
IGNITE-12708: Calcite integration. Expressions factory base implementation.

gvvinblade:


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/FilterNode.java
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ *
+ */
+public class FilterNode extends AbstractNode<Object[]> implements SingleNode<Object[]>, Upstream<Object[]> {
+    /** */
+    private final Predicate<Object[]> predicate;
+
+    /** */
+    private Deque<Object[]> inBuffer = new ArrayDeque<>(IN_BUFFER_SIZE);
```
rkondakov: Why do we need buffer here and in other nodes? I agree that we need some kind of buffering in (or before) Outbox node, but anyway, I would do it as a separate tree node, like `BufferNode` for simplicity and flexibility.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/Call.java
```diff
@@ -0,0 +1,110 @@
...
+
+    /** {@inheritDoc} */
+    @Override public <T> T accept(ExpressionVisitor<T> visitor) {
+        return visitor.visit(this);
+    }
+
+    /** {@inheritDoc} */
+    @Override public <T> T evaluate(ExecutionContext ctx, Object... args) {
+        if (opImpl == null)
+            opImpl = new ExpressionFactory(ctx.parent().typeFactory(), ctx.parent().conformance(), ctx.parent().opTable()).implement(this);
```
rkondakov: Such long lines are not readable

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/FilterNode.java
gvvinblade: A filter node requests more rows than its consumer requests from the filter because the filter filters some of them) a small buffer is used to hold the difference till a next portion of rows is requested. The same is true for other nodes, where you may request from a node input more rows then requested from the node itself.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/Upstream.java
```diff
@@ -15,25 +15,26 @@
...
 /**
  * Represents an abstract data consumer.
  *
- * <p/><b>Note</b>: except several cases (like consumer node and mailboxes), {@link Node#request()}, {@link Node#cancel()},
- * {@link Node#reset()}, {@link Sink#push(Object)} and {@link Sink#end()} methods should be used from one single thread.
+ * <p/><b>Note</b>: except several cases (like consumer node and mailboxes), {@link Node#request(int)}, {@link Node#cancel()},
+ * {@link Upstream#push(Object)} and {@link Upstream#end()} methods should be used from one single thread.
  */
-public interface Sink<T> {
+public interface Upstream<T> {
```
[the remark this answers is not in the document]
gvvinblade: Yes, you right, it's my mistake, of course "downstream" should be used.

PR #7516, opened 2020-03-11
IGNITE-12774 Fix transaction hanging after too many open files exception.

antonovsergey93:


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -3766,6 +3775,40 @@ else if (rcvCnt < 0)
...
+    }
+
+    /**
+     * Closing connections to node.
+     * NOTE: It is recommended only for tests.
+     *
+     * @param nodeId Node for which to close connections.
+     * @throws IgniteCheckedException If occurs.
+     */
+    void closeConnections(UUID nodeId) throws IgniteCheckedException {
```
macrergate: Hi, Sergey! Is this good approach adding a **testing** method to this class? I see it already has thousands of lines. Maybe better to move it to the test class and make the same job via reflection?

PR #7517, opened 2020-03-11
IGNITE-12761 Add ability to disable check crc sums of stored pages du…

fi1ipx:
…e to invalidate_indexes


modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CacheValidateIndexes.java
```diff
@@ -198,9 +216,9 @@ public UUID nodeId() {
             }
 
             if (!errors)
-                logger.info("no issues found.");
+                logger.severe("no issues found.");
```
sk0x50: Could you please describe why the level of the message was changed?

fi1ipx: Because message about the result of the validations is more important than other messages. Hope my logic is okay.

sk0x50: To be honest, I don't expect that the log should contain **SEVERE** message when there are no real issues.
I think a severe message indicates that there is a serious issue. https://docs.oracle.com/javase/8/docs/api/java/util/logging/Level.html#SEVERE

fi1ipx: Okay. The severe level needs to indicate failure then you are right. Thanks for your explanation!

PR #7520, opened 2020-03-12
IGNITE-12632: Support of cancel command for resource-consuming entities.

nizhikov:
This PR contains the support of cancel by id command for resource-consuming entities.
Cluster administrator supposed the user for these commands.

Cancellable entities:

  * Scan query.
  * SQL query.
  * Continuous query.
  * Transaction.
  * Compute task
  * Service

Proposed syntax:
control.sh:
```
>./control.sh --kill scan b2d221ca-ab08-4544-b8dc-8475538ed42f default 63
>./control.sh --kill compute 58b48c3e071-b2d221ca-ab08-4544-b8dc-8475538ed42f
>./control.sh --kill tx 30058c3e071-00000000-0bac-6d33-0000-000000000003
>./control.sh --kill sql b2d221ca-ab08-4544-b8dc-8475538ed42f_7
>./control.sh --kill service my-svc
>./control.sh --kill continuous bfca668d-5df8-4879-b97f-bd1712ad01c9
```
SQL:
```
KILL SCAN_QUERY 'b2d221ca-ab08-4544-b8dc-8475538ed42f' 'default' 63
KILL TX '30058c3e071-00000000-0bac-6d33-0000-000000000003'
KILL CONTINUOUS_QUERY 'bfca668d-5df8-4879-b97f-bd1712ad01c9'
KILL COMPUTE_TASK '58b48c3e071-b2d221ca-ab08-4544-b8dc-8475538ed42f'
KILL SERVICE 'my-svc'
```


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryManager.java
```diff
@@ -1962,6 +1962,21 @@ public String cacheName() {
         return cacheName;
     }
 
+    /**
+     * Cancel scan query.
+     *
+     * @param origNodeId Originating node id.
+     * @param qryId Query id.
+     */
+    public boolean cancelScanQuery(UUID origNodeId, long qryId) {
```
rkondakov: `GridCacheQueryManager#removeQueryResult`?

modules/core/src/main/java/org/apache/ignite/internal/visor/query/VisorContinuousQueryCancelTask.java
```diff
@@ -0,0 +1,79 @@
...
+         */
+        protected VisorContinuousQueryCancelJob(@Nullable VisorContinuousQueryCancelTaskArg arg, boolean debug) {
+            super(arg, debug);
+        }
+
+        /** {@inheritDoc} */
+        @Override protected Boolean run(@Nullable VisorContinuousQueryCancelTaskArg arg) throws IgniteException {
+            IgniteLogger log = ignite.log().getLogger(VisorContinuousQueryCancelJob.class);
+
+            log.info("Cancelling continuous query[routineId=" + arg.getRoutineId() + ']');
```
rkondakov: `if (log.isInfoEnabled())`?

modules/core/src/main/java/org/apache/ignite/internal/visor/query/VisorScanQueryCancelTaskArg.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.ignite.internal.processors.task.GridInternal;
+import org.apache.ignite.internal.util.typedef.internal.S;
+import org.apache.ignite.internal.util.typedef.internal.U;
+import org.apache.ignite.internal.visor.VisorDataTransferObject;
+
+/**
+ * Arguments of task for cancel SCAN query.
+ */
+@GridInternal
+public class VisorScanQueryCancelTaskArg extends VisorDataTransferObject {
```
rkondakov: `VisorDataTransferObject` is deprecated.

modules/core/src/main/java/org/apache/ignite/internal/visor/query/VisorContinuousQueryCancelTaskArg.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.ignite.internal.processors.task.GridInternal;
+import org.apache.ignite.internal.util.typedef.internal.S;
+import org.apache.ignite.internal.util.typedef.internal.U;
+import org.apache.ignite.internal.visor.VisorDataTransferObject;
+
+/**
+ * Arguments of task for cancel CONTINUOUS query.
+ */
+@GridInternal
+public class VisorContinuousQueryCancelTaskArg extends VisorDataTransferObject {
```
rkondakov: `VisorDataTransferObject` is deprecated.

modules/core/src/main/java/org/apache/ignite/internal/visor/query/VisorQueryCancelTask.java
```diff
@@ -17,22 +17,25 @@
 
 package org.apache.ignite.internal.visor.query;
 
+import java.util.Collection;
```
rkondakov: Imports

modules/core/src/main/java/org/apache/ignite/internal/visor/service/VisorCancelServiceTask.java
```diff
@@ -56,12 +60,30 @@ protected VisorCancelServiceJob(VisorCancelServiceTaskArg arg, boolean debug) {
...
+            if (!svc.isPresent())
+                return false;
+
+            try {
+                services.cancel(svcName);
+            }
+            catch (IgniteException e) {
+                IgniteLogger log = ignite.log().getLogger(VisorCancelServiceTask.class);
+
+                log.warning("Error on service cance.", e);
```
rkondakov: Typo

modules/core/src/main/java/org/apache/ignite/internal/QueryMXBeanImpl.java
```diff
@@ -0,0 +1,168 @@
...
+            log.info("Killing continuous query[routineId=" + routineId + ']');
+
+        boolean res;
+
+        try {
+            IgniteClusterImpl cluster = ctx.cluster().get();
+
+            IgniteCompute compute = cluster.compute();
+
+            ClusterNode srv = U.randomServerNode(ctx);
```
rkondakov: Why random node? Why not send it to the initiator?

modules/core/src/main/java/org/apache/ignite/internal/QueryMXBeanImpl.java
```diff
@@ -0,0 +1,168 @@
...
+     * @param cacheName Cache name.
+     * @param id Scan query id.
+     */
+    public void cancelScan(UUID originNodeId, String cacheName, long id) {
+        boolean res;
+
+        try {
+            IgniteClusterImpl cluster = ctx.cluster().get();
+
+            ClusterNode srv = U.randomServerNode(ctx);
```
rkondakov: Why do we use random node here? IMO we need to send task to the origin node and run `GridCacheDistributedQueryFuture#cancelQuery` there.

modules/core/src/main/java/org/apache/ignite/internal/ServiceMXBeanImpl.java
```diff
@@ -0,0 +1,76 @@
...
+        boolean res;
+
+        try {
+            IgniteClusterImpl cluster = ctx.cluster().get();
+
+            IgniteCompute compute = cluster.compute();
+
+            ClusterNode srv = U.randomServerNode(ctx);
+
+            res = compute.execute(new VisorCancelServiceTask(),
```
rkondakov: Why can't we run this task from the current node?

modules/indexing/src/test/java/org/apache/ignite/util/KillCommandsCommandShTest.java
```diff
@@ -0,0 +1,177 @@
...
+    /** {@inheritDoc} */
+    @Override protected void beforeTestsStarted() throws Exception {
+        super.beforeTestsStarted();
+
+        srvs = new ArrayList<>();
+
+        for (int i = 0; i < SERVER_NODE_CNT; i++)
+            srvs.add(grid(i));
+
+        System.out.println("CancelCommandCommandShTest.beforeTestsStarted");
```
rkondakov: Debugging artifact?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryManager.java
nizhikov: Fixed. Thanks.

modules/core/src/main/java/org/apache/ignite/internal/visor/query/VisorQueryCancelTask.java
nizhikov: What's wrong with it?

rkondakov: Unused imports. At least my IDE thinks so.

modules/core/src/main/java/org/apache/ignite/internal/visor/query/VisorScanQueryCancelTaskArg.java
nizhikov: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/visor/query/VisorQueryCancelTask.java
nizhikov: it used in the method `jobNodes`:

```
    @Override protected Collection<UUID> jobNodes(VisorTaskArgument<VisorQueryCancelTaskArg> arg) {
        return Collections.singleton(arg.getArgument().getNodeId());
    }
```

rkondakov: Ok, might be I missed it. You are right.

modules/core/src/main/java/org/apache/ignite/internal/QueryMXBeanImpl.java
nizhikov: > Why it's not a good idea to send cancel message to the initiator

My opinion - canceling of the query is the "last line of defense".
When an administrator decides to cancel a query it means something goes wrong.
The whole initiator node can be flooded with the entries update.
So I think it's not the best choice to send a cancel request to the initiator.

> Why random node?

CQ cancel implemented via discovery message so there is no difference what specific node receives request first.

modules/core/src/main/java/org/apache/ignite/internal/QueryMXBeanImpl.java
nizhikov: Same of previous:

> Why it's not a good idea to send cancel message to the initiator

My opinion - canceling of the query is the "last line of defense".
When an administrator decides to cancel a query it means something goes wrong.
The whole initiator node can be flooded and respond slowly.
So I think it's not the best choice to send a cancel request to the initiator.

> Why random node?

Actually, scan query cancel executed on all cluster nodes it implemented in the task itself see VisorScanQueryCancelTask#jobNodes:
```
    /** {@inheritDoc} */
    @Override protected Collection<UUID> jobNodes(VisorTaskArgument<VisorScanQueryCancelTaskArg> arg) {
        return F.transform(ignite.cluster().nodes(), ClusterNode::id);
    }
```

modules/core/src/main/java/org/apache/ignite/internal/ServiceMXBeanImpl.java
nizhikov: Yes, we can.

modules/core/src/main/java/org/apache/ignite/internal/visor/query/VisorContinuousQueryCancelTaskArg.java
nizhikov: fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/continuous/StopRoutineDiscoveryMessage.java
```diff
@@ -29,11 +29,24 @@
     /** */
     private static final long serialVersionUID = 0L;
 
+    /** {@code True} if stop should be forced. */
+    private boolean force;
```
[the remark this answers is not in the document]
nizhikov: fixed.

modules/core/src/main/java/org/apache/ignite/internal/QueryMXBeanImpl.java
rkondakov: I think we need to have a single entry point for the query cancellation. And this entry point should be located on the initiator. I can name a few reasons for it:

1. Convenience: single entry point is easier to control. If we want to add extra functionality to it (i.e. track statistics),  we need to add it to the just one place.
2. Clean code: it is much easier to understand the code which has only one entry point for each operation.
3. Effectiveness: we send cancellation task to the all cluster nodes, but it might be excessive. The query can be finished on some nodes by the time of cancellation. Or query might be initially executed on a few nodes or a single node due to partition pruning. Only the initiator knows the actual query status and mapping, and therefore it should be responsible for the cancellation.

modules/core/src/main/java/org/apache/ignite/internal/QueryMXBeanImpl.java
rkondakov: I think the cancellation task should be delegated to the initiator. See the explanation below.

modules/core/src/main/java/org/apache/ignite/internal/QueryMXBeanImpl.java
nizhikov: > And this entry point should be located on the initiator.
> Convenience: single entry point is easier to control

The initiator is the reason for the issue.
Imagine a case when CPU usage on the initiator is 100% because it flooded with the query results. In that case, the initiator will handle cancel message very slowly that can lead to a whole cluster disaster. 

I think we shouldn't wait until the initiator can proceed cancel message.
We should cancel it as fast as we can.

> Effectiveness: we send cancellation task to all cluster nodes, but it might be excessive

We can try to reduce the number of nodes to the number of affinity nodes.
Anyway, cancellation of the query is not a performance-critical operation.

rkondakov: Every node in the cluster can be overloaded, even the one where we are going to send the kill task from. This is not an issue.
We should have a single straightforward protocol for starting and cancelling the tasks. Existing protocol for cancelling tasks from the initiator is a good candidate for it. If instead we have a bunch of such protocols we will end up with a messy code and numerous subtle bugs which are not very easy to cope with. 
I think that proposed approach is not bug-free and there are possible races between task start and task cancellation messages. What would happen if cancel task request arrived before start task request? I am not sure this situation will be gracefully handled.
IMO we need to stick the approach with a single cancellation protocol from the initiator. It is solid and less erroneous solution.

nizhikov: > Every node in the cluster can be overloaded, even the one where we are going to send the kill task from. 
> This is not an issue.

This is the issue I'm trying to solve with these changes.

> I think that proposed approach is not bug-free

Can you, please, clarify, what tests should be added to check the changes?

> What would happen if cancel task request arrived before start task request? 

Then we would get an error "Query not found".

rkondakov: We will end up with memory and resources leak. Because the cancel task message arrives first and we get  "Query not found" here. But later the start task message arrives, the task starts execution but no one will cancel it because cancel task will not be resent.

> Then we would get an error "Query not found".

You may use custom communication SPI to delay start task messages and reorder them with cancel task messages.
> Can you, please, clarify, what tests should be added to check the changes?

PR #7523, opened 2020-03-12
IGNITE-12759 Getting a SecurityContext from GridSecurityProcessor

dgarus:


modules/core/src/main/java/org/apache/ignite/internal/processors/security/GridSecurityProcessor.java
```diff
@@ -77,7 +89,7 @@
     public SecuritySubject authenticatedSubject(UUID subjId) throws IgniteCheckedException;
```
glukos: Shouldn't we mention that actually not only node subjects can be retrieved here?

modules/core/src/main/java/org/apache/ignite/internal/processors/security/IgniteSecurityProcessor.java
```diff
@@ -55,7 +55,17 @@
...
- * Default IgniteSecurity implementation.
+ * Default {@code IgniteSecurity} implementation.
+ * <p>
+ * {@code IgniteSecurityProcessor} serves here as a facade with is exposed to Ignite internal code,
+ * while {@code GridSecurityProcessor} is hidden and managed from {@code IgniteSecurityProcessor}.
+ * <p>
+ * This implementation of {@code IgniteSecurity} is responsible for:
+ * <ul>
+ *     <li>Keeping and propagating authenticated security contexts for cluster nodes;</li>
+ *     <li>Delegating calls for all aforementioned actions to {@code GridSecurityProcessor};</li>
```
glukos: I don't think that "aforementioned" suits here unless there's another list of actions above :)

modules/core/src/main/java/org/apache/ignite/internal/processors/security/GridSecurityProcessor.java
```diff
@@ -77,7 +89,7 @@
...
 
     /**
-     * Gets security context.
+     * Gets security context for authenticated nodes and thin clients.
```
glukos: Let's at least extend javadoc of SecurityContext in order to explain its area of responsibility.
It deserved something more than this:
```
/**
 * Security context.
 */
public interface SecurityContext {
```

modules/core/src/main/java/org/apache/ignite/internal/processors/security/GridSecurityProcessor.java
dgarus: I thought we will do that in the process of security reworking. 
Currently, these methods don't use to get a security subject of thin clients.

modules/core/src/main/java/org/apache/ignite/internal/processors/security/IgniteSecurityProcessor.java
dgarus: yes

modules/core/src/main/java/org/apache/ignite/internal/processors/security/GridSecurityProcessor.java
dgarus: ok

PR #7529, opened 2020-03-13
COVID-19 GridWorkers are now at risk to get coronavirus.

Jokser:
it's not fair that humans are susceptible to coronavirus disease while databases are not.
Now GridWorkers have a chance to get coronavirus infection and even die of it.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/worker/GridWorker.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/worker/GridWorker.java
@@ -19,6 +19,9 @@ package org.apache.ignite.internal.util.worker;
 
 import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+
 import org.apache.ignite.IgniteInterruptedException;
 import org.apache.ignite.IgniteLogger;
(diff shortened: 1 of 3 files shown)


korlov42:
workers from common pool should infects each other

modules/core/src/main/java/org/apache/ignite/internal/util/worker/GridWorker.java
```diff
@@ -97,6 +106,36 @@ protected GridWorker(@Nullable String igniteInstanceName, String name, IgniteLog
...
+            do {
+                try {
+                    Thread.sleep(TimeUnit.MINUTES.toMillis(1));
+                }
+                catch (InterruptedException e) {
+                    U.warn(log, "Sorry, I'm in quarantine and can't be interrupted.");
+                }
+                finally {
+                    if (ThreadLocalRandom.current().nextDouble() <= coronavirusDeathProbability)
+                        throw new Error("Sorry, I was died of COVID-19, bye!");
```
MaxLosevskoy: Typo: `was` not needed

AMashenkov:
Should a node drop messages is gotten from infected remote node?

agura:
Guys, don't panic! Failure handler just will kill nodes with infected workers.

iartiukhov:
> Guys, don't panic! Failure handler will just kill nodes with infected workers.

Unless it is not infected itself.

agura:
It doesn't matter. It will also die.

PR #7542, opened 2020-03-17
IGNITE-12768: MetricRegistryMBean doesn't show histogram values in case when histogram name contains underscore character

nizhikov:
…se when histogram name contains underscore character

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/spi/metric/jmx/MetricRegistryMBean.java
+++ b/modules/core/src/main/java/org/apache/ignite/spi/metric/jmx/MetricRegistryMBean.java
@@ -184,7 +184,14 @@ public class MetricRegistryMBean extends ReadOnlyDynamicMBean {
         long[] bounds = h.bounds();
         long[] values = h.value();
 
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/spi/metric/jmx/MetricRegistryMBean.java
```diff
@@ -155,12 +155,26 @@ else if (metric instanceof ObjectMetric)
      * @see MetricUtils#histogramBucketNames(HistogramMetric, Map)
      */
     public static Long searchHistogram(String name, ReadOnlyMetricRegistry mreg) {
-        Scanner sc = new Scanner(name).useDelimiter("_");
+        int highBoundIdx;
 
-        if (!sc.hasNext())
+        boolean isInf = name.endsWith(INF);
+
+        if (isInf) {
```
NSAmelchev: Unnecessary brackets

nizhikov: fixed.

PR #7560, opened 2020-03-23
IGNITE-12739 : Optimistic serializable transactions may fail infinitely when read-through is enabled

Vladsz83:


Vladsz83:
Optimistic serializable transaction enforces checking of backup version for not-sync cache updates.

Vladsz83:
Better solution provided: https://github.com/apache/ignite/pull/7575.

PR #7563, opened 2020-03-24
IGNITE-12749 Fixed unsupported operation exception on node stop if collisionspi not null

ymolochkov:


modules/core/src/main/java/org/jsr166/ConcurrentLinkedHashMap.java
```diff
@@ -1536,10 +1536,49 @@ public V replacex(K key, V oldVal, V newVal) {
...
+        modCnt.increment();
+    }
+
+    /**
+     * Tries to lock all segments in order which they appear to be in an underlying array.
+     */
+    @SuppressWarnings("LockAcquiredButNotSafelyReleased")
+    private void lockWriteLocks() {
+        for (Segment<K, V> segment : segments) {
+            segment.writeLock().lock();
```
nizhikov: What is thread will be interrupted in the middle of this loop?

modules/core/src/main/java/org/jsr166/ConcurrentLinkedHashMap.java
```diff
@@ -1536,10 +1536,49 @@ public V replacex(K key, V oldVal, V newVal) {
...
      */
     @Override public void clear() {
-        throw new UnsupportedOperationException();
+
+        try {
+            lockWriteLocks();
+            replaceOldSegments();
+
+        } finally {
+            unlockWriteLocks();
```
nizhikov: If the exception will be thrown on lock acquire we will try to unlock locks that never acquired.

PR #7575, opened 2020-03-26
IGNITE-12739 Optimistic serializable transactions may fail infinitely when read-through is enabled.

Vladsz83:


Vladsz83:
`GridPartitionedSingleGetFuture` now truely follows its flag `forcePrimary` and optimistic serializable transaction always sets `forcePrimary`.

PR #7579, opened 2020-03-27
IGNITE-12842: Cache "IGNITE_DAEMON" system properties in GirdKernalContextImpl.isDaemon call

sunnychanwork:


modules/core/src/main/java/org/apache/ignite/internal/GridKernalContextImpl.java
```diff
@@ -1047,12 +1047,14 @@ else if (helper instanceof HadoopHelper)
             comp.printMemoryStats();
     }
 
+    private static final boolean IS_IGNITE_DAEMON_SYSPROP_SET = IgniteSystemProperties.getBoolean(IGNITE_DAEMON);
```
agoncharuk: @sunnychanwork Let's make this flag non-static, this is more consistent with the way we capture other system properties during the node startup.

sunnychanwork: Fixed as suggested

PR #7580, opened 2020-03-27
IGNITE-12815: SQL query cancel command.

nizhikov:


modules/core/src/test/java/org/apache/ignite/internal/commandline/CommandHandlerParsingTest.java
```diff
@@ -532,6 +533,15 @@ public void testKillArguments() {
 
         // Service command format errors.
         assertParseArgsThrows("Expected service name.", "--kill", "service");
+
+        // Transaction command format errors.
+        assertParseArgsThrows("Expected transaction id.", "--kill", "tx");
```
rkondakov: Do we have positive parsing tests to ensure that all arguments are parsed well? As well as tests that ensure the wrong argument types are handled gracefully?

nizhikov: Positive cases are testsed under `KillCommandsCommandShTest`

rkondakov: Ok. LGTM

PR #7607, opened 2020-04-01
IGNITE-11073: Create consistent partitions copy on each cluster node

Mmuzaf:


modules/platforms/dotnet/Apache.Ignite.Core.Tests/Services/ServicesTest.cs
```diff
@@ -870,20 +870,6 @@ public void TestCallJavaService()
                 binSvc.testBinaryObject(
                     Grid1.GetBinary().ToBinary<IBinaryObject>(new PlatformComputeBinarizable {Field = 6}))
                     .GetField<int>("Field"));
-
```
alamar, line 873: Why!?

Mmuzaf: Thanks, I've reverted changes. This was an incorrect merge with the master branch.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ExchangeContext.java
```diff
@@ -76,9 +77,10 @@ public ExchangeContext(GridCacheSharedContext<?, ?> cctx, boolean crd, GridDhtPa
             log.warning("Current topology does not support the PME-free switch. Please check all nodes support" +
                 " this feature and it was not explicitly disabled by IGNITE_PME_FREE_SWITCH_DISABLED JVM option.");
 
+        boolean requirePmeFree = (fut.wasRebalanced() && fut.isBaselineNodeFailed()) || startedBySnapshot(fut);
```
anton-vinogradov, line 80: just a boolean pmeFreeSwitch

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1894 @@
...
+    /** Requested snapshot from remote node. */
+    private final AtomicReference<RemoteSnapshotFuture> rmtSnpReq = new AtomicReference<>();
+
+    /** Mutex used to order cluster snapshot operation progress. */
+    private final Object snpOpMux = new Object();
+
+    /** Take snapshot operation procedure. */
+    private final DistributedProcess<SnapshotOperationRequest, SnapshotOperationResponse> startSnpProc;
+
+    /** Check previously preformed snapshot operation and delete uncompleted files if need. */
```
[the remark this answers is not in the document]
Mmuzaf: fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1894 @@
...
+        // Collection of pairs group and appropriate cache partition to be snapshot.
+        Map<Integer, Set<Integer>> parts = new HashMap<>();
+
+        for (Integer grpId : req.grpIds)
+            parts.put(grpId, null);
+
+        SnapshotFutureTask task0 = registerSnapshotTask(req.snpName,
+            req.srcNodeId,
+            parts,
+            locSndrFactory.apply(req.snpName));
```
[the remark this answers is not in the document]
Mmuzaf: This factory cannot be simply removed, but I've rechecked the source code and created the factory below. Seems it should cover all cases:

```
    Function<String, SnapshotSender> localSnapshotSenderFactory() {
        return LocalSnapshotSender::new;
    }
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1894 @@
...
+
+        return task0.chain(f -> new SnapshotOperationResponse());
+    }
+
+    /**
+     * @param id Request id.
+     * @param res Results.
+     * @param err Errors.
+     */
+    private void startLocalSnapshotResult(UUID id, Map<UUID, SnapshotOperationResponse> res, Map<UUID, Exception> err) {
```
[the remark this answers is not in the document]
Mmuzaf: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1894 @@
...
+     */
+    static String igniteCacheStoragePath(PdsFolderSettings pcfg) {
+        return Paths.get(DB_DEFAULT_FOLDER, pcfg.folderName()).toString();
+    }
+
+    /**
+     * @param cfg Ignite configuration.
+     * @return Snapshot work path.
+     */
+    static Path snapshotPath(IgniteConfiguration cfg) {
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1906 @@
...
+         */
+        public SnapshotOperationRequest(UUID srcNodeId, String snpName, List<Integer> grpIds, Set<UUID> bltNodes) {
+            this.srcNodeId = srcNodeId;
+            this.snpName = snpName;
+            this.grpIds = grpIds;
+            this.bltNodes = bltNodes;
+        }
+
+        /** {@inheritDoc} */
+        @Override public boolean equals(Object o) {
```
alex-plekhanov, line 1803: Do we really need to override `equals()` and `hashCode()`? We didn't use these methods explicitly and never use object as key for hashmaps. Did I miss something?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1906 @@
...
+
+        if (!clusterState.hasBaselineTopology()) {
+            return new IgniteFinishedFutureImpl<>(new IgniteException("Snapshot operation has been rejected. " +
+                "The baseline topology is not configured for cluster."));
+        }
+
+        GridFutureAdapter<Void> snpFut0;
+
+        synchronized (snpOpMux) {
+            if (clusterSnpFut != null && !clusterSnpFut.isDone()) {
```
alex-plekhanov, line 959: As far as I understand, if two snapshot operations with different names were requested concurrently on different nodes both will register `clusterSnpFut`, one of them will fail on `initLocalSnapshotStartStage`, but future for both operations will be completed with success and only after real snapshot operation is completed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1894 @@
...
+
+            cctx.exchange().unregisterExchangeAwareComponent(this);
+        }
+        finally {
+            busyLock.unblock();
+        }
+    }
+
+    /**
+     * Concurrently traverse the snapshot directory for given local node folder name and
```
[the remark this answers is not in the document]
Mmuzaf: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1894 @@
...
+    public static void deleteSnapshot(File snpDir, String folderName) {
+        if (!snpDir.exists())
+            return;
+
+        assert snpDir.isDirectory() : snpDir;
+
+        try {
+            List<Path> dirs = new ArrayList<>();
+
+            Files.walkFileTree(snpDir.toPath(), new SimpleFileVisitor<Path>() {
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
Mmuzaf: I think it's not true. Here are two statements which you need to take into account:
- Multiple distributed processes can run simultaneously;
- `initLocalSnapshotStartStage` runs through discovery message;

So the first snapshot operation will be able to complete successfully, but the second one must fail. We should check the request ids here, you're right.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
Mmuzaf: Removed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotMXBeanImpl.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * @param ctx Kernal context.
+     */
+    public SnapshotMXBeanImpl(GridKernalContextImpl ctx) {
+        mgr = ctx.cache().context().snapshotMgr();
+    }
+
+    /** {@inheritDoc} */
+    @Override public void createSnapshot(String snpName) {
+        mgr.createSnapshot(snpName).get();
```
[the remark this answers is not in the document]
Mmuzaf: I think the `in-progress` state must be enough for such a case. I've changed the code, please, take a look.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+
+                            DiscoveryCustomEvent customEvt = new DiscoveryCustomEvent();
+
+                            customEvt.node(evt0.node());
+                            customEvt.eventNode(evt0.eventNode());
+                            customEvt.affinityTopologyVersion(evt0.affinityTopologyVersion());
+                            customEvt.customMessage(new SnapshotStartDiscoveryMessage(discoCache, msg.processId()));
+
+                            // Handle new event inside discovery thread, so no guarantees will be violated.
+                            cctx.exchange().onDiscoveryEvent(customEvt, discoCache);
```
alex-plekhanov, line 490: I don't like the idea of injecting to the discovery and using an internal `onDiscoveryEvent` method. I think we can do it with only one event. For example,  `SnapshotStartDiscoveryMessage` can extend `InitMessage` and we can pass `InitMessage` class and `InitMessage` factory to `DistributedProcess` constructor. In this case, you don't need to make `onDiscoveryEvent` public and can use standard discovery workflow.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -1994,7 +1997,7 @@ private GridCacheContext prepareCacheContext(
      *
      * @param cctx Cache context.
      */
-    private void stopCacheSafely(GridCacheContext<?, ?> cctx) {
+    public void stopCacheSafely(GridCacheContext<?, ?> cctx) {
```
alex-plekhanov, line 2000: Used only inside this class

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -4042,6 +4047,10 @@ public void onDiscoveryEvent(
      * @return {@code True} if minor topology version should be increased.
      */
     public boolean onCustomEvent(DiscoveryCustomMessage msg, AffinityTopologyVersion topVer, ClusterNode node) {
+        if (msg instanceof InitMessage &&
```
alex-plekhanov, line 4050: Should be removed if standard discovery workflow will be used.

modules/core/src/main/java/org/apache/ignite/internal/MarshallerContextImpl.java
```diff
@@ -168,27 +170,73 @@ private void initializeCaches() {
...
+            return;
+
+        for (byte platformId = 0; platformId < mappings.size(); platformId++) {
+            Map<Integer, MappedName> attach = mappings.get(platformId);
 
-        for (Map.Entry<Integer, MappedName> e : marshallerMappings.entrySet()) {
-            int typeId = e.getKey();
-            String clsName = e.getValue().className();
+            if (attach == null)
+                return;
```
alex-plekhanov, line 221: Originally null mappings were skipped but don't break processing. I think `continue` should be there.

modules/core/src/test/java/org/apache/ignite/platform/PlatformDeployServiceTask.java
```diff
@@ -18,6 +18,11 @@
 package org.apache.ignite.platform;
 
 import java.sql.Timestamp;
+import java.util.ArrayList;
```
alex-plekhanov, line 21: No change, only imports

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+        locSnpDir = resolveSnapshotWorkDirectory(ctx.config());
+        tmpWorkDir = Paths.get(storeMgr.workDir().getAbsolutePath(), DFLT_SNAPSHOT_TMP_DIR).toFile();
+
+        U.ensureDirectory(locSnpDir, "snapshot work directory", log);
+        U.ensureDirectory(tmpWorkDir, "temp directory for snapshot creation", log);
+
+        MetricRegistry mreg = cctx.kernalContext().metric().registry(SNAPSHOT_METRICS);
+
+        mreg.register("LastSnapshotStartTime", () -> lastSeenSnpFut.startTime,
+            "The system time approximated by 10 ms when the last cluster snapshot operation has been started.");
```
alex-plekhanov, line 364: We should state in the description (for all metrics) that we are talking about snapshots started by this node. Or change implementation to show snapshots started by any node.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+        mreg.register("LastSnapshotEndTime", () -> lastSeenSnpFut.endTime,
+            "The system time approximated by 10 ms when the last cluster snapshot operation has been finished.");
+        mreg.register("LastSnapshotName", () -> lastSeenSnpFut.name, String.class,
+            "The name of last started cluster snapshot operation.");
+        mreg.register("LastSnapshotErrorMessage",
+            () -> lastSeenSnpFut.error() == null ? null : lastSeenSnpFut.error().getMessage(),
+            String.class,
+            "The error message of last started cluster snapshot operation which fail. This value will be 'null' " +
+                "if last snapshot operation completed successfully.");
+        mreg.register("localSnapshotList", this::getSnapshots, List.class,
```
alex-plekhanov, line 374: Use always upper case for the first letter or lower case for the first letter, but not mixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+
+        SnapshotFutureTask task = locSnpTasks.get(snpRq.snpName);
+
+        if (task == null)
+            return;
+
+        if (task.start()) {
+            cctx.database().forceCheckpoint(String.format("Start snapshot operation: %s", snpRq.snpName));
+
+            // schedule task on checkpoint and wait when it starts
```
alex-plekhanov, line 1095: Upcase and point

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+        }
+    }
+
+    /**
+     * @param parts Collection of pairs group and appropriate cache partition to be snapshot.
+     * @param rmtNodeId The remote node to connect to.
+     * @param partConsumer Received partition handler.
+     * @return Future which will be completed when requested snapshot fully received.
+     */
+    public IgniteInternalFuture<Void> createRemoteSnapshot(
```
alex-plekhanov, line 1111: Perhaps `requestRemoteSnapshot` is a better name. WDYT?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
Mmuzaf: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
Mmuzaf: Discussed privately. A factory added to the `DistributedProcess` configuration, so it is possible from now start a process with custom inheritor of `InitMessage`, including `SnapshotDiscoveryMessage`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
Mmuzaf: Removed, since the API of `DistributedProcess` changed.

modules/core/src/main/java/org/apache/ignite/internal/MarshallerContextImpl.java
Mmuzaf: Fixed.

modules/core/src/test/java/org/apache/ignite/platform/PlatformDeployServiceTask.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+    private ExecutorService snpRunner;
+
+    /** System discovery message listener. */
+    private DiscoveryEventListener discoLsnr;
+
+    /** Cluster snapshot operation requested by user. */
+    private ClusterSnapshotFuture clusterSnpFut;
+
+    /** Current snapshot operation on local node. */
+    private volatile SnapshotOperationRequest clusterSnpRq;
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Services/ServicesTest.cs
```diff
@@ -870,7 +870,7 @@ public void TestCallJavaService()
                 binSvc.testBinaryObject(
                     Grid1.GetBinary().ToBinary<IBinaryObject>(new PlatformComputeBinarizable {Field = 6}))
                     .GetField<int>("Field"));
-            
+
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+        return Paths.get(DB_DEFAULT_FOLDER, folderName).toString();
+    }
+
+    /**
+     * @param cfg Ignite configuration.
+     * @return Snapshot work path.
+     */
+    static File resolveSnapshotWorkDirectory(IgniteConfiguration cfg) {
+        try {
+            return cfg.getSnapshotPath() == null ?
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
Mmuzaf: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+     * @param factory Factory which produces {@link LocalSnapshotSender} implementation.
+     */
+    void setLocalSnapshotSenderFactory(Function<String, SnapshotSender> factory) {
+        locSndrFactory = factory;
+    }
+
+    /**
+     * @return Factory which produces {@link LocalSnapshotSender} implementation.
+     */
+    Function<String, SnapshotSender> localSnapshotSenderFactory() {
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
Mmuzaf: Update all metrics description.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -361,19 +356,19 @@ public static String partDeltaFileName(int partId) {
         MetricRegistry mreg = cctx.kernalContext().metric().registry(SNAPSHOT_METRICS);
 
         mreg.register("LastSnapshotStartTime", () -> lastSeenSnpFut.startTime,
-            "The system time approximated by 10 ms when the last cluster snapshot operation has been started.");
+            "The system time approximated by 10 ms of the last started cluster snapshot request on this node.");
         mreg.register("LastSnapshotEndTime", () -> lastSeenSnpFut.endTime,
-            "The system time approximated by 10 ms when the last cluster snapshot operation has been finished.");
+            "The system time approximated by 10 ms of the last started cluster snapshot request on this node.");
```
alex-plekhanov, line 361: The description is the same now. "approximated by 10 ms" can be omitted, but it's up to you.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotMXBeanTest.java
```diff
@@ -0,0 +1,66 @@
...
+    public void testCreateSnapshot() throws Exception {
+        IgniteEx ignite = startGridsWithCache(2, dfltCacheCfg, CACHE_KEYS_RANGE);
+
+        SnapshotMXBean mxBean = getMBean(ignite.name());
+
+        mxBean.createSnapshot(SNAPSHOT_NAME);
+
+        MetricRegistry mreg = ignite.context().metric().registry(SNAPSHOT_METRICS);
+
+        LongMetric endTime = mreg.findMetric("LastSnapshotEndTime");
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
Mmuzaf: Fixed but checking distributed process ids on StageResult.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotMXBeanImpl.java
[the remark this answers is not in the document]
Mmuzaf: Cluster snapshot metrics extended, `get()` removed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+        private final UUID rqId;
+
+        /** Source node id which trigger request. */
+        private final UUID srcNodeId;
+
+        /** Snapshot name. */
+        private final String snpName;
+
+        @GridToStringInclude
+        /** The list of cache groups to include into snapshot. */
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+        private final List<Integer> grpIds;
+
+        @GridToStringInclude
+        /** The list of affected by snapshot operation baseline nodes. */
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+            return S.toString(SnapshotStartDiscoveryMessage.class, this);
+        }
+    }
+
+    /** */
+    private static class ClusterSnapshotFuture extends GridFutureAdapter<Void> {
+        /** Unique snapshot request id. */
+        private final UUID rqId;
+
+        /** Snapshot name */
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotFutureTask.java
```diff
@@ -0,0 +1,881 @@
...
+import static org.apache.ignite.internal.processors.cache.persistence.snapshot.IgniteSnapshotManager.partDeltaFile;
+
+/**
+ *
+ */
+class SnapshotFutureTask extends GridFutureAdapter<Boolean> implements DbCheckpointListener {
+    /** Shared context. */
+    private final GridCacheSharedContext<?, ?> cctx;
+
+    /** Ignite logger */
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+
+        /** {@inheritDoc} */
+        @Override public boolean equals(Object o) {
+            if (this == o)
+                return true;
+
+            if (o == null || getClass() != o.getClass())
+                return false;
+
+            SnapshotStartDiscoveryMessage message = (SnapshotStartDiscoveryMessage)o;
```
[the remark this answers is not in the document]
Mmuzaf: I've removed `equals` and `hashCode` methods.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+
+    /** Snapshot operation start message. */
+    private static class SnapshotStartDiscoveryMessage implements SnapshotDiscoveryMessage {
+        /** Serial version UID. */
+        private static final long serialVersionUID = 0L;
+
+        /** Discovery cache. */
+        private final DiscoCache discoCache;
+
+        /** Snapshot request id */
```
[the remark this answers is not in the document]
Mmuzaf: I've removed it since the `id` from InitMessage is used.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+                }
+                else if (evt.type() == EVT_NODE_LEFT || evt.type() == EVT_NODE_FAILED) {
+                    SnapshotOperationRequest snpRq = clusterSnpRq;
+
+                    for (SnapshotFutureTask sctx : locSnpTasks.values()) {
+                        if (sctx.sourceNodeId().equals(leftNodeId) ||
+                            (snpRq != null &&
+                                snpRq.snpName.equals(sctx.snapshotName()) &&
+                                snpRq.bltNodes.contains(leftNodeId))) {
+                            sctx.acceptException(new ClusterTopologyCheckedException("The node which requested snapshot " +
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1986 @@
...
+                lastSeenSnpFut = snpFut0;
+            }
+
+            List<Integer> grps = cctx.cache().persistentGroups().stream()
+                .filter(g -> cctx.cache().cacheType(g.cacheOrGroupName()) == CacheType.USER)
+                .filter(g -> !g.config().isEncryptionEnabled())
+                .map(CacheGroupDescriptor::groupId)
+                .collect(Collectors.toList());
+
+            List<ClusterNode> srvNodes = cctx.discovery().serverNodes(AffinityTopologyVersion.NONE);
```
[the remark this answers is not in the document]
Mmuzaf: I've fixed it. Check baseline  nodes and cache groups prior to snapshot task created.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotFutureTask.java
```diff
@@ -0,0 +1,881 @@
...
+        try {
+            if (!started.compareAndSet(false, true))
+                return false;
+
+            tmpSnpDir = U.resolveWorkDirectory(tmpTaskWorkDir.getAbsolutePath(),
+                databaseRelativePath(cctx.kernalContext().pdsFolderResolver().resolveFolders().folderName()),
+                false);
+
+            for (Integer grpId : parts.keySet()) {
+                CacheGroupContext gctx = cctx.cache().cacheGroup(grpId);
```
alex-plekhanov, line 308: Cache context sometimes is absent on some nodes (if `NodeFilter` used). I think it's safer here to skip groups if we can't find context for them and fail only if partitions for these groups ware requested explicitly. Also, we should cover this behavior by test (check that we can restore from such a snapshot)

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
```diff
@@ -0,0 +1,734 @@
...
+                ignite = startGrid(cfg);
+            }
+
+            ignite.cluster().baselineAutoAdjustEnabled(false);
+            ignite.cluster().state(ACTIVE);
+
+            for (int i = 0; i < CACHE_KEYS_RANGE; i++)
+                ignite.cache(DEFAULT_CACHE_NAME).put(i, i);
+
+            forceCheckpoint();
```
alex-plekhanov, line 574: Why do we need explicit checkpoint here?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
```diff
@@ -0,0 +1,734 @@
...
+    @Test
+    public void testClusterSnapshotMetrics() throws Exception {
+        String newSnapshotName = SNAPSHOT_NAME + "_new";
+        CountDownLatch deltaApply = new CountDownLatch(1);
+        CountDownLatch deltaBlock = new CountDownLatch(1);
+        IgniteEx ignite = startGridsWithCache(2, dfltCacheCfg, CACHE_KEYS_RANGE);
+
+        MetricRegistry mreg0 = ignite.context().metric().registry(SNAPSHOT_METRICS);
+
+        LongMetric startTime = mreg0.findMetric("LastSnapshotStartTime");
```
alex-plekhanov, line 602: Also please check all metrics before first snapshot attempt.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
```diff
@@ -0,0 +1,734 @@
...
+import static org.apache.ignite.internal.processors.cache.persistence.snapshot.IgniteSnapshotManager.SNP_NODE_STOPPING_ERR_MSG;
+import static org.apache.ignite.internal.processors.cache.persistence.snapshot.IgniteSnapshotManager.isSnapshotOperation;
+import static org.apache.ignite.internal.processors.cache.persistence.snapshot.IgniteSnapshotManager.resolveSnapshotWorkDirectory;
+import static org.apache.ignite.testframework.GridTestUtils.assertThrowsAnyCause;
+import static org.apache.ignite.testframework.GridTestUtils.assertThrowsWithCause;
+
+/**
+ * Cluster-wide snapshot test.
+ */
+public class IgniteClusterSnapshotSelfTest extends AbstractSnapshotSelfTest {
```
alex-plekhanov, line 85: There should be test added for transactional consistency, i.e.:
1. Start the concurrent transactional activity with some checkable invariant (for example set random balance for accounts on init, concurrently transfer random sums from one account two another, invariant: total sum for all accounts at any time should be the same)
2. Take snapshot.
3. Stop transactional activity.
4. Restore from the snapshot and check invariant.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManagerSelfTest.java
```diff
@@ -0,0 +1,770 @@
...
+
+        IgniteInternalFuture<?> snpFut = startLocalSnapshotTask(cctx0,
+            SNAPSHOT_NAME,
+            F.asMap(CU.cacheId(DEFAULT_CACHE_NAME), null),
+            new DelegateSnapshotSender(log, mgr.snapshotExecutorService(), mgr.localSnapshotSenderFactory().apply(SNAPSHOT_NAME)) {
+                @Override public void sendPart0(File part, String cacheDirName, GroupPartitionId pair, Long length) {
+                    try {
+                        U.await(cpLatch);
+
+                            delegate.sendPart0(part, cacheDirName, pair, length);
```
alex-plekhanov, line 654: Wrong indent

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManagerSelfTest.java
```diff
@@ -0,0 +1,770 @@
...
+        /** Serial version. */
+        private static final long serialVersionUID = 0L;
+
+        /** Order key. */
+        private final int key;
+
+        /** Order value. */
+        private final int value;
+
+        public TestOrderItem(int key, int value) {
```
alex-plekhanov, line 741: Comment is absent

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManagerSelfTest.java
```diff
@@ -0,0 +1,770 @@
...
+    /** */
+    private static class TestOrderItem implements Serializable {
+        /** Serial version. */
+        private static final long serialVersionUID = 0L;
+
+        /** Order key. */
+        private final int key;
+
+        /** Order value. */
+        private final int value;
```
alex-plekhanov, line 739: Abbreviation should be used

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
```diff
@@ -0,0 +1,734 @@
...
+            }
+        });
+
+        // Start cache load
+        IgniteInternalFuture<Long> loadFut = GridTestUtils.runMultiThreadedAsync(() -> {
+            try {
+                U.await(loadLatch);
+
+                while (!Thread.currentThread().isInterrupted()) {
+                    int txIdx = R.nextInt(grids);
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotFutureTask.java
Mmuzaf: I've fixed this behavior. Tests for cluster and remote snapshot have been added too.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
```diff
@@ -0,0 +1,734 @@
...
+
+                    // zero out the sign bit
+                    grid(txIdx).cache(txCcfg.getName()).put(txKey.incrementAndGet(), R.nextInt() & Integer.MAX_VALUE);
+
+                    int atomicIdx = R.nextInt(grids);
+
+                    grid(atomicIdx).cache(DEFAULT_CACHE_NAME).put(atKey.incrementAndGet(), R.nextInt() & Integer.MAX_VALUE);
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
Mmuzaf: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
Mmuzaf: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManagerSelfTest.java
Mmuzaf: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManagerSelfTest.java
Mmuzaf: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManagerSelfTest.java
Mmuzaf: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManagerSelfTest.java
```diff
@@ -0,0 +1,770 @@
...
+
+                if (!processed.isEmpty())
+                    ldrLatch.countDown();
+            }
+        });
+
+        try {
+            snpFutTask.start();
+
+            // Change data before snapshot creation which must be included into it witch correct value multiplier.
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
Mmuzaf: Test added.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
```diff
@@ -83,15 +84,12 @@
...
-    /** Random instance. */
-    private static final Random R = new Random();
-
     /** Time to wait while rebalance may happen. */
     private static final long REBALANCE_AWAIT_TIME = GridTestUtils.SF.applyLB(10_000, 3_000);
 
     /** Cache configuration for test. */
-    private static CacheConfiguration<Integer, Integer> txCcfg = new CacheConfiguration<Integer, Integer>("txCacheName")
-        .setAtomicityMode(CacheAtomicityMode.TRANSACTIONAL)
+    private static CacheConfiguration<Integer, Integer> atomicCcfg = new CacheConfiguration<Integer, Integer>("txCacheName")
```
alex-plekhanov, line 91: `txCacheName` for atomic cache looks confusing

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
```diff
@@ -286,6 +293,134 @@ public void testSnapshotPrimaryBackupsTheSame() throws Exception {
...
+            startGrid(optimize(getConfiguration(getTestIgniteInstanceName(i)).setCacheConfiguration(eastCcfg, westCcfg)));
+
+        grid(0).cluster().state(ACTIVE);
+
+        Ignite client = startClientGrid(grids);
+
+        IgniteCache<Integer, Account> eastCache = client.cache(eastCcfg.getName());
+        IgniteCache<Integer, Account> westCache = client.cache(westCcfg.getName());
+
+        // Create clients with zero balance.
```
alex-plekhanov, line 321: With initial balance

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
```diff
@@ -286,6 +293,134 @@ public void testSnapshotPrimaryBackupsTheSame() throws Exception {
...
+
+                        try (Transaction tx = ignite.transactions().txStart()) {
+                            Integer id = rnd.nextInt(clients);
+
+                            Account acc0 = east.get(id);
+                            Account acc1 = west.get(id);
+
+                            acc0.balance -= amount;
+
+                            txStarted.countDown();
```
alex-plekhanov, line 354: Can be placed to the beginning of tx try block for better readability with the same effect

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
```diff
@@ -286,6 +293,134 @@ public void testSnapshotPrimaryBackupsTheSame() throws Exception {
...
+        IgniteCache<Integer, Account> westCache = client.cache(westCcfg.getName());
+
+        // Create clients with zero balance.
+        for (int i = 0; i < clients; i++) {
+            eastCache.put(i, new Account(i, balance));
+            westCache.put(i, new Account(i, balance));
+        }
+
+        assertEquals("The initial summary value in all caches is not correct.",
+            total, sumAllCacheValues(client, clients, eastCcfg.getName(), westCcfg.getName()));
```
alex-plekhanov, line 328: `clients` here seems like we want to pass some collection to `sumAllCacheValues`. Let's rename this variable to `clientsCnt` or use scan query inside `sumAllCacheValues` and remove `clients` parameter

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
```diff
@@ -286,6 +293,134 @@ public void testSnapshotPrimaryBackupsTheSame() throws Exception {
...
+
+        IgniteEx ig0 = grid(0);
+
+        ig0.cluster().baselineAutoAdjustEnabled(false);
+        ig0.cluster().state(ACTIVE);
+
+        for (int i = 0; i < CACHE_KEYS_RANGE; i++)
+            ig0.getOrCreateCache(ccfg).put(i, i);
+
+        forceCheckpoint();
```
alex-plekhanov, line 413: Redundant

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManagerSelfTest.java
```diff
@@ -354,6 +354,42 @@ public void testSnapshotCreateLocalCopyPartitionFail() throws Exception {
...
+            startGrid(optimize(getConfiguration(getTestIgniteInstanceName(i)).setCacheConfiguration()));
+
+        IgniteEx ig0 = grid(0);
+        ig0.cluster().baselineAutoAdjustEnabled(false);
+        ig0.cluster().state(ACTIVE);
+
+        for (int i = 0; i < CACHE_KEYS_RANGE; i++)
+            ig0.getOrCreateCache(ccfg).put(i, i);
+
+        forceCheckpoint();
```
alex-plekhanov, line 374: Redundant

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
Mmuzaf: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
Mmuzaf: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
Mmuzaf: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
Mmuzaf: Renamed.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
Mmuzaf: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManagerSelfTest.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotFutureTask.java
```diff
@@ -0,0 +1,881 @@
...
+        if (stopping())
+            return;
+
+        // Snapshot task is now started since checkpoint write lock released.
+        if (!startedFut.onDone())
+            return;
+
+        assert !processed.isEmpty() : "Partitions to process must be collected under checkpoint mark phase";
+
+        wrapExceptionIfStarted(() -> snpSndr.init(processed.values().stream().mapToInt(Set::size).sum()))
```
alex-plekhanov, line 464: Do we really need to start other futures in case of failure on init?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotSender.java
```diff
@@ -0,0 +1,234 @@
...
+ *
+ */
+abstract class SnapshotSender {
+    /** Busy processing lock. */
+    private final ReadWriteLock lock = new ReentrantReadWriteLock();
+
+    /** Executor to run operation at. */
+    private final Executor exec;
+
+    /** {@code true} if sender is currently working */
```
alex-plekhanov, line 43: Point

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/FileReceiver.java
```diff
@@ -82,7 +82,8 @@ public FileReceiver(
             fileIo.position(meta.offset());
         }
         catch (IOException e) {
-            throw new IgniteException("Unable to open destination file. Receiver will will be stopped", e);
+            throw new IgniteException("Unable to open destination file. Receiver will will be stopped: " +
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -1404,7 +1419,7 @@ private ExchangeType onServerNodeEvent(boolean crd) throws IgniteCheckedExceptio
     /**
      * @return Exchange type.
      */
-    private ExchangeType onExchangeFreeSwitch() {
+    private ExchangeType onExchangeFreeSwitchOnLeft() {
```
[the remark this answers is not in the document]
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -156,7 +156,6 @@
     /** First eviction was warned flag. */
     private volatile boolean firstEvictWarn;
 
-
```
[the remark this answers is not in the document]
Mmuzaf: Changes reverted

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -937,6 +939,19 @@ else if (msg instanceof WalStateAbstractMessage)
             for (PartitionsExchangeAware comp : cctx.exchange().exchangeAwareComponents())
                 comp.onInitAfterTopologyLock(this);
 
+            // For pme-free exchanges onInitAfterTopologyLock must be
+            // invoked prior to onDoneBeforeTopologyUnlock
```
alex-plekhanov, line 943: Point

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1944 @@
...
+
+            cctx.exchange().unregisterExchangeAwareComponent(this);
+        }
+        finally {
+            busyLock.unblock();
+        }
+    }
+
+    /**
+     * @param snpDir Snapshot dire
```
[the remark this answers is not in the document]
Mmuzaf: fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotFutureTask.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotSender.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1944 @@
...
+
+                U.ensureDirectory(dbDir, "snapshot work directory", log);
+            }
+            catch (IgniteCheckedException e) {
+                throw new IgniteException(e);
+            }
+            finally {
+                cctx.database().checkpointReadUnlock();
+            }
+
```
alex-plekhanov, line 1659: Redundant NL

Mmuzaf: Fixed.

modules/core/src/main/java/org/apache/ignite/IgniteSnapshot.java
```diff
@@ -0,0 +1,48 @@
...
+ * grantee data consistency between them.</li>
+ * <li>Snapshot must be resorted manually on the switched off cluster by copying data
+ * to the working directory on each cluster node.</li>
+ * </ul>
+ */
+public interface IgniteSnapshot {
+    /**
+     * @return List of all known snapshots.
+     */
+    public List<String> getSnapshots();
```
agoncharuk, line 39: Should we return a object instead of a String if we add more information about snapshot in future?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotFutureTask.java
```diff
@@ -0,0 +1,882 @@
...
+                wrapExceptionIfStarted(() -> {
+                    List<File> ccfgs = storeMgr.configurationFiles(gctx.config());
+
+                    if (ccfgs == null)
+                        return;
+
+                    for (File ccfg0 : ccfgs)
+                        snpSndr.sendCacheConfig(ccfg0, cacheDirName(gctx.config()));
+                }),
+                snpSndr.executor())
```
agoncharuk, line 518: Looks like there is a race here: we read the configuration file from snapshot executor thread while the files can be modified by SQL create/drop column commands, for example.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1943 @@
...
+    /** {@inheritDoc} */
+    @Override public List<String> getSnapshots() {
+        if (cctx.kernalContext().clientNode())
+            throw new UnsupportedOperationException("Client and daemon nodes can not perform this operation.");
+
+        synchronized (snpOpMux) {
+            return Arrays.stream(locSnpDir.listFiles(File::isDirectory))
+                .map(File::getName)
+                .collect(Collectors.toList());
+        }
```
agoncharuk, line 969: If a baseline node was offline during the snapshot creation, the result of this call on the node will not return the snapshot which was successfully created on other nodes, which will be really confusing for the user. We also need to add a test with snapshot creation with offline nodes. 

I would rework this logic to collect information from all nodes in the cluster and merge it. Also, why not support clients by sending a compute to server nodes?

If we want to keep this behavior, we should rename the method to getSnapshotsLocal, because the current combination of methods in the interface has different semantics.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1943 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override public IgniteFuture<Void> createSnapshot(String name) {
+        A.notNullOrEmpty(name, "name");
+
+        try {
+            if (cctx.kernalContext().clientNode())
+                throw new UnsupportedOperationException("Client and daemon nodes can not perform this operation.");
+
```
agoncharuk, line 979: Why not support snapshot trigger from client via compute?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
```diff
@@ -0,0 +1,895 @@
...
+
+        snpIg0.cache(DEFAULT_CACHE_NAME).query(new ScanQuery<>(null))
+            .forEach(e -> assertTrue("Snapshot must contains only negative values " +
+                "[cache=" + DEFAULT_CACHE_NAME + ", entry=" + e +']', (Integer)e.getValue() < 0));
+
+        snpIg0.cache(atomicCcfg.getName()).query(new ScanQuery<>(null))
+            .forEach(e -> assertTrue("Snapshot must contains only negative values " +
+                "[cache=" + atomicCcfg.getName() + ", entry=" + e + ']', (Integer)e.getValue() < 0));
+    }
+
```
agoncharuk, line 225: We need to add tests that check that SQL indexes are successfully restored and used, including statically configured indexes and dynamically created indexes.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
```diff
@@ -0,0 +1,895 @@
...
+        stopAllGrids();
+
+        IgniteEx snp = startGridsFromSnapshot(grids,
+            cfg -> resolveSnapshotWorkDirectory(cfg.setCacheConfiguration()).getAbsolutePath(),
+            SNAPSHOT_NAME,
+            true);
+
+        awaitPartitionMapExchange();
+
+        assertSnapshotCacheKeys(snp.cache(ccfg.getName()));
```
agoncharuk, line 423: Additionally check that node filter is preserved.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
```diff
@@ -0,0 +1,895 @@
...
+
+        fut.get();
+
+        stopAllGrids();
+
+        IgniteEx snp = startGridsFromSnapshot(3, SNAPSHOT_NAME);
+
+        awaitPartitionMapExchange();
+
+        assertSnapshotCacheKeys(snp.cache(dfltCacheCfg.getName()));
```
agoncharuk, line 686: Need to check that the rebalancing actually finished and affinity reassignment is actually triggered after snapshot restore.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1943 @@
...
+    }
+
+    /**
+     * @param snpName Snapshot name.
+     * @return Local snapshot directory for snapshot with given name.
+     */
+    public File snapshotLocalDir(String snpName) {
+        assert locSnpDir != null;
+
+        return new File(locSnpDir, snpName);
```
agoncharuk, line 777: From the API standpoint there is no indication that snapshot name will be used as a folder name. A proper path escaping and translation logic should be added, with the corresponding test.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -0,0 +1,1943 @@
...
+                lastSeenSnpFut = snpFut0;
+            }
+
+            List<Integer> grps = cctx.cache().persistentGroups().stream()
+                .filter(g -> cctx.cache().cacheType(g.cacheOrGroupName()) == CacheType.USER)
+                .filter(g -> !g.config().isEncryptionEnabled())
+                .map(CacheGroupDescriptor::groupId)
+                .collect(Collectors.toList());
+
+            List<ClusterNode> srvNodes = cctx.discovery().serverNodes(AffinityTopologyVersion.NONE);
```
agoncharuk, line 1015: At some point during the snapshot creation, we need to check the partitions availability (offline baseline nodes, LOST partitions) and prohibit snapshot creation if at least one partition is unavailable.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
Mmuzaf: Good point. Since the restore procedure is fully manual, I think it's better to add a pattern for checking snapshot name. So users will see a snapshot directory on disk with exactly the same name with which they requested a snapshot creation.

`[a-zA-Z_0-9]` pattern added for snapshot name on `createSnapshot`.

modules/core/src/main/java/org/apache/ignite/IgniteSnapshot.java
Mmuzaf: For the initial create snapshot procedure I'd like to focus only on locally available information about snapshots on each node. So, I've renamed this method to `getSnapshotNamesLocal` which will return all snapshot names on the local node. 

If we decide to work further and additional snapshot meta needs to be collected prior to a snapshot creation. The API can be extended with, for instance: 
- `public List<SnapshotMeta> getSnapshots();` -- globally known snapshots.
- `snapshots` view with global snapshots

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
Mmuzaf: Added.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
Mmuzaf: `awaitPartitionMapExchange` will do such check (that `CacheAffinityChangeMessage` happened).
An additional check that all partitions for cache in the snapshot will have an `OWNING` state added.

modules/core/src/main/java/org/apache/ignite/IgniteSnapshot.java
Mmuzaf: I've decided to remove from java API this method since it has no sense. I think it's better to extend API when the snapshots view and cluster-wide information about snapshot will be implemented.

`LocalSnapshotNames` are available through the JMX.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotFutureTask.java
Mmuzaf: You right. I've completely missed that cache configuration files can be changed during SQL add/drop a column, create/drop an index.

I've fixed it and add an appropriate test for such a case.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotSelfTest.java
Mmuzaf: Agree. I've added such a test.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
Mmuzaf: Alexey, yes we can it support. Let's do it in a separate JIRA minor issue.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
Mmuzaf: 1. I've removed the `getSnapshots()` method since it returns only local results and I agree with you it's confusing behaviour for users. I will create a separate issue which will allow getting cluster snapshots info via java API and via `snapshots` view.

2. I've added test for snapshot creation when some of BLT nodes offline.

3. I'll create a separate issue to allow sending compute from the client node for snapshot creation.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
Mmuzaf: Alexey, these guarantees currently we've implemented it this PR:
1. If the affected BLT node left the cluster the snapshot process will be interrupted, intermediate changes will be deleted. Users will get an error.
2. Cache stop request currently blocked if a snapshot operation `in-progress`.
3. We are adding to snapshot only `OWNING` partitions if some of the partitions missed users will see `warn` message in the log which partitions (index) were missed.
4. If for of requested cache groups `CacheGroupContext` is null the snapshot operation will be interrupted.
5. If partition files are not accessible the snapshot operation will be interrupted and the user will see an error.

Please let me know if I'm missing something.

Mmuzaf:
@agoncharuk 

These JIRA issues I'm planning to complete after the initial PR will be merged (probably not all of them, depend on user-list activity):

- support snapshot creation from a client node
- add starting snapshot via control.sh
- add system snapshot view with all available snapshots on the cluster
- write snapshot meta-file on each node with additional snapshot info (baseline, caches, partitions missed during snapshot operation)
- support encryption caches for a snapshot
- remove `reject cache group stop` requirement for a snapshot operation

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
agoncharuk: I was thinking about the following case: 4 nodes, 1 backup. Say, partition 0 is located on nodes A and B. Nodes A and B stop before snapshot is initiated. At this point we will have partition either on LOST state, or we will not have this partition in partitions map. If the snapshot is triggered after A and B went down, the snapshot creation will not hit any of the conditions you've mentioned: no nodes will leave the cluster during the snapshot, there will be no partitions in non-OWNING state (the rebalance will not start because baseline did not change).

Mmuzaf: In my view, a snapshot procedure just `freezes` the current cluster state. If the cluster has `LOST` partitions (not all baseline nodes are online) why we should interrupt the snapshot procedure? The process creates a cluster snapshot `as-is`.

Your case is mostly related to the recovery procedure. The additional information for the created snapshot will be added further and, for example, in snapshot-view users will see that the snapshot has `LOST` partitions and it's the point of decision - will he start recovery from this snapshot or not. 

Any way we can add a warning message. 
WDYT?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotMXBeanImpl.java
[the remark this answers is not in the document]
sk0x50: Hello @Mmuzaf,
Do we really need to publish this functionality through JMX? What is the reason to do that?
We already have activation/deactivation, stop txs, killing queries and tasks, etc.
Why do we try to duplicate **all** java API's here? Is it really needed and useful?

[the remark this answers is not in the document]
Mmuzaf: @sk0x50 Hello, 

The `control.sh` procedures currently are not implemented, so it's the only chance for an administrator to start a snapshot operation for the cluster without writing java code. 
So I think yes, we should publish it like for the other tools.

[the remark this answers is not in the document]
sk0x50: Hi Maxim,

Hmm, IMHO we should add the ability to create a snapshot into the right place/tool right away. I don't think, that it requires a huge effort.
The command-line utility is a more appropriate instrument that is used by administrators.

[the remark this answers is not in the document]
Mmuzaf: Slava,

I don't see any disadvantages of having service JMX beans. As you mentioned before - we already have such beans for txs, deactivation, etc.

I agree with you, that command-line utility is 'must-have' for the snapshot process, but I'd like to keep this PR as simple as possible, and implementing such the utility in this PR will overcomplicate the review process, it requires a much more code changes and actually doesn't serves us to bring MVP of this improvement with minimal changes.

`control.sh' commands for snapshot will be implemented in a separate JIRA issue.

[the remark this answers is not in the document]
sk0x50: Hi Maxim,

Please don't get me wrong, I don't try to argue. I just want to understand the reason.

> As you mentioned before - we already have such beans for txs, deactivation, etc.
Yes, and this fact doesn't shed the light on the reason why all these beans try to duplicate control the control utility and Java API. :) Anyway, it is up to you. 

> `control.sh' commands for snapshot will be implemented in a separate JIRA issue.
OK, this is a good idea. Please add that ticket to the IEP epic.

Thanks,
Slava.

Mmuzaf:
> @agoncharuk
> 
> These JIRA issues I'm planning to complete after the initial PR will be merged (probably not all of them, depends on the user-list activity):
> 
> * support snapshot creation from a client node
> * add starting snapshot via control.sh
> * add system snapshot view with all available snapshots on the cluster
> * write snapshot meta-file on each node with additional snapshot info (baseline, caches, partitions missed during snapshot operation)
> * support encryption caches for a snapshot
> * remove `reject cache group stop` requirement for a snapshot operation

@sk0x50 

Please, take a look at these 'next-step' issues. I'll create an appropriate JIRAs for them right after current changes will be merged into the master branch.

agoncharuk:
@Mmuzaf should we close the PR as snapshots are merged to master?

Mmuzaf:
> @Mmuzaf should we close the PR as snapshots are merged to master?

Sure, I've closed the PR.

PR #7615, opened 2020-04-02
IGNITE-12764 Regression: Thin JDBC streaming fails/BatchUpdateException if function is used

x-kreator:
…on if function is used


modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/JdbcSqlStreamingTest.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+public class JdbcSqlStreamingTest extends AbstractIndexingCommonTest {
+    /** */
+    private static final int ROW_NUM = 1600;
+
+    /**
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testStreamingWithFunctionValues() throws Exception {
```
nizhikov, line 42: Can you, please, rewrite this test using Ignite code style?

https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -691,36 +691,58 @@ private long streamQuery0(String qry, String schemaName, IgniteDataStreamer stre
...
 
-            List<List<?>> planRows = plan.createRows(args != null ? args : X.EMPTY_OBJECT_ARRAY);
+            Object[] params = args != null ? args : X.EMPTY_OBJECT_ARRAY;
+
+            Iterator<List<?>> iter0;
+
+            if (!F.isEmpty(plan.selectQuery())) {
+                SqlFieldsQuery selectQry = new SqlFieldsQuery(plan.selectQuery())
+                    .setArgs(params)
+                    .setLocal(true);
```
nizhikov, line 701: Why this query is always `true`?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -691,36 +691,58 @@ private long streamQuery0(String qry, String schemaName, IgniteDataStreamer stre
         try {
             UpdatePlan plan = dml.plan();
 
-            List<List<?>> planRows = plan.createRows(args != null ? args : X.EMPTY_OBJECT_ARRAY);
+            Object[] params = args != null ? args : X.EMPTY_OBJECT_ARRAY;
```
nizhikov, line 694: It seems this piece of code should be rewritten as follows to minimize changes in the code and improve readability:

```
    private long streamQuery0(String qry, String schemaName, IgniteDataStreamer streamer, QueryParserResultDml dml,
        final Object[] args) throws IgniteCheckedException {
        Long qryId = runningQryMgr.register(qry, GridCacheQueryType.SQL_FIELDS, schemaName, true, null);

        Exception failReason = null;

        try {
            UpdatePlan plan = dml.plan();

            Iterator<List<?>> iter = updateQueryRows(schemaName, plan, args);

            if (!iter.hasNext())
                return 0;

            IgniteBiTuple t = plan.processRow(iter.next());

            if (!iter.hasNext()) {
                streamer.addData(t.getKey(), t.getValue());

                return 1;
            }
            else {
                Map<Object, Object> rows = new LinkedHashMap<>(plan.rowCount());

                rows.put(t.getKey(), t.getValue());

                while (iter.hasNext()) {
                    List<?> row = iter.next();

                    t = plan.processRow(row);

                    rows.put(t.getKey(), t.getValue());
                }

                streamer.addData(rows);

                return rows.size();
            }
        }
        catch (IgniteException | IgniteCheckedException e) {
            failReason = e;

            throw e;
        }
        finally {
            runningQryMgr.unregister(qryId, failReason);
        }
    }

    /**
     * Calculates rows for update query.
     *
     * @param schemaName Schema name.
     * @param plan Update plan.
     * @param args Statement arguments.
     * @return Rows for update.
     */
    Iterator<List<?>> updateQueryRows(String schemaName, UpdatePlan plan, Object[] args) throws IgniteCheckedException {
        Object[] params = args != null ? args : X.EMPTY_OBJECT_ARRAY;

        if (!F.isEmpty(plan.selectQuery())) {
            SqlFieldsQuery selectQry = new SqlFieldsQuery(plan.selectQuery())
                .setArgs(params)
                .setLocal(true);

            QueryParserResult selectParseRes = parser.parse(schemaName, selectQry, false);

            GridQueryFieldsResult res = executeSelectLocal(
                selectParseRes.queryDescriptor(),
                selectParseRes.queryParameters(),
                selectParseRes.select(),
                null,
                null,
                null,
                false,
                0
            );

            return res.iterator();
        } else
            return plan.createRows(params).iterator();
    }
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/JdbcSqlStreamingTest.java
x-kreator: Done.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
x-kreator: Done. I had same idea about updateQueryRows method, but I refrained from implementing it earlier due to the fact that some reviewers don't like methods used only once.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
x-kreator: That behavior was implemented before was dropped at commit 6c9a17b4 (pr #6096 by vozerov), so I decided to return it with minimal differences.

nizhikov: Why do you think that changes in 6c9a17b is wrong?

x-kreator: Not all but condition removed from line 816 and assertion added to line 814.
That assertion was also removed at 1619dd56 (pr #6112).
Assume that adding of assertion is right, then it's logical to expect that we should meet the changes which fix that assertion failures happening when we use expressions at streaming inserts. But all further changes keep the bug described above till 1619dd56. After 1619dd56 we became to meet the bug described in our issue instead of assertion error.

PR #7638, opened 2020-04-07
IGNITE-12871: Calcite integration. Broadcast to hash conversion without exchange

gvvinblade:
…ut exchange.

### DIFF
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteHashFilter.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteHashFilter.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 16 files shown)


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteHashFilter.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.ignite.internal.processors.query.calcite.prepare.RelTargetAware;
+import org.apache.ignite.internal.processors.query.calcite.trait.DistributionTraitDef;
+
+import static org.apache.calcite.rel.RelDistribution.Type.BROADCAST_DISTRIBUTED;
+import static org.apache.calcite.rel.RelDistribution.Type.HASH_DISTRIBUTED;
+
+/**
+ *
+ */
+public class IgniteHashFilter extends SingleRel implements IgniteRel, RelTargetAware {
```
rkondakov, line 35: IMO it is a poor name for this process. It can be easily confused with `IgniteFilter` despite it is a kind of exchange. I think we need to rename it to something more meaningful. For example in the [SQL Server PDW paper](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.473.9943&rep=rep1&type=pdf) this node is called "Trim move". What if we call it in the same manner: `TreamExchange`? Or simply `BroadcastToHashExchange`?

gvvinblade: You right the name is ugly, I tried to differ the rel node from Exchange since IgniteExchange is a special logical node, a cutting point for a splitter while IgniteHashFilter describes a physical operation. TrimExchange sounds better, it at least describes what actually is being done here.

PR #7639, opened 2020-04-08
IGNITE-12438 Inverse TCP communication via Discovery implemented.

ibessonov:


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -2323,6 +2318,14 @@ private void dumpInfo(StringBuilder sb, UUID dstNodeId) {
             throw new IgniteSpiException("Failed to initialize TCP server: " + locHost, e);
         }
 
+        EnvironmentType envType = ignite.configuration().getEnvironmentType();
+
+        if (usePairedConnections) {
+            if (envType == EnvironmentType.VIRTUALIZED)
+                throw new IgniteSpiException("Node using paired connections " +
+                    "is not allowed to start in virtualized environment.");
```
agoncharuk, line 2326: We need to suggest which configuration properties a user has to change in order to resolve this error

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -3491,6 +3500,19 @@ protected GridCommunicationClient createTcpClient(ClusterNode node, int connIdx)
      * @throws IgniteCheckedException If establish connection fails.
      */
     private GridNioSession createNioSession(ClusterNode node, int connIdx) throws IgniteCheckedException {
+        boolean locNodeIsSrv = !getLocalNode().isClient() && !getLocalNode().isDaemon();
+
+        if (!(Thread.currentThread() instanceof IgniteDiscoveryThread) && locNodeIsSrv) {
+            if (node.isClient() && startedInVirtualizedEnvironment(node)) {
+                String msg = "Failed to connect to node " + node.id() +
+                    " because it is started n virtualized environment; inverse connection will be requested.";
```
agoncharuk, line 3508: Typo: n->in

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -3570,6 +3578,35 @@ public IgniteConfiguration setSqlSchemas(String... sqlSchemas) {
         return this;
     }
 
+    /**
+     * <b>This is an experimental feature. Envronment awareness approac may be changed.</b>
+     * <p>
+     *
+     * Configured environment type.
```
agoncharuk, line 3585: I suggest to expand javadoc here and describe that the environment type provides a hint to Ignite whether the Ignite instance is launched on a standalone machine or inside a virtualized environment. Describe what differences are important for virtualized environment (currently, NAT).

PR #7647, opened 2020-04-09
IGNITE-12879 Refactors test configuration of discovery messages interceptors.

ololo3000:


modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TestTcpDiscoverySpi.java
```diff
@@ -44,4 +52,19 @@
...
+        super.setListener(lsnr ==  null || discoveryHooks == null ? lsnr : wrap(lsnr, discoveryHooks));
+    }
+
+    /**
+     * Sets interceptors of discovery messages.
+     *
+     * @param discoveryHooks Interceptors of discovery messages.
+     */
+    @SuppressWarnings("AssignmentOrReturnOfFieldWithMutableType")
+    public void discoveryHooks(DiscoveryHook... discoveryHooks) {
```
NSAmelchev, line 67: Why array needed?

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TestTcpDiscoverySpi.java
```diff
@@ -44,4 +52,19 @@
...
         super.simulateNodeFailure();
     }
+
+    /** {@inheritDoc} */
+    @Override public void setListener(@Nullable DiscoverySpiListener lsnr) {
+        super.setListener(lsnr ==  null || discoveryHooks == null ? lsnr : wrap(lsnr, discoveryHooks));
+    }
+
+    /**
+     * Sets interceptors of discovery messages.
```
NSAmelchev, line 62: I think we should add javadoc and assert that hook should be set before Discovery started.

modules/core/src/test/java/org/apache/ignite/internal/processors/continuous/IgniteContinuousQueryMetadataUpdateTest.java
```diff
@@ -57,11 +55,26 @@
     @Override protected IgniteConfiguration getConfiguration(String igniteInstanceName) throws Exception {
         IgniteConfiguration cfg = super.getConfiguration(igniteInstanceName);
 
-        TestDiscoverySpi discoSpi = new TestDiscoverySpi();
```
NSAmelchev, line 60: Why this changed?

modules/core/src/test/java/org/apache/ignite/internal/processors/continuous/IgniteContinuousQueryMetadataUpdateTest.java
```diff
@@ -57,11 +55,26 @@
...
+        TestTcpDiscoverySpi discoSpi = (TestTcpDiscoverySpi)cfg.getDiscoverySpi();
 
-        discoSpi.setIpFinder(IP_FINDER);
+        DiscoveryHook discoveryHook = new DiscoveryHook() {
+            @Override public void beforeDiscovery(DiscoverySpiCustomMessage msg) {
+                DiscoveryCustomMessage customMsg = msg == null ?
```
NSAmelchev, line 62: I think we should provide `customMsg` as argument  (or instead of `DiscoverySpiCustomMessage msg`) and do not duplicate:
```
DiscoveryCustomMessage customMsg = msg == null ?
    null : (DiscoveryCustomMessage)IgniteUtils.field(msg, "delegate");
```

modules/core/src/test/java/org/apache/ignite/testframework/GridTestUtils.java
```diff
@@ -164,32 +169,89 @@ public void ignite(IgniteEx ignite) {
...
+            @Nullable DiscoverySpiCustomMessage spiCustomMsg
+        ) {
+            for (DiscoveryHook hook : hooks)
+                hook.beforeDiscovery(spiCustomMsg);
 
             IgniteFuture<?> fut = delegate.onDiscovery(type, topVer, node, topSnapshot, topHist, spiCustomMsg);
 
-            hook.afterDiscovery(spiCustomMsg);
+            for (DiscoveryHook hook : hooks)
+                hook.afterDiscovery(spiCustomMsg);
```
NSAmelchev, line 254: The `afterDiscovery` will be called not after discovery notifing listeners. `delegate.onDiscovery` is async.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/binary/BinaryMetadataUpdatesFlowTest.java
```diff
@@ -209,10 +197,7 @@ public void testFlowNoConflictsWithClients() throws Exception {
             return;
 
         discoveryHook = new DiscoveryHook() {
-            @Override public void handleDiscoveryMessage(DiscoverySpiCustomMessage msg) {
-                DiscoveryCustomMessage customMsg = msg == null ? null
```
nizhikov, line 213: Why this deletion?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/binary/GridCacheBinaryObjectMetadataExchangeMultinodeTest.java
```diff
@@ -129,13 +115,10 @@ public void testReadRequestBlockedOnUpdatingMetadata() throws Exception {
         discoveryHook = new DiscoveryHook() {
             private volatile IgniteEx ignite;
 
-            @Override public void handleDiscoveryMessage(DiscoverySpiCustomMessage msg) {
+            @Override public void beforeDiscovery(DiscoveryCustomMessage customMsg) {
                 if (finishFut.isDone())
                     return;
 
-                DiscoveryCustomMessage customMsg = msg == null ? null
```
nizhikov, line 136: Why this deletion?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/binary/BinaryMetadataUpdatesFlowTest.java
ololo3000: This logic was moved to DiscoveryHook class  to avoid duplication.

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TestTcpDiscoverySpi.java
```diff
@@ -44,4 +54,24 @@
     @Override public void simulateNodeFailure() {
         super.simulateNodeFailure();
     }
+
+    /** {@inheritDoc} */
+    @SuppressWarnings("ZeroLengthArrayAllocation")
+    @Override public void setListener(@Nullable DiscoverySpiListener lsnr) {
+        super.setListener(lsnr ==  null || discoHooks.isEmpty() ?
```
nizhikov, line 61: Is the following flow possible? If no, why?

1. addDiscoveryHook
2. setListener
3. addDiscoveryHook <- this hook will be ignored.

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TestTcpDiscoverySpi.java
```diff
@@ -31,6 +38,9 @@
     /** */
     public boolean ignorePingResponse;
 
+    /** Interceptors of discovery messages. */
+    private List<DiscoveryHook> discoHooks = new ArrayList<>();
```
nizhikov, line 42: Let's allocate discoHooks on the first hook addition.

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TestTcpDiscoverySpi.java
nizhikov: It seems to me that we can just pass discoHooks into the listener.
Therefore list always be up to date inside the listener.

ololo3000: addDiscoveryHook method asserts that it must be called only before TcpDiscoverySpi start. So it's impossible to use addDiscoveryHook after listener is set.

nizhikov: OK then.

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TestTcpDiscoverySpi.java
ololo3000: Thanks. Done.

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TestTcpDiscoverySpi.java
ololo3000: > It seems to me that we can just pass discoHooks into the listener.
> Therefore list always be up to date inside the listener.

It's not clear for me in what cases hooks are needed to be modified after SPI start. I think that the current approach is more robust.

PR #7663, opened 2020-04-12
Ignite 12702

kartiksomani:
This code prints out warning log on call of put in Ignite cache, if AffinityKeyMapped is used on an value class attribute instead of a key class attribute.


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheProxyImpl.java
```diff
@@ -1288,9 +1293,26 @@ private void validate(Query qry) {
...
+                        "AffinityKeyMapped annotation should be used on key and not value type");
+            }
+
+        }
+    }
+
     /** {@inheritDoc} */
     @Override public void put(K key, V val) {
         IgniteInternalCache<K, V> delegate = getDelegateSafe();
+        checkIneffectiveAnnotations(val.getClass());
```
Mikhail1988, line 1315: Reflection use on each put, really? it will kill performance. Please move this check to new type registration method, I believe you will find it in BinaryMarshaller or somewhere around.

kartiksomani: Thank you, I called out the performance cost of reflection on the [issue](https://issues.apache.org/jira/browse/IGNITE-12702?jql=project%20%3D%20IGNITE%20AND%20labels%20in%20(newbie)%20and%20status%20%3D%20OPEN) itself almost a month back. Didn't know about BinaryMarshaller, will use it

kartiksomani: @Mikhail1988 Where can I find new type registration method? I did see binary type configurator

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheProxyImpl.java
```diff
@@ -45,10 +46,9 @@
 import javax.cache.processor.EntryProcessor;
 import javax.cache.processor.EntryProcessorException;
 import javax.cache.processor.EntryProcessorResult;
-import org.apache.ignite.IgniteCache;
-import org.apache.ignite.IgniteCacheRestartingException;
-import org.apache.ignite.IgniteCheckedException;
-import org.apache.ignite.IgniteException;
+
+import org.apache.ignite.*;
```
AMashenkov, line 50: Wildcard imports are not allowed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheProxyImpl.java
```diff
@@ -1288,9 +1293,26 @@ private void validate(Query qry) {
         }
     }
 
+    /**
+     * Warns if an ineffective annotation is used
+     *
+     * @param valueCls Value class
+     */
+    private void checkIneffectiveAnnotations(Class valueCls) {
+        for (Field f:valueCls.getDeclaredFields()) {
```
AMashenkov, line 1302: Missed spaces before and after ":"

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheProxyImpl.java
```diff
@@ -1288,9 +1293,26 @@ private void validate(Query qry) {
...
+            AffinityKeyMapped[] annotations = f.getDeclaredAnnotationsByType(AffinityKeyMapped.class);
+            if (annotations.length > 0) {
```
AMashenkov, line 1304: Potential NPE here. 'annotations' can be null.

PR #7673, opened 2020-04-14
IGNITE-12864: Calcite integration. UNION support.

gvvinblade:


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/trait/DistributionTrait.java
```diff
@@ -139,14 +139,34 @@
...
+            int source = mapping.getSourceOpt(i);
+            int keyPos = map[source] - 1;
+
+            if (keyPos == -1)
+                continue;
+
+            res[keyPos] = i;
+
+            if (++found == keys.size())
+                return IgniteDistributions.hash(ImmutableIntList.of(res), function);
```
rkondakov, line 164: isn't it the same as `Mappings.apply2`?

gvvinblade: Actually almost the same, but TargetMapping and Mapping are different interfaces

PR #7677, opened 2020-04-16
IGNITE-12343 Remote filter and transformer of ContinuousQueries have to run inside the Ignite Sandbox

dgarus:
[https://issues.apache.org/jira/browse/IGNITE-12343](https://issues.apache.org/jira/browse/IGNITE-12343)


modules/core/src/test/java/org/apache/ignite/internal/processors/security/sandbox/AbstractContinuousQuerySandboxTest.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * Getting the key that is contained on primary partition on passed node for cache.
+     *
+     * @param ignite Node.
+     * @param cacheName Cache name.
+     * @return Key.
+     */
+    private Integer primaryKey(IgniteEx ignite, String cacheName) {
```
alex-plekhanov, line 126: Why can't we use `GridCommonAbstractTest#primaryKey`?

modules/core/src/test/java/org/apache/ignite/internal/processors/security/sandbox/AbstractContinuousQuerySandboxTest.java
```diff
@@ -0,0 +1,132 @@
...
+    private static volatile AccessControlException error;
+
+    /** Runs control action for CQ tests. */
+    protected static final Runnable CONTROL_ACTION_RUNNER = new Runnable() {
+        @Override public void run() {
+            try {
+                controlAction();
+            }
+            catch (AccessControlException e) {
+                error = e;
```
alex-plekhanov, line 56: We should rethrow the exception, also, we should check that this exception don't break the "put" thread.

dgarus: Ok

modules/core/src/test/java/org/apache/ignite/internal/processors/security/sandbox/AbstractContinuousQuerySandboxTest.java
dgarus: I need to get a primary key for a specific node.
GridCommonAbstractTest#primaryKey isn't an appropriate method because it finds a primary key for a local node.

alex-plekhanov: `primaryKey(ignite.cache(cacheName))` will solve the problem

modules/core/src/test/java/org/apache/ignite/internal/processors/security/sandbox/AbstractContinuousQuerySandboxTest.java
```diff
@@ -0,0 +1,147 @@
...
+            cur.getAll();
+        }
+        catch (Exception e) {
+            if (init)
+                throw e;
+
+            fail(e.getMessage());
+        }
+
+        if (!init) {
```
alex-plekhanov, line 128: I think `cache.get(key) == 100` should be true in any case

modules/core/src/test/java/org/apache/ignite/internal/processors/security/sandbox/AbstractContinuousQuerySandboxTest.java
```diff
@@ -0,0 +1,147 @@
...
+        Integer key = primaryKey(grid(SRV), cache.getName());
+
+        try (QueryCursor<Cache.Entry<Integer, Integer>> cur = cache.query(q)) {
+            if (!init)
+                cache.put(key, 100);
+
+            cur.getAll();
+        }
+        catch (Exception e) {
+            if (init)
```
alex-plekhanov, line 122: I think it's more readable to wrap `cache.put(key, 100);` into try/catch block then check `init` flag.

modules/core/src/test/java/org/apache/ignite/internal/processors/security/sandbox/AbstractContinuousQuerySandboxTest.java
dgarus: ok

modules/core/src/test/java/org/apache/ignite/internal/processors/security/sandbox/AbstractContinuousQuerySandboxTest.java
dgarus: ok

modules/core/src/test/java/org/apache/ignite/internal/processors/security/sandbox/AbstractContinuousQuerySandboxTest.java
dgarus: ok

PR #7691, opened 2020-04-19
IGNITE-12579 Adds JDBC SQL queries authorization tests. 

ololo3000:


modules/clients/src/test/java/org/apache/ignite/jdbc/JdbcAuthorizationTest.java
```diff
@@ -0,0 +1,440 @@
...
+     * @return {@link SecurityPermissionSet} containing specified cache permissions.
+     */
+    private SecurityPermissionSet cachePermissions(String cache, SecurityPermission... perms) {
+        return create().defaultAllowAll(false).appendCachePermissions(cache, perms).build();
+    }
+
+    /**
+     * Provider of SQL statement.
+     */
+    static interface StatementProvider {
```
alex-plekhanov, line 425: StatementProvided implemented only twice, in the `of()` method and in the `testBulkInsert` method (which is not correct). I think this interface should be removed and `String sql` should be passed to `execute()` method instead.

modules/clients/src/test/java/org/apache/ignite/jdbc/JdbcAuthorizationTest.java
```diff
@@ -0,0 +1,440 @@
...
+     */
+    @Test
+    public void testDropIndex() throws Exception {
+        String table = "test_table_drop_index";
+
+        createTable(table);
+
+        execute(of("CREATE INDEX test_drop_idx ON " + table + "(id ASC);"), EMPTY_PERMS_USER);
+
+        execute(of("DROP INDEX test_drop_idx ON " + table + ';'), EMPTY_PERMS_USER);
```
alex-plekhanov, line 331: This test do the same as `testCreateIndex` + one extra action. In this case, `testCreateIndex` method is redundant and can be removed (this method should be renamed to `testCreateAndDropIndex`).

Also, what do we check in these tests? It looks strange that we can't limit creating and dropping columns and indexes and check it like it's expected behavior. A test should assert some reasonable statement, I see no statement to assert here.

ololo3000: I agree that it looks strange. I have asked the corresponding question on the dev-list but there was no answer yet. Could we keep this tests to only show the current authorization behavior?

modules/clients/src/test/java/org/apache/ignite/jdbc/JdbcAuthorizationTest.java
```diff
@@ -0,0 +1,440 @@
...
+
+        srv.createCache(ccfg);
+    }
+
+    /**
+     * Tests INSERT query permissions check.
+     */
+    @Test
+    public void testInsert() throws Exception {
+        int key = UUID.randomUUID().hashCode();
```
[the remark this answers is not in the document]
ololo3000: Thanks. Done.

modules/clients/src/test/java/org/apache/ignite/jdbc/JdbcAuthorizationTest.java
```diff
@@ -0,0 +1,440 @@
...
+    /**
+     * Tests BULK INSERT query permissions check.
+     */
+    @Test
+    public void testBulkInsert() throws Exception {
+        StatementProvider stmt = conn -> {
+            PreparedStatement res = conn.prepareStatement(
+                "INSERT INTO " + TEST_DML_TABLE + "(_key, _val) VALUES (?, ?);");
+
+            for (int i = 1; i < 10; i++) {
```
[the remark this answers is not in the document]
ololo3000: Thanks. Done.

modules/clients/src/test/java/org/apache/ignite/jdbc/JdbcAuthorizationTest.java
ololo3000: Thanks. Done.

modules/clients/src/test/java/org/apache/ignite/jdbc/JdbcAuthorizationTest.java
alex-plekhanov: I think it's better to avoid checking unexpected results in tests even to show current behavior.

ololo3000: I fixed comments mentioned by you. Could you please take a look?

PR #7692, opened 2020-04-19
IGNITE-12916: Add travis-ci configuration for Ignite build

Mmuzaf:


bedlaj:
Travisfile can be tweaked a bit.

- `sudo` key is deprecated and ignored. It is recomended to remove it. See https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration
- `language` is missing, so default `ruby` is used. You should add `language: java`
- `dist` defaults to `xenial`.  Better add `dist: xenial` explicitly. Default dist can lead to unexpected results, once travis-ci upgrades to next release (There are still many failing builds across github after travis updated default image from trusty to xenial)

Mmuzaf:
@bedlaj Thank you for your comments. I've fixed all of them.

PR #7705, opened 2020-04-21
IGNITE-12080 Impl rebalance statistics

tkalkirill:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1641,17 +1723,41 @@ private synchronized void partitionDone(UUID nodeId, int p, boolean own) {
...
-                    partitionsLeft.decrementAndGet();
+            if (rmvd)
+                partitionsLeft.decrementAndGet();
 
             if (parts.isEmpty()) {
                 int remainingRoutines = remaining.size() - 1;
 
-                U.log(log, "Completed " + ((remainingRoutines == 0 ? "(final) " : "") +
+                int fullParts = fullRebalancingParts.get(nodeId).size();
+                int histParts = histRebalancingParts.get(nodeId).size();
```
ascherbakoff, line 1733: Let's get rid of fullRebalancingParts and histRebalancingParts. For example, you can get a number of full partitions for a node using information provided in `fut.rebalancingParts` and `fut.historical` using the filter like:
`rebalancingParts.get(null).stream().filter(p -> !historical.contains(p)).count();`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1641,17 +1723,41 @@ private synchronized void partitionDone(UUID nodeId, int p, boolean own) {
...
+
+                long duration = System.currentTimeMillis() - startTime;
+                long durationSec = Math.max(1, TimeUnit.MILLISECONDS.toSeconds(duration));
+
+                U.log(log, "Completed " + (remainingRoutines == 0 ? "(final) " : "") +
                     "rebalancing [grp=" + grp.cacheOrGroupName() +
                     ", supplier=" + nodeId +
+                    ", partitions=" + (fullParts + histParts) +
+                    ", entries=" + (fullEntries + histEntries) +
+                    ", duration=" + duration + "ms" +
```
ascherbakoff, line 1749: Rebalancing duration can be rather big and not human readable in millis.
Can you implement utility method durationToString to humanReadableByteCount ?
It should output human readable durations, like 1h20m14s. Check for example
`PeriodFormat.getDefault().print(duration.toPeriod())`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePreloaderAdapter.java
```diff
@@ -68,7 +68,7 @@ public GridCachePreloaderAdapter(CacheGroupContext grp) {
 
         log = ctx.logger(getClass());
 
-        finFut = new GridFinishedFuture();
+        finFut = new GridFinishedFuture<>(true);
```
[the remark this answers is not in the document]
tkalkirill: This future returns in **GridCachePreloaderAdapter#rebalanceFuture** and I think it should return **true** because rebalance is automatically completed successfully for it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1641,17 +1723,41 @@ private synchronized void partitionDone(UUID nodeId, int p, boolean own) {
...
+                partitionsLeft.decrementAndGet();
 
             if (parts.isEmpty()) {
                 int remainingRoutines = remaining.size() - 1;
 
-                U.log(log, "Completed " + ((remainingRoutines == 0 ? "(final) " : "") +
+                int fullParts = fullRebalancingParts.get(nodeId).size();
+                int histParts = histRebalancingParts.get(nodeId).size();
+
+                long fullEntries = fullReceivedKeys.get(nodeId).sum();
```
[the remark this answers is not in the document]
tkalkirill: I didn't understand.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1837,6 +1935,87 @@ public boolean compatibleWith(GridDhtPreloaderAssignments newAssignments) {
...
+
+            while (nonNull(fut)) {
+                if (fut.isInitial() || !fut.isDone() || !fut.result())
+                    return false;
+                else
+                    fut = fut.next;
+            }
+
+            return true;
+        }
```
ascherbakoff, line 1979: It still doesn't look correct to me.
Consider a rebalance chain g1 -> g2 ->g3 for topVer=N, g2 is currently rebalancing.
Topology is changed to N + 1, g3 is incompatible and cancelled.
After g2 has been finished rebalancing a chain will not be completed because g3.rebFut.result=false, but it should.
Also, no need to copy reference to _this_.

The correct implementation for me looks something like:

```
private boolean isChainFinished() {
        assert isDone() : this;

        if (isInitial() || !result())
            return false;

        RebalanceFuture fut = next;

        while (fut != null) {
            if (fut.isInitial() || !fut.isDone())
                return false;
            else
                fut = fut.next;
        }

        return true;
}
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1837,6 +1935,87 @@ public boolean compatibleWith(GridDhtPreloaderAssignments newAssignments) {
...
+
+        /**
+         * Callback when rebalance chain ends.
+         */
+        private void onChainFinish() {
```
ascherbakoff, line 1984: better onChainFinished

PR #7713, opened 2020-04-22
IGNITE-12918 Fix Travis config: separate Java jobs

ptupitsyn:


.travis.yml
```diff
@@ -17,37 +17,38 @@ matrix:
...
+        - |
+          export JVM_EXTRA_ARGS='--illegal-access=permit
+          --add-exports=java.base/jdk.internal.misc=ALL-UNNAMED
+          --add-exports=java.base/sun.nio.ch=ALL-UNNAMED
+          --add-exports=java.management/com.sun.jmx.mbeanserver=ALL-UNNAMED
+          --add-exports=jdk.internal.jvmstat/sun.jvmstat.monitor=ALL-UNNAMED
+          --add-exports=java.base/sun.reflect.generics.reflectiveObjects=ALL-UNNAMED
+          -Djdk.tls.client.protocols=TLSv1.2'
+      # ignite-scalar module currently doesn't build under jdk11
+      script: mvn package -Pall-java,licenses,lgpl,checkstyle,examples,!spark-2.4,!spark,!scala -DskipTests -Dmaven.javadoc.skip=true -B -V
```
Mmuzaf, line 46: Probably the `!` needs to be escaped like this `\!`.

.travis.yml
```diff
@@ -17,37 +17,38 @@ matrix:
...
+      cache: {directories: [$HOME/.m2/repository]}
+
+    - language: java
+      os: linux
+      dist: trusty
+      install: skip
+      jdk: openjdk11
       before_script:
         - echo $JVM_EXTRA_ARGS
         - echo $JDK_PROFILES
```
Mmuzaf, line 36: Let's remove it too.

.travis.yml
```diff
@@ -17,37 +17,38 @@ matrix:
...
       before_cache: rm -rfv $HOME/.m2/repository/org/apache/ignite
       cache: {directories: [$HOME/.m2/repository]}
-    
+
     - language: csharp
       mono: none
       dotnet: 3.1.101
       script:
-        - dotnet build modules/platforms/dotnet/Apache.Ignite.DotNetCore.sln
+        - dotnet build modules/platforms/dotnet/Apache.Ignite.DotNetCore.sln
```
Mmuzaf, line 54: Missed end line.

.travis.yml
ptupitsyn: Fixed

.travis.yml
ptupitsyn: Fixed

.travis.yml
ptupitsyn: We can also use `-` instead of `!`, right?

ptupitsyn: It works as it is, see build results

PR #7714, opened 2020-04-22
IGNITE-12930 DistributedProcess fails node if unable to send single message to coordinator

NSAmelchev:
Fix node failing if a coordinator failed


modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -257,11 +258,20 @@ private void sendSingleMessage(Process p) {
...
+
+        if (F.eq(ctx.localNodeId(), crdId))
+            onSingleNodeMessageReceived(singleMsg, crdId);
         else {
             try {
-                ctx.io().sendToGridTopic(p.crdId, GridTopic.TOPIC_DISTRIBUTED_PROCESS, singleMsg, SYSTEM_POOL);
+                ctx.io().sendToGridTopic(crdId, GridTopic.TOPIC_DISTRIBUTED_PROCESS, singleMsg, SYSTEM_POOL);
+            }
+            catch (ClusterTopologyCheckedException e) {
+                // The coordinator was failed. The single message will be sent when a new coordinator initilized.
```
Mmuzaf, line 270: initilized -> initialized

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -257,11 +258,20 @@ private void sendSingleMessage(Process p) {
...
+                if (log.isDebugEnabled()) {
+                    log.debug("Failed to send a single message to coordinator: [crdId=" + crdId +
+                        ", processId=" + p.id +", error=" + e.getMessage() + ']');
+                }
             }
             catch (IgniteCheckedException e) {
                 log.error("Unable to send message to coordinator.", e);
```
Mmuzaf, line 277: Can you please check, do we need logging here? It seems `FailureHandler` will log exactly the same message.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -367,6 +377,9 @@ private Process(UUID id) {
 
     /** Defines distributed processes. */
     public enum DistributedProcessType {
+        /** For test purpose only. */
```
Mmuzaf, line 380: Personally, I don't support such cases. 
- we can use the `DEFAULT` type of the distributed process which can be used on all nodes. It's also nice to mention that a newly created distributed process must be registered on all nodes (e.g. via compute?).
- already existing types can be also reused in tests.

modules/core/src/test/java/org/apache/ignite/internal/util/DistributedProcessCoordinatorLeftTest.java
```diff
@@ -0,0 +1,158 @@
...
+        CountDownLatch finishLatch = new CountDownLatch(NODES_CNT - 1);
+
+        HashMap<String, DistributedProcess<Integer, Integer>> processes = new HashMap<>();
+
+        int processRes = 1;
+
+        for (Ignite grid : G.allGrids()) {
+            DistributedProcess<Integer, Integer> dp = new DistributedProcess<>(((IgniteEx)grid).context(), TEST_PROCESS,
+                req -> GridTestUtils.runAsync(() -> {
+                    startLatch.countDown();
```
Mmuzaf, line 121: Should we release latch only after the `single-message` future completed? It's still possible that a new coordinator will be assigned prior to the single message sent.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
Mmuzaf, line 270: was failed -> has failed?

modules/core/src/test/java/org/apache/ignite/testsuites/IgniteUtilSelfTestSuite.java
```diff
@@ -136,7 +137,10 @@
     // control.sh
     CommandHandlerParsingTest.class,
 
-    GridCountDownCallbackTest.class
+    GridCountDownCallbackTest.class,
+
+    // Distributed process.
```
Mmuzaf, line 142: I don't think we need such comments :-)

modules/core/src/test/java/org/apache/ignite/internal/util/DistributedProcessCoordinatorLeftTest.java
NSAmelchev: The mentioned `startLatch` is to make sure that process started and we can drop coordinator.
A new coordinator can't be reassigned because the event queue is blocked by `msgSendLatch` until single messages sent.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
NSAmelchev: This is debug logging. Also, I reuse it in the test to make sure that was an attempt to sent a single message. (Another way is override distributed process and watching for calling `sendSingleMessage` (should be not private))

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
NSAmelchev: I'm against using existing types. It'll require unnecessary configuration and logic for such processes and will make tests more difficult.

modules/core/src/test/java/org/apache/ignite/internal/util/DistributedProcessCoordinatorLeftTest.java
Mmuzaf: The single message future must be completed prior to the event queue processing released. Is it true for now? Currently, this will happen at the same time and we will have a race with a new coordinator assigned to the `Process`.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
NSAmelchev: What means the `DEFAULT` process? What if we want to add test logic to the action and the finish callbacks? When it will be registered?

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
NSAmelchev: Fixed. Thank you.

NSAmelchev: Fixed. Thank you.

modules/core/src/test/java/org/apache/ignite/testsuites/IgniteUtilSelfTestSuite.java
NSAmelchev: Fixed. Thank you.

modules/core/src/test/java/org/apache/ignite/internal/util/DistributedProcessCoordinatorLeftTest.java
NSAmelchev: Yes, it's true. 'msgSendLatch' guarantees this. The coordinator will be changed at the event thread after releasing the latch.

PR #7720, opened 2020-04-24
IGNITE-12898 Fix incorrect collecting cache configuration (missing CacheConfigurationEnrichment) on coordinator.

ivandasch:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheConfigurationSplitterImpl.java
```diff
@@ -41,11 +44,32 @@
...
+    /** Grid kernal context. */
+    private final GridKernalContext ctx;
+
     /**
      * @param marshaller Marshaller.
+     * @param ctx Grid kernal context.
      */
-    public CacheConfigurationSplitterImpl(Marshaller marshaller) {
+    public CacheConfigurationSplitterImpl(Marshaller marshaller, GridKernalContext ctx) {
         this.marshaller = marshaller;
```
xtern, line 55: let's add assertions for arguments
assert marshaller != null;
assert ctx != null;

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheConfigurationSplitterImpl.java
```diff
@@ -115,4 +139,13 @@ public CacheConfigurationSplitterImpl(Marshaller marshaller) {
             throw new IgniteException("Failed to serialize field [fieldName=" + fieldName + ", value=" + val + ']', e);
         }
     }
+
+    /**
+     * @param cfg Cache configuration.
+     * @return {@code true} If local node is affinity node for cache.
+     */
+    private boolean isAffinity(CacheConfiguration cfg) {
```
xtern, line 147: Let's change to CacheConfiguration<?, ?> to prevent "unchecked" warning.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheConfigurationSplitterImpl.java
ivandasch: I strongly disagree. I am against this unnecessary asserts, this leads to bloated bytecode, nothing more

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheConfigurationSplitterImpl.java
ivandasch: I don't see any reason why I should add wildcard. Wildcards and bounds are useful when we deal with covariance or contravariance. This is not a case. Moreover, there is not any wildcard in this class, why I should add one? No reason at all, except highliting in one well-known IDE

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheConfigurationSplitterImpl.java
xtern: ok, it's just a suggestion.
For me It seems that the code with the assertions is more concise and readable and I don't see any problem with "bloated bytecode".

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheConfigurationSplitterImpl.java
xtern: 1. Because it's java compiler warning
2. Because most people use an IDE that will highlight this code.
3. Because we trying to cleanup Ignite code.

ivandasch: Only reasonable variant is @SuppressWarning. CacheConfiguration is generic just for compliance with jcache api, nothing more. Adding everywhere warning suppression is a concern, but at least areasonable variant. Adding everywhere wildcards is like a cargo cult.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheConfigurationSplitterImpl.java
zstan: +1 with Ivan here, if it`s necessary u may add @Nullable and check them, but not asserts.

PR #7736, opened 2020-04-26
IGNITE-12954: Fix checkstyle rules for ML

Mmuzaf:


zaleslaw:
I think it could be merged, @Mmuzaf

PR #7745, opened 2020-04-27
IGNITE-12886 Introduce separate SQL configuration

tledkov-gridgain:


modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -313,12 +320,6 @@
     /** Index create/rebuild pool size. */
     private int buildIdxPoolSize = DFLT_BUILD_IDX_THREAD_POOL_SIZE;
```
AMashenkov, line 321: Should this be a part of SqlConfiguration as well?

modules/core/src/main/java/org/apache/ignite/configuration/SqlConfiguration.java
```diff
@@ -0,0 +1,154 @@
...
+
+    /** SQL query history size. */
+    private int sqlQryHistSize = DFLT_SQL_QUERY_HISTORY_SIZE;
+
+    /**
+     * Defines the default query timeout.
+     *
+     * Defaults to {@link #DFLT_QRY_TIMEOUT}.
+     * {@code 0} means there is no timeout (this
+     * is a default value)
```
AMashenkov, line 53: Let's remove linefeed here to make more readable.

modules/core/src/main/java/org/apache/ignite/configuration/SqlConfiguration.java
```diff
@@ -0,0 +1,154 @@
...
+     * @return Default query timeout.
+     */
+    public long getDefaultQueryTimeout() {
+        return dfltQryTimeout;
+    }
+
+    /**
+     * Sets timeout in milliseconds for default query timeout.
+     * {@code 0} means there is no timeout (this
+     * is a default value)
```
AMashenkov, line 64: Let's remove linefeed here to make more readable.

modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
tledkov-gridgain: Not agreed. All we configure and start all executor services at the IgnitionEx.IgniteNamedInstance#start0. So it looks looks like a core feature.

modules/core/src/main/java/org/apache/ignite/configuration/SqlConfiguration.java
tledkov-gridgain: Fixed

modules/core/src/main/java/org/apache/ignite/configuration/SqlConfiguration.java
tledkov-gridgain: Fixed

PR #7747, opened 2020-04-27
IGNITE-12925 Fixes flaky tests for continuous query metrics.

ololo3000:


modules/core/src/test/java/org/apache/ignite/internal/metric/SystemViewSelfTest.java
```diff
@@ -526,7 +526,7 @@ public void testContinuousQuery() throws Exception {
             }
 
             assertEquals(0, origQrys.size());
-            assertEquals(0, remoteQrys.size());
+            assertTrue(waitForCondition(() -> remoteQrys.size() == 0, getTestTimeout()));
```
NSAmelchev, line 529: I suggesting fix all similar tests:
```
SystemViewSelfTest#testContinuousQuery
JmxExporterSpiTest#testContinuousQuery
SqlViewExporterSpiTest#testContinuousQuery
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/metric/SqlViewExporterSpiTest.java
```diff
@@ -560,7 +560,8 @@ public void testContinuousQuery() throws Exception {
         }
 
         assertTrue(execute(ignite0, "SELECT * FROM SYS.CONTINUOUS_QUERIES").isEmpty());
```
Mmuzaf, line 562: Should we use the same `waitForCondition` as well?

ololo3000: For local node QueryCursor#close gives guarantees that CQ information will be removed before method returns. See GridContinuousProcessor#stopRoutine::1094

PR #7751, opened 2020-04-28
IGNITE-10100 Implement calling .NET service from java

ivandasch:


modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProxy.java
```diff
@@ -63,6 +69,22 @@
...
+    /** */
+    private static final Method PLATFORM_SERVICE_INVOKE_METHOD;
+
+    static {
+        Method mtd;
+
+        try {
+            mtd = PlatformService.class.getMethod("invokeMethod", String.class, boolean.class, Object[].class);
+        }
+        catch (NoSuchMethodException e) {
```
ptupitsyn, line 81: Let's either remove this `catch`, or throw a different exception explaining the problem. I don't think we should continue execution in this case.

modules/core/src/main/java/org/apache/ignite/platform/PlatformServiceGetter.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation for setting mapping between java interface's method and getter of platform service property (mostly useful
+ * for .NET services). Name mapping constructed by concatenating {@link PlatformServiceGetter#prefix}
```
ptupitsyn, line 28: `is constructed`

modules/core/src/main/java/org/apache/ignite/platform/PlatformServiceMethod.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation for setting mapping between java interface's method and platform service's method. Name mapping
+ * is got from {@link PlatformServiceMethod#value}
```
ptupitsyn, line 28: `Name mapping is got from` -> `Platform service method name is defined in`

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProxy.java
ivandasch: Unfortunatelly, we cannot remove this catch, because `NoSuchMethodException` this is checked one. You suggests to fail node? To be honest, I don't think that failing node in static initializer is a good variant. Let's explain. Let's imagine situatuin, that someone mistakenly delete `invokeMethod` from code, but then he or she get failing tests and easily correct this mistake.

ivandasch: I suggest just to throw `ExceptionInInitializerError.html` [1](https://docs.oracle.com/javase/8/docs/api/java/lang/ExceptionInInitializerError.html), it's unchecked exception and created just for this case. What do you think?

ivandasch: Fixed as proposed.

ptupitsyn: Yes, we should fail the node, because this error indicates a nasty bug in our code.
Unchecked exception looks good to me.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/utils/PlatformUtils.java
```diff
@@ -833,12 +833,26 @@ else if (err instanceof IgniteException)
...
+     */
+    public static Object readInvocationResult(PlatformContext ctx, BinaryRawReaderEx reader, boolean deserialize)
+            throws IgniteCheckedException {
         // 1. Read success flag.
         boolean success = reader.readBoolean();
 
         if (success)
             // 2. Return result as is.
-            return reader.readObjectDetached();
+            return deserialize ? reader.readObject() : reader.readObjectDetached();
```
alex-plekhanov, line 855: I think it's not enough for correct deserialization sometimes (for arrays and collection of objects). We should also do `unwrapBinary()`.
Also (optionally), we can use `unwrapBinary()` directly after method invocation in `GridServiceProxy` instead of adding new methods to `PlatformUtils`, `PlatformService`, `PlatformAbstractService`.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProcessor.java
```diff
@@ -1030,11 +1031,11 @@ private CancelResult removeServiceFromCache(String name) throws IgniteCheckedExc
                 Service svc = ctx.service();
 
                 if (svc != null) {
-                    if (!srvcCls.isAssignableFrom(svc.getClass()))
+                    if (srvcCls.isAssignableFrom(svc.getClass()))
+                        return (T)svc;
+                    else if (!PlatformService.class.isAssignableFrom(svc.getClass()))
```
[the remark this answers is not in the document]
ivandasch: Fixed

modules/core/src/main/java/org/apache/ignite/platform/PlatformServiceGetter.java
```diff
@@ -0,0 +1,54 @@
...
+ * </pre>
+ * will be mapped to {@code get_SomeProperty} method name and corresponds to the following .NET property:
+ * <pre>
+ * SomeProperty &#123; get; &#125;
+ * </pre>
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD})
+public @interface PlatformServiceGetter {
```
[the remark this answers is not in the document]
ivandasch: Agree, fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/service/GridServiceProxy.java
```diff
@@ -354,6 +392,49 @@ T proxy() {
         return proxy;
     }
 
+    /**
+     * @param mtd Method to invoke.
+     */
+    String methodName(Method mtd) {
+        GridServiceMethodReflectKey mtdKey = new GridServiceMethodReflectKey(mtd.getName(), mtd.getParameterTypes());
+
+        return srvMtds.getOrDefault(mtdKey, mtd.getName());
```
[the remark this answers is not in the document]
ivandasch: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/utils/PlatformUtils.java
ivandasch: Unfortunatelly, the approach you suggested doesn't work, but is more elegant, without doubts. But these methods returns Object and need more additional works. But my approach works and tests prove this. So I suggest leave it as is.

modules/core/src/test/java/org/apache/ignite/platform/PlatformServiceCallTask.java
```diff
@@ -0,0 +1,86 @@
...
+
+                throw new RuntimeException("Expected exception, but invocation was success");
+            }
+            catch (IgniteException e) {
+                assertTrue(PlatformNativeException.class.isAssignableFrom(e.getCause().getClass()));
+
+                PlatformNativeException nativeEx = (PlatformNativeException)e.getCause();
+
+                assertTrue(nativeEx.toString().contains("Failed method"));
+            }
```
alex-plekhanov, line 83: GridTestUtils.assertThrowsAnyCause?

PR #7771, opened 2020-05-04
IGNITE-12894 Adds ability to wait for the service topology initialization.

ololo3000:


modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -823,19 +823,26 @@ else if (prj.predicate() == F.<ClusterNode>alwaysTrue())
...
+             desc = lookupInRegisteredServices(name);
 
-            if (timeout == 0 || (top != null && !top.isEmpty()))
-                return top;
+            if (timeout == 0 && desc == null)
+                return null;
 
             synchronized (servicesTopsUpdateMux) {
-                long wait = timeout - (U.currentTimeMillis() - startTime);
+                if (desc != null && desc.topologyInitialized())
```
daradurvs, line 835: This condition can be moved outside the synchronized block.

modules/core/src/test/java/org/apache/ignite/internal/processors/service/GridServiceProxyTopologyInitializationTest.java
```diff
@@ -0,0 +1,220 @@
...
+import org.junit.Test;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static org.apache.ignite.testframework.GridTestUtils.assertThrowsWithCause;
+import static org.apache.ignite.testframework.GridTestUtils.runAsync;
+
+/**
+ * Tests service invocation if requested service is not deployed on the local node and the service topology is required.
+ */
+public class GridServiceProxyTopologyInitializationTest extends GridCommonAbstractTest {
```
daradurvs, line 44: Use `ServiceDeploymentProcessAbstractTest` and `BlockingTcpDiscoverySpi` instead of latches.

daradurvs:
In addition: please add some more tests to show improved guaranties.
For example [from here](https://github.com/apache/ignite/pull/7213/files#diff-4f7cd01f46644072988d63539e6e794eR89) but without "timeout API".

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
ololo3000: I think it cannot be done that way.

If topology is updated before acquiring of servicesTopsUpdateMux but after this condiiton, we will be blocked until the next full msg.

modules/core/src/test/java/org/apache/ignite/internal/processors/service/GridServiceProxyTopologyInitializationTest.java
ololo3000: BlockingTcpDiscoverySpi gives an ability to only make the full msg disappear. In my cases it's needed to block/unblock the full msg and awaits for the completion of its processing.

ololo3000:
testServiceTopologyInitialized and testServiceTopologyInitializationDelayed do exactly what you proposed.

Current test cases:

The service which deployment fails on all nodes invokes on the remote node before/after full message received.

The service which deployment proceeded smoothly on all nodes invokes on the remote node before/after full message received.

Invokes a statically configured service on the local node after it starts.

Are there any missing cases?

modules/core/src/test/java/org/apache/ignite/internal/processors/service/GridServiceProxyTopologyInitializationTest.java
```diff
@@ -0,0 +1,241 @@
...
+    private final CountDownLatch fullMsgUnblockedLatch = new CountDownLatch(1);
+
+    /** Latch that indicated whether {@link ServiceSingleNodeDeploymentResultBatch} was received on the remote node. */
+    private final CountDownLatch fullMsgReceivedLatch = new CountDownLatch(1);
+
+    /** Latch that indicated whether {@link ServiceSingleNodeDeploymentResultBatch} was handled on the remote node. */
+    private final CountDownLatch fullMsgHandledLatch = new CountDownLatch(1);
+
+    /** Number of the test nodes. */
+    private static final int NODES_CNT = 2;
```
daradurvs, line 68: Constant should be over instance fields.

modules/core/src/test/java/org/apache/ignite/internal/processors/service/GridServiceProxyTopologyInitializationTest.java
```diff
@@ -0,0 +1,241 @@
...
+     *
+     * @throws Exception If fails.
+     */
+    @Test
+    @SuppressWarnings("Convert2MethodRef")
+    public void testServiceTopologyInitializationDelayed() throws Exception {
+        IgniteEx loc = startGrids(NODES_CNT);
+
+        IgniteEx rmt = grid(NODES_CNT - 1);
+
```
daradurvs, line 151: `assertEquals(1, fullMsgReceivedLatch.getCount());`

modules/core/src/test/java/org/apache/ignite/internal/processors/service/GridServiceProxyTopologyInitializationTest.java
```diff
@@ -0,0 +1,241 @@
...
+        deployServices(loc);
+
+        fullMsgReceivedLatch.await(getTestTimeout(), MILLISECONDS);
```
daradurvs, line 154: `assertTrue(fullMsgReceivedLatch.await(getTestTimeout(), MILLISECONDS));`

modules/core/src/test/java/org/apache/ignite/internal/processors/service/GridServiceProxyTopologyInitializationTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+        IgniteInternalFuture<Boolean> decentSvcFut = runAsync(() ->
+            rmt.services().serviceProxy(DECENT_SRVC, Invoker.class, false).invoke());
+
+        IgniteInternalFuture<Boolean> brokenSvcFut = runAsync(() ->
+            rmt.services().serviceProxy(BROKEN_SRVC, Invoker.class, false).invoke());
+
+        U.sleep(500);
+
```
daradurvs, line 163: `assertEquals(1, fullMsgUnblockedLatch.getCount());`

modules/core/src/test/java/org/apache/ignite/internal/processors/service/GridServiceProxyTopologyInitializationTest.java
```diff
@@ -0,0 +1,241 @@
...
+     * Tests service invocation after its topology is initialized on the local node.
+     *
+     * @throws Exception If fails.
+     */
+    @Test
+    public void testServiceTopologyInitialized() throws Exception {
+        IgniteEx loc = startGrids(NODES_CNT);
+
+        IgniteEx rmt = grid(NODES_CNT - 1);
+
```
daradurvs, line 126: `assertEquals(1, fullMsgUnblockedLatch.getCount());`

modules/core/src/test/java/org/apache/ignite/internal/processors/service/GridServiceProxyTopologyInitializationTest.java
```diff
@@ -0,0 +1,241 @@
...
+            cfg.setUserAttributes(Collections.singletonMap(ATTR_SKIP_DEPLOYMENT, true));
+        }
+
+        return cfg;
+    }
+
+    /**
+     * Ignores the test in case the legacy service processor is used.
+     */
+    @Before
```
daradurvs, line 104: `@BeforeClass` because all these tests are valid only for new service processor implementaion.

modules/core/src/test/java/org/apache/ignite/internal/processors/service/GridServiceProxyTopologyInitializationTest.java
```diff
@@ -0,0 +1,241 @@
...
+    public void testServiceTopologyInitialized() throws Exception {
+        IgniteEx loc = startGrids(NODES_CNT);
+
+        IgniteEx rmt = grid(NODES_CNT - 1);
+
+        fullMsgUnblockedLatch.countDown();
+
+        deployServices(loc);
+
+        fullMsgHandledLatch.await(getTestTimeout(), MILLISECONDS);
```
daradurvs, line 131: `assertTrue(fullMsgHandledLatch.await(getTestTimeout(), MILLISECONDS));`

modules/core/src/test/java/org/apache/ignite/internal/processors/service/GridServiceProxyTopologyInitializationTest.java
```diff
@@ -0,0 +1,241 @@
...
+     * @throws Exception If fails.
+     */
+    @Test
+    public void testServiceTopologyInitialized() throws Exception {
+        IgniteEx loc = startGrids(NODES_CNT);
+
+        IgniteEx rmt = grid(NODES_CNT - 1);
+
+        fullMsgUnblockedLatch.countDown();
+
```
daradurvs, line 128: `assertEquals(1, fullMsgHandledLatch.getCount());`

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -823,19 +823,26 @@ else if (prj.predicate() == F.<ClusterNode>alwaysTrue())
 
         long startTime = U.currentTimeMillis();
 
-        Map<UUID, Integer> top;
+        ServiceInfo desc;
 
         while (true) {
-            top = serviceTopology(name);
+             desc = lookupInRegisteredServices(name);
```
daradurvs, line 829: It looks like this block should be moved outside the loop
```
desc = lookupInRegisteredServices(name);

if (timeout == 0 && desc == null)
    return null;
```
and variable `desc` shoud be refreshed under mutex
```
synchronized (servicesTopsUpdateMux) {
    desc = lookupInRegisteredServices(name);
    ...
```
Because there is might be a race, when during the checks service can be unregistered and `ServiceClusterDeploymentResultBatch` processed.
That means it's possible hanging because in your case `wait` might be equal to `0` - that means "wait forever".

Refreshing `desc` under mutex - not to provide us happens-before guarantees but provide processing order guarantees: when undeployment may be processed before we start waiting for topology initialization.

Alternative solution - do not wait for `0` and use `100` for example to retry the cycle.
But it's would be better to provide processing order guarantees.

What do you think?

ololo3000: I got it. If after [1] but before [2] service will be deleted and full message with requested service topology will be received, topology of the local ServiceDescriptor reference will never be updated. Is that what you mean? 

```
while (true) {
    desc = lookupInRegisteredServices(name);

    if (timeout == 0 && desc == null) // [1]
        return null;

    synchronized (servicesTopsUpdateMux) { // [2]
        if (desc != null && desc.topologyInitialized())
            return desc.topologySnapshot();

        long wait = 0;

        if (timeout != 0) {
            wait = timeout - (U.currentTimeMillis() - startTime);

            if (wait <= 0)
                return desc == null ? null : desc.topologySnapshot();
        }

        try {
            servicesTopsUpdateMux.wait(wait);
        }
        catch (InterruptedException e) {
            throw new IgniteInterruptedCheckedException(e);
        }
    }
}
```

It seems that it will be enough to only move 

```
desc = lookupInRegisteredServices(name);

if (timeout == 0 && desc == null) // [1]
    return null;
```

inside the syncgronization block. 
In that case if service is removed before synchronization block or full message is received no matter the order, we will retrurn on 

```
 if (timeout == 0 && desc == null) 
    return null;
```

if service is removed during synchronization block execution, we anyway will receive the full message and stop waiting.

ololo3000: I updated PR according to my proposal. Could you please take a look?
Unfortunately I find it difficult to test this race condition.

daradurvs: Yes, you understood correctly.
Changes look good to me.
Please rerun tests and update TC bot visa in Jira, just to be sure.

PR #7772, opened 2020-05-06
IGNITE-12983 Logging exceptions inside IgniteSecurityProcessor#withContext(java.util.UUID)

dgarus:


modules/core/src/main/java/org/apache/ignite/internal/processors/security/IgniteSecurityProcessor.java
```diff
@@ -125,17 +130,24 @@ public IgniteSecurityProcessor(GridKernalContext ctx, GridSecurityProcessor secP
...
+
+            SecurityContext res = node != null ? secCtxs.computeIfAbsent(subjId,
+                uuid -> nodeSecurityContext(marsh, U.resolveClassLoader(ctx.config()), node))
+                : secPrc.securityContext(subjId);
 
-        SecurityContext res = node != null ? secCtxs.computeIfAbsent(subjId,
-            uuid -> nodeSecurityContext(marsh, U.resolveClassLoader(ctx.config()), node))
-            : secPrc.securityContext(subjId);
+            if (res == null)
+                throw new IllegalStateException("Failed to find security context for subject with given ID : " + subjId);
```
nizhikov, line 142: What's the point throw an Exception and catch it two lines later?

dgarus: @nizhikov 
After logging inside the catch block, this exception will be thrown outside where it can be handled.

nizhikov: We can log it before throw.

dgarus: @nizhikov fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/security/IgniteSecurityProcessor.java
```diff
@@ -125,17 +133,30 @@ public IgniteSecurityProcessor(GridKernalContext ctx, GridSecurityProcessor secP
...
+                uuid -> nodeSecurityContext(marsh, U.resolveClassLoader(ctx.config()), node))
+                : secPrc.securityContext(subjId);
 
-        SecurityContext res = node != null ? secCtxs.computeIfAbsent(subjId,
-            uuid -> nodeSecurityContext(marsh, U.resolveClassLoader(ctx.config()), node))
-            : secPrc.securityContext(subjId);
+            if (res != null)
+                return withContext(res);
+
+            error = new IllegalStateException("Failed to find security context for subject with given ID : " + subjId);
```
nizhikov, line 149:
```
if (res != null)
    return withContext(res);
}
catch (Throwable e) {
    log.error(MSG, e);
    throw e;
}

String msg = "Failed to find security context for subject with given ID : " + subjId;

log.error(msg);
throw new IllegalStateException(msg);

PR #7779, opened 2020-05-07
IGNITE-12969 TxRecovery discovery listener should implement HighPriorityListener	

anton-vinogradov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/NearTxResultHandler.java
```diff
@@ -108,6 +109,9 @@ public static GridNearTxEnlistResponse createResponse(GridDhtTxEnlistFuture fut)
         try {
             cctx.io().send(nearNodeId, res, cctx.ioPolicy());
         }
+        catch (ClusterTopologyCheckedException e) {
```
nizhikov, line 112: Why this change?

anton-vinogradov: Code speed-up increase probability of this case. 
It was a problem before, but very rarely.

PR #7784, opened 2020-05-08
IGNITE-12993: Initial ASF infra integration configuration.

xtern:


.asf.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - iot
+    - osgi
+    - network-client
+    - ignite
+    - data-management-platform
+    - big-data
+    - cloud
+    - database
+    - network-server
+    - hadoop
```
Mmuzaf, line 30: Do we remove Hadoop support from Ignite?

xtern: These are the current labels of Apache Ignite repository on github

xtern: I'm not sure we can remove this label without discussion.

IGFS and Ignite Hadoop Accelerator were discontinued, but Ignite can still be used to speed up Hadoop, as described on this page: https://ignite.apache.org/use-cases/hadoop-acceleration.html

Mmuzaf:
LGTM,

@dmagda @pavlukhin 

Folks, can you take a look too?

.asf.yaml
dmagda: Maxim, I would also add the following labels:

- cache
- in-memory computing
- in-memory database

PR #7792, opened 2020-05-12
IGNITE-12994 Move binary metadata to PDS storage folder

SammyVimes:
Binary metadata and marshaller directories are now located inside /db directory.

If on a startup any of these folders are found in legacy location, they're moved safely to new location.
Motivation: For example, on K8s deployment disk that is attached to container can not be accessible from other containers or outside of K8s. In case if support will need to drop persistence except data, there will be no way to recover due to binary metadata is required to process PDS files.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/BinaryMetadataFileStore.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/BinaryMetadataFileStore.java
@@ -18,6 +18,8 @@ package org.apache.ignite.internal.processors.cache.binary;
 
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.IOException;
+import java.nio.file.Files;
 import java.util.Map;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
@@ -27,6 +29,7 @@ import org.apache.ignite.IgniteCheckedException;
 import org.apache.ignite.IgniteException;
 import org.apache.ignite.IgniteLogger;
 import org.apache.ignite.binary.BinaryType;
+import org.apache.ignite.configuration.DataStorageConfiguration;
 import org.apache.ignite.failure.FailureContext;
 import org.apache.ignite.failure.FailureType;
 import org.apache.ignite.internal.GridKernalContext;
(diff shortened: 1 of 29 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/binary/BinaryMetadataMoveLegacyFolderTest.java
```diff
@@ -0,0 +1,225 @@
...
+
+        return marshalled;
+    }
+
+    /**
+     * Test that binary_meta directory, that was previously located in workdir's root, is successfully moving to PDS
+     * folder.
+     */
+    @Test
+    public void testBinaryMetadataDirectoryMigration() throws Exception {
```
Mmuzaf, line 123: You should use the `ignite-compatibility` framework to test that a new Ignite version able to start on the old persistence directory. 
See: https://issues.apache.org/jira/browse/IGNITE-5732

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/BinaryMetadataFileStore.java
```diff
@@ -255,6 +272,45 @@ void finishWrite(int typeId, int typeVer) {
         writer.finishWriteFuture(typeId, typeVer);
     }
 
+    /**
+     * Try looking for legacy directory with binary metadata and move it to new directory
+     */
+    private void fixLegacyFolder(String consistendId) throws IgniteCheckedException {
```
Mmuzaf, line 278: I don't think we should keep this in production source code. This method used only once for the migration procedure. I think the right behaviour should be:
1. fail the node when users try to start a new Ignite version from the previous persistence directory.
2. provide migration scripts.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -419,6 +421,12 @@ public static void deleteSnapshot(File snpDir, String folderName) {
                 }
             });
 
+            File binMetadataDfltDir = new File(snpDir, DFLT_BINARY_METADATA_PATH);
+            File marshallerDfltDir = new File(snpDir, DFLT_MARSHALLER_PATH);
+
+            U.delete(binMetadataDfltDir);
+            U.delete(marshallerDfltDir);
```
Mmuzaf, line 428: It's not correct in doing such a way since the marshaller directory is shared between all nodes (if nodes are started on the local machine). Thus you will give the `NoSuchFileException` due to `U.delete` doesn't work with directories in concurrent-way.

SammyVimes: Do you mean it's necessary to check if directory exists prior to deleting it? It's just that I looked inside `U.delete` and it handles absence of files/directories by catching all `IOExceptions`.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/binary/BinaryMetadataMoveLegacyFolderTest.java
SammyVimes: Added compatibility test in newest commit (testing against version 2.8.0)

PR #7844, opened 2020-05-25
IGNITE-12823 .NET: Fix service method calls with typed array args

ptupitsyn:
**Bug:** Service calls don't work when one of the parameters is an array other than `Object[]` - both .NET -> Java and .NET -> .NET.

**Reason:** array element type is lost when passing service arguments on Java side: `GridServiceProxy.invokeMethod()` takes `Object[] args`.

**Fix:** convert `Object[]` to `T[]` right before service invocation by looking at the service method signature to determine the type.


modules/core/src/main/java/org/apache/ignite/internal/processors/platform/services/PlatformServices.java
```diff
@@ -581,6 +582,26 @@ public Object invoke(String mthdName, boolean srvKeepBinary, Object[] args)
 
                 Method mtd = getMethod(serviceClass, mthdName, args);
 
+                // Convert Object[] to T[] when required:
+                // Ignite loses array item types when passing arguments through GridServiceProxy.
+                for (int i = 0; i < args.length; i++) {
```
kukushal, line 587: IgniteServices has a "keep binary" flag that makes the server side NOT to deserialize parameters, thus, allowing NOT having Java classes on the server. You need to do this block only if the "keep binary" flag is not set. Otherwise it will do that Array.newInstance thing even when "keep binary = true".

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Services/ServicesTest.cs
```diff
@@ -859,18 +903,28 @@ public void TestCallJavaService()
...
                 .OfType<PlatformComputeBinarizable>().Select(x => x.Field).ToArray());
-            Assert.AreEqual(new[] {11, 12, 13},
-                svc.testBinarizableArray(arr).OfType<PlatformComputeBinarizable>().Select(x => x.Field).ToArray());
+
+            Assert.AreEqual(new[] {11, 12, 13}, svc.testBinarizableArrayOfObjects(arrOfObj)
+                .OfType<PlatformComputeBinarizable>().Select(x => x.Field).ToArray());
+
+            Assert.IsNull(svc.testBinarizableArrayOfObjects(null));
+
+            Assert.AreEqual(new[] {11, 12, 13}, svc.testBinarizableArray(arr)
```
kukushal, line 917: Please do the same in "keep binary" mode.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Services/ServicesTest.cs
```diff
@@ -413,6 +413,50 @@ public void TestDuckTyping([Values(true, false)] bool local)
...
+            var cluster = Grid1.GetCluster().ForNodeIds(nodeId);
+
+            cluster.GetServices().DeployNodeSingleton(SvcName, new TestIgniteServiceArraySerializable());
+
+            var typedArray = new[] {10, 11, 12}
+                .Select(x => new PlatformComputeBinarizable {Field = x}).ToArray();
+
+            var objArray = typedArray.ToArray<object>();
+
+            var prx = Services.GetServiceProxy<ITestIgniteServiceArray>(SvcName);
```
kukushal, line 434: Can we repeat these checks in "keep binary" mode as well? (with 
var binPrx = Services.WithKeepBinary().GetServiceProxy<ITestIgniteServiceArray>(SvcName);

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Services/ServicesTest.cs
ptupitsyn: Good point, done

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Services/ServicesTest.cs
ptupitsyn: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/services/PlatformServices.java
ptupitsyn: `Array.newInstance` is called only when array type does not match. `keepBinary` mode requires the same treatment as normal mode - service may take `BinaryObject[]`, so array conversion is necessary. I've added tests for both .NET->.NET and .NET->Java calls to demonstrate this.

PR #7872, opened 2020-05-28
IGNITE-13015 : Use nano time in node failure detection.

Vladsz83:
Time of last received message turned into nanotime:
`ServerImpl.lastRingMsgReceivedTime`

This is made to give stability to failure detection. System time is not monotonic and can be adjusted, synchronized.


Vladsz83:
Time of last received message turned into nanotime:
{code:java}ServerImpl.lastRingMsgReceivedTime{code}

This is made to give stability to failure detection. System time is not monotonic and can be adjusted, synchronized.

PR #7897, opened 2020-06-03
IGNITE-13112 CacheEvent#subjectId is always null for cache events with types EVT_CACHE_STARTED and EVT_CACHE_STOPPED

dgarus:


modules/core/src/main/java/org/apache/ignite/internal/processors/security/SecurityUtils.java
```diff
@@ -153,6 +154,74 @@ public static SecurityContext nodeSecurityContext(Marshaller marsh, ClassLoader
...
+                }
+            }
+            else
+                r.runx();
+        }
+        catch (Exception e) {
+            if (e instanceof IgniteCheckedException)
+                throw (IgniteCheckedException)e;
+
+            throw new IgniteCheckedException(e);
```
nizhikov, line 182: Why do we need to change the exception here?
Can't we just throw it?

modules/core/src/main/java/org/apache/ignite/internal/processors/security/SecurityUtils.java
```diff
@@ -153,6 +154,74 @@ public static SecurityContext nodeSecurityContext(Marshaller marsh, ClassLoader
...
+            if (security.enabled() && secSubjId != null) {
+                try (OperationSecurityContext s = security.withContext(secSubjId)) {
+                    return c.call();
+                }
+            }
+
+            return c.call();
+        }
+        catch (Exception e) {
+            if (e instanceof IgniteCheckedException)
```
nizhikov, line 209: Why do we need to change the exception here?
Can't we just throw it?

modules/core/src/test/java/org/apache/ignite/internal/processors/security/events/AbstractSecurityCacheEventTest.java
```diff
@@ -0,0 +1,203 @@
...
+    protected static final AtomicInteger COUNTER = new AtomicInteger();
+
+    /** Node that registers event listeners. */
+    private static final String LISTENER_NODE = "listener_node";
+
+    /** Events latch. */
+    private static volatile CountDownLatch evtsLatch;
+
+    /** Logins in remote filters. */
+    private static final Collection<String> rmtLogins = new ConcurrentLinkedQueue<>();
```
nizhikov, line 63: Why do we need collection for `rmtLogins` and `locLogins`?
We expect and assume that it will be single login in this sets - 
```
        if (set.size() != 1) {
            fail(msgPrefix + " Expected subject: " + expLogin +
                ". Actual subjects: " + Iterators.toString(set.iterator()));
        }
```

modules/core/src/test/java/org/apache/ignite/internal/processors/security/events/AbstractSecurityCacheEventTest.java
```diff
@@ -0,0 +1,203 @@
...
+                }, evtType);
+
+        try {
+            // Execute tested operation.
+            op.accept(ccfgs);
+
+            // Waiting for events.
+            evtsLatch.await(10, TimeUnit.SECONDS);
+
+            // Check results.
```
nizhikov, line 110: This comment can be omitted.

modules/core/src/test/java/org/apache/ignite/internal/processors/security/events/AbstractSecurityCacheEventTest.java
dgarus: To get as much information as possible about the failure. A failure message will contain info about all logins inside predicates.

modules/core/src/test/java/org/apache/ignite/internal/processors/security/events/AbstractSecurityCacheEventTest.java
dgarus: ok

modules/core/src/main/java/org/apache/ignite/internal/processors/security/SecurityUtils.java
dgarus: No, we cannot throw Exception, the places when this method is using expect IgniteCheckedException.

modules/core/src/main/java/org/apache/ignite/internal/processors/security/SecurityUtils.java
dgarus: I'll use the run method that doesn't throw an exception.

modules/core/src/test/java/org/apache/ignite/internal/processors/security/events/AbstractSecurityCacheEventTest.java
nizhikov: We can just throw exception if we found second login.
Collection fields not necessary.
Please, remove it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/ChangeGlobalStateMessage.java
```diff
@@ -127,6 +144,8 @@ public ChangeGlobalStateMessage(
     void exchangeActions(ExchangeActions exchangeActions) {
         assert exchangeActions != null && !exchangeActions.empty() : exchangeActions;
 
+        exchangeActions.securitySubjectId(securitySubjectId());
```
nizhikov, line 147: Why do we want to do this action inside setter?
Let's do it outside of it.

nizhikov:
Do we have a test to cover the scenario when cache created and started from static configuration `IgniteConfiguration#setCacheConfiguration`?

PR #7902, opened 2020-06-05
IGNITE-13051 check mvcc configuration after register caches futu…

timoninmaxim:
Validate MVCC configuration after the future completion (wait while all caches are finally registered).

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
@@ -2624,9 +2624,7 @@ public class GridDhtPartitionsExchangeFuture extends GridDhtTopologyFutureAdapte
      */
     private void waitUntilNewCachesAreRegistered() {
         try {
-            IgniteInternalFuture<?> registerCachesFut = registerCachesFuture;
-
-            if (registerCachesFut != null && !registerCachesFut.isDone()) {
+            if (registerCachesFuture != null && !registerCachesFuture.isDone()) {
                 final int timeout = Math.max(1000,
                     (int)(cctx.kernalContext().config().getFailureDetectionTimeout() / 2));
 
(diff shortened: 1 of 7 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/mvcc/CacheMvccConfigurationValidationTest.java
```diff
@@ -352,6 +353,37 @@ public void testMvccEnabledForClientMode() throws Exception {
         assertTrue(node.context().coordinators().mvccEnabled());
     }
 
+    /**
+     * Check that node in client mode (filtered by AttributeNodeFilter) correctly works with MVCC.
+     */
+    @SuppressWarnings("unchecked")
+    @Test
+    public void testMvccEnabledForClientModeAfterTriggeringGlobalState() throws Exception {
```
anton-vinogradov, line 361: looks like a 90% duplication of testMvccEnabledForClientMode

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -346,7 +346,7 @@
     private final GridDhtPartitionsStateValidator validator;
 
     /** Register caches future. Initialized on exchange init. Must be waited on exchange end. */
-    private IgniteInternalFuture<?> registerCachesFuture;
+    private IgniteInternalFuture<?> registerCachesFut;
```
anton-vinogradov, line 349: Such renaming just increases changes amount without real profit

timoninmaxim: this is about abbr fix: Future -> Fut to decrease amount of warnings. I rely on this object, so picked the moment to fix this. Is there a process how to provide such fixes right way?

anton-vinogradov: Such fixes snatch code ownership (git annotations), so they should be performed only in case you really changing something.

timoninmaxim: Got it. Fixed with re-push to keep the ownership

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/mvcc/CacheMvccConfigurationValidationTest.java
timoninmaxim: Move to another class to move some code duplication, and add more tests

PR #7905, opened 2020-06-06
IGNITE-7105 .NET: Add IgniteLock

ptupitsyn:
* Add `IIgnite.GetOrCreateLock` (no `reentrant` because .NET locks are reentrant by default, this is expected)
* `IIgniteLock` methods are named according to standard `Monitor` class and existing Ignite `ICacheLock`
* Conditions and `*interruptibly` things are out of scope

### DIFF
--- a/modules/platforms/dotnet/Apache.Ignite.Core/IIgniteLock.cs
+++ b/modules/platforms/dotnet/Apache.Ignite.Core/IIgniteLock.cs
@@ -18,9 +18,12 @@
 namespace Apache.Ignite.Core
 {
     using System;
+    using System.Threading;
 
     /// <summary>
     /// Distributed re-entrant lock.
+    /// <para />
+    /// The functionality is similar to the standard <see cref="Monitor"/> class, but works across all cluster nodes.
     /// </summary>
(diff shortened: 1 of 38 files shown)


modules/platforms/dotnet/Apache.Ignite.Core/IIgniteLock.cs
```diff
@@ -0,0 +1,81 @@
...
+    using Apache.Ignite.Core.Configuration;
+
+    /// <summary>
+    /// Distributed reentrant monitor (lock).
+    /// <para />
+    /// The functionality is similar to the standard <see cref="Monitor"/> class, but works across all cluster nodes.
+    /// <para />
+    /// This API corresponds to <c>IgniteLock</c> in Java.
+    /// </summary>
+    public interface IIgniteLock
```
isapego, line 31: Method names changed to match .NET practices?

ptupitsyn: Yes, method names are similar to standard `Monitor` class, and also follow existing `ICacheLock` interface.

PR #7906, opened 2020-06-07
IGNITE-6499 : Compact NULL fields binary representation

hostettler:
To compress binary objects, this patch does the following:
In the case of standard objects (Not implementing Binarylizable or Externalizble) (though the nested objects can):
- It removes the 0x65 that encode null within the object
- It removes the offset from the foot to null objects.
- It adds a null mask at the end of the footer in which each bit is set for non null objects.


modules/core/src/test/java/org/apache/ignite/internal/binary/BinaryMarshallerSelfTest.java
```diff
@@ -3851,9 +3851,16 @@ protected boolean compactFooter() {
...
-     * @param marsh Marshaller.
-     * @return Binary context.
+     * @return  Whether to use null compaction or not.
      */
+    protected boolean compactNull() {
+        return false;
+    }
+
+        /**
+         * @param marsh Marshaller.
```
zstan, line 3861: aligning

hostettler: Done

alamar:
What happens when C++ thin client, C# thin client, Python thin client, etc, encounter such BinaryObject's? Do you have tests? What is your plan here? Will they be unaffected?

hostettler:
I must admit that I do not know. I would need the help of someone more knowledgeable about these clients

alamar:
I think this means the behavior should be off by default, to be enabled explicitly.

alamar:
Oh, I can see that you already have the configuration setting implemented. Sorry.

PR #7908, opened 2020-06-08
IGNITE-13033 Java thin client: Service invocation

alex-plekhanov:

The following pattern must be used: `IGNITE-12407: Add Cluster API support to Java thin client`

### Notes

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientServicesImpl.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientServicesImpl.java
@@ -150,8 +150,10 @@ class ClientServicesImpl implements ClientServices {
                 else {
                     writer.writeInt(nodeIds.size());
 
-                    for (UUID nodeId : nodeIds)
-                        writer.writeUuid(nodeId);
+                    for (UUID nodeId : nodeIds) {
+                        writer.writeLong(nodeId.getMostSignificantBits());
+                        writer.writeLong(nodeId.getLeastSignificantBits());
+                    }
                 }
 
                 PlatformServiceMethod ann = method.getDeclaredAnnotation(PlatformServiceMethod.class);
--- a/modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientServicesImpl.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientServicesImpl.java
@@ -1,0 +1,172 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.client.thin;
+
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+import java.util.Collection;
+import java.util.UUID;
+import org.apache.ignite.client.ClientClusterGroup;
+import org.apache.ignite.client.ClientException;
+import org.apache.ignite.client.ClientServices;
+import org.apache.ignite.internal.binary.BinaryRawWriterEx;
+import org.apache.ignite.internal.util.typedef.F;
+import org.apache.ignite.internal.util.typedef.internal.A;
+import org.apache.ignite.platform.PlatformServiceMethod;
+
+/**
+ * Implementation of {@link ClientServices}.
+ */
+class ClientServicesImpl implements ClientServices {
+    /** Channel. */
+    private final ReliableChannel ch;
+
+    /** Binary marshaller. */
+    private final ClientBinaryMarshaller marsh;
+
+    /** Utils for serialization/deserialization. */
+    private final ClientUtils utils;
+
+    /** Cluster group. */
+    private final ClientClusterGroupImpl grp;
+
+    /** Constructor. */
+    ClientServicesImpl(ReliableChannel ch, ClientBinaryMarshaller marsh, ClientClusterGroupImpl grp) {
+        this.ch = ch;
+        this.marsh = marsh;
+        this.grp = grp;
+
+        utils = new ClientUtils(marsh);
+    }
+
+    /** {@inheritDoc} */
+    @Override public ClientClusterGroup clusterGroup() {
+        return grp;
+    }
+
+    /** {@inheritDoc} */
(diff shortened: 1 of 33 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/service/ClientServiceInvokeRequest.java
```diff
@@ -0,0 +1,304 @@
...
+
+        args = new Object[argCnt];
+
+        // We can't deserialize some types (arrays of user defined types for example) from detached objects.
+        // On the other hand, deserialize should be done as part of process() call (not in constructor) for proper
+        // error handling.
+        // To overcome these issues we store binary reader reference, parse request in constructor (by reading detached
+        // objects), restore arguments starting position in input stream and deserialize arguments from input stream
+        // in process() method.
+        this.reader = reader;
```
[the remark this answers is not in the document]
alex-plekhanov: For example, if you marshal `Collections.singleton(new Person[] { person })` it will be unmarshalled as 'Collections.singleton(new Object[] { binaryObject })` by `readObjectDetached` and nothing we can do here to restore array parameter type.
It's not the only problem with detached objects. One more: you can't restore reference after `readObjectDetached`. For example, if you have `new Person[] { person, person }` (one object of "Person" class and two references to it), after unmarshalling and unwrapping binaries you will have two different objects of "Person" class.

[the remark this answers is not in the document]
ptupitsyn: > `Collections.singleton`

This returns `Set`, which does not specific element type at runtime, so what's the problem?

> you can't restore reference after `readObjectDetached`

Well, this is a separate matter. For some reason we use `readObjectDetached` for two things at the same time: 
* To read as binary
* To read as detached (don't preserve references)

We can keep references if we want to.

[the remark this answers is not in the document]
alex-plekhanov: > so what's the problem?
It's not a problem to invoke the service, but can be a problem if some user-defined service checks element types. For example, a user can write something like `for (Object item : collection) if (item instance of Person[]) ...` and such a check will fail.
> We can keep references if we want to.
We can't. If we read two objects as detached there is no information inside these objects that it was the same object before marshaling.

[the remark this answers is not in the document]
ptupitsyn: > If we read two objects as detached
Don't read as detached?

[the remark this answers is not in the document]
ptupitsyn: To summarize - I don't strongly object to this. The code is a bit dirty and fragile, but the problem is localized within a single class. So I leave this up to you.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientServicesImpl.java
```diff
@@ -0,0 +1,178 @@
...
+                writer.writeByte(FLAG_PARAMETER_TYPES_MASK); // Flags.
+                writer.writeLong(timeout);
+
+                if (nodeIds == null)
+                    writer.writeInt(0);
+                else {
+                    writer.writeInt(nodeIds.size());
+
+                    for (UUID nodeId : nodeIds)
+                        writer.writeUuid(nodeId);
```
ptupitsyn, line 154: writeUuid supports null values - it writes an additional byte which is not needed here. See `ClientClusterGroupGetNodeIdsResponse`, for example: we usually write UUIDs as 8 bytes when null does not make sense.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/service/ClientServiceInvokeRequest.java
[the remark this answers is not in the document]
alex-plekhanov: > Don't read as detached?

We have two options: `readObject` and` readObjectDetached`. In case of `readObject` we will deserialize in the constructor, it's undesirable.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientServicesImpl.java
alex-plekhanov: On the other hand, we have UUIDs with an extra byte in compute, partition awareness, and even handshake requests/responses. I think it should be somehow unified. At least with compute until it released (both should write `nodeIds` in the same format). WDYT? 
Should we change compute as well? Or should we leave services as is?

ptupitsyn: I would prefer to avoid redundant data. Not because 1 byte matters for performance (in this case it does not), but because it makes things confusing for the client-side code: If there is a leading byte with uuid, it means that the value is potentially null. How should the client handle nulls here? Does not make sense.

Unfortunately, it is very tempting to use `writeUuid`, so these things keep showing up. Let's not change existing code, dealing with compat is hard.

alex-plekhanov: I've fixed it.
But it seems inconsistent now to use different formats for writing the same entity for compute and services. 
Let's fix compute the same way (by another ticket of course)? It's not released yet, so there will be no compatibility issues.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/service/ClientServiceInvokeRequest.java
```diff
@@ -0,0 +1,324 @@
...
+
+        name = reader.readString();
+
+        flags = reader.readByte();
+
+        timeout = reader.readLong();
+
+        int cnt = reader.readInt();
+
+        nodeIds = U.newHashSet(cnt);
```
ptupitsyn, line 100: Let's change `nodeIds` to `Collection` - looks like there is no need for a hash set.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/service/ClientServiceInvokeRequest.java
```diff
@@ -0,0 +1,324 @@
...
+    private Method resolveMethod(ClientConnectionContext ctx, Class<?> cls) throws ReflectiveOperationException {
+        if (paramTypeIds != null) {
+            MethodDescriptor desc = new MethodDescriptor(cls, methodName, paramTypeIds);
+
+            Method method = methodsCache.get(desc);
+
+            if (method != null)
+                return method;
+
+            BinaryContext bctx = ((CacheObjectBinaryProcessorImpl)ctx.kernalContext().cacheObjects()).binaryContext();
```
ptupitsyn, line 231: Let's use public API and avoid type casts to internal classes when possible. This context is only needed for `ctx.typeId` call, which is available in `IgniteBinary`.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/service/ClientServiceInvokeRequest.java
alex-plekhanov: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/service/ClientServiceInvokeRequest.java
alex-plekhanov: Fixed

PR #7941, opened 2020-06-17
IGNITE-12843

xtern:

### Notes


modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupReencryption.java
```diff
@@ -0,0 +1,583 @@
...
+
+import static org.apache.ignite.IgniteSystemProperties.IGNITE_REENCRYPTION_BATCH_SIZE;
+import static org.apache.ignite.IgniteSystemProperties.IGNITE_REENCRYPTION_DISABLED;
+import static org.apache.ignite.IgniteSystemProperties.IGNITE_REENCRYPTION_THREAD_POOL_SIZE;
+import static org.apache.ignite.IgniteSystemProperties.IGNITE_REENCRYPTION_THROTTLE;
+
+/**
+ * Cache group reencryption manager.
+ */
+public class CacheGroupReencryption implements DbCheckpointListener {
```
ingvard: CacheGroupReencryption may be add name like Listener or Processor?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupReencryption.java
```diff
@@ -0,0 +1,583 @@
...
+    /** Executor to start partition scan tasks. */
+    private final IgniteThreadPoolExecutor execSvc;
+
+    /** Stop flag. */
+    private boolean stopped;
+
+    /**
+     * @param ctx Grid kernal context.
+     */
+    public CacheGroupReencryption(GridKernalContext ctx) {
```
ingvard: Maybe we want to use current classes without context?  public CacheGroupReencryption(Logger log, etc?)

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupReencryption.java
```diff
@@ -0,0 +1,583 @@
...
+                                pageIdx += 1;
+
+                                long page = pageMem.acquirePage(grpId, pageId);
+
+                                try {
+                                    // Can skip rewriting a dirty page if the checkpoint has been completed.
+                                    if (encrCtx.skipDirty() && pageMem.isDirty(grpId, pageId, page))
+                                        continue;
+
+                                    pageMem.writeLock(grpId, pageId, page, true);
```
ingvard: why it here?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -194,10 +227,28 @@
...
 
     /** Process to perform the master key change. Changes master key and reencrypt group keys. */
-    private DistributedProcess<MasterKeyChangeRequest, MasterKeyChangeResult> performMKChangeProc;
+    private DistributedProcess<MasterKeyChangeRequest, EmptyResult> performMKChangeProc;
+
+    /** Two phase distributed process, that performs cache group encryption key rotation. */
+    private GroupKeyChangeProcess grpKeyChangeProc;
+
+    /** Cache groups for which encryption key was changed, and they must be re-encrypted. */
+    private final Map<Integer, Map<Integer, Long>> reencryptGroups = new ConcurrentHashMap<>();
```
ingvard, line 239: Map<Integer /** PartId **/ etc

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -283,11 +334,17 @@ public GridEncryptionManager(GridKernalContext ctx) {
 
         performMKChangeProc = new DistributedProcess<>(ctx, MASTER_KEY_CHANGE_FINISH, this::performMasterKeyChange,
             this::finishPerformMasterKeyChange);
+
+        reencryption = new CacheGroupReencryption(ctx);
```
ingvard: could you put this dependency outside?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -811,6 +1280,67 @@ private void sendGenerateEncryptionKeyRequest(GenerateEncryptionKeyFuture fut) t
         ctx.io().sendToGridTopic(rndNode.id(), TOPIC_GEN_ENC_KEY, req, SYSTEM_POOL);
     }
 
+    /**
+     * @param grpIds Cache group IDs.
+     * @param skipDirty Dirty page skip flag.
+     * @throws IgniteCheckedException If failed.
+     */
+    private void startReencryption(Collection<Integer> grpIds, boolean skipDirty) throws IgniteCheckedException {
```
ingvard, line 1288: Could you divide the class into subclasses and use a compositional approach?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -1267,6 +1910,322 @@ private String decryptKeyName(byte[] data) {
         });
     }
 
+    /**
+     * Two phase distributed process, that performs cache group encryption key rotation.
+     */
+    private class GroupKeyChangeProcess {
```
ingvard, line 1916: Move to external class, e.g. with package visibility. GridEncryptionManager more 1000 lines are to match.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -182,7 +215,7 @@
     private volatile boolean recoveryMasterKeyName;
 
     /** Master key change future. Not {@code null} on request initiator. */
-    private MasterKeyChangeFuture masterKeyChangeFut;
+    private volatile KeyChangeFuture masterKeyChangeFut;
```
NSAmelchev, line 218: Why it should be volitile?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -627,10 +820,83 @@ public void groupKey(int grpId, byte[] encGrpKey) {
...
+
+            if (!grp.config().isEncryptionEnabled())
+                throw new IgniteException("Cache or group \"" + cacheOrGroupName + "\" is not encrypted.");
+
+            if (reencryptGroups.containsKey(grp.groupId()))
+                throw new IgniteException("Reencryption is in progress [grp=" + cacheOrGroupName + "]");
+
+            grpIds[n] = grp.groupId();
+            keyIds[n] = (byte)(groupKey(grp.groupId()).unsignedId() + 1);
+            keys[n] = getSpi().encryptKey(getSpi().create());
```
NSAmelchev, line 875: Must be under master key change read lock

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -676,6 +962,100 @@ public void onCacheGroupDestroyed(int grpId) {
...
+                    if (reencryptGroups.containsKey(grpId))
+                        continue;
+
+                    Set<Integer> keyIds = entry.getValue();
+
+                    boolean rmv = removeKeysById(grpEncKeys.get(grpId), keyIds);
+
+                    assert rmv : keyIds;
+
+                    writeToMetaStore(grpId, true, false);
```
NSAmelchev, line 1007: Seems, it should be under master key change read lock

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupEncryptionKeys.java
```diff
@@ -0,0 +1,387 @@
...
+
+/**
+ * Serves for managing encryption keys and related datastructure located in the heap.
+ */
+class CacheGroupEncryptionKeys {
+    /** Group encryption keys. */
+    private final Map<Integer, List<GroupKey>> grpKeys = new ConcurrentHashMap<>();
+
+    /** WAL segments encrypted with previous encrypted keys, mapped to cache group encryption key identifiers. */
+    private final Map<Long, Map<Integer, Set<Integer>>> trackedWalSegments = new ConcurrentSkipListMap<>();
```
alex-plekhanov, line 47: You don't need Map of Maps here. You can use plain structure with flat (walIdx, grpId, keyId) tuple (ConcurrentLinkedQueue for example). This structure will be naturally ordered, since you call reserveWalKey under the lock and ctx.cache().context().wal().currentSegment() always increasing.
This will simplify a lot all methods where trackedWalSegments is used.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
```diff
@@ -0,0 +1,567 @@
...
+                    return;
+                }
+
+                PageMemoryEx pageMem = (PageMemoryEx)grp.dataRegion().pageMemory();
+                long metaPageId = pageMem.partitionMetaPageId(grpId, partId);
+                long state = ctx.encryption().getEncryptionState(grpId, partId);
+                int batchSize = encrCfg.getReencryptionBatchSize();
+
+                int off = (int)(state >> Integer.SIZE);
+                int cnt = (int)state;
```
alex-plekhanov, line 489: Logic for extraction state is located at different classes. Perhaps there should be some helper class or helper methods in one class to extract and generate state out of offset and pages count.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
```diff
@@ -0,0 +1,567 @@
...
+                        limiter.acquire(pagesCnt);
+
+                    synchronized (this) {
+                        if (isDone())
+                            break;
+
+                        ctx.cache().context().database().checkpointReadLock();
+
+                        try {
+                            off += scanPages(pageMem, metaPageId + off, pagesCnt);
```
alex-plekhanov, line 509: Let's not rely on knowledge of pageId structure and pass just offset to scanPages and generate pageId each time from offset by page index.

modules/core/src/main/java/org/apache/ignite/internal/util/BasicRateLimiter.java
```diff
@@ -0,0 +1,154 @@
...
+     * @param permitsPerSecond Estimated number of permits per second.
+     */
+    public BasicRateLimiter(double permitsPerSecond) {
+        setRate(permitsPerSecond);
+    }
+
+    /**
+     * Updates the stable rate.
+     *
+     * @param permitsPerSecond the new stable rate of this {@code RateLimiter}
```
alex-plekhanov, line 77: Please fix all comments in this class according to Ignite codestyle (the point at the end, sentence start from the upper case)

modules/core/src/main/java/org/apache/ignite/internal/util/BasicRateLimiter.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    /**
+     * Acquires the given number of permits from this {@code RateLimiter}, blocking until the request
+     * can be granted. Tells the amount of time slept, if any.
+     *
+     * @param permits the number of permits to acquire
+     * @throws IllegalArgumentException if the requested number of permits is negative or zero
+     * @since 16.0 (present in 13.0 with {@code void} return type})
```
alex-plekhanov, line 105: It's not related anyhow to Ignite versions, please remove it.

modules/core/src/test/config/log4j-test.xml
```diff
@@ -84,6 +84,10 @@
     </category>
     -->
 
+    <category name="org.apache.ignite.internal.managers.encryption">
```
alex-plekhanov, line 87: Don't forget to remove it after testing.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
```diff
@@ -0,0 +1,567 @@
...
+
+    /**
+     * Collect current number of pages in the specified cache group.
+     *
+     * @param grp Cache group.
+     * @return Map of partitions with current page count.
+     * @throws IgniteCheckedException If failed.
+     */
+    public Map<Integer, Long> pagesCount(CacheGroupContext grp) throws IgniteCheckedException {
+        Map<Integer, Long> partStates = new HashMap<>();
```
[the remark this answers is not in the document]
xtern: Map<Integer, ReencryptState> replaced with long[grp.affinity().partitions() + 1] (+1 - for index partition)

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
xtern: Added utility class ReencryptStateUtils for this

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -627,22 +716,104 @@ public void groupKey(int grpId, byte[] encGrpKey) {
         return withMasterKeyChangeReadLock(() -> getSpi().getMasterKeyName());
     }
 
+    /** {@inheritDoc} */
+    @Override public IgniteFuture<Void> changeCacheGroupKey(Collection<String> cacheOrGrpNames) {
+        A.notEmpty(cacheOrGrpNames, "cacheOrGrpNames");
+
+        synchronized (opsMux) {
+            if (stopped) {
```
NSAmelchev, line 724: We should check `disconnected` too

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GroupKeyChangeProcess.java
```diff
@@ -0,0 +1,333 @@
...
+     * @param cacheOrGrpNames Cache or group names.
+     */
+    public IgniteFuture<Void> start(Collection<String> cacheOrGrpNames) {
+        if (ctx.clientNode())
+            throw new UnsupportedOperationException("Client and daemon nodes can not perform this operation.");
+
+        if (!IgniteFeatures.allNodesSupports(ctx.grid().cluster().nodes(), CACHE_GROUP_KEY_CHANGE))
+            throw new IllegalStateException("Not all nodes in the cluster support this operation.");
+
+        if (ctx.state().clusterState().state() != ClusterState.ACTIVE)
```
NSAmelchev, line 122: if read_only mode is supported than `state().active()` can be used

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -503,58 +549,96 @@ else if (newKeys != null) {
...
+    /**
+     * Adds new cache group encryption key.
+     *
+     * @param grpId Cache group ID.
+     * @param key Encryption key.
+     */
+    private void addGroupKey(int grpId, GroupKeyEncrypted key) {
+        try {
+            synchronized (metaStorageMux) {
+                withMasterKeyChangeReadLock(() -> grpKeys.changeActiveKey(grpId, key));
```
NSAmelchev, line 636: There is possible deadlock with the `doChangeMasterKey` method in order of locks acquire

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -182,7 +195,7 @@
     private volatile boolean recoveryMasterKeyName;
 
     /** Master key change future. Not {@code null} on request initiator. */
```
NSAmelchev, line 197: Javadoc shpuld be fixed too

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupMetricsImpl.java
```diff
@@ -175,6 +181,16 @@ public void onTopologyInitialized() {
         mreg.register("TotalAllocatedSize",
             this::getTotalAllocatedSize,
             "Total size of memory allocated for group, in bytes.");
+
+        if (ctx.config().isEncryptionEnabled()) {
+            mreg.register("ReencryptionFinished",
```
NSAmelchev, line 186: I think it is better to move metrics to a separate issue.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheOffheapManager.java
```diff
@@ -993,6 +1035,11 @@ private Metas getOrAllocateCacheMetas() throws IgniteCheckedException {
                     null,
                     null);
             }
+            catch (Throwable e) {
+                e.printStackTrace();
```
NSAmelchev, line 1039: Use `log`, please

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -503,58 +549,96 @@ else if (newKeys != null) {
...
-                    entry.getKey() + "]");
+                    grpId + ", keyId=" + rmtKey.id() + "]");
+
+                continue;
             }
+
+            U.quietAndInfo(log, "Store group key received from coordinator [grp=" + grpId +
+                ", keyId=" + rmtKey.id() + "]");
+
+            //changeActiveKey
```
alex-plekhanov: ?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -194,10 +207,22 @@
...
+    private DistributedProcess<MasterKeyChangeRequest, EmptyResult> performMKChangeProc;
+
+    /** Two phase distributed process, that performs cache group encryption key rotation. */
+    private GroupKeyChangeProcess grpKeyChangeProc;
+
+    /** Cache groups for which encryption key was changed, and they must be re-encrypted. */
+    private final Map<Integer, long[]> reencryptGroups = new ConcurrentHashMap<>();
+
+    /** Cache groups for which encryption key was changed on node join. */
+    private final Map<Integer, Integer> reencryptGroupsForced = new ConcurrentHashMap<>();
```
alex-plekhanov, line 222: Why do we only add entries to this map and never clean it up?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheOffheapManager.java
xtern: thanks, removed it - was used for debugging

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GroupKeyChangeProcess.java
```diff
@@ -0,0 +1,336 @@
...
+                "The previous change was not completed."));
+        }
+
+        int[] grpIds = new int[cacheOrGrpNames.size()];
+        byte[] keyIds = new byte[grpIds.length];
+
+        int n = 0;
+
+        for (String cacheOrGroupName : cacheOrGrpNames) {
+            CacheGroupContext grp = ctx.cache().cacheGroup(CU.cacheId(cacheOrGroupName));
```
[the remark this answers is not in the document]
xtern: Changed to descriptors, added test with node filter

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
```diff
@@ -0,0 +1,453 @@
...
+            return onDone(null, null, true);
+        }
+
+        /**
+         * Stop reencryption of the specified partition.
+         *
+         * @param partId Partition ID.
+         * @return {@code True} if reencryption was cancelled.
+         */
+        public synchronized boolean cancel(int partId) {
```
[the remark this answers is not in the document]
xtern: renamed to excludePartition

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
```diff
@@ -0,0 +1,453 @@
...
+
+    /**
+     * Collect current number of pages in the specified cache group.
+     *
+     * @param grp Cache group.
+     * @return Partitions with current page count.
+     * @throws IgniteCheckedException If failed.
+     */
+    public long[] pagesCount(CacheGroupContext grp) throws IgniteCheckedException {
+        long[] partStates = new long[grp.affinity().partitions() + 1];
```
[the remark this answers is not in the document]
xtern: added `// The last element of the array is used to store the status of the index section.`

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GroupKeyChangeProcess.java
```diff
@@ -0,0 +1,336 @@
...
+import org.apache.ignite.lang.IgniteFutureCancelledException;
+
+import static org.apache.ignite.internal.IgniteFeatures.CACHE_GROUP_KEY_CHANGE;
+import static org.apache.ignite.internal.util.distributed.DistributedProcess.DistributedProcessType.CACHE_GROUP_KEY_CHANGE_FINISH;
+import static org.apache.ignite.internal.util.distributed.DistributedProcess.DistributedProcessType.CACHE_GROUP_KEY_CHANGE_PREPARE;
+
+/**
+ * Two phase distributed process, that performs cache group encryption key rotation.
+ */
+class GroupKeyChangeProcess {
```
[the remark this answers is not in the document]
xtern: new javadoc
`
Cache group page stores scanner.
Scans a range of pages and marks them as dirty to re-encrypt them with the last encryption key on disk.
`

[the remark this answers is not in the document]
xtern: javadoc updated to
`
A two-phase distributed process that rotates the encryption keys of specified cache groups and initiates
re-encryption of those cache groups.
`

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -627,22 +716,104 @@ public void groupKey(int grpId, byte[] encGrpKey) {
...
+        startReencryption(encryptionStatus.keySet());
+    }
+
+    /**
+     * @param grpId Cache group ID.
+     * @return Future that will be completed when reencryption of the specified group is finished.
+     */
+    public IgniteInternalFuture<Void> reencryptionFuture(int grpId) {
+        if (pageScanner.disabled() && reencryptGroups.containsKey(grpId))
+            return new GridFutureAdapter<>();
```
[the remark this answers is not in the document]
xtern: Changed to throw an exception

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -811,6 +1124,53 @@ private void sendGenerateEncryptionKeyRequest(GenerateEncryptionKeyFuture fut) t
...
+    private void startReencryption(Collection<Integer> grpIds) throws IgniteCheckedException {
+        if (pageScanner.disabled())
+            return;
+
+        for (int grpId : grpIds) {
+            IgniteInternalFuture<?> fut = pageScanner.schedule(grpId);
+
+            fut.listen(f -> {
+                try {
+                    f.get();
```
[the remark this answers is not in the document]
xtern: What is the reason to change local `try/catch` to multiple  `if/else`?  Fir me it looks like changing `switch/case` to `if/else`

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
xtern: fixed

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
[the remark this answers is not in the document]
xtern: In this case, `synchronized` guarantees that after the call to cancel, the scan will not continue (we will get nasty exceptions in the log if we continue to scan the destroyed partition)

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
```diff
@@ -0,0 +1,453 @@
...
+         */
+        public GroupScanTask(CacheGroupContext grp, Set<Integer> parts) {
+            this.grp = grp;
+            this.parts = new GridConcurrentHashSet<>(parts);
+
+            pageMem = (PageMemoryEx)grp.dataRegion().pageMemory();
+        }
+
+        /** {@inheritDoc} */
+        @Override public synchronized boolean cancel() throws IgniteCheckedException {
```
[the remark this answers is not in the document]
xtern: In this case, `synchronized` guarantees that after the call to cancel, the scan will not continue

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/PagePartitionMetaIOV3.java
```diff
@@ -0,0 +1,137 @@
...
+        if (getEncryptedPageCount(pageAddr) == pagesCnt)
+            return false;
+
+        PageUtils.putInt(pageAddr, ENCRYPT_PAGE_MAX_OFF, pagesCnt);
+
+        return true;
+    }
+
+    /** {@inheritDoc} */
+    @Override protected void printPage(long pageAddr, int pageSize, GridStringBuilder sb) throws IgniteCheckedException {
```
alex-plekhanov, line 99: Let's simplify this method: for example, introduce new method to print fields which can reuse the same method from superclass and just add new fields of current class.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/PagePartitionMetaIOV3.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Upgrade page to PagePartitionMetaIOV3.
+     *
+     * @param pageAddr Page address.
+     */
+    @Override public void upgradePage(long pageAddr) {
+        assert PageIO.getType(pageAddr) == getType();
+        assert PageIO.getVersion(pageAddr) < 3;
+
```
alex-plekhanov, line 131: Let's invoke super method if version < 2

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
```diff
@@ -0,0 +1,453 @@
...
+                        return;
+                    }
+
+                    parts.add(partId);
+                }
+            });
+
+            GroupScanTask grpScan = new GroupScanTask(grp, parts);
+
+            ctx.getSystemExecutorService().submit(grpScan);
```
alex-plekhanov, line 205: Here task for each cache group can be executed by different threads. Let's use single-threaded execution here.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -2863,11 +2869,16 @@ private RestoreLogicalState applyLogicalUpdates(
 
                         break;
 
-                    case MASTER_KEY_CHANGE_RECORD:
+                    case MASTER_KEY_CHANGE_RECORD_V2:
```
alex-plekhanov, line 2872: Let's also keep `case MASTER_KEY_CHANGE_RECORD:`

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/delta/MetaPageUpdateIndexDataRecord.java
```diff
@@ -0,0 +1,105 @@
...
+    public MetaPageUpdateIndexDataRecord(int grpId, long pageId, int encryptPageIdx, int encryptPageCnt) {
+        super(grpId, pageId);
+
+        this.encryptPageIdx = encryptPageIdx;
+        this.encryptPageCnt = encryptPageCnt;
+    }
+
+    /** {@inheritDoc} */
+    @Override public void applyDelta(PageMemory pageMem, long pageAddr) throws IgniteCheckedException {
+        PageIndexMetaIO io = PageIndexMetaIO.VERSIONS.forPage(pageAddr);
```
alex-plekhanov, line 54: Perhaps here we can face with a not upgraded page (in case node failed after page upgrade and before first checkpoint)

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -627,22 +720,104 @@ public void groupKey(int grpId, byte[] encGrpKey) {
...
+     */
+    public IgniteInternalFuture<Void> reencryptionFuture(int grpId) {
+        return pageScanner.statusFuture(grpId);
+    }
+
+    /**
+     * @param grpId Cache group ID.
+     * @return {@code True} If the specified cache group should be re-encrypted.
+     */
+    public boolean reencryptionRequired(int grpId) {
```
alex-plekhanov, line 798: All usages of this method want to check that reencryption in progress. Let's rename it to something like `reencryptionInProgress`

modules/core/src/test/java/org/apache/ignite/internal/util/BasicRateLimiterTest.java
```diff
@@ -0,0 +1,119 @@
...
+import org.junit.Test;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Rate limiter tests.
+ */
+public class BasicRateLimiterTest {
```
alex-plekhanov, line 34: Not included into any test suite

modules/core/src/test/java/org/apache/ignite/internal/util/BasicRateLimiterTest.java
```diff
@@ -0,0 +1,119 @@
...
+    public void checkAcquireWithOverflow() throws IgniteInterruptedCheckedException {
+        double permitsPerSec = 0.5;
+        int permitsPerOp = 1;
+        int totalOps = 5;
+
+        BasicRateLimiter limiter = new BasicRateLimiter(permitsPerSec);
+
+        long startTime = System.currentTimeMillis();
+
+        for (int i = 0; i <= totalOps; i++)
```
alex-plekhanov, line 86: Actually, you check `totalOps + 1` operations here

modules/core/src/test/java/org/apache/ignite/internal/util/BasicRateLimiterTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+        ready.await();
+
+        long startTime = System.currentTimeMillis();
+
+        fut.get();
+
+        long timeSpent = System.currentTimeMillis() - startTime;
+
+        assertEquals(totalOps / opsPerSec, SECONDS.convert(timeSpent, MILLISECONDS));
```
alex-plekhanov, line 70: In some circumstances, I think `timeSpent` can be a little bit less than 10 seconds

modules/core/src/test/java/org/apache/ignite/internal/encryption/AbstractEncryptionTest.java
```diff
@@ -271,4 +313,185 @@ protected boolean checkMasterKeyName(String name) {
...
+                        long pageOff = pageStore.pageOffset(pageId);
+
+                        pageBuf.position(0);
+
+                        ch.position(pageOff);
+                        ch.read(pageBuf);
+
+                        pageBuf.position(realPageSize + encryptionBlockSize);
+
+                        // If crc present
```
alex-plekhanov, line 434: Point at the end.
Why there can be pages without CRC?

modules/core/src/test/java/org/apache/ignite/internal/encryption/CacheGroupKeyChangeTest.java
```diff
@@ -0,0 +1,988 @@
...
+        checkGroupKey(grpId, INITIAL_KEY_ID + 1, MAX_AWAIT_MILLIS);
+
+        stopAllGrids();
+
+        node1 = startGrid(GRID_0);
+        node2 = startGrid(GRID_1);
+
+        node1.cluster().state(ClusterState.ACTIVE);
+
+        // Previous leys must be deleted when the corresponding WAL segment is deleted.
```
alex-plekhanov, line 537: leys -> keys

modules/core/src/test/java/org/apache/ignite/internal/encryption/CacheGroupKeyChangeTest.java
```diff
@@ -0,0 +1,988 @@
...
+        startLatch.await(MAX_AWAIT_MILLIS, TimeUnit.MILLISECONDS);
+
+        int grpId = cache.context().groupId();
+
+        node0.encryption().changeCacheGroupKey(Collections.singleton(cacheName())).get(MAX_AWAIT_MILLIS);
+
+        Set<Integer> keys1 = new TreeSet<>(node0.context().encryption().groupKeyIds(grpId));
+        Set<Integer> keys2 = new TreeSet<>(node1.context().encryption().groupKeyIds(grpId));
+
+        assertEquals(keys1, keys2);
```
alex-plekhanov, line 597: Load future is still working and one of the nodes can already switch WAL segment, do checkpoint and delete key id in background, so test can become flaky.

modules/core/src/test/java/org/apache/ignite/internal/encryption/CacheGroupKeyChangeTest.java
```diff
@@ -0,0 +1,988 @@
...
+        forceCheckpoint();
+
+        int grpId = CU.cacheId(cacheName());
+
+        assertEquals(2, node1.context().encryption().groupKeyIds(grpId).size());
+        assertEquals(2, node2.context().encryption().groupKeyIds(grpId).size());
+
+        stopAllGrids();
+
+        // Cleanup wal arcive.
```
alex-plekhanov, line 695: WAL archive

modules/core/src/test/java/org/apache/ignite/internal/encryption/CacheGroupKeyChangeTest.java
```diff
@@ -0,0 +1,988 @@
...
+            stopLoad.set(true);
+
+            fut.get(MAX_AWAIT_MILLIS);
+        }
+
+        forceCheckpoint();
+
+        int grpId = CU.cacheId(cacheName());
+
+        assertEquals(2, node1.context().encryption().groupKeyIds(grpId).size());
```
alex-plekhanov, line 690: Can we temporary restrict deletions? Or here we can get size 1 in some circumstances

modules/core/src/test/java/org/apache/ignite/internal/encryption/CacheGroupKeyChangeTest.java
```diff
@@ -0,0 +1,988 @@
...
+
+        /** */
+        public void stopBlock() {
+            unlockLatch.countDown();
+        }
+    }
+
+    /** @throws Exception If failed. */
+    @Test
+    public void testChangeCacheGroupKeyWithoutWAL() throws Exception {
```
alex-plekhanov, line 971: Let's move this test before InitMessageDiscoveryHook class

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/ReencryptionStatusRecord.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+package org.apache.ignite.internal.pagemem.wal.record;
+
+import java.util.Map;
+
+/**
+ * Logical record to restart reencryption with the latest encryption key.
+ */
+public class ReencryptionStatusRecord extends WALRecord {
```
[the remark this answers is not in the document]
xtern: rernamed to "ReencryptionStartRecord"

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/PagePartitionMetaIOV3.java
```diff
@@ -0,0 +1,137 @@
...
+        sb.a("PagePartitionMeta[\n\ttreeRoot=").a(getReuseListRoot(pageAddr));
+        sb.a(",\n\tpendingTreeRoot=").a(getLastSuccessfulFullSnapshotId(pageAddr));
+        sb.a(",\n\tlastSuccessfulFullSnapshotId=").a(getLastSuccessfulFullSnapshotId(pageAddr));
+        sb.a(",\n\tlastSuccessfulSnapshotId=").a(getLastSuccessfulSnapshotId(pageAddr));
+        sb.a(",\n\tnextSnapshotTag=").a(getNextSnapshotTag(pageAddr));
+        sb.a(",\n\tlastSuccessfulSnapshotTag=").a(getLastSuccessfulSnapshotTag(pageAddr));
+        sb.a(",\n\tlastAllocatedPageCount=").a(getLastAllocatedPageCount(pageAddr));
+        sb.a(",\n\tcandidatePageCount=").a(getCandidatePageCount(pageAddr));
+        sb.a(",\n\tencryptedPageIndex=").a(getEncryptedPageIndex(pageAddr));
+        sb.a(",\n\tencryptedPageCount=").a(getEncryptedPageCount(pageAddr));
```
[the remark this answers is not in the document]
xtern: Fixed, added additional method to  print fields.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/PagePartitionMetaIOV3.java
xtern: added additional method to print fields.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
xtern: Added single threaded executor

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -483,16 +534,18 @@ public void onLocalJoin() {
         if (dataBag.isJoiningNodeClient() || dataBag.commonDataCollectedFor(ENCRYPTION_MGR.ordinal()))
             return;
 
-        HashMap<Integer, byte[]> knownEncKeys = knownEncryptionKeys();
+        HashMap<Integer, GroupKeyEncrypted> knownEncKeys = grpKeys.getAll();
```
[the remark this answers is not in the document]
xtern: These methods cannot be executed concurrently with the master key change, as the node join is rejected during the master key change
   onLocalJoin
   onJoiningNodeDataReceived
   collectGridNodeData

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -821,11 +1181,57 @@ private void writeKeysToMetaStore(boolean writeAll) throws IgniteCheckedExceptio
...
+     *
+     * @param grpId Cache group ID.
+     */
+    private void writeGroupKeysToMetaStore(int grpId) throws IgniteCheckedException {
+        assert Thread.holdsLock(metaStorageMux);
+
+        if (metaStorage == null || !writeToMetaStoreEnabled || stopped)
+            return;
+
+        List<GroupKeyEncrypted> keysEncrypted = withMasterKeyChangeReadLock(() -> grpKeys.getAll(grpId));
```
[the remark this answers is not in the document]
xtern: Locks reordered.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/WALRecord.java
```diff
@@ -233,7 +233,25 @@
...
+        ENCRYPTED_DATA_RECORD_V2(64, LOGICAL),
+
+        /** Master key change record containing multiple keys for single cache group. */
+        MASTER_KEY_CHANGE_RECORD_V2(65, LOGICAL),
+
+        /** Logical record to restart reencryption with the latest encryption key. */
+        REENCRYPTION_START_RECORD(66, LOGICAL),
+
+        /** Partition meta page delta record includes encryption status data. */
+        PARTITION_META_PAGE_UPDATE_COUNTERS_V3(67, PHYSICAL),
```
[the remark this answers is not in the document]
xtern: Renamed PARTITION_META_PAGE_UPDATE_COUNTERS_V3 => PARTITION_META_PAGE_DELTA_RECORD_V3

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
xtern: There is no compatibility issue. If we read "old" (MASTER_KEY_CHANGE_RECORD) in the serializer, we will have a new WALRecord with type V2 here.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/delta/MetaPageUpdateIndexDataRecord.java
xtern: Added type change
```
        // Upgrade meta page.
        if (PageIO.getType(pageAddr) == PageIO.T_META)
            PageIO.setType(pageAddr, PageIO.T_INDEX_META);
```

modules/core/src/test/java/org/apache/ignite/internal/util/BasicRateLimiterTest.java
xtern: FIxed (included in "Basic 1")

modules/core/src/test/java/org/apache/ignite/internal/util/BasicRateLimiterTest.java
xtern: Ratelimiter targets for an average rate of permits per second, these tests have been reworked. To prevent instability of tests, I added a 50% error probability (in general, the error does not exceed 20%).

modules/core/src/test/java/org/apache/ignite/internal/encryption/CacheGroupKeyChangeTest.java
xtern: Added WAL reservation to prevent deletion.

modules/core/src/test/java/org/apache/ignite/internal/encryption/AbstractEncryptionTest.java
xtern: CRC can be calculated as "0" - it was my mistake.
But in rare cases, when scanning a partition on a disk, we find a blank page. I added a special check for this case.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupReencryption.java
xtern: Now it's called  CacheGroupPageScanner )

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
xtern: I do not quite understand the essence of the problem with this particular method and what kind of decomposition needs to be done with it.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupReencryption.java
xtern: Context is not fully initialized here, moreover I'm not sure that greatly increasing the number of arguments will improve the code. I understand that we want to loose the coupling, but I am not sure that it's easy to do here (class requires EncryptionManager, PageStoreManager, GridCacheDatabaseSharedManager and GridCacheProcessor at least).

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupReencryption.java
xtern: Could you describe the question in more detail?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
xtern: submap was replaced with array of longs

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
xtern: Could you describe in more detail what should we do here?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -627,22 +739,104 @@ public void groupKey(int grpId, byte[] encGrpKey) {
...
+        for (int i = 0; i < grpIds.length; i++) {
+            int grpId = grpIds[i];
+            int newKeyId = keyIds[i] & 0xff;
+
+            synchronized (metaStorageMux) {
+                // Set new key as key for writing. Note that we cannot pass the encrypted key here because the master
+                // key may have changed in which case we will not be able to decrypt the cache encryption key.
+                GroupKey prevGrpKey = grpKeys.changeActiveKey(grpId, newKeyId);
+
+                List<GroupKeyEncrypted> keysEncrypted = withMasterKeyChangeReadLock(() -> grpKeys.getAll(grpId));
```
alex-plekhanov, line 782: Wrong lock order again

xtern: Thanks, fixed it.

alex-plekhanov (approved):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/PageIO.java
```diff
@@ -258,6 +258,9 @@
     /** */
     public static final short T_MARKER_PAGE = 33;
 
+    /** */
+    public static final short T_INDEX_META = 34;
```
ibessonov, line 262: Can you please explain why you needed new IO type instead of new version for existing type? You only add two fields, right? I don't get it then.

xtern: Yes, we adding two int fields into two pages.

We have 2 types of memory pages:
T_META (index partition)
T_PART_META (other partition)

We want to add 2 ints into each of them, but don't forget that IO for T_PART_META extends IO for T_META.

In my first approach I added 2 ints into T_META, and added PagePartitionMetaIOV3 for T_PART_META.
But since PagePartitionMetaIO "extends" PageMetaIO all offsets in PagePartitionMetaIO was shifted and there was a problem with inheritance (for example in PagePartitionMetaIO/V2 new methods were visible, all offsets was redefined in PagePartitionMetaIOV3), you can imagine what it looked like in commit with the second approach https://github.com/apache/ignite/pull/7941/commits/036b931bd40e2ffbd519779fde86fae7343da23e

With current approach we simply create the new IO that is only used for the "index" partition.

If you can suggest a better approach how for storing 2 extra int's itn the index meta page without losing binary compatibility, please describe.

p.s. I explained this change in wiki https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=95652384#TDE.Phase3.Cachekeyrotation.-PageMetaIOandPagePartitionMetaIO

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/EncryptedRecord.java
```diff
@@ -43,7 +43,7 @@ public EncryptedRecord(int grpId, RecordType plainRecType) {
 
     /** {@inheritDoc} */
     @Override public RecordType type() {
-        return RecordType.ENCRYPTED_RECORD;
+        return RecordType.ENCRYPTED_RECORD_V2;
```
sk0x50, line 46: Could you please describe in short why you need to change the type of this WAL record?
WAL record was not changed itself, so it looks weird to me.

xtern: For binary compatibility.
For ENCRYPTED_RECORD_V2 we reading 1 extra byte (encryption key identifier).
(see change in RecordDataV1Serializer.readEncryptedData)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/PageIO.java
ibessonov: I suggest class PageIndexIO being just T_META with version 2, see no problems with this approach. Just as now, two different classes will add two independent sets of new integers, that's fine. The only thing that's changed is that PageIndexIO will have the same type as its base class.

Types should be immutable while pages are alive. This saves us from a lot of troubles, I wouldn't abandon this tradition. That's exactly what versions are for.

BTW your current code has bugs, you don't mark meta page as dirty in "getOrAllocateCacheMetas" while upgrading, please fix it.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/EncryptedRecord.java
sk0x50: > (see change in RecordDataV1Serializer.readEncryptedData)
Yep, I saw this, but it does not explain why you modify the existing WAL record.

In my understanding, WAL records must be immutable in order to avoid compatibility issues in the future.
I would prefer adding a new WAL record (it can extend an existing one).
Perhaps, RecordDataV1Serializer implementation will be cleaner in that case.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/PageIO.java
xtern: If I understood you correctly, you suggest to add PageMetaIOV2 (instead of PageIndexMetaIO)?
PagePartitionMeta will continue to extend PageMetaIO(not v2), is this correct?

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/EncryptedRecord.java
xtern: Thanks for the suggestion, as far as I understand you are suggesting not to modify any existing WAL record classes.
Could you please give an example of a compatibility issue that we may face in the future (in theory)?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/PageIO.java
ibessonov: Yes, this is correct. Naming of the classes doesn't bother me too much so I was fine with "Indexing" instead of V2 (although that page is also referred as "cache group meta page" sometimes, like in PageMemoryEx#metaPageId, for example).

I think I understand the core problem, inheritance between MetaIO and PartitionMetaIO should have been replaced with extension of the same abstract superclass back in the day. But with current faulty design people just get confused :(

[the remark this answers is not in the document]
xtern: Thanks, I believe I've fixed it.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/EncryptedRecord.java
sk0x50: Hello @xtern,

I think changing already existing WAL records is an error-prone approach.
Today we changed a type; tomorrow we will change/add a field and skip the corresponding changes in the serializer code.
It is just good practice not to change existing records, IMHO. Anyway, it is up to you.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
```diff
@@ -0,0 +1,479 @@
...
+                    if (state == 0)
+                        continue;
+
+                    scanPartition(partId, ReencryptStateUtils.pageIndex(state), ReencryptStateUtils.pageCount(state));
+
+                    if (isDone())
+                        return;
+                }
+
+                boolean added = cpWaitGrps.add(this);
```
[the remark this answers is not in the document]
xtern: cpWaitGrps is a collection of groups waiting for a checkpoint to start.

PageScanner performs the following steps:
1. mark all pages in the group as dirty
2. wait for the start of the checkpoint (onCheckpointBegin)
3. wait for finishedStateFut() of the current checkpoint to complete

I though that after finishedStateFut finishes, all pages (dirty at the time of starting the checkpoint) are written to disk, am I missing something?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupEncryptionKeys.java
```diff
@@ -0,0 +1,376 @@
...
+ */
+class CacheGroupEncryptionKeys {
+    /** Group encryption keys. */
+    private final Map<Integer, List<GroupKey>> grpKeys = new ConcurrentHashMap<>();
+
+    /**
+     * WAL segments encrypted with previous encryption keys prevent keys from being deleted
+     * until the associated segment is deleted.
+     */
+    private final Collection<TrackedWalSegment> trackedWalSegments = new ConcurrentLinkedQueue<>();
```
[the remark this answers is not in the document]
xtern: It is possible that the old key will not be deleted until the next key change.

For example, we start with encryption key with ID = 0.

creates new WAL segment 0 (encrypted with key ID = 0)
..
creates new WAL segment 10 (encrypted with key ID = 0)

change encryption key ID = 1

creates new WAL segment 11 (encrypted with key ID = 1)
..
creates new WAL segment 15 (encrypted with key ID = 1)

change encryption key ID = 2

creates new WAL segment 16 (encrypted with key ID = 2)
...

and so on.

So, in order to read segment 3 of WAL (for example, for delta rebalancing), we need to keep the key with ID = 0 until the segment 10 is removed from the disk.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GroupKeyChangeProcess.java
```diff
@@ -0,0 +1,350 @@
...
+                if (!curMasterKeyName.equals(req.masterKeyName())) {
+                    return new GridFinishedFuture<>(new IgniteException("Cache group key change was rejected. " +
+                        "Master key has been changed."));
+                }
+
+                for (int i = 0; i < req.groupIds().length; i++) {
+                    // Store new key as inactive.
+                    GroupKeyEncrypted grpKey = new GroupKeyEncrypted(req.keyIds()[i] & 0xff, req.keys()[i]);
+
+                    ctx.encryption().addGroupKey(req.groupIds()[i], grpKey);
```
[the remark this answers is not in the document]
xtern: Because the master key may changes between prepare/perform stages and we'll not be able to decrypt received keys.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupEncryptionKeys.java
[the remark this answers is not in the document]
xtern: We only keep track of the last WAL segment encrypted with a specific encryption key, so we only store the current WAL segment index every time the encryption key is rotated.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/PageIO.java
xtern: I want to agree on naming for this change:

Type: T_META
Io: MetaPageIOV2 _(extends MetaPageIO)_

Wal record: MetaPageUpdate**Index**DataRecord 
Wal record type: **INDEX**_META_PAGE_DELTA_RECORD

is it ok?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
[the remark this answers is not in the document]
vldpyatkov: I see that we cannot guarantee which pages included in that checkpoint. Because onCheckpointBegin method is executed out of write lock, that means some dirty pages will belong to another checkpoint (next after that).

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupEncryptionKeys.java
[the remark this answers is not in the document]
vldpyatkov: Also, it is a misusing queue. I looked at the code and don't found any queue specific operation.
But I found this method releaseWalKeys
I think `NavigableMap<Long, TrackedWalSegment>` fits better here.
`trackedWalSegments.tailMap(walIdx, true)`

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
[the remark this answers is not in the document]
xtern: @vldpyatkov thanks, seems that we should use beforeCheckpointBegin instead of onCheckpointBegin, right?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupEncryptionKeys.java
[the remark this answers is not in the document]
vldpyatkov: These map looks like CheckpointHistory. Can we thought about move it there?

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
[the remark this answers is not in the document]
vldpyatkov: Now, we can get information about a written group after than that happens.
If it is ok, that is right. I still badly oriented in these patch.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupEncryptionKeys.java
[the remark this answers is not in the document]
xtern: @vldpyatkov, we cannot use `NavigableMap<Long, TrackedWalSegment>` here, we can use something like `NavigableMap<Long, Collection<Pair <GrpId, KeyId>>` as in theory we can have multiple cache groups for which the encryption key has been changed multiple times on single WAL segment.

I don't think that NavigableMap will be better than the queue in this case.

In my first approach, I used `ConcurrentSkipListMap` and the code looked more complicated, please check this review comment https://github.com/apache/ignite/pull/7941#discussion_r465759058

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/CacheGroupPageScanner.java
```diff
@@ -0,0 +1,479 @@
...
+            new OomExceptionHandler(ctx));
+
+        singleExecSvc.allowCoreThreadTimeOut(true);
+    }
+
+    /** {@inheritDoc} */
+    @Override public void onCheckpointBegin(Context cpCtx) {
+        Set<GroupScanTask> completeCandidates = new HashSet<>();
+
+        cpWaitGrps.removeIf(completeCandidates::add);
```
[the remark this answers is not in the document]
xtern: As mentioned in the previous comment, we now use beforeCheckpointBegin to wait for a checkpoint, which is called before the checkpoint mark-phase is executed (and before the write lock is acquired), so we have now synchronization between releasing the read lock (while scanning) and acquiring a write lock (on mark phase).

NSAmelchev (approved):

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/DistributedProcess.java
```diff
@@ -430,6 +430,16 @@ private Process(UUID id) {
          *
          * @see IgniteSnapshotManager
          */
-        END_SNAPSHOT
+        END_SNAPSHOT,
+
+        /**
+         * Cache group encyption key change prepare phase.
+         */
+        CACHE_GROUP_KEY_CHANGE_PREPARE,
```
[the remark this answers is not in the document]
xtern: Thanks, good point, but this patch is big enough and this change is not related to it, so I think it would be better to do it in a separate task.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/EncryptedRecord.java
xtern: I updated this patch and kept the original WAL records (EncryptedRecord and MasterKeyChangeRecord) unchanged.

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GroupKeyChangeProcess.java
```diff
@@ -0,0 +1,352 @@
...
+            grpIds[n] = grpDesc.groupId();
+            keyIds[n] = (byte)(ctx.encryption().groupKey(grpDesc.groupId()).unsignedId() + 1);
+
+            n += 1;
+        }
+
+        byte[][] keys = ctx.encryption().createKeys(grpIds.length).get1().toArray(new byte[grpIds.length][]);
+
+        ChangeCacheEncryptionRequest req =
+            new ChangeCacheEncryptionRequest(grpIds, keys, keyIds, ctx.config().getEncryptionSpi().getMasterKeyName());
```
NSAmelchev, line 170: Lets send a MK digest instead of a name

xtern: Done, thanks.

alex-plekhanov (approved):

PR #7967, opened 2020-06-26
Duck is a duck 

anton-vinogradov:

### Notes


modules/ducktests/tests/docker/Dockerfile
```diff
@@ -48,14 +48,13 @@ ARG APACHE_MIRROR="https://apache-mirror.rbc.ru/pub/apache/"
...
 # Install binary test dependencies.
-ARG IGNITE_VERSION="2.8.0"
-ARG IGNITE_NAME="ignite-$IGNITE_VERSION"
-ARG IGNITE_RELEASE_NAME="apache-ignite-$IGNITE_VERSION-bin"
-
-ADD $APACHE_ARCHIVE/ignite/$IGNITE_VERSION/$IGNITE_RELEASE_NAME.zip /opt/
-RUN cd /opt && unzip $IGNITE_RELEASE_NAME.zip && rm $IGNITE_RELEASE_NAME.zip
-RUN mv /opt/$IGNITE_RELEASE_NAME /opt/$IGNITE_NAME
-RUN chmod a+wr /opt/$IGNITE_NAME -R
+ADD $APACHE_ARCHIVE/ignite/2.7.6/apache-ignite-2.7.6-bin.zip /opt/
```
timoninmaxim, line 51: Use "RUN curl -O" or "RUN wget" instead of ADD to enable caching. Check this [issue](https://github.com/moby/moby/issues/15717)

modules/ducktests/tests/docker/Dockerfile
```diff
@@ -48,14 +48,13 @@ ARG APACHE_MIRROR="https://apache-mirror.rbc.ru/pub/apache/"
...
+RUN cd /opt && unzip apache-ignite-2.7.6-bin.zip && mv /opt/apache-ignite-2.7.6-bin /opt/ignite-2.7.6 && chmod a+wr /opt/$IGNITE_NAME -R
```
timoninmaxim, line 52: IGNITE_NAME was removed, chmod invoked

modules/ducktests/tests/ignitetest/services/ignite.py
```diff
@@ -20,79 +20,52 @@
...
 
-        self.logger.info("Waiting for Ignite to start...")
+        self.logger.info("Waiting for Ignite(s) to start...")
+
+        for node in self.nodes:
+            self.await_node_stated(node, timeout_sec)
 
     def start_cmd(self, node):
         jvm_opts = "-J-DIGNITE_SUCCESS_FILE=" + IgniteService.PERSISTENT_ROOT + "/success_file "
         jvm_opts += "-J-Dlog4j.configDebug=true"
```
timoninmaxim, line 54: add XX:+UseCGroupMemoryLimitForHeap option for JVM to enable correct cgroups enabling for Java application inside docker. Check this [article](https://developers.redhat.com/blog/2017/03/14/java-inside-docker/)

modules/ducktests/tests/ignitetest/tests/benchmarks/add_node_rebalance_test.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def setUp(self):
+        pass
+
+    def teardown(self):
+        pass
+
+    @cluster(num_nodes=NUM_NODES + 1)
+    @parametrize(version=str(DEV_BRANCH))
+    @parametrize(version=str(V_2_8_0))
```
timoninmaxim, line 53: Versions should be moved to config. Adding new versions (or ISP versions) will lead to change multiple files just to update version.

modules/ducktests/tests/docker/Dockerfile
```diff
@@ -48,14 +48,13 @@ ARG APACHE_MIRROR="https://apache-mirror.rbc.ru/pub/apache/"
...
+
+ADD $APACHE_ARCHIVE/ignite/2.8.0/apache-ignite-2.8.0-bin.zip /opt/
```
timoninmaxim, line 54: Versions and distributive sources should be configured externally: easy add new versions, custom versions from custom sources, without changing dockerfile. 

Possible just try use different images for different versions (to save time on prepare tests to start)

nizhikov: > Versions and distributive sources should be configured externally

Please, clarify, what issues do you see with the approach when we have all the required versions in this `Dockerfile`?
If we have to use a different image with some custom Ignite assembly, we, probably, should have script parameters to specify other `Dockerfile` to build image from.

modules/ducktests/tests/ignitetest/tests/benchmarks/add_node_rebalance_test.py
nizhikov: We want to build tests as stable as we can. 
So the explicit version in test parameters is a thoughtful choice.
Kafka uses approach with the `LATEST_2_8` or `LATEST_2_9` to change all tests to the new minor versions.

But with every major version, one should update tests.

modules/ducktests/tests/docker/ducker-ignite
```diff
@@ -258,7 +258,7 @@ docker_run() {
     must_do -v docker run --privileged \
         -d -t -h "${node}" --network ducknet "${expose_ports}" \
         --memory=${docker_run_memory_limit} --memory-swappiness=1 \
-        -v "${ignite_dir}:/opt/ignite-dev" --name "${node}" -- "${image_name}"
+        -v "${ignite_dir}:/opt/ignite-dev:delegated" --name "${node}" -- "${image_name}"
```
timoninmaxim, line 261: _delegated_ could be misleading. Containers are reused for running tests with ./run_tests.sh script. But docker docs says:

> If changes to the mount source directory are present on the host file system, those changes may be lost when the delegated mount synchronizes with the host source directory.

Check this [article](https://docs.docker.com/docker-for-mac/osxfs-caching/)

timoninmaxim: Possible solution is to force clean up all containers after tests have finished

modules/ducktests/tests/docker/Dockerfile
timoninmaxim: I see at least 3 issues there:

1. Some versions could be incompatible with each other. For example, one version could work on java11 without issues, while others are not (there is also issue with our custom build that works only on patched java). Accidentally there could be dependencies on different system libs, envs, directories, etc.

I think its a good practice do not mix different versions in the same env and prepare a separate image for every version. 

2. Also there should be images like ignite-X.Y.Z-java8, ignite-X.Y.Z-java11, ignite-X.Y.Z-alpine, etc. I think we can reuse this common images in tests rather than creating uber image just for tests.

3. I'm trying to write some ISE specific tests with enabled custom plugins. Also I want to run existing tests (written for original Ignite) for the custom version directly from the ducktape directory. Also I want compare multiple versions (custom and original) with each other.

I suppose that I can flexible configure my test just specify required versions rather than re-create new image for my every case.

nizhikov: > For example, one version could work on java11 without issues, while others are not

This knowledge should be put in the test itself. Kafka uses the same approach.

> Accidentally there could be dependencies on different system libs, envs, directories, etc.

All required dependencies exist in zip distribution.
Directories structure doesn't change frequently so we can put this knowledge into the service itself.

> I think its a good practice do not mix different versions

What is wrong with the different versions?
We want to run the same tests on the different versions.

> Also there should be images like ignite-X.Y.Z-java8, ignite-X.Y.Z-java11, ignite-X.Y.Z-alpine

1. We don't support alpine.
2. Yes, we can parametrize the creation of the image with the JDK version. Initial PR doesn't contain this parameter for the sake of simplicity.

> I'm trying to write some ISE specific tests with enabled custom plugins
> rather than re-create new image for my every case.

You should create different docker image to be able to use custom Ignite assembly.
You don't need several images, just one image that satisfies your requirements.

timoninmaxim: > All required dependencies exist in zip distribution.
Actually this is not true as at least Spark set up separately. And there could be issues with different versions of Ignite and Spark.

With storing dependencies in .zip you just reinvent containers without using any of their advantages. Now you need to support multiple applications in one container. But you can't start all of them in one container as you need to orchestrate apps by self: set LD_LIBRARY_PATH for different system libs, set JAVA_HOME for different jvms, use different networks, different mount directories, etc.

If you move JDK as param for container why do not do the same for Ignite version? It looks inconsistent now.
> Kafka uses the same approach
Kafka may have a motivation for that. To do the same one need clearly understand reasons behind this approach. Could you please provide discussion or design doc that describes this solution.

> We don't support alpine.
It's ok, it was just an example that there are should be different public official images that can be reused. I'm repeating there we could use public Ignite images to use them in tests instead of creating uber docker image just for tests. 

> You don't need several images, just one image that satisfies your requirements
Containers are building blocks. A container provides _a process_ and it is not a replacement of physical environment. There are additional limitations if try use containers as envs. For example in current dockerfile at least 2 problems that could lead to wasting time:
1. _delegated_ volumes I've mentioned below (check this [comment](https://github.com/apache/ignite/pull/7967#discussion_r449182086));
2. Using 'tail -f' as init process could lead to wasting Linux resources (Check this [article](https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/)).

nizhikov: > Actually this is not true as at least Spark set up separately. And there could be issues with different versions of Ignite and Spark.

We just should prepare all Spark or Ignite versions we want to tests.
I can't see any issues with this.

nizhikov: > If you move JDK as param for container why do not do the same for Ignite version? It looks inconsistent now.

Because I think we don't want to increase the complexity of the final solution with dozens of different images.

nizhikov: > It's ok, it was just an example that there are should be different public official images that can be reused

For now, we don't have those images.
Moreover, we want to use the same tests with the different deployment types - AWS, private cloud, etc.
If we go with the dozens of images we should provide a dozen scripts for each deployment type, also.

modules/ducktests/tests/docker/Dockerfile
anton-vinogradov: Done

modules/ducktests/tests/docker/Dockerfile
anton-vinogradov: Done

modules/ducktests/tests/ignitetest/tests/benchmarks/add_node_rebalance_test.py
anton-vinogradov: Agree with @nizhikov.
Fixed code to use Latest.

modules/ducktests/tests/docker/Dockerfile
anton-vinogradov: Lets, restart this thread at separate PR if necessary?

modules/ducktests/tests/docker/ducker-ignite
anton-vinogradov: Seems, we have to find a good fix here.
Left's restart this as a separate PR.

modules/ducktests/tests/ignitetest/services/ignite.py
anton-vinogradov: Done

modules/ducktests/tests/docker/ducker-ignite
mshonichev: is this MacOs only issue?

timoninmaxim: Hi! delegated hack is used only for Ignite code. The same time work directory is located under `/mnt/service` directory that is not a volume. Also docker uses different filesystem for non-volumes (by default it is overlayfs) than bare metal. This FS allows native performance but in some circumstances performance could be different (check this [note](https://docs.docker.com/storage/storagedriver/overlayfs-driver/#performance-best-practices)).

So one should not make assumptions about performance based on results in docker only. But this more about heavy benchmarks. Testing in docker helps to find programming issues that visibly slow down some operations.

timoninmaxim: Also, I've found that there is kibosh used in Dockerfile. I'm not confident how exactly it works, may be there are some penalties too (actually I don't know).

modules/ducktests/tests/docker/Dockerfile
```diff
@@ -42,28 +42,23 @@ COPY ./ssh-config /root/.ssh/config
...
+RUN cd /opt && curl -O $APACHE_ARCHIVE/ignite/2.7.6/apache-ignite-2.7.6-bin.zip && unzip apache-ignite-2.7.6-bin.zip && mv /opt/apache-ignite-2.7.6-bin /opt/ignite-2.7.6
+RUN cd /opt && curl -O $APACHE_ARCHIVE/ignite/2.8.0/apache-ignite-2.8.0-bin.zip && unzip apache-ignite-2.8.0-bin.zip && mv /opt/apache-ignite-2.8.0-bin /opt/ignite-2.8.0
+RUN cd /opt && curl -O $APACHE_ARCHIVE/ignite/2.8.1/apache-ignite-2.8.1-bin.zip && unzip apache-ignite-2.8.1-bin.zip && mv /opt/apache-ignite-2.8.1-bin /opt/ignite-2.8.1
 
-ADD $APACHE_ARCHIVE/ignite/$IGNITE_VERSION/$IGNITE_RELEASE_NAME.zip /opt/
-RUN cd /opt && unzip $IGNITE_RELEASE_NAME.zip && rm $IGNITE_RELEASE_NAME.zip
-RUN mv /opt/$IGNITE_RELEASE_NAME /opt/$IGNITE_NAME
-RUN chmod a+wr /opt/$IGNITE_NAME -R
+RUN rm /opt/apache-ignite-*-bin.zip
+RUN chmod a+wr /opt -R
```
mshonichev, line 54: That's just create two additional full-archive copy layers which increase image size [1] and may slow down file deletion operations due to layered nature of docker image [2]. 
```
$ docker images
REPOSITORY                TAG                 IMAGE ID            CREATED             SIZE
ducker-ignite-openjdk-8   latest              eb0bc05f45f0        10 minutes ago      4.31GB

ducker@ducker01:/$ sudo du -hs / 2>/dev/null
3.6G	/

```

If you absolutely need writable ignite dir, you could either glue RUN commands together, or you may set up a bunch of symlinks:
```
/opt/apache-ignite-2.8.0/bin ... original archive
/opt/ignite-2.8.0/ - real directory
/opt/ignite-2.8.0/bin -> symlink to /opt/apache-ignite-2.8.0/bin
...
```
---
[1] - https://www.datawire.io/not-engineer-running-3-5gb-docker-images/
[2] - https://docs.docker.com/storage/storagedriver/

anton-vinogradov: @mshonichev, 
AFAIU, the question about the necessity of "RUN chmod a+wr /opt -R".
Seems, it can be safely removed since other folders used for write access.
Removed this.

PR #7987, opened 2020-07-02
IGNITE-13109 Skip distributed metastorage entries that can not be unmarshalled

NSAmelchev:
Introduced the property that allows to skip distributed metastorage entries that can be unmarshalled when migrating from the previous cluster persistence data.

Entries skipped for:
- notify listeners by key prefix on startup
- node join procedure (values will not be validated)
- iterate with entry key prefix


modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1342,6 +1343,15 @@
      */
     public static final String IGNITE_DUMP_TX_COLLISIONS_INTERVAL = "IGNITE_DUMP_TX_COLLISIONS_INTERVAL";
 
+    /**
+     * Newline separated list of keys that {@link DistributedMetaStorage} will skip for any operations:
```
nizhikov, line 1347: Why do we want to use a newline to separate list items?
It may be inconvenient to specify a multiline value in the bash script.
Let's use a comma as a separator like it done in `IGNITE_TCP_DISCOVERY_ADDRESSES`

modules/core/src/main/java/org/apache/ignite/internal/processors/metastorage/persistence/DistributedMetaStorageImpl.java
```diff
@@ -406,6 +432,9 @@ private void onMetaStorageReadyForWrite(ReadWriteMetastorage metastorage) {
 
     /** {@inheritDoc} */
     @Override @Nullable public <T extends Serializable> T read(@NotNull String key) throws IgniteCheckedException {
+        if (keysToSkip.contains(key))
```
nizhikov, line 435: Why do we disallow `read` operation?
All we need is skip keys that can't be unmarshalled on the node startup.

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
NSAmelchev: Fixed with comma

modules/core/src/main/java/org/apache/ignite/internal/processors/metastorage/persistence/DistributedMetaStorageImpl.java
NSAmelchev: Fixed. Explicit operations will not be ignored

PR #7992, opened 2020-07-05
IGNITE-7369 .NET: Thin Client Transactions

gurustron:
Implemented  Thin Client Transactions


modules/platforms/dotnet/Apache.Ignite.Core/Client/Transactions/IClientTransactions.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace Apache.Ignite.Core.Client.Transactions
+{
+    using System;
+    using Apache.Ignite.Core.Transactions;
+
+    /// <summary>
+    /// Ignite Thin Client transactions facade.
+    /// </summary>
+    public interface IClientTransactions
```
ptupitsyn, line 26: Let's rename to `ITransactionsClient` to be consistent with `ICacheClient`, `IComputeClient`.

modules/platforms/dotnet/Apache.Ignite.Core/Client/Transactions/IClientTransaction.cs
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+namespace Apache.Ignite.Core.Client.Transactions
+{
+    using System;
+
+    /// <summary>
+    /// Thin client transaction.
+    /// </summary>
+    public interface IClientTransaction : IDisposable
```
ptupitsyn, line 25: Let's rename to `ITransactionClient` to be consistent with `ICacheClient`, `IComputeClient`.

modules/platforms/dotnet/Apache.Ignite.Core/Client/IIgniteClient.cs
```diff
@@ -111,6 +112,11 @@ public interface IIgniteClient : IDisposable
         [SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate", Justification = "Semantics.")]
         IBinary GetBinary();
 
+        /// <summary>
+        /// Gets Ignite transactions facade.
+        /// </summary>
+        IClientTransactions Transactions { get; }
```
ptupitsyn, line 118: Please convert to a method for consistency with `IIgnite.GetTransactions`.
Method is also a more future-proof option (see https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229054(v=vs.100)).

modules/platforms/dotnet/Apache.Ignite.Core/Client/Transactions/IClientTransactions.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// Starts new transaction with the specified concurrency, isolation and timeout.
+        /// </summary>
+        /// <param name="concurrency">Concurrency.</param>
+        /// <param name="isolation">Isolation.</param>
+        /// <param name="timeout">Timeout. TimeSpan. Zero for indefinite timeout.</param>
+        /// <returns>New transaction.</returns>
+        IClientTransaction TxStart(TransactionConcurrency concurrency, TransactionIsolation isolation, TimeSpan timeout);
+
+        /// <summary>
+        /// Returns instance of <see cref="IClientTransactions"/>> to mark a transaction with a special label.
```
ptupitsyn, line 52: Please add more explanation - why is this useful, how to retrieve this label

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Cache/CacheClient.cs
```diff
@@ -754,14 +797,26 @@ private void WriteRequest(Action<ClientRequestContext> writeAction, ClientReques
...
+
+            var tx = _ignite.Transactions.CurrentTx;
+            if (tx != null)
+            {
+                flags |= ClientCacheRequestFlag.WithTransactional;
+            }
+
+            ctx.Stream.WriteByte((byte) flags);
+
+            if (flags.HasFlag(ClientCacheRequestFlag.WithExpiryPolicy))
```
ptupitsyn, line 815: `HasFlag` is [slow on older frameworks](https://www.code4it.dev/blog/hasflag-performance-benchmarkdotnet), let's avoid it here.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/ClientFailoverSocket.cs
```diff
@@ -269,6 +278,12 @@ private ClientSocket GetAffinitySocket<TKey>(int cacheId, TKey key)
                 return null;
             }
 
+            // Transactional operation should be executed on node started the transaction.
+            if (_transactions.CurrentTx != null)
```
ptupitsyn, line 282: To be precise: transactional operation must be executed using the same connection (same `ClientSocket` instance) that started the transaction.

And the check here does not guarantee this at all - reconnect could have happened, for example.
A proper approach is to store get the socket from `ClientContextBase.Socket` property when executing `ClientOp.TxStart`, and use that socket directly for transactional operations.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Transactions/ClientTransaction.cs
```diff
@@ -0,0 +1,125 @@
...
+        private readonly int _id;
+
+        /** Ignite. */
+        private readonly IgniteClient _ignite;
+
+        /* Transactions. */
+        private readonly ClientTransactions _transactions;
+
+        /** Transaction is closed. */
+        private volatile bool _closed;
```
ptupitsyn, line 38: Do we need `volatile`? Transactions are tied to a specific thread, right?

modules/platforms/dotnet/Apache.Ignite.Core/Client/Transactions/IClientTransactions.cs
gurustron: I think it can be better compared to `IClientCluster` , `IClientClusterNode` and so on. To me `ITransactionsClient` sounds a little bit misleading and clumsy.

ptupitsyn: Actually, `IClientCluster` was missed during the review, so we have an inconsistency now, but I'd like to avoid further inconsistencies.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Transactions/ClientTransaction.cs
gurustron: In theory transaction returned by `TxStart` can be closed from any thread.

ptupitsyn: Usually, types and APIs are not thread safe, unless stated otherwise. E.g. we don't expect `List<T>` to be thread safe and have volatile fields. We explicitly state that Ignite transactions are per-thread, so we don't need to guarantee thread safety.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Transactions/ClientCacheTransactionManager.cs
```diff
@@ -0,0 +1,154 @@
...
+        {
+            if (_transactions.CurrentTx != null)
+            {
+                // Ignite transaction is already present.
+                // We have either enlisted it already, or it has been started manually and should not be enlisted.
+                // Java enlists existing Ignite tx in this case (see CacheJtaManager.java), but we do not.
+                return;
+            }
+
+            if (_enlistment.Value != null)
```
ptupitsyn, line 61: This condition seems to be unnecessary, please see my changes in #8004 .

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Transactions/ClientCacheTransactionManager.cs
```diff
@@ -0,0 +1,154 @@
...
+            {
+                _transactions.TxStart(_transactions.DefaultTxConcurrency,
+                    CacheTransactionManager.ConvertTransactionIsolation(ambientTx.IsolationLevel),
+                    _transactions.DefaultTimeout);
+
+                _enlistment.Value = ambientTx.EnlistVolatile(this, EnlistmentOptions.None);
+            }
+        }
+
+        void IEnlistmentNotification.Prepare(PreparingEnlistment preparingEnlistment)
```
ptupitsyn, line 83: Missing xmldoc

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Transactions/ITransactionClientInternal.cs
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Apache.Ignite.Core.Impl.Client.Transactions
+{
+    using Apache.Ignite.Core.Client.Transactions;
+
+    internal interface ITransactionClientInternal : ITransactionClient
```
ptupitsyn, line 22: xmldoc

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Transactions/ITransactionsClientInternal.cs
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+
+namespace Apache.Ignite.Core.Impl.Client.Transactions
+{
+    using System;
+    using Apache.Ignite.Core.Client.Transactions;
+    using Apache.Ignite.Core.Transactions;
+
+    internal interface ITransactionsClientInternal : ITransactionsClient
```
ptupitsyn, line 24: xmldoc

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Transactions/TransactionsClient.cs
```diff
@@ -0,0 +1,249 @@
...
+                    ctx.Writer.WriteString(label);
+                },
+                ctx => new TransactionClient(
+                    ctx.Reader.ReadInt(),
+                    _ignite,
+                    ctx.Socket)
+            );
+
+            _currentTx.Value = tx;
+            return _currentTx.Value;
```
ptupitsyn, line 150: `return tx`

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Transactions/TransactionsClient.cs
```diff
@@ -0,0 +1,249 @@
...
+        public void Dispose()
+        {
+            _currentTx.Dispose();
+            _txManager.Dispose();
+        }
+
+        /// <summary>
+        /// Wrapper for transactions with label.
+        /// </summary>
+        private class TransactionsClientWithLabel : ITransactionsClientInternal
```
ptupitsyn, line 174: I think it will be simpler to incorporate label into the `TransactionsClient` class.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Transactions/TransactionsClient.cs
```diff
@@ -0,0 +1,249 @@
...
+
+        /** <inheritDoc /> */
+        public ITransactionClient TxStart(TransactionConcurrency concurrency, 
+            TransactionIsolation isolation,
+            TimeSpan timeout)
+        {
+            return TxStart(concurrency, isolation, timeout, null);
+        }
+
+        private ITransactionClient TxStart(TransactionConcurrency concurrency,
```
ptupitsyn, line 124: XMLDoc

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Client/Cache/CachePartitionedTxTest.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                Assert.AreEqual(10, cache.Get(key1));
+                Assert.AreEqual(20, cache.Get(key2));
+            }
+
+            Assert.AreEqual(1, cache.Get(key1));
+            Assert.AreEqual(2, cache.Get(key2));
+        }
+
+        protected override string GetCacheName()
```
ptupitsyn, line 96: inheritdoc

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/ClientFailoverSocket.cs
```diff
@@ -247,6 +256,16 @@ public IEnumerable<IClientConnection> GetConnections()
         /// </summary>
         private ClientSocket GetSocket()
         {
+            var tx = _transactions.CurrentTx;
+            if (tx != null)
+            {
+                if (tx.Socket.IsDisposed)
+                {
+                    throw new IgniteClientException("Transaction context has been lost due to connection errors.");
```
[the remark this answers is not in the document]
gurustron: I've added one. Calling dispose o tx should take care of it.

modules/platforms/dotnet/Apache.Ignite.Core/Client/Transactions/TransactionClientConfiguration.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// Gets or sets the cache transaction timeout to use when one is not explicitly specified.
+        /// <see cref="TimeSpan.Zero"/> for infinite timeout.
+        /// </summary>
+        [DefaultValue(typeof(TimeSpan), "00:00:00")]
+        public TimeSpan DefaultTimeout { get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TransactionClientConfiguration" /> class.
+        /// </summary>
+        public TransactionClientConfiguration()
```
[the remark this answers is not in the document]
gurustron: 1.  IgniteConfigurationTest.TestDefaultValueAttributes:103 for default ctor. Will add one for non-default one
2.  CacheClientAbstractTxTest.TestClientTransactionConfiguration would not be sufficient?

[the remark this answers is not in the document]
ptupitsyn: Yep that looks good, thanks.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Transactions/TransactionClient.cs
```diff
@@ -0,0 +1,172 @@
...
+        /// <summary>
+        /// Closes the transaction. 
+        /// </summary>
+        private void Close(bool commit)
+        {
+            if (!_closed)
+            {
+                try
+                {
+                    _ignite.Socket.DoOutInOp<object>(ClientOp.TxEnd,
```
ptupitsyn, line 138: 1. `_socket` should be used instead of `_ignite.Socket`
2. Make sure we do not throw exceptions from the finalizer and `Dispose`

modules/platforms/dotnet/Apache.Ignite.Core/Client/Transactions/ITransactionClient.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// Commits this transaction.
+        /// </summary>
+        void Commit();
+
+        /// <summary>
+        /// Rolls back this transaction.
+        /// </summary>
+        void Rollback();
+        
+        TransactionConcurrency Concurrency { get; }
```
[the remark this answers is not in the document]
gurustron: Fixed.

modules/platforms/dotnet/Apache.Ignite.Core/Client/Transactions/ITransactionClient.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// </summary>
+        void Rollback();
+        
+        TransactionConcurrency Concurrency { get; } 
+        
+        TransactionIsolation Isolation { get; }
+
+        TimeSpan Timeout { get; }
+
+        string Label { get; }
```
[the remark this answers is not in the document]
gurustron: Added to `TestWithLabel`

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Transactions/TransactionClient.cs
gurustron: 1. Fixed.
2. What exceptions we shoud cover?

ptupitsyn: I think the only situation to be covered here is a socket disconnect.
We can catch `Exception`, and swallow it only if `_socket` is disconnected, otherwise `throw`.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Transactions/TransactionsClient.cs
```diff
@@ -0,0 +1,257 @@
...
+        internal TransactionClient Tx
+        {
+            get
+            {
+                var tx = _currentTx.Value;
+
+                if (tx == null)
+                    return null;
+
+                if (tx.Closed)
```
ptupitsyn, line 93: We should handle an edge case when multiple clients are used from the same thread:
check that `tx` actually belongs to the current `IgniteClient` and has the same `ClientSocket`.

If there is an active transaction for another client or socket, I think we should throw  `NotSupportedException`.

ptupitsyn (changes requested):

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Transactions/TransactionsClient.cs
ptupitsyn: Also, when socket is disconnected, we should clear all related thread locals to avoid leading memory.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Client/Cache/CacheClientAbstractTxTest.cs
```diff
@@ -0,0 +1,822 @@
...
+        /// </summary>
+        [Test]
+        public void TestDifferentClientsCanStartTransactions()
+        {
+            Assert.DoesNotThrow(() =>
+            {
+                using (Client.GetTransactions().TxStart())
+                using (GetClient().GetTransactions().TxStart())
+                {
+                    // No-op.
```
ptupitsyn, line 298: We should check that operations can be performed and that one client does not use the TX started by the other.
I think this is not the case, since we only have one thread local which is shared by all clients.

gurustron: Added test.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Transactions/TransactionClient.cs
gurustron: Fixed.

ptupitsyn (approved):

PR #8004, opened 2020-07-07
IGNITE-13214 .NET: Fix TransactionScope for read operations

ptupitsyn:
* Add `StartTxIfNeeded` to all cache operations (it was missing for Get, GetAll, TryGet)
* Add `OptimisticTransactionTest` with explicit and ambient tests
* Add exception mapping for `IgniteTxOptimisticCheckedException`
* Fix error handling in ambient transactions - close Ignite tx on Prepare failure


modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/CacheAbstractTransactionalTest.cs
```diff
@@ -862,6 +862,50 @@ public void TestTransactionScopeAllOperations()
...
+        /// <summary>
+        /// Tests that read operations lock keys in Serializable mode.
+        /// </summary>
+        [Test]
+        public void TestTransactionScopeWithSerializableIsolationLocksKeysOnRead()
+        {
+            Action<Func<ICache<int, int>, int, int>>
+                test = TestTransactionScopeWithSerializableIsolationLocksKeysOnRead;
+
+            test((cache, key) => cache[key]);
```
gurustron, line 874: Can ```ContainsKey``` and ```ContainsKeys``` trigger transnational behavior?

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/CacheAbstractTransactionalTest.cs
```diff
@@ -862,6 +862,50 @@ public void TestTransactionScopeAllOperations()
...
+            cache.Put(1, 1);
+
+            var options = new TransactionOptions {IsolationLevel = IsolationLevel.Serializable};
+
+            using (var scope = new TransactionScope(TransactionScopeOption.Required, options))
+            {
+                Assert.AreEqual(1, readOp(cache, 1));
+
+                var taskFinished = Task.Factory.StartNew(() => cache.Put(1, 2)).Wait(TimeSpan.FromSeconds(1));
+
```
gurustron, line 899: To be honest I prefer check the optimistic serializable transaction for throwing exception (as defined in [documetation](https://apacheignite.readme.io/docs/concurrency-modes-and-isolation-levels#optimistic-transactions)) on commit it is seems to me to be more predictable and no need for magic numbers and waits.

gurustron: Or just check that transaction is started.

gurustron: Or just check that transaction is started.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/CacheAbstractTransactionalTest.cs
ptupitsyn: Double checked this - no, `ContainsKey` and `ContainsKeys` do not participate in transactions.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/CacheAbstractTransactionalTest.cs
ptupitsyn: Good point, I've added current Tx check, and replaced time-based wait with a wait handle.
But I'd like to keep the logic that checks whether keys are locked.

Optimistic transactions are a different matter, we can have a separate test for them.

ptupitsyn: I've added more tests for Optimistic transactions, please have a look.

PR #8039, opened 2020-07-14
IGNITE-13257 update kubernetes plain example

timoninmaxim:
Fix issues while start kubernetes plain example

### DIFF
--- a/modules/kubernetes/DEVNOTES.txt
+++ b/modules/kubernetes/DEVNOTES.txt
@@ -9,8 +9,7 @@ feed the image to your kubernetes environment.
 Building Apache Ignite
 =========================
 
-Use the command below to assemble an Apache Ignite binary:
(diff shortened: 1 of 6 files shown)


modules/kubernetes/DEVNOTES.txt
```diff
@@ -23,7 +22,7 @@ Once this is done, make sure that Minikube sees Docker images registered locally
     eval $(minikube docker-env)
 
 Start Minikube:
-    minikube start --vm-driver=xhyve
+    minikube start
```
dmagda, line 25: What's your operating system? It was required to pass in "--vm-driver=xhyve" parameter on my Mac OS laptop. Probably, we need to mention that this parameter has to be added on that operating system.

timoninmaxim: I use MacOS too, default provider works fine

dmagda:
Thanks for the contribution, I've merged the changes to the master branch!

PR #8109, opened 2020-07-31
Json params support

anton-vinogradov:


modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/tests/DataGenerationApplication.java
```diff
@@ -34,13 +35,13 @@ public DataGenerationApplication(Ignite ignite) {
     }
 
     /** {@inheritDoc} */
-    @Override protected void run(String[] args) {
+    @Override protected void run(Map<String, String> args) {
```
timoninmaxim, line 38: Why do you use Map instead of fair JsonNode object? Now you're limited with flat structure and user responsible to cast value types.

modules/ducktests/tests/ignitetest/services/utils/ignite_aware_app.py
```diff
@@ -82,7 +84,7 @@ def app_args(self):
         args = self.java_class_name + "," + IgniteAwareApplicationService.CONFIG_FILE
 
         if self.params != "":
-            args += "," + self.params
+            args += "," + str(base64.b64encode(json.dumps(self.params).encode("UTF-8")))
```
timoninmaxim, line 87: Why do you use base64 there? Is it a workaround to avoid escape issue?

modules/ducktests/tests/ignitetest/services/utils/ignite_aware_app.py
```diff
@@ -106,7 +108,23 @@ def env(self):
         return "export MAIN_CLASS={main_class}; ".format(main_class=self.servicejava_class_name) + \
                "export EXCLUDE_TEST_CLASSES=true; " + \
                "export IGNITE_LOG_DIR={log_dir}; ".format(log_dir=self.PERSISTENT_ROOT) + \
-               "export USER_LIBS=%s:/opt/ignite-dev/modules/ducktests/target/*; " % self.user_libs
+               "export USER_LIBS=%s:/opt/ignite-dev/modules/ducktests/target/*%s; " % \
```
timoninmaxim, line 111: Looks like user_libs and service_libs should be lists and shouldn't know about colons delimeter. As colon is a part of USER_LIBS, now we have usage of the delimiter in 3 places (user_libs, service_libs, USER_LIBS)

modules/ducktests/tests/ignitetest/services/utils/ignite_path.py
```diff
@@ -56,3 +56,9 @@ def script(self, script_name):
         :return: absolute path to the specified script
         """
         return os.path.join(self._home, "bin", script_name)
+
+    def home(self):
```
timoninmaxim, line 60: just rename _home to home and use it

modules/ducktests/tests/ignitetest/services/utils/ignite_aware_app.py
```diff
@@ -106,7 +108,23 @@ def env(self):
...
+        """
+        :return: Libs required to start IgniteAwareApplication java implementation.
+        """
+        if self.version.is_dev:
+            return ""
+
+        libs = ""
+
+        for line in self.nodes[0].account.ssh_capture(
+                "ls -d %s/libs/optional/ignite-aws/* | grep jackson | tr '\n' ':' | sed 's/.$//'" % self.path.home()):
```
timoninmaxim, line 124: Why do you extract jackson jars from the AWS library? As I can see AWS isn't used in the application

modules/ducktests/tests/ignitetest/services/utils/ignite_aware_app.py
ivandasch: Yep

modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/tests/DataGenerationApplication.java
ivandasch: +1 JsonObject seems to be more universal solution and has a potential to extend

modules/ducktests/tests/ignitetest/services/utils/ignite_aware_app.py
anton-vinogradov: Yes

modules/ducktests/tests/ignitetest/services/utils/ignite_aware_app.py
anton-vinogradov: Because we should gain it somewhere, but it always presented as a dependency and always have the same version for all cases (set inside parent.pom)

timoninmaxim: Why do not keep them in pom.xml of ducktests module?

modules/ducktests/tests/ignitetest/services/utils/ignite_aware_app.py
timoninmaxim: Just use single quotes then. This is enough, check how run_tests.sh do the same:
https://github.com/apache/ignite/blob/ignite-ducktape/modules/ducktests/tests/docker/run_tests.sh#L136

ivandasch: Max, I suppose that this is more robust solution to problem. Constantly escape everything is a little bit tiresome. base64 solves this problem radically.

anton-vinogradov: For example, JSON contains "," already used as a separator at ignite_aware.py because of necessary to transfer some params where only one allowed.

modules/ducktests/tests/ignitetest/services/utils/ignite_aware_app.py
anton-vinogradov: For every old version?

modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/tests/DataGenerationApplication.java
anton-vinogradov: Done

modules/ducktests/tests/ignitetest/services/utils/ignite_path.py
anton-vinogradov: Done

modules/ducktests/tests/ignitetest/services/utils/ignite_aware_app.py
timoninmaxim: ok, understand.

modules/ducktests/tests/ignitetest/services/utils/ignite_aware_app.py
timoninmaxim: Oh, now I understand. But do we need use strict link to AWS project, as we can't guarantee that it wouldn't be changed in future? Could we just use the `find` util to locate jackson jars anywhere in optional libs?

timoninmaxim: Also there must be a comment with description of solution

modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/tests/DataGenerationApplication.java
```diff
@@ -34,13 +35,13 @@ public DataGenerationApplication(Ignite ignite) {
...
-    @Override protected void run(String[] args) {
+    @Override protected void run(JsonNode jsonNode) {
         log.info("Creating cache...");
 
-        IgniteCache<Integer, Integer> cache = ignite.createCache(args[0]);
+        IgniteCache<Integer, Integer> cache = ignite.createCache(jsonNode.get("cacheName").asText());
 
         try (IgniteDataStreamer<Integer, Integer> stmr = ignite.dataStreamer(cache.getName())) {
-            for (int i = 0; i < Integer.parseInt(args[1]); i++) {
+            for (int i = 0; i < jsonNode.get("range").asInt(); i++) {
```
timoninmaxim, line 44: Is it possible to use a DTO object instead? Generics is more clear solution

modules/ducktests/tests/ignitetest/services/utils/ignite_aware_app.py
anton-vinogradov: Comment added. Can we postpone ideal resolving (locate jackson jars anywhere in optional libs) and just use AWS as source for now?

modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/tests/DataGenerationApplication.java
anton-vinogradov: It will be unique for every app, this will lead to useless classes?

timoninmaxim: Additional classes are ok. In case of complex hierarchy I'd prefer use generics (for example smth like `obj.getList().get(0).getRange()`) instead of ugly constructions (smth like `node.get("list").asArray()[0].get("range").asInt()`). 

DTOs make code much much clearer, also it provides sanity checks out of the box.

modules/ducktests/tests/ignitetest/services/utils/ignite_aware_app.py
timoninmaxim: Ok, I'd think it should be part of infrastructure spec.

modules/ducktests/tests/ignitetest/services/utils/ignite_aware_app.py
anton-vinogradov: Done

modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/tests/DataGenerationApplication.java
anton-vinogradov: Not sure I've got an idea.
Could you provide an example related to the current tests?

timoninmaxim: Oh, I see. I thought that JsonNode is a generic, but it's not. So, there is a choose between plain JsonNode and DTO. In favor of latter:
1.  Currently there are not, but in one day could be complex json with nested objects and arrays. User will have to use `mapper.readValue(jsonNode, DTO.class)` if need it. 
2. If you want change code of your app in one day and change params, there won't help from IDE for you. There is a possibility of NPE in jsonNode.get(). DTO provides checks for such situations (code fails in compile time).
3. All tests have own package, so I don't see problems with additional classes that are isolated within a package.

anton-vinogradov: How to share java DTO to python? Seems we should add and sync DTO couple for each application. 
How to sync this automatically?
Seem overcomplicated...?

ivandasch: I think, that we should make in the bright future a generic application, that accepts broad range of generic options to load data to cache. For example, cache name, key-value schema (binary object), range for keys, number of keys, etc.

For this argument DTO is a preferrable solution, I suppose. But we can implement this a little bit later as a separate task.

timoninmaxim: There still python, shame :( There are different approaches. Python just use dict and user is responsible for everything that is running in python code. Java could provide more strict checks. 

Ok, agree, not an issue, as python won't directly use DTO object.

anton-vinogradov: Great, let's re-discuss this later

modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/utils/IgniteAwareApplicationService.java
```diff
@@ -40,8 +47,15 @@ public static void main(String[] args) throws Exception {
 
         IgniteAwareApplication app = (IgniteAwareApplication)clazz.getConstructor().newInstance();
 
-        String[] appParams = Arrays.copyOfRange(params, 1, params.length);
+        ObjectMapper mapper = new ObjectMapper();
+
+        ObjectReader reader = mapper.readerFor(Map.class);
```
ivandasch, line 52: Unused varoiable, I suppose

anton-vinogradov: Which one?

anton-vinogradov: "reader" makes next line more readable ))))

ivandasch: mapper.readTree(new String(Base64.getDecoder().decode(params[2]), UTF_8))

anton-vinogradov: Fixed

PR #8110, opened 2020-08-01
IGNITE-13293 .NET: Fix enum serialization performance

ptupitsyn:
* Don't send enum metadata more than once (main part of this fix - see `BinaryWriter` changes)
* Cache name-value map per enum type
* Add benchmark

As a result, enum field serialization is on par with int field:

```
|                Method |     Mean |    Error |   StdDev | Ratio | RatioSD |
|---------------------- |---------:|---------:|---------:|------:|--------:|
|          PutPrimitive | 45.47 us | 0.907 us | 1.330 us |  1.00 |    0.00 |
|     PutPrimitiveAsync | 66.28 us | 1.317 us | 1.617 us |  1.45 |    0.05 |
|  PutClassWithIntField | 68.34 us | 1.275 us | 1.130 us |  1.49 |    0.05 |
| PutClassWithEnumField | 71.13 us | 1.419 us | 2.734 us |  1.57 |    0.08 |
```

### DIFF
--- a/modules/platforms/dotnet/Apache.Ignite.Core.Tests/Binary/EnumsTest.cs
+++ b/modules/platforms/dotnet/Apache.Ignite.Core.Tests/Binary/EnumsTest.cs
@@ -19,6 +19,7 @@
 namespace Apache.Ignite.Core.Tests.Binary
 {
     using System;
+    using System.Linq;
     using System.Runtime.Serialization;
     using Apache.Ignite.Core.Binary;
     using Apache.Ignite.Core.Impl;
@@ -85,6 +86,11 @@ namespace Apache.Ignite.Core.Tests.Binary
                 {
                     Assert.AreEqual(string.Format("{0} [typeId={1}, enumValue={2}, enumValueName={3}]",
                         typeof(T).FullName, binRes.GetBinaryType().TypeId, binRes.EnumValue, val), binRes.ToString());
+
+                    var expectedEnumNames = Enum.GetValues(typeof(T)).OfType<T>().Select(x => x.ToString()).ToList();
+                    var actualEnumNames = binRes.GetBinaryType().GetEnumValues().Select(v => v.EnumName).ToList();
+                    
+                    CollectionAssert.AreEquivalent(expectedEnumNames, actualEnumNames);
                 }
                 else
                 {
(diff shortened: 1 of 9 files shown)


modules/platforms/dotnet/Apache.Ignite.Core.Tests/Binary/EnumsTest.cs
```diff
@@ -85,6 +86,11 @@ private static void CheckValue<T>(T val, bool isBinaryEnum = true)
                 {
                     Assert.AreEqual(string.Format("{0} [typeId={1}, enumValue={2}, enumValueName={3}]",
                         typeof(T).FullName, binRes.GetBinaryType().TypeId, binRes.EnumValue, val), binRes.ToString());
+
+                    var expectedEnumNames = Enum.GetValues(typeof(T)).OfType<T>().Select(x => x.ToString()).ToList();
+                    var actualEnumNames = binRes.GetBinaryType().GetEnumValues().Select(v => v.EnumName).ToList();
```
ashapkin, line 91: Optional: it looks like we need to check this:
Assert.AreSame(binRes.GetBinaryType().GetEnumValues(), binRes.GetBinaryType().GetEnumValues())

ptupitsyn: I think I understand what you mean - since we cache enum values, we should always return the same collection. However, other nodes may have more values for the same enum (e.g. a newer code version). So the cache we introduced in this change is only for write operations. And when the user asks for all known enum values, we don't use the cache, but load them from the cluster.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Binary/BinaryUtils.cs
```diff
@@ -1749,6 +1753,76 @@ public static TimeSpan LongToTimeSpan(long ms)
...
+
+            EnumValues.Set(enumType, res);
+            
+            return res;
+        }
+        
+        /// <summary>
+        /// Gets the enum value as int.
+        /// </summary>
+        private static int GetEnumValueAsInt(Type underlyingType, object value)
```
gurustron, line 1792: Should we support long?

PR #8118, opened 2020-08-05
IGNITE-13308 Thin cpp tx realization.

zstan:

### Notes

### DIFF
--- a/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
+++ b/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
@@ -331,9 +331,16 @@ BOOST_AUTO_TEST_CASE(TestTxOps)
 
 const std::string label = std::string("label_2_check");
 
+std::string label1 = std::string("label_2_check1");
+
 bool checkTxLabelMessage(const ignite::IgniteError& ex)
 {
     return std::string(ex.what()).find(label) != std::string::npos;
+}
+
+bool checkTxLabel1Message(const ignite::IgniteError& ex)
+{
+    return std::string(ex.what()).find("label_2_check1") != std::string::npos;
 }
 
 BOOST_AUTO_TEST_CASE(TestTxWithLabel)
@@ -351,7 +358,7 @@ BOOST_AUTO_TEST_CASE(TestTxWithLabel)
 
     transactions::ClientTransactions transactions = client.ClientTransactions();
 
-    transactions::ClientTransaction tx = transactions.withLabel(label.c_str()).TxStart(TransactionConcurrency::OPTIMISTIC, TransactionIsolation::SERIALIZABLE, TX_TIMEOUT);
+    transactions::ClientTransaction tx = transactions.withLabel(label).TxStart(TransactionConcurrency::OPTIMISTIC, TransactionIsolation::SERIALIZABLE, TX_TIMEOUT);
 
     cache.Put(1, 10);
 
@@ -370,6 +377,20 @@ BOOST_AUTO_TEST_CASE(TestTxWithLabel)
     boost::this_thread::sleep_for(boost::chrono::milliseconds(2 * TX_TIMEOUT));
 
     BOOST_CHECK_EXCEPTION(tx.Commit(), ignite::IgniteError, !checkTxLabelMessage);
+
+    tx.Close();
+
+    // Label is gone
+
+    tx = transactions.withLabel(label1).TxStart(TransactionConcurrency::OPTIMISTIC, TransactionIsolation::SERIALIZABLE, TX_TIMEOUT);
+
+    label1 = "NULL";
+
+    cache.Put(1, 10);
+
+    boost::this_thread::sleep_for(boost::chrono::milliseconds(2 * TX_TIMEOUT));
+
+    BOOST_CHECK_EXCEPTION(tx.Commit(), ignite::IgniteError, checkTxLabel1Message);
 
     tx.Close();
 }
--- a/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
+++ b/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
@@ -220,7 +220,7 @@ BOOST_AUTO_TEST_CASE(TestCacheOpsWithTx)
 
     // Test transaction with a timeout.
 
-    const uint TX_TIMEOUT = 200L;
+    const uint32_t TX_TIMEOUT = 200;
 
     tx = transactions.TxStart(TransactionConcurrency::OPTIMISTIC, TransactionIsolation::SERIALIZABLE, TX_TIMEOUT);
 
@@ -347,7 +347,7 @@ BOOST_AUTO_TEST_CASE(TestTxWithLabel)
     cache::CacheClient<int, int> cache =
         client.GetCache<int, int>("partitioned");
 
-    const uint TX_TIMEOUT = 200L;
+    const uint32_t TX_TIMEOUT = 200;
 
     transactions::ClientTransactions transactions = client.ClientTransactions();
 
--- a/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
+++ b/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
@@ -31,6 +31,8 @@
 
 using namespace ignite::thin;
 using namespace boost::unit_test;
+using namespace ignite::thin::transactions;
+using namespace ignite::common::concurrent;
 
 class IgniteTxTestSuiteFixture
 {
--- a/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
+++ b/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
@@ -27,17 +27,15 @@
 using namespace ignite::thin;
 using namespace boost::unit_test;
 
-class AuthTestSuiteFixture
+class IgniteClientTestSuiteFixture1
 {
 public:
-    AuthTestSuiteFixture()
+    IgniteClientTestSuiteFixture1()
     {
-        ignite_test::ClearLfs();
-
-        serverNode = ignite_test::StartCrossPlatformServerNode("auth.xml", "ServerNode");
+        serverNode = ignite_test::StartCrossPlatformServerNode("cache.xml", "ServerNode");
     }
 
-    ~AuthTestSuiteFixture()
+    ~IgniteClientTestSuiteFixture1()
     {
         ignite::Ignition::StopAll(false);
     }
@@ -47,27 +45,19 @@ private:
     ignite::Ignite serverNode;
 };
 
-BOOST_FIXTURE_TEST_SUITE(AuthTestSuite, AuthTestSuiteFixture)
+BOOST_FIXTURE_TEST_SUITE(IgniteClientTestSuite, IgniteClientTestSuiteFixture1)
 
-BOOST_AUTO_TEST_CASE(AuthSuccess)
-{
-    IgniteClientConfiguration cfg;
-
-    cfg.SetEndPoints("127.0.0.1:11110");
-    cfg.SetUser("ignite");
-    cfg.SetPassword("ignite");
-
-    IgniteClient::Start(cfg);
-}
-
-BOOST_AUTO_TEST_CASE(AuthReject)
+BOOST_AUTO_TEST_CASE(TestTx)
 {
     IgniteClientConfiguration cfg;
 
     cfg.SetEndPoints("127.0.0.1:11110");
 
-    BOOST_CHECK_THROW(IgniteClient::Start(cfg), ignite::IgniteError);
-}
+    IgniteClient client = IgniteClient::Start(cfg);
 
+    ignite::thin::transactions::ClientTransaction* tx = client.ClientTransactions().txStart();
+
+    tx->commit();
+}
 
 BOOST_AUTO_TEST_SUITE_END()
--- a/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
+++ b/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
@@ -24,6 +24,7 @@
 #include <ignite/thin/cache/cache_peek_mode.h>
 
 #include <test_utils.h>
+#include <ignite/ignite_error.h>
 
 using namespace ignite::thin;
 using namespace boost::unit_test;
@@ -194,6 +195,28 @@ BOOST_AUTO_TEST_CASE(TestCacheOpsWithTx)
     tx.Rollback();
 
     BOOST_CHECK_EQUAL(cache.ContainsKey(1), true);
+}
+
+BOOST_AUTO_TEST_CASE(TestTxOps)
+{
+    IgniteClientConfiguration cfg;
+
+    cfg.SetEndPoints("127.0.0.1:11110");
+
+    IgniteClient client = IgniteClient::Start(cfg);
+
+    cache::CacheClient<int, int> cache =
+        client.GetCache<int, int>("partitioned");
+
+    cache.Put(1, 1);
+
+    transactions::ClientTransactions transactions = client.ClientTransactions();
+
+    transactions::ClientTransaction tx1 = transactions.TxStart();
+
+    BOOST_CHECK_THROW( transactions.TxStart(), ignite::IgniteError );
+
+    tx1.Close();
 }
 
 BOOST_AUTO_TEST_SUITE_END()
--- a/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
+++ b/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
@@ -63,19 +63,18 @@ BOOST_AUTO_TEST_CASE(TestTx)
         client.GetCache<int, int>("partitioned");
 
     cache.Put(1, 1);
-    cache.Put(2, 2);
 
     transactions::ClientTransactions transactions = client.ClientTransactions();
 
     transactions::ClientTransaction tx = transactions.txStart();
 
     cache.Put(1, 10);
-    cache.Put(2, 20);
+
+    BOOST_CHECK_EQUAL(10, cache.Get(1));
 
     tx.rollback();
 
     BOOST_CHECK_EQUAL(1, cache.Get(1));
-    BOOST_CHECK_EQUAL(2, cache.Get(2));
 }
 
 BOOST_AUTO_TEST_SUITE_END()
--- a/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
+++ b/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
@@ -80,7 +80,7 @@ BOOST_AUTO_TEST_CASE(TestGetPut)
 
     BOOST_CHECK_EQUAL(10, cache.Get(1));
 
-    //tx.rollback();
+    tx.rollback();
 
     //BOOST_CHECK_EQUAL(1, cache.Get(1));
 }
--- a/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
+++ b/modules/platforms/cpp/thin-client-test/src/tx_test.cpp
@@ -228,7 +228,7 @@ BOOST_AUTO_TEST_CASE(TestCacheOpsWithTx)
 
     boost::this_thread::sleep_for(boost::chrono::milliseconds(2 * TX_TIMEOUT));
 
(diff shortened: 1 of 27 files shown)


modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transactions.h
```diff
@@ -0,0 +1,189 @@
...
+                    return ClientTransaction(proxy.txStart(concurrency, isolation, timeout, txSize, label));
+                }
+
+                /**
+                 * Returns instance of {@code ClientTransactions} to mark each new transaction with a specified label.
+                 *
+                 * @param label Transaction label.
+                 * @return ClientTransactions implementation.
+                 */
+                ClientTransactions withLabel(const char* lbl)
```
ivandasch, line 165: I suppose, that const std::string& is preferrable in client API (In internal it's ok to use C-style strings)

modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transactions.h
```diff
@@ -0,0 +1,189 @@
...
+                {
+                    ClientTransactions copy = ClientTransactions(proxy);
+
+                    copy.label = lbl;
+
+                    return copy;
```
ivandasch, line 171: What do you think about private constructor? (label + proxy)?

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_impl.cpp
```diff
@@ -0,0 +1,252 @@
...
+                    {
+                        std::map<int32_t, SP_TransactionImpl>::iterator it;
+
+                        {
+                            RwSharedLockGuard lock(txToIdRWLock);
+
+                            it = txToId.find(txId);
+                        }
+
+                        if (it != txToId.end())
```
ivandasch, line 87: Why this is out of lock? I suppose this is not thread-safe

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_impl.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+                    std::map<int32_t, SP_TransactionImpl>::iterator it;
+
+                    {
+                        RwSharedLockGuard lock(txToIdRWLock);
+
+                        it = txToId.find(txId);
+                    }
+
+                    if (it != txToId.end())
```
ivandasch, line 133: Same as above

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_impl.h
```diff
@@ -0,0 +1,276 @@
...
+                typedef SharedPointer<TransactionImpl> SP_TransactionImpl;
+
+                /**
+                 * Ignite transactions implementation.
+                 *
+                 * This is an entry point for Thin C++ Ignite transactions.
+                 */
+                class TransactionImpl
+                {
+                    typedef ThreadLocalInstance<int32_t> TL_TXID;
```
ivandasch, line 51: I don't understand, why not simply store pointer to Transaction instead of id? This can eliminate usage of syncrhonized map at all

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_proxy.cpp
```diff
@@ -0,0 +1,80 @@
...
+                    GetTxImpl(impl).Rollback();
+                }
+
+                void TransactionProxy::close()
+                {
+                    try
+                    {
+                        GetTxImpl(impl).Close();
+                    }
+                    catch (...)
```
ivandasch, line 73: Empty catch block? Do you think that it is good idea?

zstan: yep, close can`t throw any exceptions i hope.

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_impl.h
zstan: I believe we talking about this internally or question still actual ?

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_proxy.cpp
ivandasch: I understand why. But i  suppose, that std::set<int32_t> under locks + pointer to transaction in thread_local is better.

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_impl.h
ivandasch: I understand why. But i suppose, that std::set<int32_t> under locks + pointer to transaction in thread_local is better.__

modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transactions.h
```diff
@@ -0,0 +1,189 @@
...
+                    return ClientTransaction(proxy.txStart(concurrency, isolation, timeout, txSize, label));
+                }
+
+                /**
+                 * Returns instance of {@code ClientTransactions} to mark each new transaction with a specified label.
+                 *
+                 * @param label Transaction label.
+                 * @return ClientTransactions implementation.
+                 */
+                ClientTransactions withLabel(const std::string& lbl)
```
ivandasch, line 155: What if somebody do something like this:
```
ClientTransactions getTxWithRandomLabel(const ClientTransactions& cl_tx) {
         std::string rnd = getRandomString();
        
         return cl_tx.withLabel(rnd);
}
```
I think that something bad can happens:)

modules/platforms/cpp/thin-client/src/impl/message.h
```diff
@@ -658,7 +658,7 @@ namespace ignite
                     writer.WriteInt8(concurrency);
                     writer.WriteInt8(isolation);
                     writer.WriteInt64(timeout);
-                    writer.WriteString(label.Get());
+                    writer.WriteString(label.IsValid() ? label.Get() : "");
```
ivandasch, line 661: According to specs, you should write null, not empty string.

zstan: got it ! thanks !

modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transactions.h
```diff
@@ -0,0 +1,191 @@
...
+#ifndef TRANSACTIONS_H
+#define TRANSACTIONS_H
+
+#include <string>
+
+#include <ignite/common/concurrent.h>
+#include <ignite/impl/thin/transactions/transactions_proxy.h>
+
+using namespace ignite::impl::thin::transactions;
+using namespace ignite::common::concurrent;
```
isapego, line 27: There should never be `using` in public headers.

modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transactions.h
```diff
@@ -0,0 +1,191 @@
...
+            /**
+             * Transactions client.
+             *
+             * This is an entry point for Thin C++ Ignite transactions.
+             *
+             * This class implemented as a reference to an implementation so copying of this class instance will only
+             * create another reference to the same underlying object. Underlying object released automatically once all
+             * the instances are destructed.
+             */
+            class ClientTransactions {
```
isapego, line 111: Class should be moved to its own header.

modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transactions.h
```diff
@@ -0,0 +1,191 @@
...
+                    proxy.close();
+                }
+
+                /**
+                 * Assignment operator.
+                 *
+                 * @param other Another instance.
+                 * @return *this.
+                 */
+                ClientTransaction& operator=(const ClientTransaction& other)
```
isapego, line 86: If there is assignment operator, there should also be copy constructor.

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_impl.cpp
```diff
@@ -0,0 +1,212 @@
...
+            {
+                TransactionImpl::TL_TXID TransactionImpl::threadTx;
+
+                TransactionsImpl::TransactionsImpl(const SP_DataRouter& router) :
+                    router(router)
+                {
+                }
+
+                template<typename ReqT, typename RspT>
+                void TransactionsImpl::SyncMessage(const ReqT& req, RspT& rsp)
```
isapego, line 41: Can you re-name this method to something like `SendTxMessage`? SyncMessage should only be in network parts.

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_impl.h
```diff
@@ -0,0 +1,268 @@
...
+                typedef SharedPointer<TransactionImpl> SP_TransactionImpl;
+
+                /**
+                 * Ignite transactions implementation.
+                 *
+                 * This is an entry point for Thin C++ Ignite transactions.
+                 */
+                class TransactionImpl
+                {
+                    typedef ThreadLocalInstance<SP_TransactionImpl> TL_TXID;
```
isapego, line 51: The naming is bad, looks like a constant or macro. Also, typedef itself is not really needed - the type only used in 2 places.

modules/platforms/cpp/thin-client/src/impl/message.h
```diff
@@ -569,6 +619,110 @@ namespace ignite
                 const Writable& val3;
             };
 
+            /**
+             * Tx start request.
+             */
+            template<int32_t OpCode>
```
isapego, line 625: Why does this type needs to be parameterized if the parameter is always the same?

modules/platforms/cpp/thin-client/src/impl/message.h
```diff
@@ -569,6 +619,110 @@ namespace ignite
...
+
+                /** Tx label. */
+                ignite::common::concurrent::SharedPointer<const char> label;
+            };
+
+            /**
+             * Tx end request.
+             */
+            template<int32_t OpCode>
+            class TxEndRequest : public Request<OpCode>
```
isapego, line 685: The same as for `TxStartRequest`

modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transactions.h
```diff
@@ -0,0 +1,191 @@
...
+
+                /**
+                 * Returns instance of {@code ClientTransactions} to mark each new transaction with a specified label.
+                 *
+                 * @param label Transaction label.
+                 * @return ClientTransactions implementation.
+                 */
+                ClientTransactions withLabel(const std::string& lbl)
+                {
+                    ClientTransactions copy = ClientTransactions(proxy, lbl.c_str());
```
isapego, line 158: Why do you pass `lbl.c_str()` here, if `ClientTransactions()` takes `const std::string&`?

modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transactions.h
```diff
@@ -0,0 +1,191 @@
...
+
+                /**
+                 * Constructor.
+                 *
+                 * @param impl Implementation.
+                 */
+                ClientTransactions(TransactionsProxy& impl, const std::string& lbl) :
+                    proxy(impl)
+                {
+                    char* l = new char[lbl.size() + 1];
```
isapego, line 182: `SharedPointer` uses `delete` as a deleter by default, not `delete[]`. Use something different for a container here, for example `SharedPointer<common::FixedSizeArray<char>>`.

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_proxy.cpp
isapego: I'm not sure about this solution to be honest. This can lead to really hard-to-debug bugs.

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_proxy.cpp
```diff
@@ -0,0 +1,80 @@
...
+                    {
+                    }
```
isapego, line 75: Maybe we should add `assert(false);` here at least?

isapego (changes requested):
See my comments inline.

modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transaction_consts.h
```diff
@@ -0,0 +1,119 @@
...
+
+namespace ignite
+{
+    namespace thin
+    {
+        namespace transactions
+        {
+            #define TX_ALREADY_CLOSED "The transaction is already closed."
+            #define TX_ALREADY_STARTED "A transaction has already been started by the current thread."
+            #define TX_DIFFERENT_THREAD "You can commit transaction only from the thread it was started."
```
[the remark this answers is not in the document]
zstan: yep, they used in tests, fr example.

modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transactions.h
zstan: of course ! i miss it :(

modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transactions.h
zstan: i remove it, looks like it excessive here.

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_proxy.cpp
zstan: looks like it would bring some tests in that case ... java thin client use the same approach.

modules/platforms/cpp/thin-client/include/ignite/impl/thin/transactions/transactions_proxy.h
```diff
@@ -126,7 +127,7 @@ namespace ignite
                             ignite::thin::transactions::TransactionIsolation::Type isolation = ignite::thin::transactions::TransactionIsolation::READ_COMMITTED,
                             int64_t timeout = 0,
                             int32_t txSize = 0,
-                            ignite::common::concurrent::SharedPointer<const char> lbl = "");
+                            ignite::common::concurrent::SharedPointer<ignite::common::FixedSizeArray<char> > lbl = "");
```
ivandasch, line 130: Why empty string? default should be null string.

modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transactions.h
```diff
@@ -181,9 +111,11 @@ namespace ignite
                 ClientTransactions(ignite::impl::thin::transactions::TransactionsProxy& impl, const std::string& lbl) :
                     proxy(impl)
                 {
-                    char* l = new char[lbl.size() + 1];
+                    ignite::common::FixedSizeArray<char> *label0 = new ignite::common::FixedSizeArray<char>(lbl.size() + 1);
+
+                    strcpy(label0->GetData(), lbl.c_str());
 
-                    label = strcpy(l, lbl.c_str());
+                    label = ignite::common::concurrent::SharedPointer<ignite::common::FixedSizeArray<char> >(label0);
```
ivandasch, line 118: `ignite::common::concurrent::SharedPointer<ignite::common::FixedSizeArray<char>>`

modules/platforms/cpp/thin-client/include/ignite/impl/thin/transactions/transactions_proxy.h
zstan: thanks !

modules/platforms/cpp/thin-client-test/src/tx_test.cpp
```diff
@@ -0,0 +1,400 @@
...
+
+    cache.Put(1, 10);
+
+    BOOST_CHECK_EQUAL(10, cache.Get(1));
+
+    tx.Rollback();
+
+    BOOST_CHECK_EQUAL(1, cache.Get(1));
+
+    //---
```
isapego, line 101: A lot of code in a single test case. Why won't we break it up in a short tests, one of them testing its own scenario?

modules/platforms/cpp/thin-client-test/src/tx_test.cpp
```diff
@@ -0,0 +1,400 @@
...
+
+    return true;
+}
+
+bool checkTxTimeoutMessage(const ignite::IgniteError& ex)
+{
+    return std::string(ex.what()).find("Cache transaction timed out") != std::string::npos;
+}
+
+BOOST_AUTO_TEST_CASE(TestCacheOpsWithTx)
```
isapego, line 76: Let's add a comment with a brief explanation to every test so it would be easier for people with QA background to understand what is going on in test.

modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transaction.h
```diff
@@ -0,0 +1,92 @@
...
+                 * @param impl Implementation.
+                 */
+                ClientTransaction(ignite::impl::thin::transactions::TransactionProxy impl) :
+                    proxy(impl)
+                {}
+
+                /**
+                 * Destructor.
+                 */
+                ~ClientTransaction() {}
```
isapego, line 53: Add `// No-op.` comment to all empty blocks please.

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_impl.h
```diff
@@ -0,0 +1,264 @@
...
+
+                typedef ignite::common::concurrent::SharedPointer<TransactionsImpl> SP_TransactionsImpl;
+                typedef ignite::common::concurrent::SharedPointer<TransactionImpl> SP_TransactionImpl;
+
+                /**
+                 * Ignite transactions implementation.
+                 *
+                 * This is an entry point for Thin C++ Ignite transactions.
+                 */
+                class TransactionImpl
```
isapego, line 47: The same as for public class - let's move those classes to their own headers.

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_impl.h
```diff
@@ -0,0 +1,264 @@
...
+                        isolation(isolation),
+                        timeout(timeout),
+                        txSize(size),
+                        closed(false)
+                    {}
+                    
+                    /**
+                     * Destructor.
+                     */
+                    ~TransactionImpl() {}
```
isapego, line 79: This will cause memory leak if transaction object is destroyed on C++ side without calling `Close()`. Destructor should always release all resources it holds.

modules/platforms/cpp/thin-client/src/impl/transactions/transactions_proxy.cpp
isapego: well, Ok, let's leave it as is for now.

isapego (changes requested):
Please, see my comments inline.

modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transaction.h
```diff
@@ -0,0 +1,95 @@
...
+                }
+
+            private:
+                /** Implementation. */
+                ignite::impl::thin::transactions::TransactionProxy proxy;
+
+                /**
+                 * Default constructor.
+                 */
+                ClientTransaction() {}
```
isapego, line 89: Default constructor should be public in this case I believe, this will allow user to store transactions in collections.

modules/platforms/cpp/thin-client/include/ignite/thin/transactions/transactions.h
```diff
@@ -0,0 +1,125 @@
...
+                /** Implementation. */
+                ignite::impl::thin::transactions::TransactionsProxy proxy;
+
+                /** Transaction specific label. */
+                ignite::common::concurrent::SharedPointer<ignite::common::FixedSizeArray<char> > label;
+
+                /**
+                 * Default constructor.
+                 */
+                ClientTransactions();
```
isapego, line 104: Not sure why should it be private. May cause additional trouble for users, let's leave it public.

isapego (approved):
Overall looks good. Please, find 2 comments in PR. After fixing them we may proceed to merging this feature.

PR #8143, opened 2020-08-11
IGNITE-12492 TDE - Phase-2. Documentation.

NSAmelchev:
Add master key rotation docs.

### DIFF
--- a/docs/_docs/security/master-key-rotation.adoc
+++ b/docs/_docs/security/master-key-rotation.adoc
@@ -62,7 +62,7 @@ tab:Java[]
 
 [source, java]
 ----
-include::{javaCodeDir}/TDE.java[tags=masterKeyRotation, indent=0]
+include::{javaCodeDir}/TDE.java[tags=master-key-rotation, indent=0]
 ----
 --
 
--- a/docs/_docs/security/master-key-rotation.adoc
+++ b/docs/_docs/security/master-key-rotation.adoc
@@ -30,7 +30,7 @@ Ignite provide the ability to change the master key from the following interface
 
 === Command line tool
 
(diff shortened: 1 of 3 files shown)


docs/_docs/security/master-key-rotation.adoc
```diff
@@ -0,0 +1,119 @@
...
+|Method | Description
+|getMasterKeyName() | Gets the current master key name.
+|changeMasterKey(String masterKeyName) | Starts master key change process.
+|===
+
+=== From code
+
+The master key change process can be managed programmatically:
+
+[source,java]
```
artem-budnikov, line 59: Please move the code example to a compilable java file (code-snippets/java/src/main/java/org/apache/ignite/snippets/TDE.java) and include snippets from that file here

docs/_docs/security/master-key-rotation.adoc
```diff
@@ -0,0 +1,119 @@
...
+
+# Change the master key.
+control.sh|bat --encryption change_master_key newMasterKeyName
+----
+
+=== JMX
+
+You can also manage the master key change process via the `EncryptionMXBean` interface:
+
+[{table_opts}]
```
artem-budnikov, line 48: The `table_opts` variable is not defined anywhere in the file. You should use explicit table configuration in this case, like so: 
[cols="1,3", opts="header"]

docs/_docs/security/master-key-rotation.adoc
NSAmelchev: Moved.

docs/_docs/security/master-key-rotation.adoc
NSAmelchev: Fixed. Thanks

PR #8145, opened 2020-08-12
IGNITE-13349: Tcp discovery statistics migrated to the new metrics framework.

nizhikov:
Thank you for submitting the pull request to the Apache Ignite.

In order to streamline the review of the contribution 
we ask you to ensure the following steps have been taken:

### The Contribution Checklist
- [ ] There is a single JIRA ticket related to the pull request. 
- [ ] The web-link to the pull request is attached to the JIRA ticket.
- [ ] The JIRA ticket has the _Patch Available_ state.
- [ ] The pull request body describes changes that have been made. 
The description explains _WHAT_ and _WHY_ was made instead of _HOW_.
- [ ] The pull request title is treated as the final commit message. 
The following pattern must be used: `IGNITE-XXXX Change summary` where `XXXX` - number of JIRA issue.
- [ ] A reviewer has been mentioned through the JIRA comments 
(see [the Maintainers list](https://cwiki.apache.org/confluence/display/IGNITE/How+to+Contribute#HowtoContribute-ReviewProcessandMaintainers)) 
- [ ] The pull request has been checked by the Teamcity Bot and 
the `green visa` attached to the JIRA ticket (see [TC.Bot: Check PR](https://mtcga.gridgain.com/prs.html))

### Notes
- [How to Contribute](https://cwiki.apache.org/confluence/display/IGNITE/How+to+Contribute)
- [Coding abbreviation rules](https://cwiki.apache.org/confluence/display/IGNITE/Abbreviation+Rules)
- [Coding Guidelines](https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines)
- [Apache Ignite Teamcity Bot](https://cwiki.apache.org/confluence/display/IGNITE/Apache+Ignite+Teamcity+Bot)

If you need any help, please email dev@ignite.apache.org or ask anу advice on http://asf.slack.com _#ignite_ channel.

PR #8150, opened 2020-08-13
IGNITE-13354 Add ClusterMetrics to the new framework

NSAmelchev:
Add metrics to the new framework:
- TopologyVersion
- TotalBaselineNodes
- TotalServerNodes
- TotalClientNodes
- ActiveBaselineNodes

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsMXBeanImpl.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsMXBeanImpl.java
@@ -18,18 +18,26 @@
 package org.apache.ignite.internal;
 
 import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.UUID;
-import org.apache.ignite.cluster.BaselineNode;
 import org.apache.ignite.cluster.ClusterGroup;
 import org.apache.ignite.cluster.ClusterMetrics;
 import org.apache.ignite.cluster.ClusterNode;
+import org.apache.ignite.internal.processors.metric.MetricRegistry;
 import org.apache.ignite.internal.util.typedef.internal.S;
 import org.apache.ignite.mxbean.ClusterMetricsMXBean;
+import org.apache.ignite.spi.metric.IntMetric;
+import org.apache.ignite.spi.metric.LongMetric;
+
+import static org.apache.ignite.internal.cluster.IgniteClusterImpl.ACTIVE_BASELINE_NODES;
+import static org.apache.ignite.internal.cluster.IgniteClusterImpl.TOPOLOGY_VERSION;
+import static org.apache.ignite.internal.cluster.IgniteClusterImpl.TOTAL_BASELINE_NODES;
+import static org.apache.ignite.internal.cluster.IgniteClusterImpl.TOTAL_CLIENT_NODES;
+import static org.apache.ignite.internal.cluster.IgniteClusterImpl.TOTAL_NODES;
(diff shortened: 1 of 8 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/ClusterProcessor.java
```diff
@@ -607,6 +610,33 @@ private Boolean findLastFlag(Collection<Serializable> vals) {
                 U.error(log, "Failed to register MBean for cluster: ", e);
             }
         }
+
+        MetricRegistry reg = ctx.metric().registry(CLUSTER_METRICS);
+
+        reg.register("TopologyVersion",
```
nizhikov, line 616: Can we use this metric and all below inside `ClusterMetricsMXBeanImpl` to keep single method to calculate metric value.

modules/core/src/test/java/org/apache/ignite/internal/ClusterBaselineNodesMetricsSelfTest.java
```diff
@@ -82,9 +92,13 @@ public void testBaselineNodes() throws Exception {
         log.info(String.format(">>> State #1: topology version = %d", ignite0.cluster().topologyVersion()));
 
         assertEquals(2, mxBeanCluster.getTotalServerNodes());
+        assertEquals(2, ((IntMetric)mreg.findMetric("TotalServerNodes")).value());
```
nizhikov, line 95: Let's use `merg.<IntMetrics>findMetric("XXX").value()` to avoid type casting.

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/ClusterProcessor.java
NSAmelchev: We can use this metrics for the `ClusterMetricsMXBeanImpl` implementation. But not for the `ClusterLocalNodeMetricsMXBeanImpl`.
Moreover metrics registration should be placed between starting `GridMetricManager` and beans registration. For example, in the `IgniteKernal#registerMetrics()`

modules/core/src/test/java/org/apache/ignite/internal/ClusterBaselineNodesMetricsSelfTest.java
NSAmelchev: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/ClusterNodeMetricsSelfTest.java
```diff
@@ -310,7 +314,7 @@ public void testIoMetrics() throws Exception {
      */
     @Test
     public void testClusterNodeMetrics() throws Exception {
-        final Ignite ignite0 = grid();
+        IgniteEx ignite0 = grid();
```
nizhikov, line 317: Let's return `final` modifier.

NSAmelchev: Reverted.

modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsMXBeanImpl.java
```diff
@@ -47,13 +54,37 @@
     /** Cluster metrics update mutex. */
     private final Object clusterMetricsMux = new Object();
 
+    /** Topology version metric. */
+    private final LongMetric topVer;
```
nizhikov, line 58: Let's remove this metric from new framework.
"CurrentTopologyVersion" would be added by the #8145 in the "io.discovery" registry.

NSAmelchev: Done.

PR #8154, opened 2020-08-14
Ducktape Service Specs

timoninmaxim:
1. Specs for services (config, start command)
2. Injecting specs from globals
3. Prepare distribution for external usage


modules/ducktests/tests/setup.py
```diff
@@ -14,46 +14,22 @@
...
       license="apache2.0",
-      packages=find_packages(),
+      packages=find_packages(exclude=["ignitetest.tests", "ignitetest.tests.*"]),
       include_package_data=True,
-      install_requires=["ducktape==0.7.8", "requests==2.20.0"],
-      tests_require=["pytest", "mock", "monotonic"],
-      cmdclass={'test': PyTest}
+      install_requires=["ducktape==0.7.8", "requests==2.22.0"],
+      tests_require=["pytest", "mock", "monotonic"]
 )
```
ivandasch, line 35: As for me, monotonic is install_required, and pytest and mock is not needed at all

timoninmaxim: fixed

PR #8174, opened 2020-08-20
IGNITE-7623 Add async API to Java Thin Client

ptupitsyn:
* Add new Future interface for all thin client async operations: `IgniteClientFuture`
* Add async counterparts for operations in `IgniteClient`, `ClientCache`, `ClientCompute`

IEP-51: https://cwiki.apache.org/confluence/display/IGNITE/IEP-51%3A+Java+Thin+Client+Async+API


modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
```diff
@@ -176,6 +297,24 @@
...
+     * Removes the mapping for a key from this cache if it is present.
+     * <p>
+     * More formally, if this cache contains a mapping from key <tt>k</tt> to value <tt>v</tt> such that
+     * <code>(key==null ?  k==null : key.equals(k))</code>, that mapping is removed.
+     * (The cache can contain at most one such mapping.)
+     *
+     * <p>Returns <tt>true</tt> if this cache previously associated the key, or <tt>false</tt> if the cache
+     * contained no mapping for the key.
+     * <p>
+     * The cache will not contain a mapping for the specified key once the
```
isapego, line 310: Minor: Looks like the line break is at the different position every time :)

modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
```diff
@@ -205,13 +364,29 @@
...
      * Removes all of the mappings from this cache.
      * <p>
      * The order that the individual entries are removed is undefined.
      */
     public void removeAll() throws ClientException;
 
+    /**
+     * Removes all of the mappings from this cache.
+     * <p>
+     * The order that the individual entries are removed is undefined.
```
isapego, line 386: Maybe it needs to mention if the operation is atomic

modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
```diff
@@ -293,11 +529,36 @@
...
+     */
+    public IgniteClientFuture<Boolean> putIfAbsentAsync(K key, V val) throws ClientException;
+
     /**
      * Clears the contents of the cache.
      */
     public void clear() throws ClientException;
 
+    /**
+     * Clears the contents of the cache.
```
isapego, line 558: Probably, needs a `@return` tag and the information about how this is different from `removeAllAsync`

modules/core/src/main/java/org/apache/ignite/client/ClientCompute.java
```diff
@@ -55,9 +55,23 @@
      * @return A Future representing pending completion of the task.
      * @throws ClientException If task failed.
      * @see ComputeTask for information about task execution.
+     * @deprecated Use {@link ClientCompute#executeAsync2(String, Object)}
```
isapego, line 58: Maybe needs an explanation why is it deprecated.

modules/core/src/main/java/org/apache/ignite/client/IgniteClient.java
```diff
@@ -57,25 +72,50 @@
...
+
     /**
      * Create cache.
      *
      * @param name Cache name.
      */
     public <K, V> ClientCache<K, V> createCache(String name) throws ClientException;
 
+    /**
+     * Create cache.
```
isapego, line 98: Probably, it worth to mention that the default configuration is going to be used.

modules/core/src/main/java/org/apache/ignite/configuration/ClientConfiguration.java
```diff
@@ -115,6 +118,9 @@
     /** Reconnect throttling retries. See {@code reconnectThrottlingPeriod}. */
     private int reconnectThrottlingRetries = 3;
 
+    /** Executor for async operations continuations. */
+    private Executor asyncContinuationExecutor;
```
isapego, line 122: Is it `null` by default? Does it mean, user will get an exception in runtime if it uses continuation with no executor set?

Maybe, we should set some implementation here as a default value? I'm pretty much sure, it is going to be fine for the most users, at least at the beginning of the development.

isapego (commented):
Public API reviewed. I'll submit internal implementation review later today.

modules/core/src/main/java/org/apache/ignite/configuration/ClientConfiguration.java
ptupitsyn: This is a private property, please see public setter and getter, there is a detailed javadoc.

modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
```diff
@@ -88,6 +128,17 @@
...
     public int size(CachePeekMode... peekModes) throws ClientException;
 
+    /**
+     * Gets the number of all entries cached across all nodes. By default, if {@code peekModes} value isn't provided,
+     * only size of primary copies across all nodes will be returned. This behavior is identical to calling
+     * this method with {@link CachePeekMode#PRIMARY} peek mode.
+     * <p>
+     * NOTE: this operation is distributed and will query all participating nodes for their cache sizes.
+     *
+     * @param peekModes Optional peek modes. If not provided, then total cache size is returned.
```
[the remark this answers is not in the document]
ptupitsyn: Fixed.

modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
```diff
@@ -77,6 +111,12 @@
      */
     public ClientCacheConfiguration getConfiguration() throws ClientException;
 
+    /**
+     * Gets the cache configuration asynchronously.
+     * @return The cache configuration.
```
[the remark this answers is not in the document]
ptupitsyn: Fixed

modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
```diff
@@ -117,6 +178,25 @@
...
+     * <p>
+     * The order in which the individual puts occur is undefined.
+     * <p>
+     * The behavior of this operation is undefined if entries in the cache
+     * corresponding to entries in the map are modified or removed while this
+     * operation is in progress. or if map is modified while the operation is in
+     * progress.
+     * <p>
+     *
+     * @param map Mappings to be stored in this cache.
```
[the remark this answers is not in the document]
ptupitsyn: Fixed

modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
```diff
@@ -117,6 +178,25 @@
...
+     * <p>
+     * The effect of this call is equivalent to that of calling
+     * {@link #put(Object, Object) put(k, v)} on this cache once for each mapping
+     * from key <tt>k</tt> to value <tt>v</tt> in the specified map.
+     * <p>
+     * The order in which the individual puts occur is undefined.
+     * <p>
+     * The behavior of this operation is undefined if entries in the cache
+     * corresponding to entries in the map are modified or removed while this
+     * operation is in progress. or if map is modified while the operation is in
```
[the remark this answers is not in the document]
ptupitsyn: Fixed

modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
ptupitsyn: I just copied the javadoc from synchronous operations :)

modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
ptupitsyn: Fixed

modules/core/src/main/java/org/apache/ignite/client/ClientCompute.java
ptupitsyn: Fixed

modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
ptupitsyn: Fixed

modules/core/src/main/java/org/apache/ignite/client/IgniteClient.java
ptupitsyn: Fixed

modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
```diff
@@ -138,6 +218,27 @@
...
+     * This is equivalent to:
+     * <pre><code>
+     * if (cache.containsKey(key) &amp;&amp; equals(cache.get(key), oldValue)) {
+     *  cache.put(key, newValue);
+     * return true;
+     * } else {
+     *  return false;
+     * }
+     * </code></pre>
+     * except that the action is performed atomically.
```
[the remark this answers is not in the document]
ptupitsyn: Good point, fixed here and in other places

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientComputeImpl.java
```diff
@@ -115,6 +119,11 @@
         return executeAsync0(taskName, arg, dfltGrp, (byte)0, 0L);
     }
 
+    /** {@inheritDoc} */
+    @Override public <T, R> IgniteClientFuture<R> executeAsync2(String taskName, @Nullable T arg) throws ClientException {
```
isapego, line 123: Do we need `executeAsync` at all?

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
```diff
@@ -210,14 +213,95 @@
...
+                                          AtomicInteger chIdx,
+                                          ClientChannel ch,
+                                          T res,
+                                          Throwable err) {
+        if (err == null) {
+            fut.complete(res);
+            return null;
+        }
+
+        if (err instanceof ClientConnectionException) {
```
isapego, line 251: I think, only those operations that do not modify cache state should be retried or we can get unexpected results, if the operation was successfully performed, but we were unable to receive response.

Also, it is OK to retry if failure happens on send.

isapego (approved):
Overall looks good. Please, see my comments inline

modules/core/src/main/java/org/apache/ignite/client/ClientCompute.java
```diff
@@ -55,9 +55,25 @@
      * @return A Future representing pending completion of the task.
      * @throws ClientException If task failed.
      * @see ComputeTask for information about task execution.
+     * @deprecated Use {@link #executeAsync2(String, Object)} instead.
+     * This method calls {@link #executeAsync2(String, Object)} internally, but returns a more limited
+     * Future interface.
      */
+    @Deprecated
```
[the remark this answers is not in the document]
ptupitsyn: As I understand, this still breaks binary compatibility, so we have to create a new method.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientComputeImpl.java
ptupitsyn: See above, I think we do, because of binary compatibility.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
ptupitsyn: I agree, but this behavior is consistent with sync API, I don't think we should have different behavior for sync and async API.

isapego: Well, OK then

PR #8191, opened 2020-08-27
IGNITE-13388 apache-ignite deb package depends on a non-existent package and can't be installed on Debian 10

vveider:


packaging/deb/control
```diff
@@ -8,7 +8,7 @@ Package: apache-ignite
 Architecture: all
 Section: misc
 Priority: optional
-Depends: openjdk-8-jdk | oracle-java8-installer, systemd, passwd
+Depends: openjdk-8-jre | openjdk-11-jre | default-jre | java-runtime, systemd, passwd
```
alamar, line 11: JRE dependency does not have javac & stuff, which may lead to problems in deployments.
Can we change it to `openjdk-8-jdk | openjdk-11-jdk | default-jdk | java-sdk`?

alamar (changes requested):

packaging/deb/control
vveider: Of course, I was sure that JRE runtime will be enough for Apache Ignite running.

alamar (approved):

PR #8193, opened 2020-08-27
IGNITE-13380 Output IgniteSystemProperties via ignite.sh

NSAmelchev:
The new command added:
`ignite.sh -systemProps`


modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -53,30 +55,43 @@
...
      */
+    @IgniteSystemProperty(description = "If this system property is present the Ignite will include instance name " +
+        "into verbose log.", type = String.class)
     public static final String IGNITE_LOG_INSTANCE_NAME = "IGNITE_LOG_INSTANCE_NAME";
 
     /**
      * This property is used internally to pass an exit code to loader when
      * Ignite instance is being restarted.
      */
+    @IgniteSystemProperty(description = "This property is used internally to pass an exit code to loader when Ignite " +
```
nizhikov, line 73: Please, remove all boilerplate text like "This property", "System property".
Also, please, use the spellchecker to ensure correct english in the description.

nizhikov:
Please, add `String defaultValue` to the annotation.

nizhikov:
We should add 2 columns in the output:

1. Default value.
2. Actual value.

PR #8196, opened 2020-08-29
IGNITE-13392 [ML]: fix Incorrect Vector::kNorm evaluation for odd powers

mrk-andreev:
Current implementation of `Vector::kNorm` is incorrect. 

Current formula is (`org.apache.ignite.ml.math.primitives.vector.AbstractVector:882`):
```
(\sum_{i}{x^p})^{1/p}
```
But correct formula is:
```
(\sum_{i}{|x|^p})^{1/p}
```
We can verify this using lectures (https://www.math.usm.edu/lambers/mat610/sum10/lecture2.pdf) or using Wolfram Mathematica:
```
> Norm[{x, y, z}, p]
(Abs[x]^p+Abs[y]^p+Abs[z]^p)^(1/p)
```

Issue: https://issues.apache.org/jira/browse/IGNITE-13392

### DIFF
--- a/modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/AbstractVector.java
+++ b/modules/ml/src/main/java/org/apache/ignite/ml/math/primitives/vector/AbstractVector.java
@@ -879,7 +879,8 @@ public abstract class AbstractVector implements Vector {
             return nonZeroElements();
         else
             // Default case.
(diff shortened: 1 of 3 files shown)


zaleslaw:
Looks good for me, I'll run TC yet once to attach to the ticket and will merge it next week.

PR #8205, opened 2020-09-01
IGNITE-13393

sanpwc:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/atomic/GridDhtAtomicCache.java
```diff
@@ -1038,91 +1050,108 @@ private IgniteInternalFuture updateAll0(
         final GridCacheOperation op,
         boolean async
     ) {
-        assert ctx.updatesAllowed();
+        try (MTC.TraceSurroundings ignored =
+                 MTC.support(ctx.kernalContext().tracing().create(CACHE_API_UPDATE, MTC.span()))) {
+            MTC.span().addTagOrLog("cache", CACHE_API_UPDATE, () -> Objects.toString(cacheCfg.getName()));
 
-        ctx.checkSecurity(SecurityPermission.CACHE_PUT);
+            Map<? extends K, ? extends V> m = map;
```
ascherbakoff, line 1057: map can be declared as final in the argument list. better to log keys count only once:

```
if (map != null)
MTC.span().addTagOrLog("keys.count", CACHE_API_UPDATE,
                () -> String.valueOf(map.size()));

if (invokeMap!= null)
MTC.span().addTagOrLog("keys.count", CACHE_API_UPDATE,
                () -> String.valueOf(invokeMap.size()));
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/atomic/GridDhtAtomicCache.java
```diff
@@ -1038,91 +1050,108 @@ private IgniteInternalFuture updateAll0(
...
+            MTC.span().addTagOrLog("keys.count", CACHE_API_UPDATE,
+                () -> m == null ? "0" : String.valueOf(m.size()));
 
-        final CacheOperationContext opCtx = ctx.operationContextPerCall();
+            Map<? extends K, ? extends EntryProcessor> invokeM = invokeMap;
```
ascherbakoff, line 1061: invokeMap can be declared as final in the argument list.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/atomic/GridDhtAtomicCache.java
sanpwc: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/atomic/GridDhtAtomicCache.java
sanpwc: Fixed.

PR #8235, opened 2020-09-10
IGNITE-13425 Add task name to the log messages related to this compute task

Wuchte:

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskProcessor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskProcessor.java
@@ -83,6 +83,7 @@ import org.apache.ignite.lang.IgniteUuid;
 import org.apache.ignite.marshaller.Marshaller;
 import org.apache.ignite.plugin.security.SecurityPermission;
 import org.apache.ignite.spi.systemview.view.ComputeTaskView;
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import static org.apache.ignite.events.EventType.EVT_MANAGEMENT_TASK_STARTED;
@@ -1142,7 +1143,7 @@ public class GridTaskProcessor extends GridProcessorAdapter implements IgniteCha
 
             if (stopping && !waiting) {
                 U.warn(log, "Received job execution response while stopping grid (will ignore): " + msg
-                    + (task != null && task.getSession() != null ? (", " + task.getSession().getTaskName()) : ""));
+                    + tryResolveTaskName(task));
 
                 return;
             }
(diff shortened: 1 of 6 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskProcessor.java
```diff
@@ -1138,14 +1138,15 @@ public void processJobExecuteResponse(UUID nodeId, GridJobExecuteResponse msg) {
         lock.readLock();
 
         try {
+            GridTaskWorker<?, ?> task = tasks.get(msg.getSessionId());
+
             if (stopping && !waiting) {
-                U.warn(log, "Received job execution response while stopping grid (will ignore): " + msg);
+                U.warn(log, "Received job execution response while stopping grid (will ignore): " + msg
+                    + (task != null && task.getSession() != null ? (", " + task.getSession().getTaskName()) : ""));
```
ascherbakoff, line 1145: The similar code block repeats 5 times. It should be refactored to separate method, like tryResolveTaskName(sesId)

Wuchte: Good idea.
However, I think it will be more clear to use task as parameter. Otherwise, here will be one more place which works with tasks map.

modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskProcessor.java
```diff
@@ -1138,14 +1139,15 @@ public void processJobExecuteResponse(UUID nodeId, GridJobExecuteResponse msg) {
         lock.readLock();
 
         try {
+            GridTaskWorker<?, ?> task = tasks.get(msg.getSessionId());
```
ascherbakoff, line 1142: Why this is not inlined in method argument, like

tryResolveTaskName(tasks.get(msg.getSessionId()) ?

The same for other calls.

modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskProcessor.java
```diff
@@ -1530,4 +1536,9 @@ else if (msg instanceof GridTaskSessionRequest)
             }
         }
     }
+
+    /** Tries to get task name in appended form(after ', '). */
+    @NotNull private static String tryResolveTaskName(GridTaskWorker<?, ?> task) {
```
ascherbakoff, line 1541: Missing _param_.
Missing _Nullable_ annotation on method arg.

modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskProcessor.java
```diff
@@ -1530,4 +1536,9 @@ else if (msg instanceof GridTaskSessionRequest)
...
+        return task != null && task.getSession() != null ? (", " + task.getSession().getTaskName()) : "";
```
ascherbakoff, line 1542: This looks dirty.
I suggest to create a method like
formatMessage(String msg, Nullable GridTaskWorker<?, ?> task)

PR #8262, opened 2020-09-21
IGNITE-13465 : Ignite cluster falls apart if two nodes segmented sequentially

Vladsz83:
Fixed the timeout resetting.


Vladsz83:
Behavior changes:
The connection recovery timeout is sliced into smaller parts in order to check several next nodes in the ring while recovering conenction. This avoids case when connectionRecoveryTimeout is spent on only one next node while 2 or more next nodes failed what causes segmentaion of previous alive nodes.

PR #8277, opened 2020-09-24
IGNITE-13431 Fix NPE when PRIMITIVE strategy is used with BLOB value.

alamar:


modules/cassandra/store/src/main/java/org/apache/ignite/cache/store/cassandra/persistence/PersistenceSettings.java
```diff
@@ -544,14 +544,14 @@ private void readObject(java.io.ObjectInputStream in) throws IOException, ClassN
      */
     protected List<F> enrichFields(List<F> src) {
         if (src != null) {
-            List<F> enriched = new ArrayList<>(src.size());
+            List<F> enriched = new ArrayList<>();
```
AMashenkov, line 547: Why are you drop ArrayList initial size to default one?

alamar: This list is only created once per cache initialization, and is expected to have small size (<1000 entries), making initial size specification superfluous in my opinion. Do you have any specific reason why this exact ArrayList needs initial size?

PR #8283, opened 2020-09-25
IGNITE-13487 Move logging of delegated errors to debug level.

dmekhanikov:
Logging of exceptions happening in the Compute Engine is moved to the debug level in cases when the error is being delegated to caller site.


modules/core/src/main/java/org/apache/ignite/internal/processors/job/GridJobWorker.java
```diff
@@ -484,7 +483,8 @@ boolean initialize(GridDeployment dep, Class<?> taskCls) {
             job = SecurityUtils.sandboxedProxy(ctx, ComputeJob.class, job);
         }
         catch (IgniteCheckedException e) {
-            U.error(log, "Failed to initialize job [jobId=" + ses.getJobId() + ", ses=" + ses + ']', e);
+            if (log.isDebugEnabled())
+                U.error(log, "Failed to initialize job [jobId=" + ses.getJobId() + ", ses=" + ses + ']', e);
```
nizhikov, line 487: I think we should use the debug level here and in all other changed places.

nizhikov: Besides LGTM.

dmekhanikov: There is no implementation of `log.debug` that accepts exceptions. Is there a good way to print stacktraces to log in debug level?

nizhikov: Makes sense. Let's keep it as for now .

PR #8311, opened 2020-10-03
Represent in the documenttion affection of several node addresses on failure detection.

Vladsz83:
We should emphasize that TcpDiscoverySpi prolongs detection of node failure if several IP addresses are set for a node. Actual failure detection delay is: `failureDetectionTimeout * addressesNumber`.


Vladsz83:
Wrong branch

PR #8336, opened 2020-10-08
IGNITE-13531 [ML]: Code cleanup in Util classes

mrk-andreev:
- add a private constructorto Util classes in ignite ml.
- add Serializable to Blas class
- add final modifier to static inner fields in utils class
- inefficient use of keySet iterator instead of entrySet
- can be replaced with single Arrays.fill method call

Issue: https://issues.apache.org/jira/browse/IGNITE-13531


zaleslaw:
LGTM!

PR #8347, opened 2020-10-12
IGNITE-13534 fixing of "using deprecated active() and uninformative message in console"

aries-3:
- uninformative messages corrected
 - each using .active() changed to .state() != ClusterState.INACTIVE
 - each using .active(boolean b) changed to .state(b ? ClusterState.ACTIVE :
ClusterState.INACTIVE)

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
@@ -2158,8 +2158,8 @@ public class IgniteKernal implements IgniteEx, IgniteMXBean, Externalizable {
         }
 
         if (ctx.state().clusterState().state() == ClusterState.INACTIVE) {
-            U.quietAndInfo(log, ">>> Ignite cluster is not active (limited functionality available). " +
-                "Use control.(sh|bat) script or IgniteCluster interface to activate.");
+            U.quietAndInfo(log, ">>> Ignite cluster is in INACTIVE state (limited functionality available). " +
+                "Use control.(sh|bat) script or IgniteCluster.state(ClusterState.ACTIVE) to change the state.");
         }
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -4490,7 +4490,7 @@ private void registerMetrics() {
         reg.register("longJVMPauseLastEvents", this::getLongJVMPauseLastEvents, Map.class,
             LONG_JVM_PAUSE_LAST_EVENTS_DESC);
 
-        reg.register("active", () -> ctx.state().clusterState().active()/*this::active*/, Boolean.class,
+        reg.register("active", () -> ctx.state().clusterState().state() != ClusterState.INACTIVE, Boolean.class,
```
alamar, line 4493: Can we do ctx.state().clusterState().state().active() here?

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -3795,7 +3795,7 @@ private void checkNearCacheStarted(IgniteCacheProxy<?, ?> cache) throws IgniteCh
 
     /** {@inheritDoc} */
     @Override public void active(boolean active) {
-        cluster().active(active);
+        cluster().state(active ? ClusterState.ACTIVE : ClusterState.INACTIVE);
```
alamar, line 3798: Let's withhold this change.  Both Ignite.active() and IgniteCluster.active() are deprecated. So it's OK for one to map to the former. If there's a warning inside IgniteCluster.active(), it's even better that way.

alamar (changes requested):
Please see comments

PR #8395, opened 2020-10-26
IGNITE-13620 : Bind ignite node to 1 address in the ducktests.

Vladsz83:
Ignite node should use one ip address in rcp discovery to make failure detection delay clear.


Vladsz83:
Better version is rolling up.

PR #8398, opened 2020-10-27
IGNITE-13620 : Bind ignite node to 1 address in the ducktests.

Vladsz83:
Iginite node should use 1 address to make diascovery test results clear.


Vladsz83:
wrong branch

PR #8424, opened 2020-11-03
IGNITE-13663 : Represent in the documenttion affection of several node addresses on failure detection v2.

Vladsz83:
We should document that TcpDiscoverySpi prolongs detection of node failure if node has several addresses. By default, all available addresses are assigned to node and node listens any address (0.0.0.0). Actual failure detection delay of this node is: `failureDetectionTimeout * addressesNumber + connRecoveryTimeout`. 

Often, middleware runs in environments with several IP addresses (virtualizations, containers, different networks). Node sends all obtained addresses with other node info to the cluster. Connection to node is established to first of its addresses. But if lost, other addresses are attempted to reconnect sequentially. If addresses do not belong to assumed node network, do not represent existing physical connection, processing them is just waste of time.


dmagda:
@Vladsz83, should we encourage the developer to initialize the TcpCommunicationSpi.localAddresses as well? So that the nodes use only reachable network interfaces for communication needs.

docs/_docs/clustering/network-configuration.adoc
```diff
@@ -49,10 +49,18 @@ tab:C++[unsupported]
 The following table describes some most important properties of `TcpDiscoverySpi`.
 You can find the complete list of properties in the javadoc:org.apache.ignite.spi.discovery.tcp.TcpDiscoverySpi[] javadoc.
 
+[CAUTION]
+====
+You should assign `IgniteConfiguration.localHost` or `TcpDiscoverySpi.localAddress`.
```
dmagda, line 54: Vlad,

Good explanation, crystal clear and straight to the point. I made some tweaks. Please use this final version below and see if we need to mention `TcpCommunicationSpi.localAddress` along with `TcpDiscoverySpi.localAddress`

_You should initialize the `IgniteConfiguration.localHost` or `TcpDiscoverySpi.localAddress` parameter only with those network interfaces that will be used for inter-node communication. By default, a node binds to and listens on all available IP addresses of an environment it's running on. It can prolong node failures detection if some of the node's addresses are not reachable from other cluster nodes._

docs/_docs/clustering/network-configuration.adoc
```diff
@@ -49,10 +49,18 @@ tab:C++[unsupported]
...
+By default, node obtains all available IP addresses. Several node addresses can prolong detection of node failure.
+You should not allow node bind to multiple ip addresses until they represent some real physical connections which can
+give you more reliability.
+====
+
 [cols="1,2,1",opts="header"]
 |===
 |Property | Description| Default Value
-| `localAddress`| Local host IP address used for discovery. | By default, the node uses the first non-loopback address it finds. If there is no non-loopback address available, then `java.net.InetAddress.getLocalHost()` is used.
+| `localAddress`| Local host IP address used for discovery. Prevails over `IgniteConfiguration.localHost`. | By default, all available addresses are assigned to node. If there is no non-loopback address available, then `java.net.InetAddress.getLocalHost()` is used.
```
dmagda, line 63: Please adjust to the following:

_... If set, overrides the `IgniteConfiguration.localHost` setting. | By default, a node binds to all available network addresses. If there is a non-loopback address available, then `java.net.InetAddress.getLocalHost()` is used._

docs/_docs/clustering/network-configuration.adoc
```diff
@@ -111,7 +119,7 @@ You can find the list of all properties in the javadoc:org.apache.ignite.spi.com
 [cols="1,2,1",opts="header"]
 |===
 |Property | Description| Default Value
-| `localAddress` | The local address for the communication SPI to bind to. |
+| `localAddress` | The local address for the communication SPI to bind to. Prevails over `IgniteConfiguration.localHost`. |
```
dmagda, line 122: Please update to:

_... If set, overrides `IgniteConfiguration.localHost` setting._

Vladsz83:
> @Vladsz83, should we encourage the developer to initialize the TcpCommunicationSpi.localAddresses as well? So that the nodes use only reachable network interfaces for communication needs.

We should encourage user to set `IgniteConfiguration.localHost`. This one is read buy other componentc. ZooKeeperSpi as an example. No reason to setup localAddress everywhere.

docs/_docs/clustering/network-configuration.adoc
Vladsz83: Fixed with the exception we can bind only to one net. interface currently. Or automatically to all.

docs/_docs/clustering/network-configuration.adoc
Vladsz83: Fixed.

docs/_docs/clustering/network-configuration.adoc
Vladsz83: Fixed.

PR #8430, opened 2020-11-06
IGNITE-13662 : Describe soLinger setting in TCP Discovery and SSL issues.

Vladsz83:
Describes `soLinger` setting of `TcpDiscoverySpi`. Notes of known SSL issues of socket deadlock in the SSL configuraton.


docs/_docs/clustering/network-configuration.adoc
```diff
@@ -56,6 +56,7 @@ You can find the complete list of properties in the javadoc:org.apache.ignite.sp
 | `localPort`  | The port that the node binds to. If set to a non-default value, other cluster nodes must know this port to be able to discover the node. | `47500`
 | `localPortRange`| If the `localPort` is busy, the node attempts to bind to the next port (incremented by 1) and continues this process until it finds a free port. The `localPortRange` property defines the number of ports the node will try (starting from `localPort`).
    | `100`
+| `soLinger`| Setting linger-on-close can help with socket deadlocks of SSL issues like JDK-8219658. But costs longer detection of node failure. | `0`
```
dmagda, line 59: I guess the SSL issue is not the only reason of why developers might want to tweak this flag. How about this description? It's enough to mention the SSL issue on the SSL page only and leave the description in this place generic.

_Specifies a linger-on-close timeout. This option disables/enables immediate return from a close() of a TCP Socket. Setting the timeout to `0` closes the socket immediately. See the Javadoc for more details._

docs/_docs/security/ssl-tls.adoc
```diff
@@ -32,6 +32,12 @@ To enable SSL/TLS for cluster nodes, configure an `SSLContext` factory in the no
 You can use the `org.apache.ignite.ssl.SslContextFactory`, which is the default factory that uses a configurable keystore to initialize the SSL context.
 //You can also implement your own `SSLContext` factory.
 
+[NOTE]
+====
```
dmagda, line 36: I would put it this way:

_Consider setting the `TcpDiscovery.soLinger` parameter to X to avoid SSL-related deadlocks on certain versions of JRE. Alternatively, update your JRE version to the latest one._

And do the following with this text:

1. Turn the "SSL-related deadlock" phrase into an external link: https://github.com/apache/ignite/tree/master/docs#links-to-external-resources
2. The callout needs to be of the "caution" type: https://github.com/apache/ignite/tree/master/docs#callouts

docs/_docs/clustering/network-configuration.adoc
Vladsz83: I don't think so because socket is an internal implementation of the SPI. User shouldn't know about it. As example, there is no linger option for the communication SPI. Its internals are NIO. Linger was introduced only to have a workaround for SSL deadlock issues. It seems there was no need and are no need to configure it.

dmagda: The linger setting is a TCP parameter that changes how the protocol behaves. Even if the initial motivation for adding it to the discovery SPI was to address the SSL issue, the setting still needs to be explained generically because it helps you to control the TCP behavior for all sorts of needs.

The workarounds are usually handled via system properties in Ignite. As long as you've already added a new API method to the public interface it needs to be explained generically. Otherwise, you can remove the linger setting from the API and create a system property for the workaround instead.

dmagda: Moreover, you already have this workaround mentioned on the SSL page below that is a proper place. To recap, we need a generic description of this soLinger parameter on this page and have the workaround added as a warning callout on the SSL page.

Vladsz83: > The linger setting is a TCP parameter that changes how the protocol behaves.

linger only sets socket behavior on socket closing with rest of the data in TCP buffers. But this doesn’t affect behavior of Discovery because it has own message acknowledgement. The exception is the delay of Discovery. That is why I think duplicating of Java API documentation is useless for Ignite user. This parameter doesn’t change general logic and gives no benefits to user.

> The workarounds are usually handled via system properties in Ignite. As long as you've already added a new API method to the public interface .

I haven't. This API had been added in 2.8 as hotfix of raised SSL issue. But hadn't been documented. We just changed it's default value to disabled like before 2.8.

dmagda: The value of the parameter [is passed into the Socket via its Java API](https://github.com/apache/ignite/blob/master/modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoverySpi.java#L1687), thus the description on this page needs to sound generic like in Java. Our [JavaDoc follows this guideline](https://ignite.apache.org/releases/latest/javadoc/org/apache/ignite/spi/discovery/tcp/TcpDiscoverySpi.html#setSoLinger-int-) and doesn't mention any workarounds for specific use cases.

We can put everything together this way mentioning the SSL issue as a reason why this parameter set to 0 for Ignite discovery sockets:

_Specifies a linger-on-close timeout. This option disables/enables immediate return from a close() of a TCP Socket. The option defaults to `0` for Ignite TcpDiscovery sockets to avoid [potential deadlocks with SSL connections](https://bugs.openjdk.java.net/browse/JDK-8219658)._ 

Please take this all into consideration and send another pull-request. We've already beat this to death. The initial pull-request was merged skipping a review of technical documentation committers and maintainers.

Vladsz83:
@dmagda , new doc improvement PR: https://github.com/apache/ignite/pull/8452 . Please review. Didn't only manage to make a caution within the table.

PR #8431, opened 2020-11-06
IGNITE-13633 Fix ServiceDescriptor.serviceClass failure in case of de…

alex-plekhanov:
…ployed via DeploymentSPI service

- [x] There is a single JIRA ticket related to the pull request. 
- [x] The web-link to the pull request is attached to the JIRA ticket.
- [x] The JIRA ticket has the _Patch Available_ state.
- [x] The pull request body describes changes that have been made. 
- [x] The pull request title is treated as the final commit message. 
- [x] A reviewer has been mentioned through the JIRA comments 
- [x] The pull request has been checked by the Teamcity Bot and 

### Notes

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceInfo.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceInfo.java
@@ -21,6 +21,8 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.UUID;
 import org.apache.ignite.IgniteException;
+import org.apache.ignite.internal.GridKernalContext;
+import org.apache.ignite.internal.managers.deployment.GridDeployment;
 import org.apache.ignite.internal.util.tostring.GridToStringInclude;
 import org.apache.ignite.internal.util.typedef.internal.S;
 import org.apache.ignite.lang.IgniteUuid;
@@ -36,6 +38,9 @@ import org.jetbrains.annotations.Nullable;
 public class ServiceInfo implements ServiceDescriptor {
     /** */
     private static final long serialVersionUID = 0L;
+
+    /** Context. */
+    private final GridKernalContext ctx;
 
     /** Origin node ID. */
     private final UUID originNodeId;
@@ -53,13 +58,16 @@ public class ServiceInfo implements ServiceDescriptor {
     @GridToStringInclude
     private volatile Map<UUID, Integer> top;
 
+    /** Service class. */
+    private volatile Class<? extends Service> srvcCls;
+
     /**
      * @param originNodeId Initiating node id.
      * @param srvcId Service id.
      * @param cfg Service configuration.
      */
-    public ServiceInfo(@NotNull UUID originNodeId, @NotNull IgniteUuid srvcId, @NotNull ServiceConfiguration cfg) {
-        this(originNodeId, srvcId, cfg, false);
+    public ServiceInfo(GridKernalContext ctx, @NotNull UUID originNodeId, @NotNull IgniteUuid srvcId, @NotNull ServiceConfiguration cfg) {
+        this(ctx, originNodeId, srvcId, cfg, false);
     }
 
     /**
@@ -68,8 +76,9 @@ public class ServiceInfo implements ServiceDescriptor {
      * @param cfg Service configuration.
      * @param staticCfg Statically configured flag.
      */
-    public ServiceInfo(@NotNull UUID originNodeId, @NotNull IgniteUuid srvcId, @NotNull ServiceConfiguration cfg,
+    public ServiceInfo(@NotNull GridKernalContext ctx, @NotNull UUID originNodeId, @NotNull IgniteUuid srvcId, @NotNull ServiceConfiguration cfg,
         boolean staticCfg) {
+        this.ctx = ctx;
         this.originNodeId = originNodeId;
         this.srvcId = srvcId;
         this.cfg = cfg;
@@ -116,15 +125,28 @@ public class ServiceInfo implements ServiceDescriptor {
     }
 
     /** {@inheritDoc} */
-    @SuppressWarnings("unchecked")
     @Override public Class<? extends Service> serviceClass() {
         if (cfg instanceof LazyServiceConfiguration) {
+            if (srvcCls != null)
+                return srvcCls;
+
             String clsName = ((LazyServiceConfiguration)cfg).serviceClassName();
 
             try {
-                return (Class<? extends Service>)Class.forName(clsName);
+                srvcCls = (Class<? extends Service>)Class.forName(clsName);
+
+                return srvcCls;
             }
             catch (ClassNotFoundException e) {
+                GridDeployment srvcDep = ctx.deploy().getDeployment(clsName);
+
+                if (srvcDep != null) {
+                    srvcCls = (Class<? extends Service>)srvcDep.deployedClass(clsName);
+
+                    if (srvcCls != null)
+                        return srvcCls;
+                }
+
                 throw new IgniteException("Failed to find service class: " + clsName, e);
             }
         }
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceInfo.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceInfo.java
@@ -40,7 +40,7 @@ public class ServiceInfo implements ServiceDescriptor {
     private static final long serialVersionUID = 0L;
 
     /** Context. */
-    private final transient GridKernalContext ctx;
+    private transient volatile GridKernalContext ctx;
 
     /** Origin node ID. */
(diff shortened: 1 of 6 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/service/ServiceHotRedeploymentViaDeploymentSpiTest.java
```diff
@@ -61,12 +66,6 @@
         return cfg;
     }
 
-    /** */
-    @BeforeClass
```
daradurvs, line 65: I think that we shouldn't remove this check because this test is relevant only for the new Service Grid implementation.
We can remove such checks with the legacy SG-processor in future.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceInfo.java
```diff
@@ -37,6 +39,9 @@
     /** */
     private static final long serialVersionUID = 0L;
 
+    /** Context. */
+    private final GridKernalContext ctx;
```
daradurvs, line 43: `ServiceInfo` is a serializable unit.
I'm not sure that there is a sense to transfer `GridKernalContext` instance.

Maybe it's better to make the field transient and initialize it after deserialization.
What do you sink?

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceInfo.java
```diff
@@ -116,15 +125,28 @@ public IgniteUuid serviceId() {
...
             String clsName = ((LazyServiceConfiguration)cfg).serviceClassName();
 
             try {
-                return (Class<? extends Service>)Class.forName(clsName);
+                srvcCls = (Class<? extends Service>)Class.forName(clsName);
+
+                return srvcCls;
             }
             catch (ClassNotFoundException e) {
+                GridDeployment srvcDep = ctx.deploy().getDeployment(clsName);
```
daradurvs, line 141: We have similar logic here `IgniteServiceProcessor#copyAndInject`

Can't we reuse it?

modules/core/src/test/java/org/apache/ignite/internal/processors/service/ServiceHotRedeploymentViaDeploymentSpiTest.java
alex-plekhanov: New test added to this class `testServiceDeploymentViaDeploymentSpi` which also relevant to legacy implementation, so I've moved this check to `serviceHotRedeploymentTest` to skip legacy implementation only for `testServiceHotRedeploymentNode` and `testServiceHotRedeploymentThinClient`

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceInfo.java
alex-plekhanov: Ok, I will fix it.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceInfo.java
alex-plekhanov: It's only one line of code can be reused, I think it's not such a good idea to make a new method to reuse only `ctx.deploy().getDeployment(clsName)`.

daradurvs: I didn't mean copy/paste :)

I thought that there is another way to load className.

For example, load and set a className externally (without KernalContext field in `ServiceInfo`).
Not sure if `IgniteServiceProcessor#copyAndInject` is a suitable place for that or not.

modules/core/src/test/java/org/apache/ignite/internal/processors/service/ServiceHotRedeploymentViaDeploymentSpiTest.java
daradurvs: Got it.
Thank you for your clarification.

daradurvs: Got it.
Thank you for your clarification.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceInfo.java
alex-plekhanov: I've tried to implement class resolving logic in `IgniteServiceProcessor` before `ServiceInfo` constructor and pass it as an argument, but I think it's risky if we fail in `processServicesChangeRequest`, so I've moved class resolving logic to  `ServiceInfo.serviceClass()` method.

modules/core/src/main/java/org/apache/ignite/internal/processors/service/ServiceInfo.java
alex-plekhanov: We can't get context after deserialization, because we don't know Ignite instance at this stage. I've marked the field as transient and added a check for null to avoid NPE. I think this is the best we can do. The class resolving logic will not work for services with lazy configuration and deployed via deployment SPI, after serialization/deserialization of `ServiceInfo`. But I think it's not such a common case - `ServiceInfo` not serialized by Ignite and serialization of `ServiceInfo` by the user looks strange. I think this class should not be `Serializable` at all.

daradurvs: We can handle it in `IgniteServiceProcessor` in places where `registeredServices#put`is called.
Adding to `registeredServices` is the first place when Ignite's node receives `ServiceInfo`

We can introduce a new method:
```
private void registerService(ServiceInfo desc) {
    desc.setContext(ctx);
    registeredServices.put(desc.serviceId(), desc);
}
```

What do you think?

alex-plekhanov: I have missed the point. `ctx` already passed to `ServiceInfo` constructor right before `registeredServices#put`. And this works fine, the only problem will show up if the user will serialize and deserialize `ServiceInfo` items in some circumstances. But this problem is not solved by your proposal either.

daradurvs: I've tried to cover the case with statically configured services, which will be received during node joining.

In this case server (not user) will deserialize `ServiceInfo` without calling the new constructor.

Look at:
https://github.com/apache/ignite/blob/8f824ca5822d4d7d0b9eb12aa726c32a56776bbc/modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java#L350

I think we need a test for this case too.

alex-plekhanov: Thanks, fixed it, and added a check to the test.

I found one more problem with services using classes deployed by deployment SPI: such services can't be deployed to the joined node. It can be reproduced if we change in `testServiceDeploymentViaDeploymentSpi` from `deployClusterSingleton` to `deployNodeSingleton`. We can't register class in deployment SPI before node starts, so, after new node joins it receives services information from the cluster and tries to start service locally, but can't find deployment SPI classes, so a registration of such services fails. This issue is not related to this thicket. It looks conceptual and I don't have any ideas on how to solve it right now. 

Also, looks like deployment SPI is not supported by the old service processor implementation, so I've reverted my changes related to GridServiceProcessor. I don't know why TC bot visa was green, I've double-checked it manually, and test `testServiceDeploymentViaDeploymentSpi` failed. This issue can be easily fixed, if we copy-paste `copyAndInject` method from the new implementation, but since the old implementation is deprecated I think we can live with these known issues.

daradurvs: Looks like `desc.context(ctx)` should be called here as well:
1. https://github.com/apache/ignite/blob/1ce95c603e15aa81d665c0be5b2b89466982d0fb/modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java#L389

2. https://github.com/apache/ignite/blob/1ce95c603e15aa81d665c0be5b2b89466982d0fb/modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java#L1519

I'd suggest to introduce a new method:
```
    private void registerService(ServiceInfo desc) {
        desc.context(ctx);
        
        registeredServices.put(desc.serviceId(), desc);
    }
```
It'll allow us to not miss such places in future.

What do you think?

alex-plekhanov: Ok, fixed

daradurvs (approved):

PR #8468, opened 2020-11-17
IGNITE-13705 : Another node fails with failure of target node.

Vladsz83:
Scenario:
The nodes have relative places in the ring: N and N+1. Node N detects failure of node N+1. Node N tries to connect to node N+2. Node N+2 checks backward connection to node N+1.

Problem: 
Node N can fail too.

Cause: 
The timeout on node N to recover connection to node N+2 appears shorter than timeout on node N+2 to check connection to N+1.

Fix:
Introduced a fundamental timeout value to check/recover connection based on current configuration. The mentioned timeouts have been turned relative. The timeout of backward connection check is now generally shorter than the timeout to recover connection.

Additions:
Brought some logs to have diagnostoc ability. It was hard to realize the issue without them.


Vladsz83:
Better solution in: https://github.com/apache/ignite/pull/8484

PR #8499, opened 2020-11-26
IGNITE-13760 .NET: Fix NullPointerException in GetAffinity on client nodes

ptupitsyn:
Cache context does not exist on client nodes unless `GetCache` was called, so `GridCacheAffinityManager` retrieval in `PlatformAffinity` constructor can cause NPE.

`GridCacheAffinityManager` was added to `PlatformAffinity` solely for Platform Cache needs and is not used otherwise, so it makes sense to move this functionality to a dedicated `PlatformAffinityManager` class. This fixes the bug, because cache context always exists by the time Platform Cache is created.


isapego (approved):
Looks good to me.

PR #8503, opened 2020-11-27
IGNITE-13680 maxSwappiness declared as int in suggestions

nyshubin:
IGNITE-13680 maxSwappiness declared as int in OsConfigurationSuggestions.java

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/suggestions/OsConfigurationSuggestions.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/suggestions/OsConfigurationSuggestions.java
@@ -78,10 +78,10 @@ public class OsConfigurationSuggestions {
 
             if ((value = readVmParam(SWAPPINESS)) != null) {
                 try {
-                    double maxSwappiness = 10.0;
+                    int maxSwappiness = 10;
 
-                    if (Float.parseFloat(value) > maxSwappiness)
-                        suggestions.add(String.format("Reduce pages swapping ratio (set vm.%s=%f or less)", SWAPPINESS,
+                    if (Integer.parseInt(value) > maxSwappiness)
+                        suggestions.add(String.format("Reduce pages swapping ratio (set vm.%s=%d or less)", SWAPPINESS,
                                                       maxSwappiness));
                 }
                 catch (NumberFormatException ignored) {
--- a/modules/core/src/main/java/org/apache/ignite/internal/suggestions/OsConfigurationSuggestions.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/suggestions/OsConfigurationSuggestions.java
@@ -61,11 +61,11 @@ public class OsConfigurationSuggestions {
         List<String> suggestions = new ArrayList<>();
 
         if (U.isRedHat()) {
-            String value;
-            String expected = "500";
+            String val;
+            String exp = "500";
 
-            boolean dwcParamFlag = (value = readVmParam(DIRTY_WRITEBACK_CENTISECS)) != null && !value.equals(expected);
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/ignite/internal/suggestions/OsConfigurationSuggestions.java
```diff
@@ -61,11 +61,11 @@
         List<String> suggestions = new ArrayList<>();
 
         if (U.isRedHat()) {
```
alamar, line 63: Can we perhaps do this for all Linux VMs and not just Red Hat? I have Ubuntu but all of these VM params are there.

modules/core/src/main/java/org/apache/ignite/internal/suggestions/OsConfigurationSuggestions.java
```diff
@@ -74,28 +74,28 @@
...
             }
 
-            if ((value = readVmParam(ZONE_RECLAIM_MODE)) != null && !value.equals(expected = "0"))
+            if ((val = readVmParam(ZONE_RECLAIM_MODE)) != null && !val.equals(exp = "0"))
                 suggestions.add(String.format("Disable NUMA memory reclaim (set vm.%s=%s)", ZONE_RECLAIM_MODE,
-                    expected));
+                    exp));
 
-            if ((value = readVmParam(EXTRA_FREE_KBYTES)) != null && !value.equals(expected = "1240000"))
+            if ((val = readVmParam(EXTRA_FREE_KBYTES)) != null && !val.equals(exp = "1240000"))
```
alamar, line 96: Can we also get rid of this check (and corresponding constant?) It seems to be a proprietary Red Hat property abandoned a long time ago, and the hard-coded value is extremely suspicious.

modules/core/src/main/java/org/apache/ignite/internal/suggestions/OsConfigurationSuggestions.java
```diff
@@ -60,12 +60,12 @@
...
-        if (U.isRedHat()) {
-            String value;
-            String expected = "500";
+        if (U.isLinux()) {
+            String val;
+            String exp = "500";
 
-            boolean dwcParamFlag = (value = readVmParam(DIRTY_WRITEBACK_CENTISECS)) != null && !value.equals(expected);
-            boolean decParamFlag = (value = readVmParam(DIRTY_EXPIRE_CENTISECS)) != null && !value.equals(expected);
+            boolean dwcParamFlag = (val = readVmParam(DIRTY_WRITEBACK_CENTISECS)) != null && !val.equals(exp);
```
alamar, line 67: Can we parse the value as int and do a "less than" comparison? val < exp (500)
Otherwise we may be telling people to slow down their writeback. Applies for both.

alamar (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/suggestions/OsConfigurationSuggestions.java
nyshubin: Cast exp and val to int and replaced .equals comparison with val > exp (500)
The gridgain guide suggests: As a solution, you can try changing the page flushing interval from the default 30 seconds to 5 seconds

modules/core/src/main/java/org/apache/ignite/internal/suggestions/OsConfigurationSuggestions.java
nyshubin: replaced this check with int comparison, but did not remove it completely, because gridgain doc suggests:
To work around the effects caused by page memory reclamation on Linux, add extra bytes between wmark_min and wmark_low with /proc/sys/vm/extra_free_kbytes

PR #8509, opened 2020-11-29
IGNITE-10075 .NET: Avoid binary configurations of Ignite Java service params

nizhikov:
This PR contains changes to avoid explicit type registration while call java service from .Net.
The type will be automatically registered if FQN equals for .Net type and for Java companion.

### Notes


modules/core/src/main/java/org/apache/ignite/internal/processors/platform/binary/PlatformBinaryProcessor.java
```diff
@@ -126,16 +128,28 @@ public PlatformBinaryProcessor(PlatformContext platformCtx) {
...
+                        String typeName = platformContext().kernalContext().marshallerContext()
+                            .getClassName(platformId, typeId);
+
+                        writer.writeString(typeName);
+
+                        err = null;
+
+                        break;
+                    }
+                    catch (ClassNotFoundException e) {
```
ptupitsyn, line 144: Let's not use exceptions for control flow - may be introduce an overload for `MarshallerContext.getClassName` that does not throw exceptions?

modules/core/src/main/java/org/apache/ignite/internal/MarshallerContextImpl.java
```diff
@@ -368,12 +369,27 @@ public void onMappingAccepted(final MarshallerMappingItem item) {
 
     /** {@inheritDoc} */
     @Override public Class getClass(int typeId, ClassLoader ldr) throws ClassNotFoundException, IgniteCheckedException {
-        String clsName = getClassName(JAVA_ID, typeId);
+        String clsName;
+
+        ClassNotFoundException err = null;
+
+        for (byte platformId : new byte[] {JAVA_ID, DOTNET_ID}) {
```
ptupitsyn, line 376: `JAVA_ID` should come last - in most cases, platform type will be resolved, not a Java type.

ptupitsyn (changes requested):
Changes look good to me in general - please see a couple comments.

However, this seems to cover a very rare use case when Java and .NET full type names match. Naming conventions are different, so a user has to go out of their way to break them. Are we sure that this is useful?

modules/core/src/main/java/org/apache/ignite/internal/MarshallerContextImpl.java
nizhikov: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/binary/PlatformBinaryProcessor.java
nizhikov: I see your point, but if we introduce a method that not throws then we should return a tuple from it - T2(className, errorMsg).
Moreover, the second value of the tuple required in rare cases when we can't resolve the class name from type id.

AFAIU `getClassName` executed several times on every interaction between DotNet and Java so the creation of an extra object on each invocation will hurt the performance.

So I propose to keep changes as is. What do you think?

ptupitsyn: 1. The cost of throwing an exception is **a lot** higher than a tuple allocation.
2. `getClassName` is called from .NET once per type, then cached

I'm not concerned about performance too much here because of (2). Exception-based control flow is just dirty.

nizhikov: Fixed according to your proposal.

nizhikov:
> Are we sure that this is useful?

I and @kukushal have a request for a feature from one of the Ignite users that building an application that uses Java services from the DotNet platform.

ptupitsyn (approved):

nizhikov:
Hello, @ptupitsyn 

I've reworked this patch.
After some discussion with @kukushal I came to the decision that the initial patch (searching java type with .net name during deserialization) is bad.
Because we enhance global logic but not the logic that relates to the .net -> java service execution.

So I changed the algorithm to the following: during .net -> java service execution we register a new binary type both for java and .net platforms. We expect that there is java analog for each .Net type.

What do you think?
Can you, please, review the patch one more time.

ptupitsyn (changes requested):
The idea of `registerSameJavaType` is fine with me. I agree that it is a better solution.

I think we should simplify the implementation - instead of passing the flag around in all those places, use a thread local flag in `Marshaller`, then in `ServiceProxySerializer.WriteProxyMethod` do something like
```
try
{
    writer.Marshaller.RegisterSameJavaTypes = true;
    // Write args
}
finally
{
    writer.Marshaller.RegisterSameJavaTypes = false;
}
```

nizhikov:
@ptupitsyn 

> use a thread local flag in Marshaller

Implemented as suggested.
Please, take a look one more time.

ptupitsyn:
@nizhikov thanks, one more adjustment please:

I think it is a bit cleaner to pass the thread local flag as a parameter to the `IBinaryProcessor.RegisterType`, which is called from a single place in `Marshaller`. That way we reduce the global state usage.

nizhikov:
> I think it is a bit cleaner to pass the thread local flag as a parameter to the IBinaryProcessor.RegisterType

done.

ptupitsyn (approved):

ptupitsyn:
Looks good to me, thanks!

PR #8533, opened 2020-12-02
IGNITE-10655 .NET: Add IgniteConfiguration.JavaPeerClassLoadingEnabled

ashapkin:


modules/platforms/dotnet/Apache.Ignite.Core/IgniteConfiguration.cs
```diff
@@ -1712,5 +1717,20 @@ public int SqlQueryHistorySize
         /// </summary>
         [SuppressMessage("Microsoft.Usage", "CA2227:CollectionPropertiesShouldBeReadOnly")]
         public ICollection<ExecutorConfiguration> ExecutorConfiguration { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether peer class loading is enabled for <b>Java</b> side.
+        /// <para/>
+        /// <b>Important!</b> Do not mix this property with <see cref="PeerAssemblyLoadingMode"/>.
```
ptupitsyn, line 1724: `Do not mix` -> `Do not confuse`

ptupitsyn: I would add something like `Peer assembly loading in .NET and peer class loading in Java are two distinct and independent features.`

ptupitsyn (changes requested):

PR #8546, opened 2020-12-06
IGNITE-8884 .NET: Fix async key-val operations - use WriteObjectDetached

ptupitsyn:
Fix async cache operations when key and value objects reference each other or have references to the same object.
Async key-val operations used `WriteObject` instead `WriteObjectDetached`, so references to the same inner object were shared in the binary stream (referenced object is written once). However, cache stores key and val binary objects separately, so the reference to the inner object gets broken.

`WriteObjectDetached` disables reference sharing and writes both object independently.


isapego (approved):
Looks good to me.

PR #8566, opened 2020-12-11
Improve parrallel start of Ignite nodes in discovery ducktape test to reduce tests duration.

Vladsz83:
Parrallel start of Ignite nodes is slightly improoved to reduce tests time.


Vladsz83:
Wrong source branch

PR #8577, opened 2020-12-16
IGNITE-13856 make linear performance for DirectByteBufferStreamImplV2.writeString

aries-3:


modules/core/src/main/java/org/apache/ignite/internal/direct/stream/v2/DirectByteBufferStreamImplV2.java
```diff
@@ -301,6 +302,9 @@
     /** */
     protected boolean lastFinished;
 
+    /** map for cashing byte-array representations of strings */
+    private Map<String, byte[]> stringsMap;
```
akalash, line 306: Why are you using a map here? A simple byte array is enough, isn't it?

modules/core/src/main/java/org/apache/ignite/internal/direct/stream/v2/DirectByteBufferStreamImplV2.java
```diff
@@ -584,7 +588,29 @@ public DirectByteBufferStreamImplV2(MessageFactory msgFactory) {
...
+            if (buf.capacity() < val.length()) {
+                if (stringsMap == null)
+                    stringsMap = new HashMap<>();
+
+                byte[] bytes = stringsMap.computeIfAbsent(val, s -> s.getBytes());
+
+                try {
+                    writeByteArray(bytes);
+                }
+                catch (Exception e) {
```
akalash, line 601: I believe if an exception appears here, this whole object will be invalid, for example, arrOff would be incorrect, so perhaps it's not necessary to handle this exception here.

modules/core/src/main/java/org/apache/ignite/internal/direct/stream/v2/DirectByteBufferStreamImplV2.java
```diff
@@ -584,7 +587,21 @@ public DirectByteBufferStreamImplV2(MessageFactory msgFactory) {
 
     /** {@inheritDoc} */
     @Override public void writeString(String val) {
-        writeByteArray(val != null ? val.getBytes() : null);
+        if (val != null) {
+            if (buf.capacity() < val.length()) {
```
akalash, line 591: I don't think that this condition is right. for example if 'buf.capacity() > val.length()' and 'buf.remaining() < val.length()' that it still requires to cache the bytes but in your case this won't happen. So maybe it is better to remove that condition at all, I don't think that writing to the local variable is too expensive so you can do it every time even it isn't required for certain case. Or you should change you condition to the proper one.

modules/core/src/main/java/org/apache/ignite/internal/direct/stream/v2/DirectByteBufferStreamImplV2.java
```diff
@@ -301,6 +301,9 @@
     /** */
     protected boolean lastFinished;
 
+    /** byte-array representation of string */
+    private byte[] byteArr;
```
alamar, line 305: Can we perhaps make it a `private volatile` and name smth like `curStrBackingArr`? I'm not confident that it will be called from the same thread.

modules/core/src/main/java/org/apache/ignite/internal/direct/stream/v2/DirectByteBufferStreamImplV2.java
```diff
@@ -584,7 +587,17 @@ public DirectByteBufferStreamImplV2(MessageFactory msgFactory) {
...
+            if (byteArr == null)
+                byteArr = val.getBytes();
+
+            writeByteArray(byteArr);
+
+            if (lastFinished)
+                byteArr = null;
+        }
+        else
+            writeByteArray(null);
```
alamar, line 600: can we perhaps declare byteArr a local variable, and only assign it to curStrBackingArr if not lastFinished? This way we will not touch this variable at all when string fits one frame.

alamar (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/direct/stream/v2/DirectByteBufferStreamImplV2.java
alamar: Actually, we need to reset it to null if lastFinished. Need to figure out the logic, maybe it's too cumbersome and confusing to implement.

alamar: Disregard this, we are going to read curStrBackingArr value either way. I think we may skip this change entirely, just rename var / mark volatile.

modules/core/src/main/java/org/apache/ignite/internal/direct/stream/v2/DirectByteBufferStreamImplV2.java
alamar: @akalash says volatile is not needed. Can you please just rename this field as you see fit?

alamar (changes requested):

PR #8581, opened 2020-12-16
IGNITE-11110 fixed stop case with collision SPI enabled.

antkr:


modules/core/src/main/java/org/apache/ignite/internal/processors/job/GridJobProcessor.java
```diff
@@ -377,7 +377,10 @@ public GridJobProcessor(GridKernalContext ctx) {
     /** {@inheritDoc} */
     @Override public void stop(boolean cancel) {
         // Clear collections.
-        activeJobs.clear();
+        if (jobAlwaysActivate)
+            activeJobs.clear();
```
alamar, line 381: Can we please put this logic in a method, do not use clear in both cases?
I.e. line 309 and stop() should use the same logic to create a fresh map.

alamar (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/job/GridJobProcessor.java
antkr: Makes sense, pushed an update.

PR #8627, opened 2020-12-29
IGNITE-13883 .NET: Simplify and optimize TryReadSystemType

ptupitsyn:
Replace dictionary+interfaces with a simple switch statement in `BinarySystemHandlers.TryReadSystemType`:
* Less code
* Simpler
* Faster

Benchmark added: `BinarySystemTypeReadBenchmark`; results on Core i7-9700K, Ubuntu 20.04, .NET Core 3.1:

```
|        |     Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |---------:|----------:|----------:|-------:|------:|------:|----------:|
| Before | 1.080 us | 0.0030 us | 0.0025 us | 0.0725 |     - |     - |     456 B |
|  After | 849.4 ns |   4.44 ns | 4.16 ns   | 0.0725 |     - |     - |     456 B |
```

(~20% improvement)


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Binary/BinaryUtils.cs
```diff
@@ -1170,6 +1170,38 @@ public static T[] ReadArray<T>(BinaryReader ctx, bool typed)
...
+        /// </summary>
+        /// <param name="stream">Stream</param>
+        /// <returns>String array.</returns>
+        public static string[] ReadStringArray(IBinaryStream stream)
+        {
+            var len = stream.ReadInt();
+            var res = new string[len];
+
+            for (var i = 0; i < len; i++)
+                res[i] = stream.ReadByte() == HdrNull ? null : ReadString(stream);
```
isapego, line 1184: Should not we at least assert here that header is `HdrString`, if not `HdrNull`?

isapego (approved):

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Binary/BinaryUtils.cs
ptupitsyn: Good point, added header checks to this and related methods.

PR #8636, opened 2021-01-08
IGNITE-13963 default constructor for reflection needs for Platf…

antkr:
…ormDotNetSessionLockResult.

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/platform/BinaryContextPlatformDotNetSessionLockResultPredefinedTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/platform/BinaryContextPlatformDotNetSessionLockResultPredefinedTest.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


modules/core/src/test/java/org/apache/ignite/platform/BinaryContextPlatformDotNetSessionLockResultPredefinedTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.ignite.internal.binary.TestCachingMetadataHandler;
+import org.apache.ignite.internal.processors.platform.websession.PlatformDotNetSessionLockResult;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Tests for registering PlatformDotNetSessionLockResult class.
+ */
+public class BinaryContextPlatformDotNetSessionLockResultPredefinedTest {
+
```
alamar, line 34: Blank line not needed

modules/core/src/test/java/org/apache/ignite/platform/BinaryContextPlatformDotNetSessionLockResultPredefinedTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.ignite.internal.binary.BinaryContext;
+import org.apache.ignite.internal.binary.TestCachingMetadataHandler;
+import org.apache.ignite.internal.processors.platform.websession.PlatformDotNetSessionLockResult;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Tests for registering PlatformDotNetSessionLockResult class.
+ */
+public class BinaryContextPlatformDotNetSessionLockResultPredefinedTest {
```
alamar, line 33: Can we also add a generic test that no constructors are created during default node startup and that ReflectionFactory is not used? In this case the test may be renamed to BinaryContextConstructorGenerationTest perhaps.

Can you also add a custom class with no public empty constructor, add it to binary context, and make sure that for it we generate two distict constructors as per ReflectionFactory?

alamar (changes requested):

PR #8638, opened 2021-01-09
IGNITE-13754 .NET: Fix LINQ provider for queries with JOIN and GROUP BY combined

ptupitsyn:
Fix projection traversal in `AliasDictionary.GetQuerySource` and `CacheQueryExpressionVisitor.VisitMember`: use `memberHint` to understand where the given member of the projected type comes from to emit correct table and field name.


modules/platforms/dotnet/Apache.Ignite.Linq/Impl/CacheQueryExpressionVisitor.cs
```diff
@@ -676,5 +683,20 @@ private void VisitArguments(IEnumerable<Expression> arguments)
...
             }
         }
+
+        /// <summary>
+        /// Determines whether specified type is an anonymous type.
+        /// <para />
+        /// There is no 100% way to identify an anonymous type, since it is up to the compiler implementation,
+        /// so we rely on some well-known anonymous type properties.
+        /// </summary>
+        private static bool IsAnonymousType(Type type)
```
gurustron, line 693: Maybe add check for `CompilerGeneratedAttribute` ?

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/Query/Linq/CacheLinqTest.GroupBy.cs
```diff
@@ -0,0 +1,253 @@
...
+
+            // Ordering and count
+            var res1 =
+                from p in persons
+                orderby p.Value.Name
+                group p by p.Value.OrganizationId
+                into gs
+                orderby gs.Key
+                where gs.Count() > 10
+                select new { Count = gs.Count(), OrgId = gs.Key, AvgAge = gs.Average(x => x.Value.Age) };
```
gurustron, line 104: This throws if we change it to:

```
        public class MyClass
        {
            public int Count { get; set; }
            public int OrgId { get; set; }
            public double AvgAge { get; set; }
        }

        .....
       select new MyClass { Count = ....

```

Should we support such scenario also?

ptupitsyn: Good point, we should not actually limit our projection support to anonymous types.
* Anonymous type check removed
* Custom class projection test added
* Member init expression support added

gurustron: @ptupitsyn looks good to me. 
```

PR #8648, opened 2021-01-12
IGNITE-13805

xtern:

### Notes


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreCacheGroupProcess.java
```diff
@@ -0,0 +1,647 @@
...
+            opCtx0.interrupt(reason);
+    }
+
+    /**
+     * Ensures that a cache with the specified name does not exist locally.
+     *
+     * @param name Cache name.
+     * @throws IllegalStateException If cache with the specified name already exists.
+     */
+    private void ensureCacheAbsent(String name) throws IllegalStateException {
```
Mmuzaf, line 242: I don't think we should declare `IllegalStateException` in the method header and it can be removed. The method is too small to check the and we can see all that is going on.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreCacheGroupProcess.java
```diff
@@ -0,0 +1,647 @@
...
+                    continue;
+
+                snapshotMgr.restoreCacheGroupFiles(snpName, grpName, grpDir, this::interrupted, grps.get(grpName));
+            }
+        }
+
+        /**
+         * Rollback changes made by process in specified cache group.
+         */
+        protected void rollback() {
```
Mmuzaf, line 633: We also should not place such methods in the context. Let's pass the `ctx` into this method if we should do any actions.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -1022,6 +1056,91 @@ private SnapshotMetadata readSnapshotMetadata(File smf) {
...
+        }
+        catch (IOException e) {
+            throw new IgniteCheckedException("Unable to copy file [snapshot=" + snpName + ", grp=" + grpName + ']', e);
+        }
+    }
+
+    /**
+     * @param files Collection of files to delete.
+     */
+    protected void rollbackRestoreOperation(Collection<File> files) {
```
Mmuzaf, line 1124: Let's move it to `SnapshotRestoreCacheGroupProcess`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreCacheGroupProcess.java
```diff
@@ -0,0 +1,647 @@
...
+    private void finishPrepare(UUID reqId, Map<UUID, SnapshotRestoreEmptyResponse> res, Map<UUID, Exception> errs) {
+        GridFutureAdapter<Void> fut0 = fut;
+
+        if (fut0.isDone() || !reqId.equals(opCtx.requestId()))
+            return;
+
+        Exception failure = F.first(errs.values());
+
+        if (failure != null) {
+            opCtx.rollback();
```
Mmuzaf, line 341: Deleting a large number of files may be a heavy operation.  We should use atomic operations like deleting the whole directory at once or perform such operations on executor service.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreCacheStartRequest.java
```diff
@@ -0,0 +1,52 @@
...
+package org.apache.ignite.internal.processors.cache.persistence.snapshot;
+
+import java.io.Serializable;
+import java.util.UUID;
+import org.apache.ignite.internal.util.typedef.internal.S;
+
+/**
+ * Request to start restored cache group.
+ */
+public class SnapshotRestoreCacheStartRequest implements Serializable {
```
Mmuzaf, line 27: Can you use directly the `UUID` instead of the creation `SnapshotRestoreCacheStartRequest` class?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+     */
+    protected void restore(SnapshotRestoreContext opCtx, boolean updateMeta) throws IgniteCheckedException {
+        BooleanSupplier stopChecker = () -> opCtx.err.get() != null;
+        String pdsFolderName = ctx.pdsFolderResolver().resolveFolders().folderName();
+
+        if (updateMeta) {
+            File binDir = binaryWorkDir(
+                ctx.cache().context().snapshotMgr().snapshotLocalDir(opCtx.snpName).getAbsolutePath(), pdsFolderName);
+
+            if (stopChecker.getAsBoolean())
```
Mmuzaf, line 483: I think this check makes sense only on heavy-background operations. Do we really need it here?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+                return;
+
+            // Check binary metadata compatibility.
+            ctx.cacheObjects().checkMetadata(binDir);
```
Mmuzaf, line 487: Is it possible to unite the `checkMetadata` and `updateMetadata` methods? They are used both each time.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+            assert opCtx0 != null : ctx.localNodeId();
+
+            Map<Integer, StoredCacheData> globalCfgs = new HashMap<>();
+
+            for (List<StoredCacheData> storedCfgs : res.values()) {
+                if (storedCfgs == null)
+                    continue;
+
+                for (StoredCacheData cacheData : storedCfgs)
+                    globalCfgs.put(CU.cacheId(cacheData.config().getName()), cacheData);
```
Mmuzaf, line 575: I think you can put directly to `opCtx0.cfgs`, use the `putIfAbsent` instead and enrich the configurations that you already have on the local node, isn't it?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+
+            opCtx0.cfgs = globalCfgs;
+
+            if (U.isLocalNodeCoordinator(ctx.discovery()))
+                cacheStartProc.start(reqId, reqId);
+
+            return;
+        }
+
+        if (opCtx0 == null)
```
Mmuzaf, line 586: if `opCtx0 == null && failure != null` you may return from this method earlier, isn't it? I think you should place this condition upper in the method body.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+            finishProcess(failure);
+        else // Remove files asynchronously.
+            ctx.cache().context().snapshotMgr().snapshotExecutorService().execute(() -> {
```
Mmuzaf, line 589: Probably this should be performed under the `rollback` distributed process since in case of any failures you won't receive successful/failure rollback results on other nodes.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+    /**
+     * Check the response for probable failures.
+     *
+     * @param errs Errors.
+     * @param opCtx Snapshot restore operation context.
+     * @param respNodes Set of responding topology nodes.
+     * @return Error, if any.
+     */
+    private Exception checkFailure(Map<UUID, Exception> errs, SnapshotRestoreContext opCtx, Set<UUID> respNodes) {
+        Exception err = F.first(errs.values());
```
Mmuzaf, line 698: If you return only the first exception to the user (the `restore` future will be completed with this exception, right?) probably you should log other exceptions which you are skipping.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+
+        GridFutureAdapter<Boolean> retFut = new GridFutureAdapter<>();
+
+        try {
+            Collection<StoredCacheData> ccfgs = opCtx0.cfgs.values();
+
+            // Ensure that shared cache groups has no conflicts before start caches.
+            for (StoredCacheData cfg : ccfgs) {
+                if (!F.isEmpty(cfg.config().getGroupName()))
+                    ensureCacheAbsent(cfg.config().getName());
```
Mmuzaf, line 635: I think the message which is thrown here is incorrect since you already have a guarantee that cache doesn't exist here. And if it does it means you violate your process guarantees - probably it's better the `IgniteIllegalStateException` throw here.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+            }
+
+            if (log.isInfoEnabled()) {
+                log.info("Starting restored caches " +
+                    "[requestID=" + opCtx0.reqId + ", snapshot=" + opCtx0.snpName +
+                    ", caches=" + F.viewReadOnly(ccfgs, c -> c.config().getName()) + ']');
+            }
+
+            ctx.cache().dynamicStartCachesByStoredConf(ccfgs, true, true, false, null, true, opCtx0.nodes).listen(
```
Mmuzaf, line 644: Can you return the `IgniteInternalFuture` directly here or `chain` the results if you need any additional result transformation?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+                finishProcess(failure);
+            });
+    }
+
+    /**
+     * @param reqId Request ID.
+     * @return Result future.
+     */
+    private IgniteInternalFuture<Boolean> cacheStart(UUID reqId) {
+        SnapshotRestoreContext opCtx0 = opCtx;
```
Mmuzaf, line 601: You should skip `client` nodes here and throw an exception if the `opCtx0` is `null` here, isn't it? The nullable context on cache start means incorrect restore process state.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+            return new GridFinishedFuture<>(
+                new IgniteCheckedException("Unknown snapshot restore operation was rejected."));
+        }
+
+        if (!U.isLocalNodeCoordinator(ctx.discovery()))
+            return new GridFinishedFuture<>();
+
+        DiscoveryDataClusterState state = ctx.state().clusterState();
+
+        if (state.state() != ClusterState.ACTIVE || state.transition())
```
Mmuzaf, line 616: I think you should not check the sate of the cluster each time you run the next stage. If the state is changed it might be better to set the correct exception for the `opCtx.err` field and check it each time you want to proceed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+     */
+    private IgniteInternalFuture<Boolean> cacheStart(UUID reqId) {
+        SnapshotRestoreContext opCtx0 = opCtx;
+
+        if (opCtx0 == null)
+            return new GridFinishedFuture<>();
+
+        if (!reqId.equals(opCtx0.reqId)) {
+            return new GridFinishedFuture<>(
+                new IgniteCheckedException("Unknown snapshot restore operation was rejected."));
```
Mmuzaf, line 608: The `restore` procedure state is incorrect if this condition occurred. I don't think another concurrent restore operation can be executed here, so it's better to fix the error messsage.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+
+        if (state.state() != ClusterState.ACTIVE || state.transition())
+            return new GridFinishedFuture<>(new IgniteCheckedException(OP_REJECT_MSG + "The cluster should be active."));
+
+        Throwable err = opCtx0.err.get();
+
+        if (err != null)
+            return new GridFinishedFuture<>(err);
+
+        if (!allNodesInBaselineAndAlive(opCtx0.nodes))
```
Mmuzaf, line 624: The same thing - is it better to set `opCtx.err` if any of the affected baseline nodes leave the cluster.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+        if (isSnapshotRestoring()) {
+            throw new IgniteCheckedException(OP_REJECT_MSG +
+                "The previous snapshot restore operation was not completed.");
+        }
+
+        GridCacheSharedContext<?, ?> cctx = ctx.cache().context();
+
+        SnapshotMetadata meta = F.first(cctx.snapshotMgr().readSnapshotMetadatas(req.snapshotName()));
+
+        if (meta != null && meta.consistentId().equals(cctx.localNode().consistentId().toString())
```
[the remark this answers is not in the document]
xtern: added explicit `meta==null` processing 
(`meta == null` means we should create "empty" context, `throw exception` means - don't create context at allt)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+            && meta.pageSize() != cctx.database().pageSize()) {
+            throw new IgniteCheckedException("Incompatible memory page size " +
+                "[snapshotPageSize=" + meta.pageSize() +
+                ", local=" + cctx.database().pageSize() +
+                ", snapshot=" + req.snapshotName() +
+                ", nodeId=" + cctx.localNodeId() + ']');
+        }
+
+        String pdsFolderName = ctx.pdsFolderResolver().resolveFolders().folderName();
```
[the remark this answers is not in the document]
xtern: done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+
+            ((FilePageStoreManager)cctx.pageStore()).readCacheConfigurations(snpCacheDir, cfgsByName);
+
+            File cacheDir = U.resolveWorkDirectory(ctx.config().getWorkDirectory(),
+                Paths.get(databaseRelativePath(pdsFolderName), snpCacheDir.getName()).toString(), false);
+
+            if (!cacheDir.exists())
+                cacheDir.mkdir();
+            else if (cacheDir.list().length > 0) {
+                throw new IgniteCheckedException("Unable to restore cache group, directory is not empty " +
```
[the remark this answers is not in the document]
xtern: Currently, cache destroy doesn't remove the local cache directory.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+            return new GridFinishedFuture<>();
+
+        if (log.isInfoEnabled()) {
+            log.info("Starting local snapshot restore operation [requestID=" + req.requestId() +
+                ", snapshot=" + req.snapshotName() + ", group(s)=" + req.groups() + ']');
+        }
+
+        GridFutureAdapter<ArrayList<StoredCacheData>> retFut = new GridFutureAdapter<>();
+
+        ctx.cache().context().snapshotMgr().snapshotExecutorService().execute(() -> {
```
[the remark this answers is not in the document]
xtern: added handling for `RejectedExecutionException`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+
+    /**
+     * Copy partition files and update binary metadata.
+     *
+     * @param opCtx Snapshot restore operation context.
+     * @param updateMeta Update binary metadata flag.
+     * @throws IgniteCheckedException If failed.
+     */
+    protected void restore(SnapshotRestoreContext opCtx, boolean updateMeta) throws IgniteCheckedException {
+        BooleanSupplier stopChecker = () -> opCtx.err.get() != null;
```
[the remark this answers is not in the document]
xtern: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
xtern: Isn't cluster-wide metadata registering a long enough operation?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
xtern: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
xtern: We reading and writing `cfgs` from different threads, so we should use ConcurrentMap instead of HashMap for this. But in this case, it is not rational (compared to a couple of volatile reads/writes of object reference).

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
xtern: It seems we should log local failures locally, I think it's better to add additional logging on each exception catching, what do you think?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
xtern: Changed to `IgniteIllegalStateException`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
xtern: added assrtion `opCtx != null`for server nodes

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
xtern: How to handle the transition from/to `read-only` mode?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
xtern: removed allNodesInBaselineAndAlive
(it seemed to me that sometimes `node left` callback notify later than such direct checking that's why I kept it)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
xtern: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+                return;
+
+            // Check binary metadata compatibility.
+            ctx.cacheObjects().checkMetadata(binDir);
+
+            // Cluster-wide update binary metadata.
+            ctx.cacheObjects().updateMetadata(binDir, stopChecker);
+        }
+
+        for (File cacheDir : opCtx.dirs) {
```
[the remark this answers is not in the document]
xtern: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
xtern: done, this check has been removed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,799 @@
...
+        // Collect cache configuration(s) and verify cache groups page size.
+        for (File snpCacheDir : cctx.snapshotMgr().snapshotCacheDirectories(req.snapshotName(), pdsFolderName)) {
+            String grpName = FilePageStoreManager.cacheGroupName(snpCacheDir);
+
+            if (!req.groups().contains(grpName))
+                continue;
+
+            ((FilePageStoreManager)cctx.pageStore()).readCacheConfigurations(snpCacheDir, cfgsByName);
+
+            File cacheDir = U.resolveWorkDirectory(ctx.config().getWorkDirectory(),
```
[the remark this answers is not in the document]
xtern: done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
xtern: done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,777 @@
...
+
+                    return true;
+                }
+
+                return false;
+            };
+
+            GridFutureAdapter<ArrayList<StoredCacheData>> retFut = new GridFutureAdapter<>();
+
+            restoreAsync(opCtx0.snpName, opCtx0.dirs, updateMeta, stopChecker, errHnd).thenAccept(res -> {
```
Mmuzaf, line 410: I think it would be more readable if you extract the functionality related to the `updateMeta` from this method.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestorePrepareRequest.java
```diff
@@ -0,0 +1,108 @@
...
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.Set;
+import java.util.UUID;
+import org.apache.ignite.internal.util.typedef.internal.S;
+
+/**
+ * Request to prepare cache group restore from the snapshot.
+ */
+public class SnapshotRestorePrepareRequest implements Serializable {
```
Mmuzaf, line 29: I think we can simplify it to > `SnapshotRestoreRequest`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestorePrepareRequest.java
```diff
@@ -0,0 +1,108 @@
...
+    private final String snpName;
+
+    /** Baseline node IDs that must be alive to complete the operation. */
+    private final Set<UUID> nodes;
+
+    /** List of cache group names to restore from the snapshot. */
+    private final Collection<String> grps;
+
+    /** Node ID from which to update the binary metadata. */
+    private final UUID updateMetaNodeId;
```
Mmuzaf, line 46: Do we need this property or maybe it's better to use the coordinator node always?

modules/core/src/main/java/org/apache/ignite/internal/processors/cacheobject/IgniteCacheObjectProcessor.java
```diff
@@ -306,6 +307,15 @@ public void updateMetadata(int typeId, String typeName, @Nullable String affKeyF
      */
     public void saveMetadata(Collection<BinaryType> types, File dir);
 
+    /**
+     * Merge the binary metadata files stored in the specified directory.
+     *
+     * @param metadataDir Directory containing binary metadata files.
+     * @param stopChecker Prcoess interrupt checker.
```
Mmuzaf, line 314: Prcoess > Process

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,777 @@
...
+        }
+        else if (log.isInfoEnabled()) {
+            log.info("Successfully restored cache group(s) from the snapshot" + (opCtx0 == null ? "" :
+                " [reqId=" + opCtx0.reqId + ", snapshot=" + opCtx0.snpName + ']'));
+        }
+
+        GridFutureAdapter<Void> fut0 = fut;
+
+        if (fut0 != null)
+            fut0.onDone(null, err);
```
Mmuzaf, line 304: Should we use the system pool from performing the user future completion and avoiding doing that in the discovery thread?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,777 @@
...
+    /**
+     * Finish local cache group restore process.
+     *
+     * @param err Error, if any.
+     */
+    private void finishProcess(@Nullable Throwable err) {
+        SnapshotRestoreContext opCtx0 = opCtx;
+
+        if (err != null) {
+            log.error("Failed to restore snapshot cache group" + (opCtx0 == null ? "" :
```
Mmuzaf, line 293: Can you please fix the Intellij IDEA suggestions here?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestorePrepareRequest.java
xtern: Coordinator may not have a local snapshot

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,777 @@
...
+
+        assert opCtx0 != null || failure != null : ctx.localNodeId();
+
+        if (opCtx0 == null) {
+            finishProcess(failure);
+
+            return;
+        }
+
+        if (failure == null)
```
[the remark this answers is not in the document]
xtern: We cannot rely on single node errors during the "finish"-phase of the distributed process. For example, one of the nodes may observe a node failure, and the other may not, in which case we will have different behavior on different nodes at the completion stage, so the decision is to track only those errors that are visible for all nodes.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,777 @@
...
+            failure = checNodeLeft(opCtx0.nodes, res.keySet());
```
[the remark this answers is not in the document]
xtern: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,777 @@
...
+            return new IgniteFinishedFutureImpl<>(new IgniteException(OP_REJECT_MSG + "The cluster should be active."));
+
+        if (!clusterState.hasBaselineTopology()) {
+            return new IgniteFinishedFutureImpl<>(
+                new IgniteException(OP_REJECT_MSG + "The baseline topology is not configured for cluster."));
+        }
+
+        IgniteSnapshotManager snpMgr = ctx.cache().context().snapshotMgr();
+
+        if (snpMgr.isSnapshotCreating()) {
```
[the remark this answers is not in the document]
xtern: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,777 @@
...
+
+        Collection<StoredCacheData> ccfgs = opCtx0.cfgs.values();
+
+        if (log.isInfoEnabled()) {
+            log.info("Starting restored caches " +
+                "[reqId=" + opCtx0.reqId + ", snapshot=" + opCtx0.snpName +
+                ", caches=" + F.viewReadOnly(ccfgs, c -> c.config().getName()) + ']');
+        }
+
+        return ctx.cache().dynamicStartCachesByStoredConf(ccfgs, true, true, false, null, true, opCtx0.nodes);
```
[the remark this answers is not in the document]
xtern: comment added
```
// We set the topology node IDs required to successfully start the cache, if any of the required nodes leave
// the cluster during the cache startup, the whole procedure will be rolled back.
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,777 @@
...
+            log.info("Successfully restored cache group(s) from the snapshot" + (opCtx0 == null ? "" :
+                " [reqId=" + opCtx0.reqId + ", snapshot=" + opCtx0.snpName + ']'));
+        }
+
+        GridFutureAdapter<Void> fut0 = fut;
+
+        if (fut0 != null)
+            fut0.onDone(null, err);
+
+        opCtx = null;
```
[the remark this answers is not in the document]
xtern: changed, but from my point of view there is no difference

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,777 @@
...
+        }
+    }
+
+    /**
+     * Copy partition files and update binary metadata.
+     *
+     * @param snpName Snapshot name.
+     * @param dirs Cache directories to restore from the snapshot.
+     * @param updateMeta Update binary metadata flag.
+     * @param stopChecker Prcoess interrupt checker.
```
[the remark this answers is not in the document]
xtern: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestorePrepareRequest.java
xtern: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cacheobject/IgniteCacheObjectProcessor.java
xtern: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
xtern: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
xtern: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,836 @@
...
+                if (!F.isEmpty(cfg.config().getGroupName()))
+                    ensureCacheAbsent(cfg.config().getName());
+            }
+
+            if (log.isInfoEnabled()) {
+                log.info("Starting local snapshot restore operation [reqId=" + req.requestId() +
+                    ", snapshot=" + req.snapshotName() + ", group(s)=" + req.groups() + ']');
+            }
+
+            boolean updateMeta = ctx.localNodeId().equals(req.updateMetaNodeId());
```
Mmuzaf, line 426: Let's in-line the `updateMeta` variable.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,832 @@
...
+            opCtx0.err.compareAndSet(null, new ClusterTopologyCheckedException(OP_REJECT_MSG +
+                "Required node has left the cluster [nodeId=" + leftNodeId + ']'));
+        }
+    }
+
+    /**
+     * Abort the currently running restore procedure (if any).
+     */
+    public void stop() {
+        interrupt(new NodeStoppingException("Node is stopping."), true);
```
Mmuzaf, line 326: The `cancel` flag should be used as an input parameter from the method above.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -869,6 +895,37 @@ else if (f1.error() instanceof IgniteSnapshotVerifyException)
...
+        kctx0.task().setThreadContext(TC_SUBGRID, bltNodes);
+
+        return kctx0.task().execute(SnapshotMetadataCollectorTask.class, name);
+    }
+
+    /**
+     * @param metas Nodes snapshot metadata.
+     * @return Future with the verification results.
+     */
+    IgniteInternalFuture<IdleVerifyResultV2> runSnapshotVerfification(Map<ClusterNode, List<SnapshotMetadata>> metas) {
```
Mmuzaf, line 920: `runSnapshotVerfification` > `runSnapshotVerification`

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
```diff
@@ -0,0 +1,774 @@
...
+import org.junit.Test;
+
+import static org.apache.ignite.internal.util.distributed.DistributedProcess.DistributedProcessType.RESTORE_CACHE_GROUP_SNAPSHOT_PREPARE;
+import static org.apache.ignite.internal.util.distributed.DistributedProcess.DistributedProcessType.RESTORE_CACHE_GROUP_SNAPSHOT_START;
+import static org.apache.ignite.testframework.GridTestUtils.runAsync;
+
+/**
+ * Snapshot restore tests.
+ */
+public class IgniteClusterSnapshotRestoreSelfTest extends AbstractSnapshotSelfTest {
```
Mmuzaf, line 67: Let's add a test, that unable to create a snapshot during the restore procedure.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
```diff
@@ -0,0 +1,774 @@
...
+    private static final long TIMEOUT = 15_000;
+
+    /** Binary type name. */
+    private static final String BIN_TYPE_NAME = "customType";
+
+    /** Static cache configurations. */
+    protected CacheConfiguration<?, ?>[] cacheCfgs;
+
+    /** Cache value builder. */
+    protected Function<Integer, Object> valBuilder = new IndexedValueBuilder();
```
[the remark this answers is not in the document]
xtern: JUnit creates a new instance of the class for each test run, so this variable will always have its default value.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
```diff
@@ -0,0 +1,774 @@
...
+        IgniteFuture<Void> fut =
+            ignite.snapshot().restoreSnapshot(SNAPSHOT_NAME, Collections.singleton(dfltCacheCfg.getName()));
+
+        GridTestUtils.assertThrowsAnyCause(
+            log, () -> fut.get(TIMEOUT), IgniteException.class, "The cluster should be active");
+    }
+
+    /** @throws Exception If failed. */
+    @Test
+    public void testClusterSnapshotRestoreDiffTopology() throws Exception {
```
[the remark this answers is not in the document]
xtern: renamed to `testClusterSnapshotRestoreOnBiggerTopology`

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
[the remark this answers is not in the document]
xtern: Done `testStartClusterSnapshotRestoreMultipleThreadsSameNode`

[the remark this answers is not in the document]
xtern: Done, `testStartClusterSnapshotRestoreMultipleThreadsDiffNode`

xtern: added `testCreateSnapshotDuringRestore`

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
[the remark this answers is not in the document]
xtern: Added test `testClusterSnapshotRestoreOnSmallerTopology`

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
```diff
@@ -0,0 +1,774 @@
...
+
+    /** @throws Exception If failed. */
+    @Test
+    public void testNodeFail() throws Exception {
+        checkTopologyChange(true);
+    }
+
+    /** @throws Exception If failed. */
+    @Test
+    public void testNodeJoin() throws Exception {
```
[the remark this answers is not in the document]
xtern: This set of tests has the name IgniteCluster**SnapshotRestore**SelfTest do we really need to duplicate `OnRestoreInProgress` for each test?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
```diff
@@ -0,0 +1,774 @@
...
+    }
+
+    /**
+     * @param state Cluster state.
+     * @param procType The type of distributed process on which communication is blocked.
+     * @param exCls Expected exception class.
+     * @param expMsg Expected exception message.
+     * @throws Exception if failed.
+     */
+    private void checkClusterStateChange(
```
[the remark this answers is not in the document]
xtern: It is not very clear why parameterization is better than the current approach.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
```diff
@@ -0,0 +1,774 @@
...
+    public void testParallelCacheStartWithTheSameNameOnStart() throws Exception {
+        checkCacheStartWithTheSameName(RESTORE_CACHE_GROUP_SNAPSHOT_START, CacheExistsException.class,
+            "Failed to start cache (a cache with the same name is already started):");
+    }
+
+    /**
+     * @param procType The type of distributed process on which communication is blocked.
+     * @throws Exception if failed.
+     */
+    private void checkCacheStartWithTheSameName(
```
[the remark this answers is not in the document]
xtern: It is not very clear why parameterization is better than the current approach.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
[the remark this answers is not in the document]
xtern: This is exactly what `checkClusterStateChange` is doing:
`testClusterStateChangeActiveReadonlyOnPrepare`
`testClusterStateChangeActiveReadonlyOnCacheStart`
`testClusterDeactivateOnPrepare`
`testClusterDeactivateOnCacheStart`

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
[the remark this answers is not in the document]
xtern: `testNodeJoin` -> `testNodeJoinDuringRestore`
`testNodeFail` -> `testNodeFailDuringRestore`

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
```diff
@@ -0,0 +1,774 @@
...
+        discoSpi.unblock();
+
+        GridTestUtils.assertThrowsAnyCause(log, () -> fut.get(TIMEOUT), ClusterTopologyCheckedException.class, null);
+
+        ensureCacheDirEmpty(2, dfltCacheCfg);
+    }
+
+    /** @throws Exception If failed. */
+    @Test
+    public void testBasicClusterSnapshotRestore() throws Exception {
```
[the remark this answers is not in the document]
xtern: Done, now we also check if the index can be used (after each test).

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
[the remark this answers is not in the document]
xtern: Why we shouldn't check SQL and indexing?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
```diff
@@ -0,0 +1,774 @@
...
+
+            ensureCacheDirEmpty(3, dfltCacheCfg);
+
+            fut0.get(TIMEOUT);
+
+            awaitPartitionMapExchange();
+
+            dfltCacheCfg = null;
+
+            GridTestUtils.assertThrowsAnyCause(
```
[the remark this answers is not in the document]
xtern: added `testNodeFailDuringFilesCopy`

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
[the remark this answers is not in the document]
xtern: I reviewed the test suite and came to the conclusion that some of the tests do not check keys at all, so it makes no sense to test them with SQL/indexing.
So, I split the test suite into `IgniteClusterSnapshotRestoreWithIndexingTest` (SQL/indexing) and `IgniteClusterSnapshotRestoreSelfTest` (core)

modules/core/src/main/java/org/apache/ignite/IgniteSnapshot.java
```diff
@@ -48,4 +50,13 @@
      * @return Future which will be completed when cancel operation finished.
      */
     public IgniteFuture<Void> cancelSnapshot(String name);
+
+    /**
+     * Restore cache group(s) from the snapshot.
```
Mmuzaf, line 55: Let's add to the descriptoin that caches that are planning to be restored must be destroyed manually by the user.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
```diff
@@ -0,0 +1,916 @@
...
+import static org.apache.ignite.internal.util.distributed.DistributedProcess.DistributedProcessType.RESTORE_CACHE_GROUP_SNAPSHOT_PREPARE;
+import static org.apache.ignite.internal.util.distributed.DistributedProcess.DistributedProcessType.RESTORE_CACHE_GROUP_SNAPSHOT_ROLLBACK;
+import static org.apache.ignite.internal.util.distributed.DistributedProcess.DistributedProcessType.RESTORE_CACHE_GROUP_SNAPSHOT_START;
+
+/**
+ * Distributed process to restore cache group from the snapshot.
+ */
+public class SnapshotRestoreProcess {
+    /** Temporary cache directory prefix. */
+    public static final String TMP_CACHE_DIR_PREFIX = ".tmp.snp.restore.";
```
Mmuzaf, line 82: I think it's better to use the `_` instead of `.` for the cache directory prefix. I also think that a dot should be at the first place of the temparary directory name due to we are not hiding temparary directories.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -5126,6 +5126,12 @@ public void onNodeLeft(final ClusterNode node) {
 
                             if (crd0 == null)
                                 finishState = new FinishState(null, initialVersion(), null);
+
+                            if (dynamicCacheStartExchange() &&
+                                exchActions.cacheStartRequiredAliveNodes().contains(node.id())) {
+                                exchangeGlobalExceptions.put(cctx.localNodeId(), new ClusterTopologyCheckedException(
```
Mmuzaf, line 5132: I think we should also set the `exchangeLocE` param.
Let's add a dedicated test for starting caches with the required alive nodes which is not related to the snapshot procedure.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreBaseTest.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.ignite.IgniteCache;
+import org.apache.ignite.binary.BinaryObjectBuilder;
+import org.apache.ignite.internal.IgniteEx;
+
+/**
+ * Snapshot restore test base.
+ */
+public abstract class IgniteClusterSnapshotRestoreBaseTest extends AbstractSnapshotSelfTest {
+    /** Timeout. */
+    protected static final long TIMEOUT = 15_000;
```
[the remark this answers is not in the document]
xtern: I don't think this is a good idea. This is a short timeout to fail tests faster.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreSelfTest.java
```diff
@@ -0,0 +1,774 @@
...
+            try {
+                startLatch.await(TIMEOUT, TimeUnit.MILLISECONDS);
+
+                grid(nodeIdxSupplier.getAsInt()).snapshot().restoreSnapshot(
+                    SNAPSHOT_NAME, Collections.singleton(DEFAULT_CACHE_NAME)).get(TIMEOUT);
+
+                successCnt.incrementAndGet();
+            }
+            catch (Exception ignore) {
+                // Expected exception.
```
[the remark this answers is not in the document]
xtern: There possible 2 exceptions, first about another process started, second (rare) - cache exists (if the second process will be delayed).
Do you suggest checking for exceptions manually?

modules/core/src/main/java/org/apache/ignite/IgniteSnapshot.java
xtern: description changed to
```
Restore cache group(s) from the snapshot.
<p>
<b>NOTE:</b> Cache groups to be restored from the snapshot must not present in the cluster, if they present, they must be destroyed by the user before starting this operation.
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotRestoreProcess.java
xtern: dots replaced with underscore

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreWithIndexingTest.java
```diff
@@ -0,0 +1,209 @@
...
+
+        ignite.context().cacheObjects().removeType(typeId);
+
+        forceCheckpoint();
+
+        ignite.snapshot().restoreSnapshot(SNAPSHOT_NAME, Collections.singleton(DEFAULT_CACHE_NAME)).get(TIMEOUT);
+
+        IgniteCache<Object, Object> cache = ignite.cache(DEFAULT_CACHE_NAME).withKeepBinary();
+
+        assertTrue(cache.indexReadyFuture().isDone());
```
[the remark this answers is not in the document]
xtern: Done

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotRestoreWithIndexingTest.java
```diff
@@ -0,0 +1,209 @@
...
+
+        IgniteEx client = startGridsWithSnapshot(2, CACHE_KEYS_RANGE, true);
+
+        grid(0).snapshot().restoreSnapshot(SNAPSHOT_NAME, Collections.singleton(DEFAULT_CACHE_NAME)).get(TIMEOUT);
+
+        IgniteCache<Object, Object> cache = client.cache(DEFAULT_CACHE_NAME);
+
+        assertTrue(cache.indexReadyFuture().isDone());
+
+        checkCacheKeys(cache, CACHE_KEYS_RANGE);
```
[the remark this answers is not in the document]
xtern: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
xtern: added DynamicCacheStartFailsOnNodeLeftTest

Mmuzaf (approved):

Mmuzaf (approved):

PR #8662, opened 2021-01-13
ducktape-better-params

Vladsz83:
Introduces soLinger for Ignite 2.8.1 to reduce node failure detection. + other minor params.

### DIFF
--- a/modules/ducktests/tests/ignitetest/services/utils/templates/discovery_macro.j2
+++ b/modules/ducktests/tests/ignitetest/services/utils/templates/discovery_macro.j2
@@ -39,6 +39,7 @@
     <bean class="org.apache.ignite.spi.discovery.zk.ZookeeperDiscoverySpi">
         <property name="zkConnectionString" value="{{ spi.connection_string }}"/>
         <property name="zkRootPath" value="{{ spi.root_path }}"/>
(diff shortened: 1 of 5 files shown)


modules/ducktests/tests/ignitetest/services/utils/templates/discovery_macro.j2
```diff
@@ -39,6 +39,7 @@
     <bean class="org.apache.ignite.spi.discovery.zk.ZookeeperDiscoverySpi">
         <property name="zkConnectionString" value="{{ spi.connection_string }}"/>
         <property name="zkRootPath" value="{{ spi.root_path }}"/>
+        <property name="sessionTimeout" value="{{ spi.session_timeout }}"/>
```
anton-vinogradov, line 42: What profit do we have here comparing to zookeeper server min_session_timeout setup?

Vladsz83: The documentation [1] says: 
"The following parameters are required (other parameters are optional):

zkConnectionString - keeps the list of addresses of ZooKeeper servers.

sessionTimeout - specifies the time after which an Ignite node is considered disconnected if it doesn’t react to events exchanged via Discovery SPI.​"

[1] https://ignite.apache.org/docs/latest/clustering/zookeeper-discovery

PR #8668, opened 2021-01-15
IGNITE-13560 Add node attribute colocated affinity backup filter

alex-plekhanov:

### Notes

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/cache/affinity/rendezvous/ClusterNodeAttributeColocatedBackupFilter.java
+++ b/modules/core/src/main/java/org/apache/ignite/cache/affinity/rendezvous/ClusterNodeAttributeColocatedBackupFilter.java
@@ -46,7 +46,7 @@ import org.apache.ignite.lang.IgniteBiPredicate;
  * attributes.
  *
  * Note: All nodes should have not an empty co-location attribute value. The absence of the attribute on some nodes
- * can lead to improper partitions co-location.
+ * will trigger the failure handler.
  *
  * Note: Node attributes persisted in baseline topology at the time of baseline topology change. If the co-location
  * attribute of some node was updated, but the baseline topology wasn't changed, the outdated attribute value can be
@@ -117,6 +117,9 @@ public class ClusterNodeAttributeColocatedBackupFilter implements IgniteBiPredic
         String primaryAttrVal = previouslySelected.get(0).attribute(attrName);
         String candidateAttrVal = candidate.attribute(attrName);
 
-        return (primaryAttrVal == null || candidateAttrVal == null) || primaryAttrVal.equals(candidateAttrVal);
+        if (primaryAttrVal == null || candidateAttrVal == null)
+            throw new IllegalStateException("Empty co-location attribute value");
+
+        return primaryAttrVal.equals(candidateAttrVal);
     }
 }
--- a/modules/core/src/main/java/org/apache/ignite/cache/affinity/rendezvous/ClusterNodeAttributeColocatedBackupFilter.java
+++ b/modules/core/src/main/java/org/apache/ignite/cache/affinity/rendezvous/ClusterNodeAttributeColocatedBackupFilter.java
@@ -18,7 +18,6 @@
 package org.apache.ignite.cache.affinity.rendezvous;
 
 import java.util.List;
-import java.util.Objects;
 import org.apache.ignite.cluster.ClusterNode;
 import org.apache.ignite.internal.util.typedef.internal.A;
 import org.apache.ignite.lang.IgniteBiPredicate;
@@ -46,8 +45,8 @@ import org.apache.ignite.lang.IgniteBiPredicate;
  * Note: All cluster nodes, on startup, automatically register all the environment and system properties as node
  * attributes.
  *
- * Note: If the attribute is absent on the node, it will be treated as {@code null} attribute value and copies of
- * partitions assigned to this node will be co-located on other nodes with {@code null} attribute value.
+ * Note: All nodes should have not an empty co-location attribute value. The absence of the attribute on some nodes
+ * can lead to improper partitions co-location.
  *
  * Note: Node attributes persisted in baseline topology at the time of baseline topology change. If the co-location
  * attribute of some node was updated, but the baseline topology wasn't changed, the outdated attribute value can be
@@ -103,9 +102,10 @@ public class ClusterNodeAttributeColocatedBackupFilter implements IgniteBiPredic
     /**
      * Defines a predicate which returns {@code true} if a node is acceptable for a backup
      * or {@code false} otherwise. An acceptable node is one where its attribute value
-     * is exact match with previously selected nodes.  If an attribute does not
-     * exist on exactly one node of a pair, then the attribute does not match.  If the attribute
-     * does not exist both nodes of a pair, then the attribute matches.
+     * is exact match with previously selected nodes. If an attribute does not
+     * exist on candidate node, then the attribute matches any attribute values of previously
+     * selected nodes. If the attribute does not exist on primary node, then the attribute matches
+     * any attribute value of candidate node.
      *
      * @param candidate          A node that is a candidate for becoming a backup node for a partition.
      * @param previouslySelected A list of primary/backup nodes already chosen for a partition.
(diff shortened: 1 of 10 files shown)


modules/core/src/main/java/org/apache/ignite/cache/affinity/rendezvous/ClusterNodeAttributeColocatedBackupFilter.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param attrName The attribute name for the attribute to compare.
+     */
+    public ClusterNodeAttributeColocatedBackupFilter(String attrName) {
+        this.attrName = attrName;
+    }
+
+    /**
+     * Defines a predicate which returns {@code true} if a node is acceptable for a backup
+     * or {@code false} otherwise. An acceptable node is one where its attribute value
+     * is exact match with previously selected nodes.  If an attribute does not
```
ascherbakoff, line 100: I strongly suggest to deny null attributes in this filter and add the validation of attribute value existence, otherwise it's possible to lose data if a node was restarted without an attribute due to loss of owners.

Checking not null value and throwing an error to trigger failure handler in case of absence will do.

Javadoc should be modified to reflect a requirement of the attribute.

Even better would be to replace attribute filter closure with something like this for correct attribute validation:
```
/**
 * A basic class for node attribute aware backup filter implementations.
 * Listed attributes are required to be configured on new cluster node joining or first activation.
 */
public abstract class NodeAttributeAwareBackupFilter implements IgniteBiPredicate<ClusterNode, List<ClusterNode>> {
    /** Attribute names. */
    protected final Set<String> attributeNames;

    /**
     * @param attributeNames The list of attribute names for the set of attributes to compare. Must be at least one.
     */
    protected NodeAttributeAwareBackupFilter(String... attributeNames) {
        A.ensure(attributeNames.length > 0, "attributeNames.length > 0");

        this.attributeNames = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(attributeNames)));
    }

    /**
     * @return The set of used cluster node attributes for this filter.
     */
    public Set<String> attributeNames() {
        return attributeNames;
    }
}
```

Ideally attribute values for backup filter should be persisted in the node's metastore to avoid accidental changes (without direct user request), or maybe persisted in cache configuration after cache creation.

These improvements can be done in separate tickets.

modules/core/src/main/java/org/apache/ignite/cache/affinity/rendezvous/ClusterNodeAttributeColocatedBackupFilter.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param candidate          A node that is a candidate for becoming a backup node for a partition.
+     * @param previouslySelected A list of primary/backup nodes already chosen for a partition.
+     *                           The primary is first.
+     */
+    @Override public boolean apply(ClusterNode candidate, List<ClusterNode> previouslySelected) {
+        for (ClusterNode node : previouslySelected)
+            return Objects.equals(candidate.attribute(attrName), node.attribute(attrName));
+
+        return true;
+    }
```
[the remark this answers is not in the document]
alex-plekhanov: Fixed

modules/core/src/main/java/org/apache/ignite/cache/affinity/rendezvous/ClusterNodeAttributeColocatedBackupFilter.java
alex-plekhanov: If a node was restarted without an attribute value it will be treated as another cell with an empty attribute value, so, there will be no data lost (but sometimes rebalance). If we will shutdown the nodes by failure handler in case of a node with an empty attribute value joined then all cluster nodes will be failed one by one in some, unexpected to user moment (when the coordinator is changed, or when a new cache is started, or when baseline topology is changed, but not when the node with wrong attribute value joined the cluster). I think it's even worse than allow null attribute values.

I didn't get how the code in this snippet will help to avoid null attribute values. It checks attribute names at construction time, but not attribute values of joined nodes. 

If we will store attribute value in the metastore it will be unclear to the user how to reconfigure it (even with documentation it will be complicated and risky).

ascherbakoff: 1. Not agreed on this. I've attached a test reproducing data loss then null attribute is allowed. I want to have some way of protection from this. Data loss or stopped grid, obviously second is "better". Javadoc must contain clear statement on avoiding null values. Not setting it always a user mistake. You should add some tests for FH triggering. It must trigger FH for any case of affinity recalculation on affected cache. Of course, better to have some way for eager validation , see p2.
2. This filter provides a method _attributeNames_, which can be used for attribute value validation on node joining or grid activation.
3. Not true. We have a tool for managing grid called _control.sh_. Attribute management can be implemented here in clear way. 
4. I've attached a test reproducing broken affinity calculation due to absence of proper attribute lifecycle.

I'm ok if 2, 3, 4 will be addressed later, in separate tickets.

alex-plekhanov: Agreed, with null attribute value data loss can happen in some circumstances. But null attribute value it's only one special case of misconfiguration, there are much more such cases exist when misconfiguration can lead to data loss (wrong attribute value for example). 
Also, I think we should not shut down the cluster at some unexpected moment due to misconfigured node joined some time ago. For example, the node with the empty attribute can join during maintenance time and nothing will happen. Later, during high load time coordinator can leave the cluster and at this time all cluster nodes will be stopped by the failure handler one by one.
If we want to handle only a special case of "null attribute value" I have another proposal: allow collocation on null attribute value nodes and any other attribute value nodes, for example, like this:
```
        String primaryAttrVal = previouslySelected.get(0).attribute(attrName);
        String candidateAttrVal = candidate.attribute(attrName);

        return (primaryAttrVal == null || candidateAttrVal == null) || primaryAttrVal.equals(candidateAttrVal);
``` 
In this case, there will be no data loss and the cluster will be available. What do you think?

ascherbakoff: This solution doesn't look good to me because it ruins the whole idea of cells.
A user may detect broken data placement later and will be forced to stop grid and reload data to fix it.
From my point of view, the grid stopping is quite predictable, and should only happen on grid activation and node joining (with automatic baseline change) or node addition to baseline. For me, better fail fast in this case when pretend everything is good.

alex-plekhanov: Of course, we should note about avoiding such misconfiguration in javadoc, but, I think, delayed cluster failure is even worse than ruining the cell. It's not necessary to reload the data in case of broken data placement, data will be rebalanced after fixing the configuration. AFAIK node joining without cluster deactivation and baseline change it's a common case, in this case, there will be no affinity recalculation and node failure due to failure handler, but nodes will fail when the coordinator is suddenly changed. Coordinator failure is unpredictable (it can be hardware or software failure), and the whole cluster failure caused by one node failure it's a very undesirable case. Also, there are other cases exist. For example, if there was no attribute defined for some nodes in baseline, but it was defined for online nodes, when such nodes left the grid and coordinator changed there will be the whole cluster failure again. So, if it's happened during high load hours I think it's better to have unexpected rebalancing than unexpected cluster failure. WDYT?

ascherbakoff: "AFAIK node joining without cluster deactivation and baseline change it's a common case, in this case, there will be no affinity recalculation" - this statement doesn't look correct to me.

Overall, I see no point in arguing further. 
For me, better fail fast when have a grid in unpredictable state.

My statement is the grid will fail fast due to FH on activation if some nodes are misconfigured, or if a new misconfigured node is added to baseline. 

Can you validate this statement? If it's true better stick with my proposal, if not - I'll be ok to have misconfigured nodes grouped with others.

alex-plekhanov: I have just consulted with some Ignite administrators and they confirmed that usually they join nodes to the cluster without deactivation. So at least on some deployments it's possible to get delayed cluster failure.

alex-plekhanov: I've done some tests and found that affinity calculation is also called on a joined node. In case of an empty attribute, it will be detected by the new node first.
So, I've changed the implementation to trigger failure handler in case of the empty attribute value, it will shut down the joined node and there will be no delayed cluster failure.
Please have a look again.

PR #8684, opened 2021-01-21
Ignite-13980 : Remove duplicated ping: status check message.

Vladsz83:


Vladsz83:
Split into #8696 and #8697

PR #8688, opened 2021-01-22
IGNITE-14039 Add warning to runtime, javadoc and documentation on WAL enable/disable.

alamar:
Also fixed WAL being incorrectly considered disabled after cache was destroyed and recreated.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/IgniteCluster.java
+++ b/modules/core/src/main/java/org/apache/ignite/IgniteCluster.java
@@ -527,6 +527,12 @@ public interface IgniteCluster extends ClusterGroup, IgniteAsyncSupport {
      * when all dirty pages are prepared for checkpoint, but not necessarily flushed to disk.
      * <p>
      * WAL state can be changed only for persistent caches.
+     * <p>
+     * <b>NOTE:</b>
+     * Currently, this method should only be called on a stable topology when no nodes are leaving or joining cluster,
+     * and all baseline nodes are present.
+     * Cache may be stuck in inconsistent state due to violation of these conditions. It is advised to destroy
+     * such cache.
      *
      * @param cacheName Cache name.
      * @return Whether WAL disabled by this call.
@@ -545,6 +551,12 @@ public interface IgniteCluster extends ClusterGroup, IgniteAsyncSupport {
      * when all data is persisted to disk.
      * <p>
      * WAL state can be changed only for persistent caches.
+     * <p>
+     * <b>NOTE:</b>
+     * Currently, this method should only be called on a stable topology when no nodes are leaving or joining cluster,
+     * and all baseline nodes are present.
+     * Cache may be stuck in inconsistent state due to violation of these conditions. It is advised to destroy
+     * such cache.
      *
      * @param cacheName Cache name.
      * @return Whether WAL enabled by this call.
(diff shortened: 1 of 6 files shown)


modules/core/src/main/java/org/apache/ignite/IgniteCluster.java
```diff
@@ -527,6 +527,12 @@
      * when all dirty pages are prepared for checkpoint, but not necessarily flushed to disk.
      * <p>
      * WAL state can be changed only for persistent caches.
+     * <p>
+     * <b>NOTE:</b>
+     * Currently, this method should only be called on a stable topology when no nodes are leaving or joining cluster,
```
Mmuzaf, line 532: @alamar should we also add todo comment with the reference to the original issue?

modules/core/src/main/java/org/apache/ignite/IgniteCluster.java
```diff
@@ -527,6 +527,12 @@
...
+     * and all baseline nodes are present.
+     * Cache may be stuck in inconsistent state due to violation of these conditions. It is advised to destroy
+     * such cache.
```
Mmuzaf, line 535:
```suggestion
     * such cache.
    TODO: IGNITE-13976 WAL disable/enable with node restarts results in mismatching state, data loss
```

modules/core/src/main/java/org/apache/ignite/IgniteCluster.java
```diff
@@ -545,6 +551,12 @@
      * when all data is persisted to disk.
      * <p>
      * WAL state can be changed only for persistent caches.
+     * <p>
+     * <b>NOTE:</b>
+     * Currently, this method should only be called on a stable topology when no nodes are leaving or joining cluster,
+     * and all baseline nodes are present.
+     * Cache may be stuck in inconsistent state due to violation of these conditions. It is advised to destroy
+     * such cache.
```
Mmuzaf, line 559:
```suggestion
     * such cache.
      TODO: IGNITE-13976 WAL disable/enable with node restarts results in mismatching state, data loss
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/WalStateManager.java
```diff
@@ -304,6 +305,12 @@ public boolean prohibitWALDisabling() {
             String.format("Cache WAL mode cannot be changed within lock or transaction " +
                     "[cacheNames=%s, walEnabled=%s]", cacheNames, enabled));
```
Mmuzaf, line 307:
```suggestion
     // TODO: IGNITE-13976 WAL disable/enable with node restarts results in mismatching state, data loss
```

Mmuzaf (changes requested):

modules/core/src/main/java/org/apache/ignite/IgniteCluster.java
alamar: I don't think this is appropriate place for a TODO. We have this link in developer warning, where it is more convenient in my opinion.

TODOs usually reside in comments which are not javadoc.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/WalStateManager.java
alamar: There's a link in the text below. But I can also add this comment. What do you think?

Mmuzaf: Ilya, from my point of view we should add such comment due to we can forget to remove warnings when the issue will be fixed.

Mmuzaf: But as you wish. Let's proceed anyway.

alamar: I think somebody will point out that the warning is still shown even through the issue is resolved. I will proceed.

PR #8706, opened 2021-01-26
IGNITE-14061 Exclude extdata modules from deployment

vveider:
* Disable deployment by default in parent pom
* Enable deployment explicitly for eligible modules

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -708,7 +708,6 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-deploy-plugin</artifactId>
                 <version>2.8.2</version>
-                <inherited>false</inherited>
                 <configuration>
                     <skip>true</skip>
                 </configuration>
--- a/modules/aop/pom.xml
+++ b/modules/aop/pom.xml
@@ -132,6 +132,14 @@
                 </configuration>
             </plugin>
 
+            <plugin>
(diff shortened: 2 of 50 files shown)


parent/pom.xml
```diff
@@ -708,7 +708,6 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-deploy-plugin</artifactId>
                 <version>2.8.2</version>
-                <inherited>false</inherited>
```
alamar, line 711: Maybe we could specify skip=true for the modules that we do not want to publish? Keep the default true. Otherwise we will eventually ship version with some of new modules not in Maven.

vveider: The whole idea was in changing default policy to `skip`, rather than `include`.
So that we did not push something we should not (like `extdata` in this issue).
Otherwise most of the changes in this PR make no sense.

However, I think we need some place in documentation, where this should be described.
@dmagda, can you help us?

slukyano: I don't think this is about documentation - at least not the user documentation. Maybe just add it as a comment to the parent pom.xml.

slukyano:
LGTM

parent/pom.xml
vveider: I ment "How to contribute" (or alike) doc in Apache Ignite Wiki.

PR #8722, opened 2021-01-28
IGNITE-14095 : Try to fasten cluster start in the ducktests with decreasing 'spi.reconnectDelay'

Vladsz83:
Try to speed up cluster start ath the ducktests with decreasing TcpDiscoverySpi.reconnectDelay


Vladsz83:
Won't work. Same results. No start speed up.

PR #8724, opened 2021-01-28
IGNITE-13639 .NET: Fix detached semantics for array and collection elements

ptupitsyn:
Arrays and non-generic collections are written as Java-compatible types, and elements of those collections are deserialized individually on Java side, even in binary mode. Therefore we should detach every collection element, so that handles to the same object instances are not shared.


modules/platforms/dotnet/Apache.Ignite.Core.Tests/Binary/BinarySelfTest.cs
```diff
@@ -1645,18 +1645,37 @@ public void TestBinaryField()
...
+            {
+                new NestedList {Inner = new List<object>()},
+                new NestedList {Inner = new List<object>()}
+            };
+
+            var bytes = _marsh.Marshal(list);
+            var res = _marsh.Unmarshal<NestedList[]>(bytes);
+
+            Assert.AreEqual(2, res.Length);
+            Assert.AreEqual(0, res[0].Inner.Count);
```
isapego, line 1676: Let's check the second one as well?

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Binary/BinaryUtils.cs
```diff
@@ -1084,7 +1084,7 @@ public static void WriteArray(Array val, BinaryWriter ctx, int? elemTypeId = nul
             stream.WriteInt(val.Length);
 
             for (int i = 0; i < val.Length; i++)
-                ctx.Write(val.GetValue(i));
+                ctx.WriteObjectDetached(val.GetValue(i), parentCollection: val);
```
isapego, line 1087: Don't you think that using `Write` instead of `WriteObjectDetached` is very error-prone and sooner or later produce a hard-to-find errors? Are there cases where `Write` really gives a good performance gain?

isapego (approved):
See minor comments in PR.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Binary/BinaryUtils.cs
ptupitsyn: `Write` supports circular references, so we can't write everything in detached mode.

Yes, this feature introduces a lot of complexity, and possibly degrades performance due to handle tracking. Hopefully we'll get rid of this on 3.0, but for 2.x we have to maintain the compatibility.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Binary/BinarySelfTest.cs
ptupitsyn: Done

PR #8730, opened 2021-01-29
IGNITE-14064 .NET: Fix SQL table name for generic query types

ptupitsyn:
* Fix `BinaryNameMapper.simpleName` to strip namespaces from generic types
* Fix `QueryUtils.typeName` to strip generic part

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/binary/BinaryBasicNameMapper.java
+++ b/modules/core/src/main/java/org/apache/ignite/binary/BinaryBasicNameMapper.java
@@ -116,6 +116,9 @@ public class BinaryBasicNameMapper implements BinaryNameMapper {
         if (idx < 0)
             idx = clsName.lastIndexOf('.');
 
+
+        // TODO: We should string namespaces from .NET generics here
+        // Foo.Bar.GenericTest2`1[[System.String]] => GenericTest2`1[[String]]
         return idx >= 0 ? clsName.substring(idx + 1) : clsName;
     }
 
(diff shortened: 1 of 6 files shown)


modules/core/src/main/java/org/apache/ignite/binary/BinaryBasicNameMapper.java
```diff
@@ -87,6 +87,18 @@ public BinaryBasicNameMapper setSimpleName(boolean isSimpleName) {
...
+        int genericIdx = clsName.indexOf("[[");
+
+        if (genericIdx > 0)
+            clsName = clsName.substring(0, genericIdx + 2) + simpleName(clsName.substring(genericIdx + 2));
+
+        genericIdx = clsName.indexOf("],[", genericIdx);
+
+        if (genericIdx > 0)
+            clsName = clsName.substring(0, genericIdx + 3) + simpleName(clsName.substring(genericIdx + 3));
+
```
isapego, line 101: Consider extract this as a separate function that can be tested separately on simple strings.

isapego (approved):

modules/core/src/main/java/org/apache/ignite/binary/BinaryBasicNameMapper.java
ptupitsyn: Good point, I've extracted a method. However, I don't think we should test it separately - it is an implementation detail. Corresponding tests added to `BinaryBasicNameMapperSelfTest`.

PR #8744, opened 2021-02-02
IGNITE-13019 Fail on non-colocated join

timoninmaxim:
https://issues.apache.org/jira/browse/IGNITE-13019


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/GridSqlQuerySplitter.java
```diff
@@ -1269,6 +1279,10 @@ private void splitSelect(GridSqlAst parent, int childIdx) throws IgniteCheckedEx
         map.hasSubQueries(SplitterUtils.hasSubQueries(mapQry));
         map.hasOuterJoinReplicatedPartitioned(SplitterUtils.hasOuterJoinReplicatedPartitioned(mapQry.from()));
 
+        // TODO: join in subqueries? Subqueries through this process too.
```
slukyano, line 1282: Need to get rid of the TODO before merging of course

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/SplitterUtils.java
```diff
@@ -286,6 +290,255 @@ public static boolean hasPartitionedTables(GridSqlAst ast) {
         return false;
     }
 
+    /**
+     * Traverse AST while join operation isn't found. Check it if found.
+     */
+    public static void lookForPartitionedJoin(GridSqlAst ast, GridSqlAst upWhere, IgniteLogger log) {
```
slukyano, line 296: I'm really struggling to get a sense of completeness from this code. The test cases list seems extensive, so this code probably works, but I'm not sure I understand why...

I think structuring might solve it.

How about the following structure?

1. We always start from a SELECT - so let's process it first
2. In a SELECT, we want to check the subselects in the columns (yes!), the FROM, and the subselects in the where
3. In a FROM, we may see either a single table (not interesting), a subselect (go to step 1), or a JOIN

... and so on. I don't insist on a particular structure but I think we need to be able to understand from the code that nothing is missing. BTW splitting into several methods (lookForPartitionJoinInSelect/From/Where/etc) could help.

modules/indexing/src/test/java/org/apache/ignite/sqltests/CheckWarnJoinPartitionedTables.java
```diff
@@ -0,0 +1,408 @@
...
+
+    /** */
+    @Test
+    public void joinSameTableWithPrimaryKey() {
+        execute(crd, new SqlFieldsQuery(
+            "CREATE TABLE A (ID INT PRIMARY KEY, TITLE VARCHAR);"));
+
+        checkSameTableWithJoinType("LEFT JOIN");
+        checkSameTableWithJoinType("RIGHT JOIN");
+        checkSameTableWithJoinType("INNER JOIN");
```
slukyano, line 66: Aren't INNER JOIN and JOIN the same thing?

modules/indexing/src/test/java/org/apache/ignite/sqltests/CheckWarnJoinPartitionedTables.java
```diff
@@ -0,0 +1,408 @@
...
+
+        checkLogListener(false,
+            "SELECT a1.* FROM A a1 where a1.ID in (select a2.ID from A a2 " + joinType + " A a3 on a2.ID = a3.ID);");
+
+        checkLogListener(false,
+            "SELECT a1.* FROM A a1 " + joinType + " (select a2.ID from A a2 " + joinType + " A a3 on a2.ID = a3.ID) t" +
+                " on a1.ID = t.ID;");
+
+        // Ignite doesn't support this.
+        if (!joinType.contains("RIGHT"))
```
slukyano, line 128: Ignite doesn't support two RIGHT JOINs in the same statement or what?

modules/indexing/src/test/java/org/apache/ignite/sqltests/CheckWarnJoinPartitionedTables.java
```diff
@@ -0,0 +1,408 @@
...
+
+        checkLogListener(false,
+            "SELECT a.* FROM A a " + joinType + " B b on b.ID = a.ID and b.PRICE > 100;");
+
+        checkLogListener(false,
+            "SELECT a.* FROM A a " + joinType + " B b on b.ID = a.ID and b.PRICE != a.ID;");
+
+        // Some wrong queries.
+
+        checkLogListener(true,
```
slukyano, line 353: Need at least some tests with WHERE instead of ON

modules/indexing/src/test/java/org/apache/ignite/sqltests/CheckWarnJoinPartitionedTables.java
timoninmaxim: Actually don't find a docs about equality of JOIN and INNER JOIN. So test them both.

modules/indexing/src/test/java/org/apache/ignite/sqltests/CheckWarnJoinPartitionedTables.java
timoninmaxim: Yes, SQL engine write a message about unsupported clause.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/SplitterUtils.java
timoninmaxim: I added more comments in this method. Now it should be more straightforward. WDYT?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/QueryPartitionsChecker.java
```diff
@@ -0,0 +1,355 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+import org.apache.ignite.IgniteLogger;
+import org.apache.ignite.internal.processors.query.h2.opt.GridH2Table;
+
+/**
+ * Traverse over query AST to find info about partitioned table usage.
+ */
+class QueryPartitionsChecker {
```
AMashenkov, line 28: Irrelevant class name.

timoninmaxim: Rename to SqlAstTraverser.

AMashenkov (approved):

PR #8801, opened 2021-02-15
IGNITE-14173 .NET: Fix partition awareness auto-disable on old servers

ptupitsyn:
* Fix failing `TestReconnectToOldNodeDisablesPartitionAwareness`: when thin client reconnects to a new "default" node, and that node does not support partition awareness, the feature should be disabled automatically. 
* Enable the test on .NET Core
* Fix compatibility test flakiness: add port range to `JavaServer.GetClientConfiguration`


isapego (approved):
Looks good to me

PR #8811, opened 2021-02-18
IGNITE-14206 Bug: 2.8.1 inlines POJO but ignores it since 2.9.0

timoninmaxim:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/InlineObjectBytesDetector.java
```diff
@@ -0,0 +1,196 @@
...
+import org.apache.ignite.IgniteLogger;
+import org.apache.ignite.internal.pagemem.PageUtils;
+import org.apache.ignite.internal.processors.cache.persistence.tree.BPlusTree;
+import org.apache.ignite.internal.processors.cache.persistence.tree.io.BPlusIO;
+import org.apache.ignite.internal.processors.query.h2.opt.H2Row;
+import org.h2.value.Value;
+import org.h2.value.ValueNull;
+
+/**
+ * This class helps detects wheter tree contains inline JO type.
```
korlov42, line 31: wheter -> whether

korlov42 (approved):

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/InlineObjectBytesDetector.java
timoninmaxim: @korlov42 hi! I've fixed your comments

.travis.yml
```diff
@@ -57,6 +57,6 @@ matrix:
       dist: trusty
       install: skip
       jdk: openjdk8
-      script: mvn test -Pcheck-test-suites -B -V
+      script: mvn test -Pcheck-test-suites,all-java,all-scala,scala -B -V
```
Mmuzaf, line 60: Why do we need this change here?

timoninmaxim: @Mmuzaf I've add a compatibility test for indexes and forget add it to suite. But the check was OK. So I've found that ignite-compatibility module is excluded from pom.xml by default. For me this module is important one and the check must cover it. So I've add "all-java" profile to the check. This profile enables the ignite-examples module that requires the "scala" profile to be enabled.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/InlineObjectBytesDetector.java
korlov42: looks like misspell is still here

timoninmaxim: @korlov42 because there are misspells in 2 places where "whether" is used. Sorry, will fix that

PR #8820, opened 2021-02-22
IGNITE-2399: Implement acquireAndExecute In IgniteSemaphore

atris:
This commit introduces a new method which takes in a semaphore and a Callable task, acquires the
semaphore, executes the task and returns a Future immediately.

The semaphore is released when the Future completes.


modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -12051,4 +12055,25 @@ public static long uncompressedSize(File zip) throws IOException {
...
+    public static <T> IgniteFuture<T> acquireAndExecute(IgniteSemaphore semaphore,
+                                          IgniteCallable<IgniteFuture<T>> callable, int numPermits) throws Exception {
+
+        semaphore.acquire(numPermits);
+
+        IgniteFuture<T> future = callable.call();
+
+        future.listen(new IgniteInClosure<IgniteFuture<T>>() {
+            @Override public void apply(IgniteFuture<T> IgniteFuture) {
+                if (IgniteFuture.isCancelled() || IgniteFuture.isDone()) {
```
SammyVimes, line 12071: Not sure if these checks needed here: future's listen callback only executed when future is completed

modules/core/src/test/java/org/apache/ignite/internal/util/IgniteUtilsSelfTest.java
```diff
@@ -1455,6 +1470,51 @@ public void testUncompressedSize() throws Exception {
         }
     }
 
+    /**
+     * Test to verify the {@link U#acquireAndExecute(IgniteSemaphore, IgniteCallable, int)}.
+     *
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testAcquireAndExecute() throws Exception {
```
SammyVimes, line 1479: You can also add a negative scenario, e.g. what will happen if there is an error during callable's execution

SammyVimes (changes requested):

nizhikov:
Hello, Why do we need this method in the first place?
Currently, this PR adds some internal method without usages.

Those methods should be added only when they really required.

atris:
> Hello, Why do we need this method in the first place?
> Currently, this PR adds some internal method without usages.
> 
> Those methods should be added only when they really required.

As noted in the ticket, this is an utility method that 1. Captures a very common usage pattern of semaphores and 2. Does not pollute the upper layer APIs, hence causing no impact to any component of the system with no chances of functional/performance regressions.

The opening of thr ticket itself, IMO, points to a usecase :)

nizhikov:
@atris 
If we introduce some internal method then we should use it through the codebase.
But, for now, there are no usages of this new method.

vkulichenko:
@atris What Nikolay is saying is that the `IgniteUtils` class is not considered to be a part of public API, as it's located in the `internal` package.

Actually, I can't think of a good location for such a method right now, so this probably should be discussed separately.

atris:
> @atris What Nikolay is saying is that the `IgniteUtils` class is not considered to be a part of public API, as it's located in the `internal` package.
> 
> Actually, I can't think of a good location for such a method right now, so this probably should be discussed separately.

Ah, understood it now, thanks for clarifying and apologies, Nikolay!

@vkulichenko Should this not be a part of IgniteSemaphore class, since this is really a semaphore acquisition method available as a part of public API like acquire and tryAcquire methods?

atris:
@nizhikov @SammyVimes @vkulichenko Moved the method to the public API for IgniteSemaphore. Please review

atris:
@nizhikov @SammyVimes @vkulichenko Please let me know. Also, should I move IGNITE-14225 to a separate PR?

atris:
@alamar please see

atris:
Please let me know what the outcome of this PR is. If the method is redundant, I will move IGNITE-14225 to a separate PR and close this one.

atris:
I have moved IGNITE-14225 out of this PR

atris:
@SammyVimes @nizhikov @vkulichenko Please let me know your thoughts on this PR.

modules/core/src/main/java/org/apache/ignite/internal/processors/datastructures/GridCacheSemaphoreImpl.java
```diff
@@ -662,6 +665,28 @@ private void initializeSemaphore() throws IgniteCheckedException {
...
+    /** {@inheritDoc} */
+    @Override public <T> IgniteFuture<T> acquireAndExecute(IgniteCallable<IgniteFuture<T>> callable,
+                                                           int numPermits) throws Exception {
+        acquire(numPermits);
+
+        IgniteFuture<T> future = null;
+        try {
+            future = callable.call();
+        } finally {
+            release(numPermits);
```
vkulichenko, line 677: Is this release required? Since `call()` here is asynchronous, permits will be released before completion, which seems wrong. Also, there is another `release` invocation, which happens within the future listener - that one looks correct.

modules/core/src/main/java/org/apache/ignite/IgniteSemaphore.java
```diff
@@ -244,6 +247,16 @@ public boolean tryAcquire(int permits, long timeout, TimeUnit unit)
...
+    /**
+     * Acquires the given semaphore, executes the given callable and schedules the release of permits asynchronously
+     *
+     * @param callable the callable to execute
+     * @param numPermits the number of permits to acquire
+     * @throws Exception if the callable throws an exception
+     */
+    public <T> IgniteFuture<T> acquireAndExecute(IgniteCallable<IgniteFuture<T>> callable,
+                                                 int numPermits) throws Exception;
+
```
vkulichenko, line 259: I'm not sure I understand why callable returns `IgniteFuture<T>` rather than just `T`. I saw a comment in the ticket, but it doesn't really provide justification. It probably makes sense to have a separate thread on the dev list regarding this - we should have a discussion if there are different opinions.

vkulichenko (changes requested):
@atris Please take a look at my comments. Generally, I'm not sure I agree with the API, but I'm happy to hear why the current design is correct. We should probably have a separate discussion on the dev list.

atris:
Thanks @vkulichenko. I will start the discussion.

modules/core/src/main/java/org/apache/ignite/IgniteSemaphore.java
atris: Agreed, fixed the same

modules/core/src/main/java/org/apache/ignite/internal/processors/datastructures/GridCacheSemaphoreImpl.java
```diff
@@ -662,6 +667,31 @@ private void initializeSemaphore() throws IgniteCheckedException {
         }
     }
 
+    /** {@inheritDoc} */
+    @Override public <T> IgniteFuture<T> acquireAndExecute(IgniteCallable<T> callable,
+                                                           int numPermits) throws Exception {
+        acquire(numPermits);
+
+        T value = callable.call();
```
vkulichenko, line 675: The method signature implies that it's asynchronous. The implementation, however, executes the provided callable synchronously. It should be submitted to the local public thread pool instead.

vkulichenko (changes requested):

vkulichenko (approved):

PR #8827, opened 2021-02-25
IGNITE-14208 .NET: Enable examples multi-targeting

ptupitsyn:
Use `<TargetFrameworks>netcoreapp2.1;netcoreapp3.1;net5</TargetFrameworks>` so that examples can be built with any current SDK.


ptupitsyn:
Won't fix - a change like that requires ALL specified frameworks to be present, which is even worse

PR #8828, opened 2021-02-25
IGNITE-14243 .NET: Extend ConfigurationManager dependency version range

ptupitsyn:
Test Ignite with recent `System.Configuration.ConfigurationManager` versions (including 6.x preview), and extend supported version range accordingly.


isapego (approved):
Looks good to me.

PR #8845, opened 2021-03-02
IGNITE-12508: Add Missing Cache Remove Instance in ClusterCachesInfo

atris:
In IGNITE-12508, we added a new cache mapping. This commit adds a missing remove operation where an entry gets deleted from the main cache but not the mapping.


atris:
@alex-plekhanov @alamar

PR #8866, opened 2021-03-09
IGNITE-14293 .NET: Fix AffinityKey metadata in QueryEntity.KeyType

ptupitsyn:
`AffinityKey` system type, when used as a `QueryEntiry` key, got overwritten by `UnmanagedCallbacks.BinaryTypeGet` call, which broke the mapping to a corresponding Java type.

* Add missing check for existing registered type (actual fix in Marshaller.cs)
* Add check for system type overwrite
* Add test for `AffinityKey` + `QueryEntity`
* Improve examples tests - SQL and LINQ examples were affected by this bug
* Change cache names in examples so different examples do not interfere


modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/Affinity/AffinityTest.cs
```diff
@@ -121,6 +121,45 @@ public void TestAffinityKeyMappedWithQueryEntitySpringXml()
...
+            var aff = ignite.GetAffinity(cache.Name);
+
+            var ignite2 = Ignition.GetIgnite("grid-1");
+            var cache2 = ignite2.GetOrCreateCache<AffinityKey, QueryEntityValue>(cacheCfg);
+            var aff2 = ignite2.GetAffinity(cache2.Name);
+
+            // Check mapping.
+            for (var i = 0; i < 100; i++)
+            {
+                Assert.AreEqual(aff.GetPartition(i), aff.GetPartition(new AffinityKey("foo", i)));
```
isapego, line 149: I think we test is going to be more robust if we add variability to the string field as well. How about `"foo" + i`?

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/Affinity/AffinityTest.cs
```diff
@@ -121,6 +121,45 @@ public void TestAffinityKeyMappedWithQueryEntitySpringXml()
...
+                Assert.AreEqual(aff.GetPartition(i), aff2.GetPartition(i));
+            }
+
+            // Check put/get.
+            var key = new AffinityKey("x", 123);
+            cache[key] = new QueryEntityValue {Name = "y", AffKey = 321};
+
+            var val = cache2[key];
+            Assert.AreEqual("y", val.Name);
+            Assert.AreEqual(321, val.AffKey);
```
isapego, line 160: Not critical, but there is duplication.
```suggestion
            var key = new AffinityKey("x", 123);
            var expected = new QueryEntityValue {Name = "y", AffKey = 321};
            cache[key] = expected;

            var val = cache2[key];
            Assert.AreEqual(expected.Name, val.Name);
            Assert.AreEqual(expected.AffKey, val.AffKey);
```

isapego (approved):
See my minor comments.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/Affinity/AffinityTest.cs
ptupitsyn: Fixed

PR #8867, opened 2021-03-10
IGNITE-14299 .NET: Fix passing array of custom type during service method invocation

nizhikov:
Currently, `BinaryMarshaller` loses the type of the array component after write -> read in binary mode -> write.
This leads to the situation when we can't perform a service call .Net (client node) -> .Net (server node) for a method that contains typed array arguments.

PR fixes this with the new `BinaryArrayWrapper` - special binary object implementation that keeps the information about array component type.

We can't use this new wrapper all over the code due to the back compatibility.
So we use it only during the platform service calls.

### Notes


nizhikov:
Hello, @ptupitsyn 
Can you, please, take a look at my changes?

nizhikov:
Hello, @alex-plekhanov 

Can you, please, take a look at my changes?

modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryArrayWrapper.java
```diff
@@ -0,0 +1,204 @@
...
+        if (ldr != null)
+            GridBinaryMarshaller.USE_CACHE.set(Boolean.FALSE);
+
+        try {
+            Class compType = BinaryUtils.resolveClass(ctx, compTypeId, compClsName, resolveLdr, false);
+
+            Object[] res = (Object[])Array.newInstance(compType, arr.length);
+
+            for (int i = 0; i < arr.length; i++)
+                    res[i] = arr[i] instanceof BinaryObject ? ((BinaryObject)arr[i]).deserialize(ldr) : arr[i];
```
alex-plekhanov, line 106: 1. Wrong indent.
2. Perhaps objects should be deserialized recursively since there can be collections and maps inside the array.

modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryArrayWrapper.java
```diff
@@ -0,0 +1,204 @@
...
+    /**
+     * @return Component class name.
+     */
+    public String componentClassName() {
+        return compClsName;
+    }
+
+    /** {@inheritDoc} */
+    @Override public BinaryObject clone() throws CloneNotSupportedException {
+        return new BinaryArrayWrapper(ctx, compTypeId, compClsName, arr.clone(), size);
```
alex-plekhanov, line 138: IDEA show a warning about using constructor in clone() method, can we use super.clone() here?

modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryClassDescriptor.java
```diff
@@ -757,7 +761,10 @@ void write(Object obj, BinaryWriterExImpl writer) throws BinaryObjectException {
 
                 case OBJECT_ARR:
                     writer.doWriteObjectArray((Object[])obj);
+                    break;
```
alex-plekhanov, line 764: NL

modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryClassDescriptor.java
```diff
@@ -757,7 +761,10 @@ void write(Object obj, BinaryWriterExImpl writer) throws BinaryObjectException {
...
 
+                case OBJECT_ARR_WRAPPER:
+                    writer.doWriteBinaryArrayWrapper((BinaryArrayWrapper)obj);
```
alex-plekhanov, line 767: NL

modules/core/src/main/java/org/apache/ignite/internal/binary/GridBinaryMarshaller.java
```diff
@@ -150,9 +150,16 @@
...
-    /** Binary enum */
+    /** Binary enum. */
     public static final byte BINARY_ENUM = 38;
 
+    /**
+     * Binary wrapper for {@code Object[]}.
+     * This wrapper used to store array component type id during serde process.
+     * @see BinaryUtils#doReadObjectArrayWrapper(BinaryInputStream, BinaryContext, ClassLoader, BinaryReaderHandlesHolder, boolean, boolean)
+     */
+    public static final byte OBJ_ARR_WRAPPER = 39;
```
alex-plekhanov, line 161: Do we really need a dedicated type for the wrapper? The wrapper is only a temporary internal representation of the array after unmarshalling, if we marshall it again we can use the same OBJ_ARR type.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Services/Model.cs
```diff
@@ -123,7 +123,7 @@ public override int GetHashCode()
     // ReSharper disable once InconsistentNaming
     public enum ACL
     {
-        Allow, Deny
+        ALLOW, DENY
```
alex-plekhanov, line 126: Why it's changed to uppercase here? AFAIK pascal case should be used for enum values in C# (https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/4x252001(v=vs.71)?redirectedfrom=MSDN). And for binary protocol enum order is used, so uppercase is not required to match dotnet and java enums.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Services/ServicesTypeAutoResolveTest.cs
```diff
@@ -155,39 +157,59 @@ public void TestCallJavaServiceLocal()
...
+        public void TestCallJavaServiceThin()
+        {
+            DoTestPlatformService(name => _thinClient.GetServices().GetServiceProxy<IJavaService>(name));
+        }
 
-            DoTestService(svc);
+        /// <summary>
+        /// Tests .Net service invocation.
+        /// </summary>
+        public void DoTestPlatformService(Func<String, IJavaService> svc, bool isPlatform = false)
```
alex-plekhanov, line 176: The method name (and comment) is confusing, it can invoke java service as well as .net service, but according to method name and comment it can invoke only .net service.

modules/core/src/main/java/org/apache/ignite/internal/binary/GridBinaryMarshaller.java
nizhikov: Yes. It passed through the async compute call in `GridServiceProxy`.
So ser-deser happens in a different thread. If we don't use dedicated typeid we lose array information.
```
                        // Execute service remotely.
                        return ctx.closure().callAsyncNoFailover(
                            GridClosureCallMode.BROADCAST,
                            new ServiceProxyCallable(methodName(mtd), name, mtd.getParameterTypes(), args, BinaryUtils.USE_ARRAY_WRAPPER.get()),
                            Collections.singleton(node),
                            false,
                            waitTimeout,
                            true).get();

``

File: modules/platforms/dotnet/Apache.Ignite.Core.Tests/Services/Model.cs (hunk 6 above)
Comment (nizhikov) [reply modules/platforms/dotnet/Apache.Ignite.Core.Tests/Services/Model.cs:126]: Actually, it's kind of one more minor .Net -> Java integration issue.
With previous enum declaration we have an exception in `ServicesTypeAutoResolveTest#TestCallPlatformServiceRemote`

```
Apache.Ignite.Core.Services.ServiceInvocationException : Proxy method invocation failed with an exception. Examine InnerException for details.
  ----> Apache.Ignite.Core.Common.IgniteException : Conflicting enum values. Name 'Allow' uses ordinal value (0) that is also used for name 'ALLOW' [typeName='org.apache.ignite.platform.model.ACL']
  ----> Apache.Ignite.Core.Common.JavaException : class org.apache.ignite.IgniteException: Conflicting enum values. Name 'Allow' uses ordinal value (0) that is also used for name 'ALLOW' [typeName='org.apache.ignite.platform.model.ACL']
```

modules/core/src/main/java/org/apache/ignite/internal/binary/GridBinaryMarshaller.java
ptupitsyn: Agree with Alex, it would be great to avoid a new system type.
Otherwise we'll have to add the support for this type to .NET and C++, right?

nizhikov: This type exists only during transfer from a caller to a callee.
.Net and C++ receives typed array.

ptupitsyn:
Summarizing our private discussion with @nizhikov :

*Problem*
Service arguments are deserialized "in the middle" when pass `Object[] args` around in `GridServiceProxy`:
* This causes array element type info loss
* This is totally unnecessary and reduces performance

*Possible fix*
Do not perform unnecessary argument des/ser, retain original binary representation along the call chain.
E.g. for .NET -> .NET service call, the bytes are written on .NET side on one node will be received exactly as is on .NET side on another node.

This can be achieved by wrapping the arguments in some way - `byte[]` or `BinaryObject`, and passing this around instead of `Object[]`.

PR #8868, opened 2021-03-11
IGNITE-8635: Add a Method to Inspect BinaryObject Size

atris:
This commit introduces a method to BinaryObject interface allowing users to get an estimate of the size
of the object.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryEnumObjectImpl.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryEnumObjectImpl.java
@@ -476,6 +476,10 @@ public class BinaryEnumObjectImpl implements BinaryObjectEx, Externalizable, Cac
 
     /** {@inheritDoc} */
     @Override public int size() {
+        if (valBytes == null) {
+            return -1;
+        }
+
         return BinaryPrimitives.readInt(valBytes, ord + GridBinaryMarshaller.TOTAL_LEN_POS);
     }
 }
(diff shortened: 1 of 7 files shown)


atris:
Teamcity Run: https://ci.ignite.apache.org/buildConfiguration/IgniteTests24Java8_RunAll/5909703

alamar:
Please test this new API, by adding new tests or modifying a few existing tests so that they check this API call's results on various cache objects.

atris:
TeamCity run:

https://ci.ignite.apache.org/project/IgniteTests24Java8?mode=builds

modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryEnumObjectImpl.java
```diff
@@ -473,4 +473,13 @@ public BinaryEnumObjectImpl(BinaryContext ctx, byte[] arr) {
 
         return reader.afterMessageRead(BinaryEnumObjectImpl.class);
     }
+
+    /** {@inheritDoc} */
+    @Override public int size() {
+        if (valBytes == null) {
```
alamar, line 479: I don't think that returning -1 is meaningful here. We should marshal if size() requested and valBytes is null, such as in valueBytes, e.g.:

        if (valBytes != null)
            return valBytes;

        valBytes = U.marshal(ctx.marshaller(), this);

        return valBytes;

Also, please note that if() with single-line body should not be brace-indented.

alamar (changes requested):

atris:
@alamar Updated, please see and let me know

modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryEnumObjectImpl.java
```diff
@@ -473,4 +473,17 @@ public BinaryEnumObjectImpl(BinaryContext ctx, byte[] arr) {
...
+            try {
+                valBytes = U.marshal(ctx.marshaller(), this);
+            } catch (IgniteCheckedException e) {
```
alamar, line 482: Please note that catch ( should start on a new line according to https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines

I will fix that during merge.

atris: Sorry about that, will ensure this henceforth :)

PR #8870, opened 2021-03-11
IGNITE-12033 Move async continuations away from striped pool

ptupitsyn:
User continuations should not run on striped pool - this can cause unexpected, hard to diagnose performance issues and deadlocks.

* Add `IgniteConfiguration#asyncContinuationExecutor`. Defaults to `null`, which means `ForkJoinPool#commonPool` should be used.
* Use default executor for Cache async operations
* Add .NET and Java tests

IEP: https://cwiki.apache.org/confluence/display/IGNITE/IEP-70%3A+Async+Continuation+Executor


modules/core/src/main/java/org/apache/ignite/internal/util/future/IgniteFutureImpl.java
```diff
@@ -68,7 +80,10 @@ public IgniteFutureImpl(IgniteInternalFuture<V> fut) {
     @Override public void listen(IgniteInClosure<? super IgniteFuture<V>> lsnr) {
         A.notNull(lsnr, "lsnr");
 
-        fut.listen(new InternalFutureListener(lsnr));
+        if (defaultExecutor != null && !isDone())
+            fut.listen(new InternalFutureListener(new AsyncFutureListener<>(lsnr, defaultExecutor)));
```
isapego, line 84: So the `defaultExecutor` overwrites explicitly set `lsnr`? Does not seem right to me.

modules/core/src/main/java/org/apache/ignite/internal/util/future/IgniteFutureImpl.java
```diff
@@ -68,7 +80,10 @@ public IgniteFutureImpl(IgniteInternalFuture<V> fut) {
...
+        else
+            fut.listen(new InternalFutureListener(lsnr));
```
isapego, line 86: Also, it is not clean, why `lsnr` used if future `isDone()`

isapego (commented):
Take a look at my comments please.

modules/benchmarks/src/main/java/org/apache/ignite/internal/benchmarks/jmh/cache/JmhCacheAsyncListenBenchmark.java
```diff
@@ -0,0 +1,161 @@
...
+ * Benchmark                          Mode  Cnt      Score      Error  Units
+ * JmhCacheAsyncListenBenchmark.get  thrpt   10  82052.664 ± 2891.182  ops/s
+ * JmhCacheAsyncListenBenchmark.put  thrpt   10  77859.584 ± 2071.196  ops/s
+ *
+ * With ForkJoinPool async continuation executor:
+ * Benchmark                          Mode  Cnt      Score      Error  Units
+ * JmhCacheAsyncListenBenchmark.get  thrpt   10  76008.272 ± 1506.928  ops/s
+ * JmhCacheAsyncListenBenchmark.put  thrpt   10  73393.986 ± 1336.420  ops/s
+ */
+@SuppressWarnings({"unchecked", "rawtypes"})
```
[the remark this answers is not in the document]
ptupitsyn: Yes, as expected, because we schedule the listener on another thread instead of executing it directly.
This is noted in the IEP and on the dev list.

modules/core/src/main/java/org/apache/ignite/internal/util/future/IgniteFutureImpl.java
ptupitsyn: We don't overwrite `lsnr`. We wrap `lsnr` into `AsyncFutureListener` which uses the provided executor to run the delegate.

modules/core/src/main/java/org/apache/ignite/internal/util/future/IgniteFutureImpl.java
ptupitsyn: Some async operations complete synchronously, like `cache.get` on a primary key. In this case we run the continuation immediately on the current user thread - no need to schedule it somewhere else.

```
// Key 1 is available locally (primary or backup)
// Striped pool is not involved, cache operation and doSomething are executed on the current thread
cache.get(1).listen(doSomething) 
```

PR #8877, opened 2021-03-15
IGNITE-14314: GridDhtLockFuture#onComplete Should Be Aware of Cache C…

atris:
…leanup

The current structure of the method only considers node stoppage as a parameter when
deciding if cache context methods should be invoked or not. It should also consider if the
cache is cleaned up or not.


alamar:
Is it possible to write a test which will show the impact of this commit?

PR #8892, opened 2021-03-17
IGNITE-14335 Merges APIs of IgniteAuthenticationProcessor and IgniteSecurity.

ololo3000:
1. Removed an IgniteAuthenticationProcessor that is now treated as an independent processor.
2. Moved the logic of IgniteAuthenticationProcessor into the implementation of the security plugin that is used if authentication is enabled via configuration.
3. Remove duplication of security checks and leave IgniteSecurity as a single security API. 
4. Removed the AuthorizationContext completely.
5. Extended GridSecurityPlugin interface with methods that provide the ability to manage security users.

### Notes


modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
```diff
@@ -310,19 +301,16 @@ public AuthorizationContext authenticate(String login, String passwd) throws Ign
...
-
-                return new AuthorizationContext(User.create(login));
-            }
+            } while (fut.retry());
         }
         else
-            return new AuthorizationContext(authenticateOnServer(login, passwd));
+            authenticateOnServer(login, passwd);
+
+        return new SecurityContextImpl(authCtx.subjectId(), login);
```
dgarus, line 313: We should feel fields of `SecuritySubject` like `SecuritySubjectType`, `InetSocketAddress`, and so on.

dgarus (changes requested):
I left a detailed comment with suggestions in the jira ticket.

modules/clients/src/test/java/org/apache/ignite/common/RunningQueryInfoCheckInitiatorTest.java
```diff
@@ -227,7 +227,7 @@ public void testJdbcV2InitiatorId() throws Exception {
 
             GridTestUtils.runAsync(() -> {
                     try (Connection conn = DriverManager.getConnection(
-                        CFG_URL_PREFIX + "nodeId=" + grid0NodeId + "@modules/clients/src/test/config/jdbc-config.xml")) {
+                        CFG_URL_PREFIX + "nodeId=" + grid0NodeId + "@modules/clients/src/test/config/jdbc-security-config.xml")) {
```
timoninmaxim, line 230: Why do you change config only for this test? Is it possible to create a different test, or make a inherited class that overrides config?

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
```diff
@@ -212,53 +230,33 @@ else if (msg instanceof UserAuthenticateResponseMessage)
...
-        sharedCtx = ctx.cache().context();
-    }
-
     /** {@inheritDoc} */
     @Override public void stop(boolean cancel) throws IgniteCheckedException {
-        if (!isEnabled)
-            return;
-
-        ctx.io().removeMessageListener(GridTopic.TOPIC_AUTH, ioLsnr);
+        if (ioLsnr != null)
```
timoninmaxim, line 235: Why did this check for NULL appear?

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
```diff
@@ -971,6 +888,91 @@ private static boolean isNodeHoldsUsers(ClusterNode n) {
...
+        return new SecurityContextImpl(
+            node.id(),
+            node.attribute(ATTR_IGNITE_INSTANCE_NAME),
+            REMOTE_NODE,
+            new InetSocketAddress(F.first(node.addresses()), 0));
+    }
+
+    /** {@inheritDoc} */
+    @Override public SecuritySubject authenticatedSubject(UUID subjId) throws IgniteCheckedException {
+        return null;
```
timoninmaxim, line 902: Why NULL? Let's make throw Exception or return a real value.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
```diff
@@ -971,6 +888,91 @@ private static boolean isNodeHoldsUsers(ClusterNode n) {
...
+    }
+
+    /** {@inheritDoc} */
+    @Override public boolean isGlobalNodeAuthentication() {
+        return false;
+    }
+
+    /** {@inheritDoc} */
+    @Override public void authorize(String name, SecurityPermission perm, SecurityContext securityCtx) throws SecurityException {
```
timoninmaxim, line 916: Looks like this method should do a work of checkUserOperation?

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
```diff
@@ -73,11 +85,22 @@
 import static org.apache.ignite.events.EventType.EVT_NODE_JOINED;
 import static org.apache.ignite.events.EventType.EVT_NODE_LEFT;
 import static org.apache.ignite.internal.GridComponent.DiscoveryDataExchangeType.AUTH_PROC;
+import static org.apache.ignite.internal.IgniteNodeAttributes.ATTR_AUTHENTICATION_ENABLED;
+import static org.apache.ignite.internal.IgniteNodeAttributes.ATTR_IGNITE_INSTANCE_NAME;
+import static org.apache.ignite.internal.IgniteNodeAttributes.ATTR_SECURITY_CREDENTIALS;
+import static org.apache.ignite.internal.processors.authentication.User.DFAULT_USER_NAME;
```
timoninmaxim, line 91: s/DFAULT/DEFAULT/

ololo3000: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
ololo3000: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
ololo3000: Frankly, this method is not used in any way.  I have replaced it with UnsupportedException.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
ololo3000: This component can be stopped before ioLsnr is initialized - for example in case when exception is thrown during node startup by other components. And we will face NPE. As I see null check is a common practice.

modules/clients/src/test/java/org/apache/ignite/common/RunningQueryInfoCheckInitiatorTest.java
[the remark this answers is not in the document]
ololo3000: Because only this test checks JDBC connection to the cluster for which Ignite authentication is enabled through Ignite node.
As were mentioned in the dev list discussion:
"The current implementation of IgniteAuthenticationProcessor can enable authentication itself if the current node connects to the cluster with authentication enabled - this functionality will not be supported. The user can easily overcome this by explicitly enabling authentication in the configuration on all nodes."

So in this test case with respect to IgniteAuthenticationProcessor changes by this PR we must explicitly enable authentication for the node that JDBC uses to connect to the cluster.

That was my motivation to create the new configuration file.

This test class is all about security but previously it works because the authentication on joining node was enabled automatically.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
```diff
@@ -836,14 +792,11 @@ private void onAuthenticateResponseMessage(UserAuthenticateResponseMessage msg)
      * Local node joined to topology. Discovery cache is available but no discovery custom message are received.
      * Initial user set and initial user operation (received on join) are processed here.
      */
-    public void onLocalJoin() {
-        if (coordinator() == null)
+    private void onLocalJoin() {
+        if (ctx.isDaemon() || ctx.clientDisconnected() || coordinator() == null)
```
timoninmaxim, line 796: Can we replace clientDisconnected with ctx.clientNode(), and then remove a check below. Client disconnected looks weird, as there no actions on reconnected, and it's misleading how a node should fill initial user state.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
```diff
@@ -971,6 +893,81 @@ private static boolean isNodeHoldsUsers(ClusterNode n) {
...
+    /** {@inheritDoc} */
+    @Override public void onSessionExpired(UUID subjId) {
+        // No-op.
+    }
+
+    /** {@inheritDoc} */
+    @Override public SecurityContext securityContext(UUID subjId) {
+        User user = users.get(subjId);
+
+        return user == null ? null : new SecurityContextImpl(subjId, user.name(), REMOTE_CLIENT, null);
```
timoninmaxim, line 950: Why does it always return REMOTE_CLIENT? Docs say it applies for nodes too: "Gets security context for authenticated nodes and thin clients". Looks like we can support a cache for contexts, also it will fix @dgarus comment (I agree with him).

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
```diff
@@ -971,6 +893,81 @@ private static boolean isNodeHoldsUsers(ClusterNode n) {
...
+    }
+
+    /**
+     * Gets the user with the specified ID and login. It is necessary to check the login to make sure that there was
+     * no collision when calculating the user ID.
+     */
+    private User findUser(UUID subjId, String login) {
+        User user = users.get(subjId);
```
timoninmaxim, line 958: The users variable guarded with the mux lock in other methods. Actually this lock is tricky, I'm not fully understand what is a purpose of that. Could you please should we use this lock there too?

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/ClientListenerAbstractConnectionContext.java
```diff
@@ -104,33 +95,12 @@ public GridKernalContext kernalContext() {
...
-    protected AuthorizationContext authenticate(GridNioSession ses, String user, String pwd)
-        throws IgniteCheckedException {
-        if (ctx.security().enabled())
-            authCtx = authenticateExternal(ses, user, pwd).authorizationContext();
-        else if (ctx.authentication().enabled()) {
-            if (F.isEmpty(user))
-                throw new IgniteAccessControlException("Unauthenticated sessions are prohibited.");
-
-            authCtx = ctx.authentication().authenticate(user, pwd);
+    protected void authenticate(GridNioSession ses, String user, String pwd) throws IgniteCheckedException {
```
timoninmaxim, line 100: There is a weird behavior on last line of this method. We do a check of secCtx for NULL, if NULL that throws an exception. But actually AuthProcessor never returns NULL, it just throws exception on auth failure. Also docs for GridSecurityProcessor are misleading:
```

     * @return {@code True} if succeeded, {@code false} otherwise.
    public SecurityContext authenticate(AuthenticationContext ctx) throws IgniteCheckedException;
```
Let's fix declaration of return value in the docs (true / false -> SecurityContext), also there should be a description for auth fail behavior - should it return NULL or throw an exception.

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
timoninmaxim: Also it returns all permissions for any subjId (default behavior for SecurityContextImpl), even it fails authentication. Is it OK?

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/ClientListenerAbstractConnectionContext.java
ololo3000: We should not consider AuthProcessor as the only security implementation. It can be replaced with the custom implementation by the user and  nothing restricts null to be returned. So the check is justified.
I have fixed java doc.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
ololo3000: My opinion is - we should not use lock there. 
1. We just read from user map that is already ConcurrentMap.
2. As I see the main purpose of this mux is to to keep user operations consistent in case they should be send to a new joining node. But frankly it does not always work as expected see  https://issues.apache.org/jira/browse/IGNITE-14301

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
ololo3000: SecurityContext associated with the node is stored in node attributes and is obtained automatically by Ignite using the node ID (see IgniteSecurityProcessor#withContext(java.util.UUID)). As we use node ID as subject ID during node authentication, this method makes sense only for the REMOTE_CLIENT.

As far as I could I have fixed @dgarus comment. sc_1 == sc_2  equality strict implementation affects performance heavily.
So now we provide these invariants sc_1.subjec,id == sc_2.subjec.id and  sc_1.subjec,login == sc_2.subjec.login
for both clients and nodes.

Permissions and permission checks provided by the SecurityContext are orphaned. All (except one - check on node join 
 - but IgniteAuthenticationProcessor do not support node authentication) permission checks are performed through IgniteSecurity#authorize.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
ololo3000: Please, see changes related to GridDiscoveryManager#start. My motivation to add this check:
It was called by GridDiscoveryManager previously and to keep IgntieAuthenticationProcessor behavior intact I duplicated it.

It might be weird but It makes review easier. And it works exactly as before.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
```diff
@@ -1335,4 +1332,99 @@ private RefreshUsersStorageWorker(ArrayList<User> usrs) {
...
+        }
+
+        /** {@inheritDoc} */
+        @Override public InetSocketAddress address() {
+            return addr;
+        }
+
+        /** {@inheritDoc} */
+        @Override public SecurityPermissionSet permissions() {
+            return null;
```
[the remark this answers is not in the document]
ololo3000: Permissions and permission checks provided by the SecurityContext are orphaned (except one - check on node join 
 - but IgniteAuthenticationProcessor do not support node authentication). All permission checks are performed through IgniteSecurity#authorize. Also current Ignite Authentication Processor implementation do not implement any kind of Ignite cache/task/etc operation authorization.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
```diff
@@ -314,15 +314,14 @@ public AuthorizationContext authenticate(String login, String passwd) throws Ign
                 }
 
                 fut.get();
```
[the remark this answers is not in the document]
ololo3000: You are right. I have filed a ticket with the problem that shows that we can hang forever -  https://issues.apache.org/jira/browse/IGNITE-14301
But I think this problem is not related to the current ticket, lets leave it as is.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
```diff
@@ -136,8 +160,8 @@
     /** Node activate future. */
     private final GridFutureAdapter<Void> activateFut = new GridFutureAdapter<>();
 
-    /** Validate error. */
-    private String validateErr;
+    /** User management operations. */
+    EnumSet<SecurityPermission> userOps = EnumSet.of(CREATE_USER, DROP_USER, ALTER_USER);
```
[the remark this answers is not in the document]
ololo3000: Thanks. Done.

modules/clients/src/test/java/org/apache/ignite/common/RunningQueryInfoCheckInitiatorTest.java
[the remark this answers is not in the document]
timoninmaxim: Got it. Thanks for the clarification.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
```diff
@@ -677,8 +635,6 @@ private ClusterNode coordinator() {
                     U.warn(log, "Cannot find the server coordinator node. "
                         + "Possible a client is started with forceServerMode=true. " +
                         "Security warning: user authentication will be disabled on the client.");
-
-                    isEnabled = false;
```
timoninmaxim, line 681: I see there 2 possible issues:
1. a misleading comment on the previous line, as user authentication isn't disabled now explicitly;
2. as there no check for `isEnabled` in ioLsnr then for the case described in the warning below (client is in forceServerMode=true, cfg.isAuthenticationEnabled=true and there no other nodes) we can hang on this chain:

`ioLsnr -> onAuthenticateRequestMessage -> authenticateOnServer -> readyForAuthFut.get();`

Also for `discoLsnr` there can be NPE for `onNodeLeft -> coordinator().id()`.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
timoninmaxim: I see, thanks.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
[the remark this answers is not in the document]
timoninmaxim: OK.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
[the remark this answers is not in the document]
timoninmaxim: I see a method IgniteAuthenticationProcessor#authenticateNode, that is used as authenticator in DiscoverySpi. Maybe there is an if-check that makes this authentication unuseful, but it's not obvious. For me default behavior that returns NULL is dangerous and lead to NPE, also because method is not documented as Nullable.

If it's not used let's replace it with exception like Unsupported, or make all methods consistent (if allowPermissions are true then return all permissions there, if allowPermissions are false then return empty list there). WDYT?

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
timoninmaxim: > this method only makes sense for REMOTE_CLIENT

I see, thanks. Let's fix the javadoc then, WDYT?

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
timoninmaxim: I'm OK with that there is the ticket 14301 for discovering that.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
ololo3000: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
ololo3000: 1. I have fixed comment.
2. 
Mentioned by you problem is related only to client node with no server nodes are alive.
So onAuthenticateRequestMessage or onNodeLeft can't be called - they are require at leas one server node to be alive.
The problem appears during authentication method execution:
previously if no server node is found security was tuned off on client node and all authentication or user operations were skipped but this client node can't join the cluster with authentication enabled.
Now we can't turn off security so I propose to explicitly throw exception during authentication if no server nodes are alive.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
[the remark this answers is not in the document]
ololo3000: Fixed. ALLOW_ALL permissions set is returned now.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
timoninmaxim: OK, thanks!

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -1568,6 +1567,11 @@ private static DecimalFormat doubleFormat() {
     private GridProcessor securityProcessor() throws IgniteCheckedException {
         GridSecurityProcessor prc = createComponent(GridSecurityProcessor.class, ctx);
 
+        if (cfg.isAuthenticationEnabled() && !(prc instanceof IgniteAuthenticationProcessor)) {
```
nizhikov, line 1570: Why do we need this check?

modules/core/src/main/java/org/apache/ignite/internal/processors/security/GridSecurityProcessor.java
```diff
@@ -133,4 +134,39 @@ public void authorize(String name, SecurityPermission perm, SecurityContext secu
...
+
+    /**
+     * Creates user with the specified login and password.
+     *
+     * @param login Login of the user to be created.
+     * @param pwd User password.
+     * @throws IgniteCheckedException If error occurred.
+     */
+    @IgniteExperimental
+    public default void createUser(String login, String pwd) throws IgniteCheckedException {
```
nizhikov, line 146: We must use `char[] pwd` instead of `String` here and below.

modules/core/src/main/java/org/apache/ignite/internal/processors/security/GridSecurityProcessor.java
```diff
@@ -133,4 +134,39 @@ public void authorize(String name, SecurityPermission perm, SecurityContext secu
...
     }
+
+    /**
+     * Creates user with the specified login and password.
+     *
+     * @param login Login of the user to be created.
+     * @param pwd User password.
+     * @throws IgniteCheckedException If error occurred.
+     */
+    @IgniteExperimental
```
nizhikov, line 145: We don't need IgniteExperimental for internal classes.

modules/core/src/main/java/org/apache/ignite/internal/processors/security/IgniteSecurity.java
```diff
@@ -126,4 +127,33 @@ public default void authorize(SecurityPermission perm) throws SecurityException
...
     public boolean enabled();
+
+    /**
+     * Creates user with the specified login and password.
+     *
+     * @param login Login of the user to be created.
+     * @param pwd User password.
+     * @throws IgniteCheckedException If error occurred.
+     */
+    @IgniteExperimental
```
nizhikov, line 138: We don't need IgniteExperimental for internal classes.

ololo3000: Reverted.

modules/core/src/main/java/org/apache/ignite/internal/processors/security/GridSecurityProcessor.java
ololo3000: Reverted.

modules/core/src/main/java/org/apache/ignite/internal/processors/security/GridSecurityProcessor.java
ololo3000: Done.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
ololo3000: To prevent configuration of external security plugin and Ignite authentication at the same time.

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationPluginProvider.java
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.ignite.plugin.ExtensionRegistry;
+import org.apache.ignite.plugin.IgnitePlugin;
+import org.apache.ignite.plugin.PluginConfiguration;
+import org.apache.ignite.plugin.PluginContext;
+import org.apache.ignite.plugin.PluginProvider;
+import org.apache.ignite.plugin.PluginValidationException;
+import org.jetbrains.annotations.Nullable;
+
+/** Represents Ignite Authentication Plugin provider. */
+public class IgniteAuthenticationPluginProvider implements PluginProvider<PluginConfiguration> {
```
nizhikov, line 38: Let's do the following:

1. Remove this.
2. Create an instance of `IgniteAuthenticationPlugin` in `GridKernal#createComponent`
3. Make a callback to it just before plugin start.

ololo3000: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/security/NoOpIgniteSecurityProcessor.java
```diff
@@ -40,6 +41,9 @@
  * No operation IgniteSecurity.
  */
 public class NoOpIgniteSecurityProcessor extends GridProcessorAdapter implements IgniteSecurity {
+    /** Error message that occurs when trying to perform security operations if security is disabled. */
+    public static final String SECURITY_DISABLED_ERROR_MSG = "Operation cannot be performed: Ignite security is disabled.";
```
nizhikov, line 45: typo: "Ignite security disabled"

modules/core/src/main/java/org/apache/ignite/internal/processors/authentication/IgniteAuthenticationProcessor.java
```diff
@@ -136,8 +161,8 @@
     /** Node activate future. */
     private final GridFutureAdapter<Void> activateFut = new GridFutureAdapter<>();
 
-    /** Validate error. */
-    private String validateErr;
+    /** User management operations. */
+    private final EnumSet<SecurityPermission> userOps = EnumSet.of(CREATE_USER, DROP_USER, ALTER_USER);
```
nizhikov, line 165: This set used only one time so let's remove it and use explicit check like 

`if (op == CREATE_USER || op == ALTER_USER || op == DROP_USER)`

ololo3000: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/security/NoOpIgniteSecurityProcessor.java
ololo3000: Is it proposed to remove "is" and write "Operation cannot be performed: Ignite security disabled"?

nizhikov: Yes.

ololo3000: Done.

nizhikov (approved):

PR #8898, opened 2021-03-19
IGNITE-14347: Fix Node Failure on Receiving Data of Unknown Class via…

atris:
… Distributed Metastorage

Original author of this patch is Sam Vimes


atris:
https://ci.ignite.apache.org/project/IgniteTests24Java8?mode=builds

PR #8903, opened 2021-03-19
IGNITE-13399 Fix access right issues in computation of system metrics.

dmekhanikov:
Previously certain Ignite metrics were calculated based on private fields of some internal JVM MBeans. These private fields were accessed via reflection, which lead to errors on Java 9+, thus making the metrics have negative values.
This patch changes the way the required values are retrieved, so that no reflection is used.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
@@ -781,7 +781,8 @@ public class GridMetricManager extends GridManagerAdapter<MetricExporterSpi> imp
      */
     private long totalSysMemory() {
         try {
-            return U.<Long>property(os, "totalPhysicalMemorySize");
+            com.sun.management.OperatingSystemMXBean sunOs = (com.sun.management.OperatingSystemMXBean) os;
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/metric/GridMetricManager.java
```diff
@@ -840,9 +841,10 @@ private double getCpuLoad() {
             long cpuTime;
 
             try {
-                cpuTime = U.<Long>property(os, "processCpuTime");
+                com.sun.management.OperatingSystemMXBean sunOs = (com.sun.management.OperatingSystemMXBean) os;
+                cpuTime = sunOs.getProcessCpuTime();
             }
-            catch (IgniteException ignored) {
+            catch (ClassCastException ignored) {
```
alievmirza, line 847: What if `sunOs.getProcessCpuTime` throws RE?

dmekhanikov: I don't think it ever happens, but I don't mind replacing `ClassCastException` with `Exception`.

alievmirza (approved):

PR #8952, opened 2021-03-31
IGNITE-14452 : Add checking of the iptables settings applied.

Vladsz83:
Sometimes, we lack settings of iptables for unknows reason. Let's monitor this issue.


Vladsz83:
Migrated to https://github.com/apache/ignite/pull/8954

PR #8960, opened 2021-04-01
IGNITE-14402 Java thin client: Continuous queries support

alex-plekhanov:

### Notes

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
+++ b/modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
@@ -22,9 +22,11 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import javax.cache.configuration.CacheEntryListenerConfiguration;
+import javax.cache.event.CacheEntryListener;
 import javax.cache.expiry.ExpiryPolicy;
-
 import org.apache.ignite.cache.CachePeekMode;
+import org.apache.ignite.cache.query.ContinuousQuery;
 import org.apache.ignite.cache.query.FieldsQueryCursor;
 import org.apache.ignite.cache.query.Query;
 import org.apache.ignite.cache.query.QueryCursor;
@@ -703,13 +705,28 @@ public interface ClientCache<K, V> {
     public <K1, V1> ClientCache<K1, V1> withExpirePolicy(ExpiryPolicy expirePlc);
 
     /**
-     * Queries cache. Supports {@link ScanQuery} and {@link SqlFieldsQuery}.
+     * Queries cache. Supports {@link ScanQuery}, {@link SqlFieldsQuery} and {@link ContinuousQuery}.
+     * <p>
+     * NOTE: For continuous query listeners there is no failover in case of client channel failure, this event should
+     * be handled on the user's side. Use {@link #query(ContinuousQuery, ClientDisconnectListener)} method to get
+     * notified about client disconnected event via {@link ClientDisconnectListener} interface if you need it.
      *
      * @param qry Query.
      * @return Cursor.
-     *
      */
     public <R> QueryCursor<R> query(Query<R> qry);
+
+    /**
+     * Start {@link ContinuousQuery} on the cache.
+     * <p>
+     * NOTE: There is no failover in case of client channel failure, this event should be handled on the user's side.
+     * Use {@code disconnectListener} to handle this.
+     *
+     * @param qry Query.
+     * @param disconnectListener Listener of client disconnected event.
+     * @return Cursor.
+     */
+    public <R> QueryCursor<R> query(ContinuousQuery<K, V> qry, ClientDisconnectListener disconnectListener);
 
     /**
      * Convenience method to execute {@link SqlFieldsQuery}.
@@ -718,4 +735,43 @@ public interface ClientCache<K, V> {
      * @return Cursor.
      */
     public FieldsQueryCursor<List<?>> query(SqlFieldsQuery qry);
+
+    /**
+     * Registers a {@link CacheEntryListener}. The supplied {@link CacheEntryListenerConfiguration} is used to
+     * instantiate a listener and apply it to those events specified in the configuration.
+     * <p>
+     * NOTE: There is no failover in case of client channel failure, this event should be handled on the user's side.
+     * Use {@link #registerCacheEntryListener(CacheEntryListenerConfiguration, ClientDisconnectListener)} method to get
+     * notified about client disconnected event via {@link ClientDisconnectListener} interface if you need it.
+     *
+     * @param cacheEntryListenerConfiguration a factory and related configuration for creating the listener.
+     * @throws IllegalArgumentException is the same CacheEntryListenerConfiguration is used more than once or
+     *          if some unsupported by thin client properties are set.
+     * @see CacheEntryListener
+     */
+    public void registerCacheEntryListener(CacheEntryListenerConfiguration<K, V> cacheEntryListenerConfiguration);
+
+    /**
+     * Registers a {@link CacheEntryListener}. The supplied {@link CacheEntryListenerConfiguration} is used to
+     * instantiate a listener and apply it to those events specified in the configuration.
+     * <p>
+     * NOTE: There is no failover in case of client channel failure, this event should be handled on the user's side.
+     * Use {@code disconnectListener} to handle this.
+     *
+     * @param cacheEntryListenerConfiguration a factory and related configuration for creating the listener.
+     * @param disconnectListener Listener of client disconnected event.
+     * @throws IllegalArgumentException is the same CacheEntryListenerConfiguration is used more than once or
+     *          if some unsupported by thin client properties are set.
+     * @see CacheEntryListener
+     */
+    public void registerCacheEntryListener(CacheEntryListenerConfiguration<K, V> cacheEntryListenerConfiguration,
+        ClientDisconnectListener disconnectListener);
+
+    /**
+     * Deregisters a listener, using the {@link CacheEntryListenerConfiguration} that was used to register it.
+     *
+     * @param cacheEntryListenerConfiguration the factory and related configuration that was used to create the
+     *         listener.
+     */
+    public void deregisterCacheEntryListener(CacheEntryListenerConfiguration<K, V> cacheEntryListenerConfiguration);
 }
--- a/modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenerHandler.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenerHandler.java
@@ -1,0 +1,265 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.client.thin;
+
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import javax.cache.Cache;
+import javax.cache.configuration.Factory;
+import javax.cache.event.CacheEntryEvent;
+import javax.cache.event.CacheEntryEventFilter;
+import javax.cache.event.CacheEntryUpdatedListener;
+import javax.cache.event.EventType;
+import org.apache.ignite.cache.query.ContinuousQuery;
+import org.apache.ignite.client.ClientDisconnectListener;
+import org.apache.ignite.client.ClientException;
+import org.apache.ignite.internal.binary.GridBinaryMarshaller;
+import org.apache.ignite.internal.binary.streams.BinaryByteBufferInputStream;
+import org.apache.ignite.internal.binary.streams.BinaryInputStream;
+import org.apache.ignite.internal.binary.streams.BinaryOutputStream;
+import org.apache.ignite.internal.util.typedef.T2;
+import org.apache.ignite.internal.util.typedef.internal.U;
+
+import static org.apache.ignite.internal.client.thin.ClientNotificationType.CONTINUOUS_QUERY_EVENT;
+import static org.apache.ignite.internal.client.thin.TcpClientCache.JAVA_PLATFORM;
+
+/**
+ * Handler for {@link ContinuousQuery} listeners and JCache cache entry listeners.
+ */
+public class ClientCacheEntryListenerHandler<K, V> implements NotificationListener, AutoCloseable {
+    /** "Keep binary" flag mask. */
+    private static final byte KEEP_BINARY_FLAG_MASK = 0x01;
+
+    /** */
+    private final Cache<K, V> jCacheAdapter;
+
+    /** */
+    private final ReliableChannel ch;
+
+    /** */
+    private final boolean keepBinary;
+
+    /** */
+    private final ClientUtils utils;
+
+    /** */
+    private volatile CacheEntryUpdatedListener<K, V> locLsnr;
+
+    /** */
+    private volatile ClientDisconnectListener disconnectLsnr;
+
+    /** */
+    private volatile ClientChannel clientCh;
+
+    /** */
+    private volatile Long rsrcId;
+
+    /** */
+    ClientCacheEntryListenerHandler(
+        Cache<K, V> jCacheAdapter,
+        ReliableChannel ch,
+        ClientBinaryMarshaller marsh,
+        boolean keepBinary
+    ) {
+        this.jCacheAdapter = jCacheAdapter;
+        this.ch = ch;
+        this.keepBinary = keepBinary;
+        utils = new ClientUtils(marsh);
+    }
+
+    /**
+     * Send request to the server and start
+     */
+    public synchronized void startListen(
+        CacheEntryUpdatedListener<K, V> locLsnr,
+        ClientDisconnectListener disconnectLsnr,
+        Factory<? extends CacheEntryEventFilter<? super K, ? super V>> rmtFilterFactory,
+        int pageSize,
+        long timeInterval,
+        boolean includeExpired
+    ) {
+        assert locLsnr != null;
+
+        if (clientCh != null)
+            throw new IllegalStateException("Listener was already started");
+
+        this.locLsnr = locLsnr;
+        this.disconnectLsnr = disconnectLsnr;
+
+        Consumer<PayloadOutputChannel> qryWriter = payloadCh -> {
+            BinaryOutputStream out = payloadCh.out();
+
+            out.writeInt(ClientUtils.cacheId(jCacheAdapter.getName()));
+            out.writeByte(keepBinary ? KEEP_BINARY_FLAG_MASK : 0);
+            out.writeInt(pageSize);
+            out.writeLong(timeInterval);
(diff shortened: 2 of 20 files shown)


modules/core/src/main/java/org/apache/ignite/client/ClientChannelDisconnectListener.java
```diff
@@ -0,0 +1,33 @@
...
+
+import javax.cache.configuration.CacheEntryListenerConfiguration;
+import org.apache.ignite.cache.query.ContinuousQuery;
+
+/**
+ * This interface can be implemented by {@link ContinuousQuery} local listeners or by cache entry listeners
+ * registered via {@link ClientCache#registerCacheEntryListener(CacheEntryListenerConfiguration)} to get notification
+ * about client channel disconnected event.
+ */
+public interface ClientChannelDisconnectListener {
```
ptupitsyn, line 28: `ClientDisconnectListener` seems to be a better name - `channel` sounds like an implementation detail.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenerHandler.java
```diff
@@ -0,0 +1,266 @@
...
+import org.apache.ignite.client.ClientException;
+import org.apache.ignite.internal.binary.GridBinaryMarshaller;
+import org.apache.ignite.internal.binary.streams.BinaryByteBufferInputStream;
+import org.apache.ignite.internal.binary.streams.BinaryInputStream;
+import org.apache.ignite.internal.binary.streams.BinaryOutputStream;
+import org.apache.ignite.internal.util.typedef.T2;
+import org.apache.ignite.internal.util.typedef.internal.U;
+
+/**
+ *
```
ptupitsyn, line 40: Missing javadoc

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenerHandler.java
```diff
@@ -0,0 +1,266 @@
...
+            out.writeByte(keepBinary ? KEEP_BINARY_FLAG_MASK : 0);
+            out.writeInt(pageSize);
+            out.writeLong(timeInterval);
+            out.writeBoolean(includeExpired);
+
+            if (rmtFilterFactory == null)
+                out.writeByte(GridBinaryMarshaller.NULL);
+            else {
+                utils.writeObject(out, rmtFilterFactory);
+                out.writeByte((byte)1); // Java platform
```
ptupitsyn, line 111: `ClientPlatform.JAVA`

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenerHandler.java
```diff
@@ -0,0 +1,266 @@
...
+
+    /** {@inheritDoc} */
+    @Override public void acceptNotification(
+            ClientChannel ch,
+            ClientOperation op,
+            long rsrcId,
+            ByteBuffer payload,
+            Exception err
+    ) {
+        if (op == ClientOperation.QUERY_CONTINUOUS_EVENT_NOTIFICATION && rsrcId == this.rsrcId) {
```
ptupitsyn, line 156: nit: Invert condition to reduce nesting

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenerHandler.java
```diff
@@ -0,0 +1,266 @@
...
+            if (err == null && payload != null) {
```
ptupitsyn, line 157: Combine with the previous condition?

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenerHandler.java
```diff
@@ -0,0 +1,266 @@
...
+            }
+        }
+    }
+
+    /** {@inheritDoc} */
+    @Override public synchronized void close() {
+        ch.removeChannelCloseListener(chCloseLsnr);
+
+        if (clientCh != null) {
+            clientCh.removeNotificationListener(this);
```
ptupitsyn, line 186: Even though we never set `clientCh` from non-null to null, and this code seems to be safe right now,
the field is still marked `volatile`, so this use-after-nullcheck looks suspicious.

To reduce cognitive load, copy `clientCh` to a temp variable, and then check for null.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenerHandler.java
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    /** */
+    private EventType eventType(byte evtTypeByte) {
+        switch (evtTypeByte) {
+            case 0: return EventType.CREATED;
+            case 1: return EventType.UPDATED;
+            case 2: return EventType.REMOVED;
+            case 3: return EventType.EXPIRED;
+            default: return null;
```
ptupitsyn, line 207: Should we log an error instead? Other event types are not valid.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenersRegistry.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import javax.cache.configuration.CacheEntryListenerConfiguration;
+
+/**
+ * Per-cache cache entry listeners registry. Listeners can't be stored inside ClientCache instance, since there can be
+ * several such instances per one cache.
+ */
+public class ClientCacheEntryListenersRegistry {
```
ptupitsyn, line 28: I think we should do the following:
* Refactor existing `TcpClientChannel.notificationLsnrs` to a `Map<Long, NotificationListener>`
* Instead of iterating over all `notificationLsnrs` and checking the id in every listener, get the listener by id from map and invoke it without inner checks
* Use this mechanism for continuous query notifications, avoid creating a new registry

This approach seems to work fine in .NET thin client.

ptupitsyn (changes requested):

modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
```diff
@@ -703,11 +705,14 @@
     public <K1, V1> ClientCache<K1, V1> withExpirePolicy(ExpiryPolicy expirePlc);
 
     /**
-     * Queries cache. Supports {@link ScanQuery} and {@link SqlFieldsQuery}.
+     * Queries cache. Supports {@link ScanQuery}, {@link SqlFieldsQuery} and {@link ContinuousQuery}.
+     * <p>
+     * NOTE: For continuous query listeners there is no failover in case of client channel failure, this event should
+     * be handled on the user's side. Listeners will be notified about client disconnect event if they implement
+     * {@link ClientChannelDisconnectListener} interface.
```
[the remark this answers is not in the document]
alex-plekhanov: Ok

modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
```diff
@@ -718,4 +723,27 @@
...
+     * NOTE: There is no failover in case of client channel failure, this event should be handled on the user's side.
+     * Listeners will be notified about client disconnect event if they implement {@link ClientChannelDisconnectListener}
+     * interface.
+     *
+     * @param cacheEntryListenerConfiguration a factory and related configuration for creating the listener.
+     * @throws IllegalArgumentException is the same CacheEntryListenerConfiguration is used more than once or
+     *          if some unsupported by thin client properties are set.
+     * @see CacheEntryListener
+     */
+    public void registerCacheEntryListener(CacheEntryListenerConfiguration<K, V> cacheEntryListenerConfiguration);
```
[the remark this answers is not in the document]
alex-plekhanov: But why not provide this ability to the user if it's relatively low cost and no changes needed on server-side? The user can use these methods in thick client, and having these methods in thin client makes migration to thin client easier.

modules/core/src/main/java/org/apache/ignite/client/ClientChannelDisconnectListener.java
alex-plekhanov: Ok

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenerHandler.java
alex-plekhanov: `ClientPlatform' it's a server-side class, perhaps it's better to introduce some client-side constant and refactor `scanQuery` as well.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenerHandler.java
alex-plekhanov: We don't have logger on the client-side

modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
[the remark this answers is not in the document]
ptupitsyn: Ok, no objections.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenerHandler.java
ptupitsyn: > introduce some client-side constant

Sounds good to me. Let's avoid magic numbers.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenersRegistry.java
alex-plekhanov: This will require creating new listener for each compute task and refactoring of existing compute functionality (I think if we agreed to do this, it should be done by another ticket). Also, this change will add new limitations: we can't use notifications without resource id in the future.
To reduce count of listeners, I think for continuous queries we can act the same way as for compute: register only one listener for QUERY_CONTINUOUS_EVENT_NOTIFICATION and find handles in this listener by resource id using some map. WDYT?

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenerHandler.java
ptupitsyn: Then we should close the channel with a `Throwable cause` that explains the problem.
Invalid event type means some kind of protocol breakdown, we should not silently continue.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientCacheEntryListenersRegistry.java
ptupitsyn: > new limitations: we can't use notifications without resource id in the future

As I understand, this limitation already exists - all messages have an ID in any case.

> find handles in this listener by resource id using some map

We could do this, but again, why have `Collection<NotificationListener>`, then separate maps in Compute and Continuous Query (and, possibly, more places in future), when a single map will:
* Simplify the implementation
* Make it more efficient (single lookup, single collection)

ptupitsyn: >  it should be done by another ticket

Ok, let's do this first, and then get back to the current ticket?

alex-plekhanov: > As I understand, this limitation already exists - all messages have an ID in any case.
There is no requirement for the uniqueness of this ID. We can use 0 for some events (for example, we have topology change event implemented as special flag of every response, but it can be implemented with notification and no additional resource ID required here). Also, perhaps in the future events initiated by server-side for some resource will be needed, in this case, client should be able to listen some type of notification with any resource ID. 
> why have Collection<NotificationListener>, then separate maps in Compute and Continuous Query
We can replace the first collection with `EnumMap` (Listener per ClientOperation), it is very lightweight and fast, and there will be no iterations. There will be at most one listener per one operation type, and each listener can implement its own strategy for resource ID mapping. This change does not require to refactor compute a lot, so, I think, it can be implemented in the current ticket. 
Or we can even provide two API on client channel level: listener for client operation and any resource id (refactor compute to use this API) and listener for client operation and exactly one resource Id. WDYT?

ptupitsyn: > There is no requirement for the uniqueness of this ID

Right now this ID is required to match either a compute task ID, or a continuous query ID.
A combination of `resId`+`notificationCode` is required to be "unique". We can use `IgniteBiTuple` as a map key.
If we are going to introduce predefined ids (like 0 or -1) for general-purpose events - that will work fine as well.

> each listener can implement its own strategy for resource ID mapping

This looks like code duplication to me.

alex-plekhanov: > If we are going to introduce predefined ids (like 0 or -1) for general-purpose events - that will work fine as well.

There is still a case with resource IDs initiated from the server, but ok, we don't have it now, if we will need it we can change implementation.
I've tried to change notification listeners and listen to resource ID, but unfortunately, there is one critical problem: we can process notification before the notification listener is registered. For compute it's a blocker since notification can be lost and the future will never complete. This happens because the earliest place when we can register listener - it's payload reader, payload reader executed in user's thread or in `asyncContinuationExecutor` thread. Concurrently the new message (notification) from the channel can be read by NIO thread and executed by `asyncContinuationExecutor` without notifying listeners.

ptupitsyn: The same is true for Continuous Query - you can get a notification before you know the query ID and register the listener (and you can have multiple active queries).
This is another piece of logic to be reused - that's why we need a centralized notification handling mechanism.

We can queue "unknown" notifications per resourceId and handle them as soon as listener is registered - we'll have to do it one way or another. For Compute the queue will have at most 1 item, but for CQ it can have any number of events.

ptupitsyn: Please have a look at [C# implementation](https://github.com/apache/ignite/blob/master/modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/ClientNotificationHandler.cs) - hopefully I did not miss anything there.

The only potential problem is that `requestId` is assumed to be unique across Compute, CQ and other future notification types - we should use a combined key instead (notificationType + requestId).

alex-plekhanov: > The same is true for Continuous Query - you can get a notification before you know the query ID and register the listener (and you can have multiple active queries).

For CQ we don't need such strong guarantees. We can skip some events from the start and CQ still will be valid (it's like if we start listening a little bit later).

> This is another piece of logic to be reused.
> We can queue "unknown" notifications per resourceId and handle them as soon as listener is registered

The difference between CQ and compute: compute can't skip notifications, compute has exactly one notification, taking into account this logic compute was implemented in such way, that unregistering listener performed when receiving notification. For CQ there can be many notifications per listener and listener unregistering should be performed when we close CQ. And if we will use some queue for pending notifications there will be notifications leaks after CQ closing, some notifications can stay in queue forever (as far as I understand C# implementation has the same problem). We can use some flag for notification type (use pending notification queue for compute and don't use it for CQ) to solve this problem, but I'm not sure that it's better than solution with different logic for different components.

alex-plekhanov: @ptupitsyn I've refactored notification listeners and partially move changes to the ticket 
https://issues.apache.org/jira/browse/IGNITE-14492. When this ticket will be in the master branch I will rebase the current ticket.

alex-plekhanov: @ptupitsyn, all comments fixed, ticket rebased, can you please have a look again?

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientCache.java
```diff
@@ -747,6 +775,92 @@ else if (qry instanceof SqlFieldsQuery)
         ));
     }
 
+    /** {@inheritDoc} */
+    @Override public <R> QueryCursor<R> query(ContinuousQuery<K, V> qry, ClientDisconnectListener disconnectLsnr) {
+        A.ensure(!(qry.getInitialQuery() instanceof ContinuousQuery), "Initial query for continuous query " +
+            "can't be an instance of another continuous query");
+        A.notNull(qry.getLocalListener(), "Local listener");
+        A.ensure(!qry.isLocal(), "Unsupported Local flag value");
```
ptupitsyn, line 783: What about a message like `Local query is not supported by thin client`?

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientCache.java
```diff
@@ -747,6 +775,92 @@ else if (qry instanceof SqlFieldsQuery)
...
+        A.ensure(qry.isAutoUnsubscribe(), "Unsupported AutoUnsubscribe flag value");
```
ptupitsyn, line 784: Similarly, `AutoUnsubscribe flag is not supported by thin client`.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientCache.java
```diff
@@ -747,6 +775,92 @@ else if (qry instanceof SqlFieldsQuery)
...
+        A.ensure(qry.getRemoteFilterFactory() == null || qry.getRemoteFilter() == null,
+            "Should be used either RemoterFilter or RemoteFilterFactory.");
```
ptupitsyn, line 786: `RemoteFilter and RemoteFilterFactory can't be used together`

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientCache.java
```diff
@@ -947,4 +1061,73 @@ private void writeEntries(Map<? extends K, ? extends V> map, PayloadOutputChanne
                     serDes.writeObject(out, e.getValue());
                 });
     }
+
+    /**
+     * Adapter to convert CQ listener calls to JCache listener calls.
+     */
+    private static class JCacheEntryListenerAdapter<K, V> implements CacheEntryUpdatedListener<K, V> {
```
ptupitsyn, line 1068: The file is too big already, can we move those classes to separate files?

modules/core/src/test/java/org/apache/ignite/internal/client/thin/CacheEntryListenersTest.java
```diff
@@ -0,0 +1,708 @@
...
+import org.apache.ignite.internal.util.typedef.internal.U;
+import org.junit.Test;
+
+import static org.apache.ignite.testframework.GridTestUtils.assertThrowsWithCause;
+import static org.apache.ignite.testframework.GridTestUtils.waitForCondition;
+
+/**
+ * Thin client cache entry listeners test.
+ */
+public class CacheEntryListenersTest extends AbstractThinClientTest {
```
ptupitsyn, line 59: Let's add a test that combines one or more continuous queries with compute calls: since we use a shared notification mechanism, it would be good to check this scenario.

ptupitsyn (changes requested):

ptupitsyn:
@alex-plekhanov looks good to me in general, please see a few comments.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientCache.java
alex-plekhanov: I've moved `JCacheEntryListenerAdapter` to the separate file, but `JCacheDisconnectListenerAdapter` is not static and relies on `TcpClientCache` fields, so I've inlined it as an anonymous class.

PR #9001, opened 2021-04-12
IGNITE-14516 Always log to not archived segment

nizhikov:
This PR fixes concurrent bug on the node stop.
But scenario:

* `DataStorageConfiguration#getWalAutoArchiveAfterInactivity()} > 0`
* Node is gracefully stopping using `G#stop(String, boolean)`.
* T0: `Checkpointer#doCheckpoint()` execute last checkpoint on stop and freeze.
* T1: Rollover segment after inactivity timeout.
* T2: Archive segment.

After restart WAL should log in to the next segment.

This bug fixed with the check on node start.
Ignite will continue to log into the next, not archived segment.

### Notes


ivandasch (approved):
Looks good to me.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/wal/WalRolloverOnStopTest.java
```diff
@@ -0,0 +1,169 @@
...
+            // No-op.
+        }
+
+        /** {@inheritDoc} */
+        @Override public void beforeCheckpointBegin(Context ctx) {
+            // No-op.
+        }
+    }
+}
+
```
ivandasch, line 169: Please, remove extra new line.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/wal/WalRolloverOnStopTest.java
```diff
@@ -0,0 +1,168 @@
...
+            AtomicLong cntr = new AtomicLong(0);
+
+            db.addCheckpointListener(new WaitOnLastCheckpoint(ign, waitAfterCp, cntr, getTestTimeout()));
+
+            for (int j = i * 3; j < (i + 1) * 3; j++)
+                cache.put(j, j);
+
+            long cutIdx = aware.curAbsWalIdx();
+
+            runAsync(() -> {
```
Mmuzaf, line 94: We should wait for the completion of this runnable prior to checking the `cntr` it increments.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/wal/WalRolloverOnStopTest.java
```diff
@@ -0,0 +1,168 @@
...
+                finally {
+                    waitAfterCp.countDown();
+                }
+            });
+
+            G.stop(ign.name(), false);
+
+            // Checkpoint will happens two time because of segment archivation.
+            assertEquals("Should successfully wait for current segment archivation", 3, cntr.get());
+        }
```
Mmuzaf, line 112: I think we should also check:
1. the logical WAL record is in the archive after stop
2. the node after start uses the `currIdx` segment number

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/wal/WalRolloverOnStopTest.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /** */
+    private static class WaitOnLastCheckpoint implements CheckpointListener {
```
Mmuzaf, line 116: I think the anonymous class can be used as a test listener. This will simplify the code itself.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -1325,6 +1325,14 @@ private FileWriteHandle rollOver(FileWriteHandle cur, @Nullable WALRecord rec) t
     private FileWriteHandle restoreWriteHandle(@Nullable WALPointer lastReadPtr) throws StorageException {
         long absIdx = lastReadPtr == null ? 0 : lastReadPtr.index();
 
+        FileDescriptor[] walArchiveFiles = walArchiveFiles();
+
+        if (archiver != null &&
+            !F.isEmpty(walArchiveFiles) &&
+            absIdx <= walArchiveFiles[walArchiveFiles.length - 1].idx) {
+            absIdx = walArchiveFiles[walArchiveFiles.length - 1].idx + 1;
```
Mmuzaf, line 1333: Is it correct to set the segment index based on the wal-archive directory? As far as I understand, the issue about the rolling over a segment that we are losing the node stop.

So, in the test `testWallRollover` the following line returns different segment index on the second iteration (after th node restart):
`long cutIdx = aware.curAbsWalIdx();`

1. if the WAL archive enabled `cutIdx=0` on the first iteration and `cutIdx=1` on the second iteration (this is correct)
2. if the WAL archive disabled `cutIdx=0` on the first iteration and `cutIdx=0` on the second iteration, however, the rollover already happened and after restart, the node will log entries to the incorrect WAL segment (this should be fixed)

ivandasch: This is not true, "wal archive disabled" basically means that `walArchiveDir` and `walWorkDir` are equal. So `walArchiveFiles()` will work as it expected in that case.

nizhikov: > if the WAL archive disabled `cutIdx=0` on the first iteration and `cutIdx=0` on the second iteration, however, the rollover already happened and after restart, the node will log entries to the incorrect WAL segment (this should be fixed)

It's not an issue to continue to log into some segment if "rollover" happens before the restart, because there is nothing to lose(next segment is empty) and we just "write-ahead" to log.

The issue happens if we already copy the current state of the segment to archive. 
If we continue to log into copied segment then we lose newly logged records, because co

tkalkirill: It seems to me that this is not quite the right decision, perhaps it is worth adding a marker record to the new segment so that everything works as it should. @sergey-chugunov-1985 WDYT?

sergey-chugunov-1985: @tkalkirill, from user perspective this change seems working as it fixes wrong behavior. But from code perspective I agree with you: we add some logic that increments `absIdx` based on checks that seems absolutely random. Adding a marker record with proper documentation would make this code much more readable and easy to understand.

nizhikov: @tkalkirill 
@sergey-chugunov-1985 

What marker record do you have in mind guys? I don't understand your proposal.

tkalkirill: You can create a new record like **org.apache.ignite.internal.pagemem.wal.record.SwitchSegmentRecord**.

nizhikov: For what purpose?

tkalkirill: So that when the automatic rollover happens, we write the record and then the segment would not be empty and after restarting we would continue to write to it, and not to the previous one.

nizhikov: > segment would not be empty and after restarting we would continue to write to it

It sounds very hacky to me :)
I slightly reworked the solution, please, take a look one more time.

It seems right now code understandability OK.

Mmuzaf: @tkalkirill 

Hello, can you please clarify the `SwitchSegmentRecord` purpose? Will it be written to the new segment each time the rollover procedure happened?

tkalkirill: @Mmuzaf 
Hello, i gave the **SwitchSegmentRecord** as an example of marker WAL record, you can make a similar one and write it (new marker WAL record) after rollover (or only after automatic rollover). Then, after restarting the node, there will be no problem that occurs in this ticket.

nizhikov: The root issue is the following: Right now Ignite continues to write to rollovered segment on restart.
There is no way to distinguish between the already rollovered and "current" segment on node failover.

Your proposal doesn't fix the root issue - we still can fall into a scenario when the segment rollovered and archived but the marker record not written in the next segment.

tkalkirill: @nizhikov 
We define a segment into which we will continue to write based on the last read WAL record. In the normal case, everything works well because after rollover, we write a record to a new segment, see **FileWriteAheadLogManager#log**.

In the root cause of the current ticket, we do not write any WAL record to the new segment, and at the same time archive the previous segment, so after restarting the node, we will try to read the WAL records in a new segment, but they are not there, so we will ignore it, so we continue to write to the previous segment. If we had written a record to the auto-rolloved segment, then there would be no problem since we would have read this WAL record in a new segment and continued writing to a new segment.

nizhikov:
@sergey-chugunov-1985 
@tkalkirill 
@Mmuzaf 
@ivandasch 

I reworked the PR based on your comments and proposals.
I think we should do the following:

1. Switch to the next segment in `FileWriteAheadLogManager#restoreWriteHandle` if failover iteration stops on `SWITCH_SEGMENT_RECORD`.
2. Create a flag `WalIterator#switchSegmentRecordReached` to implement 1.

Looks like we have 2 cases when we stop iteration, not on `SWITCH_SEGMENT_RECORD`:

1. Segment is full and `SWITCH_SEGMENT_RECORD` doesn't fill in it - then it doesn't matter when archivation happens, we will have a consistent segment in the archive.
2. Node fails with the unfilled segment - it's OK to continue to log into it.

Please, take a look.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
nizhikov: @tkalkirill 

I see your point

> If we had written a record to the auto-rolloved segment

Write something just to make segment "nonempty" is the wrong approach - we shouldn't do spare writes.
Moreover, spare write can not happen, also due to the write thread feeze.
So with some "fake" record we just make this issue less probable.
I think we should make a decision based on the currently existing `SwitchSegmentRecord`. 

Please, take a look at PR and my [comment](https://github.com/apache/ignite/pull/9001#issuecomment-820201450)

> In the root cause of the current ticket, we do not write any WAL record to the new segment

My understanding is different - the root cause of the issue is the failover algorithm that, currently, can't distinguish between full and nonfull segments during node start - this leads to the wrong decision on what segment Ignite should continue to log into.

tkalkirill:
@nizhikov I disagree with this case:
`Segment is full and SWITCH_SEGMENT_RECORD doesn't fill in it - then it doesn't matter when archivation happens, we will have a consistent segment in the archive.`

This can still be a problem if we did a rollover without recording SWITCH_SEGMENT_RECORD and had time to archive it.

PR #9009, opened 2021-04-15
IGNITE-13691 Support of EXCEPT operator

alex-plekhanov:

### Notes

### DIFF
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/RowHandler.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/RowHandler.java
@@ -41,9 +41,6 @@ public interface RowHandler<Row> {
     int columnCount(Row row);
 
     /** */
-    Object[] getColumns(Row row);
-
-    /** */
     default RowFactory<Row> factory(IgniteTypeFactory typeFactory, RelDataType rowType) {
         if (rowType.isStruct())
             return factory(typeFactory, RelOptUtil.getFieldTypeList(rowType));
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/RowHandler.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/RowHandler.java
@@ -41,6 +41,9 @@ public interface RowHandler<Row> {
     int columnCount(Row row);
 
     /** */
+    Object[] getColumns(Row row);
+
+    /** */
     default RowFactory<Row> factory(IgniteTypeFactory typeFactory, RelDataType rowType) {
         if (rowType.isStruct())
             return factory(typeFactory, RelOptUtil.getFieldTypeList(rowType));
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/MinusNode.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/MinusNode.java
@@ -63,14 +63,14 @@ public class MinusNode<Row> extends AbstractNode<Row> {
      * @param ctx Execution context.
      */
     public MinusNode(ExecutionContext<Row> ctx, RelDataType rowType, AggregateType type, boolean all,
-        RowFactory<Row> rowFactory, int sourcesCnt) {
+        RowFactory<Row> rowFactory) {
         super(ctx, rowType);
 
         this.all = all;
         this.type = type;
         this.rowFactory = rowFactory;
 
-        grouping = new Grouping(sourcesCnt);
+        grouping = new Grouping();
     }
 
     /** {@inheritDoc} */
@@ -111,6 +111,14 @@ public class MinusNode<Row> extends AbstractNode<Row> {
         assert curSrcIdx == idx;
 
         checkState();
+
+        if (type == AggregateType.SINGLE && grouping.isEmpty()) {
+            requested = 0;
+
+            downstream().end();
+
+            return;
+        }
 
         curSrcIdx++;
 
@@ -194,18 +202,17 @@ public class MinusNode<Row> extends AbstractNode<Row> {
 
     /** */
     private class Grouping {
-        /** */
+        /**
+         * Value in this map will always have 2 elements, first - count of keys in the first set, second - count of
+         * keys in all sets except first.
+         */
         private final Map<GroupKey, int[]> groups = new HashMap<>();
 
         /** */
         private final RowHandler<Row> hnd;
 
-        /** Count of sets (input sources). */
-        private final int setsCnt;
-
         /** */
-        private Grouping(int setsCnt) {
-            this.setsCnt = setsCnt;
+        private Grouping() {
             hnd = context().rowHandler();
         }
 
@@ -216,8 +223,10 @@ public class MinusNode<Row> extends AbstractNode<Row> {
 
                 addOnReducer(row);
             }
-            else
+            else if (type == AggregateType.MAP)
                 addOnMapper(row, setIdx);
+            else
+                addOnSingle(row, setIdx);
         }
 
         /**
@@ -231,21 +240,56 @@ public class MinusNode<Row> extends AbstractNode<Row> {
             else if (type == AggregateType.MAP)
                 return getOnMapper(cnt);
             else
-                return getOnReducer(cnt);
+                return getOnSingleOrReducer(cnt);
+        }
+
+        /** */
+        private GroupKey key(Row row) {
+            int size = hnd.columnCount(row);
+
+            Object[] fields = new Object[size];
+
+            for (int i = 0; i < size; i++)
+                fields[i] = hnd.get(i, row);
+
+            return new GroupKey(fields);
+        }
+
+        /** */
+        private void addOnSingle(Row row, int setIdx) {
+            int[] cntrs;
+
+            GroupKey key = key(row);
+
+            if (setIdx == 0) {
+                cntrs = groups.computeIfAbsent(key, k -> new int[2]);
+
+                cntrs[0]++;
+            }
+            else {
+                cntrs = groups.get(key);
+
+                if (cntrs != null) {
+                    cntrs[1]++;
+
+                    if (cntrs[1] >= cntrs[0])
+                        groups.remove(key);
+                }
+            }
         }
 
         /** */
         private void addOnMapper(Row row, int setIdx) {
-            int[] cntrs = groups.computeIfAbsent(new GroupKey(hnd.getColumns(row)), k -> new int[setsCnt]);
+            int[] cntrs = groups.computeIfAbsent(key(row), k -> new int[2]);
 
-            cntrs[setIdx]++;
+            cntrs[setIdx == 0 ? 0 : 1]++;
         }
 
         /** */
         private void addOnReducer(Row row) {
             GroupKey grpKey = (GroupKey)hnd.get(0, row);
 
-            int[] cntrs = groups.computeIfAbsent(grpKey, k -> new int[setsCnt]);
+            int[] cntrs = groups.computeIfAbsent(grpKey, k -> new int[2]);
 
             int[] cntrsMap = (int[])hnd.get(1, row);
 
@@ -260,10 +304,15 @@ public class MinusNode<Row> extends AbstractNode<Row> {
             int amount = Math.min(cnt, groups.size());
             List<Row> res = new ArrayList<>(amount);
 
-            for (int i = 0; i < amount; i++) {
+            while (amount > 0 && it.hasNext()) {
                 Map.Entry<GroupKey, int[]> entry = it.next();
 
-                res.add(rowFactory.create(entry.getKey(), entry.getValue()));
+                // Skip row if it doesn't affect the final result.
+                if (entry.getValue()[0] != entry.getValue()[1]) {
+                    res.add(rowFactory.create(entry.getKey(), entry.getValue()));
+
+                    amount--;
+                }
 
                 it.remove();
             }
@@ -272,7 +321,7 @@ public class MinusNode<Row> extends AbstractNode<Row> {
         }
 
         /** */
-        private List<Row> getOnReducer(int cnt) {
+        private List<Row> getOnSingleOrReducer(int cnt) {
             Iterator<Map.Entry<GroupKey, int[]>> it = groups.entrySet().iterator();
 
             List<Row> res = new ArrayList<>(cnt);
@@ -280,7 +329,15 @@ public class MinusNode<Row> extends AbstractNode<Row> {
             while (it.hasNext() && cnt > 0) {
                 Map.Entry<GroupKey, int[]> entry = it.next();
 
-                Row row = rowFactory.create(entry.getKey().fields());
+                GroupKey key = entry.getKey();
+
+                Object[] fields = new Object[key.fieldsCount()];
+
+                for (int i = 0; i < fields.length; i++)
+                    fields[i] = key.field(i);
+
+                Row row = rowFactory.create(fields);
+
                 int[] cntrs = entry.getValue();
 
                 int availableRows = availableRows(entry.getValue());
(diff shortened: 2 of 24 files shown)


modules/calcite/src/test/java/org/apache/ignite/testsuites/PlannerTestSuite.java
```diff
@@ -20,13 +20,15 @@
...
 import org.apache.ignite.internal.processors.query.calcite.planner.CorrelatedNestedLoopJoinPlannerTest;
+import org.apache.ignite.internal.processors.query.calcite.planner.ExceptPlannerTest;
 import org.apache.ignite.internal.processors.query.calcite.planner.HashAggregatePlannerTest;
 import org.apache.ignite.internal.processors.query.calcite.planner.HashIndexSpoolPlannerTest;
 import org.apache.ignite.internal.processors.query.calcite.planner.JoinColocationPlannerTest;
 import org.apache.ignite.internal.processors.query.calcite.planner.PlannerTest;
 import org.apache.ignite.internal.processors.query.calcite.planner.SortAggregatePlannerTest;
 import org.apache.ignite.internal.processors.query.calcite.planner.SortedIndexSpoolPlannerTest;
 import org.apache.ignite.internal.processors.query.calcite.planner.TableSpoolPlannerTest;
+import org.apache.ignite.internal.processors.query.calcite.planner.UnionPlannerTest;
```
korlov42, line 31: worth to add this to suite too

korlov42 (commented):
please add unit tests for both types (single and two-phase) of execution nodes. See `org.apache.ignite.internal.processors.query.calcite.exec.rel` for examples

modules/calcite/src/test/java/org/apache/ignite/testsuites/PlannerTestSuite.java
alex-plekhanov: I was thinking about it and add the test to the suite, but found that there no actually any checks in this test, only plan output.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/MinusNode.java
```diff
@@ -0,0 +1,341 @@
...
+import org.apache.ignite.internal.processors.query.calcite.exec.RowHandler;
+import org.apache.ignite.internal.processors.query.calcite.exec.RowHandler.RowFactory;
+import org.apache.ignite.internal.processors.query.calcite.exec.exp.agg.AggregateType;
+import org.apache.ignite.internal.processors.query.calcite.exec.exp.agg.GroupKey;
+import org.apache.ignite.internal.util.typedef.F;
+
+/**
+ * Execution node for MINUS (EXCEPT) operator.
+ */
+public class MinusNode<Row> extends AbstractNode<Row> {
```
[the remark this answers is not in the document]
korlov42: I may not really understand how the EXCEPT is supposed to work... So let's find out it first to sure we are on the same page. My understanding is matches the [Postgres definition](https://www.postgresql.org/docs/7.4/sql-select.html) (sorry, but there is no anchor for exact place).

Current implementation of the execution node says that only leftmost query defines a top bound of row count, every next stage of EXCEPT could only shorten the resultset. Hence a conclusion: we could drop following stages as soon as current stage returns empty resultset. It's true that we have to complete gathering rows for current stage, but I see no reason to proceed the work a result of which we are no more interested in.


> Did you mean [all=true] (not DISTINCT)?

No, I meant exactly what I said. Frankly I have no clue how a query with EXCEPT ALL could be easily rewritten with ANTI JOIN since it requires a counting rows on both sides. On the other hand a query like `SELECT f1 FROM t1 EXCEPT SELECT f1 FROM t2` could be simply rewritten to `SELECT DISTINCT f1 FROM t1 ANTI JOIN t2 USING (f1)`.


> Calcite uses about the same approach

For some (unclear to me) reasons Calcite doesn't like SEMI and ANTI joins. And we already faced a [problem](https://issues.apache.org/jira/browse/IGNITE-14277) because of this

modules/calcite/src/test/java/org/apache/ignite/testsuites/PlannerTestSuite.java
korlov42: hw, looks like we need to revise these tests. Could you please file a ticket for this?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/set/IgniteMapMinus.java
```diff
@@ -0,0 +1,147 @@
...
+        RelTraitSet nodeTraits,
+        List<RelTraitSet> inputTraits
+    ) {
+        Set<IgniteDistribution> distributions = inputTraits.stream()
+            .map(TraitUtils::distribution)
+            .collect(Collectors.toSet());
+
+        ImmutableList.Builder<Pair<RelTraitSet, List<RelTraitSet>>> b = ImmutableList.builder();
+
+        for (IgniteDistribution distribution : distributions) {
```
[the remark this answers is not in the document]
alex-plekhanov: In this case broadcast input can be planned on the MAP phase to processed together with random input on each node, which produces wrong results. `CalciteQueryProcessorTest#testExceptMerge` reproduces this issue.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/RowHandler.java
```diff
@@ -40,6 +40,9 @@
     /** */
     int columnCount(Row row);
 
+    /** */
+    Object[] getColumns(Row row);
```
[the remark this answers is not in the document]
alex-plekhanov: It's not only a shortcut but there is also a lot of overhead when we copy each field of row one by one. And GroupKey class still needs to be changed (at least `fieldsCount` method required). I think the current change doesn't break any abstraction but produces less garbage and works faster, but I can rewrite it to use one-by-one fields copy methods if you are not agreed. WDYT?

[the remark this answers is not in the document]
korlov42: Please don't be fooled by the fact we have only one row representation -- array of object. You should treat a row as an abstract row. I don't like the idea to extend the interface that way because this change is too implementation specific (in terms of reducing garbage).

> And GroupKey class still needs to be changed

I'm OK with this. Moreover I would rather generify a GroupKey. It should be a wrapper for the Row, and there is nothing wrong to return underlying row. But perhaps it would be better to do under separate ticket.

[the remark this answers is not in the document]
alex-plekhanov: > Please don't be fooled by the fact we have only one row representation -- array of object.

There can be any row representation. As far as `RowHandler` can return `Object` by the index of column and count of columns it can return `Object[]` - the array of columns values as well. The `getColumns()` method of `RowHandler` doesn't mean "give me internals of row", it means "make an array from columns values", but in some cases (`ArrayRowHandler`) it can be optimized and return just internals.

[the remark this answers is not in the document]
korlov42: >  As far as RowHandler can return Object by the index of column and count of columns it can return Object[]

should we provided similar methods for `List<Object>` and other collections then? And what about `BinaryObject getColumns(Row row)`?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/MinusNode.java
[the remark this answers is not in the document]
alex-plekhanov: > My understanding is matches the Postgres definition

Yes, the current implementation matches the Postgres definition too. I got the idea with this optimization, I will implement it. But as I said before, is not always possible, for example, we still need to process all sources on the MAP phase.

> On the other hand a query like `SELECT f1 FROM t1 EXCEPT SELECT f1 FROM t2` could be simply rewritten to `SELECT DISTINCT f1 FROM t1 ANTI JOIN t2 USING (f1)`

I'm not sure it's more optimal than the current approach. As far as I understand there are some limitations where this approach is applicable: tables `t1` and `t2` should be colocated, there should be an index on `t2` for `f1` columns, `t1` should be sorted by `f1` or streaming mode will be not possible with DISTINCT, etc.
I propose to start with the current implementation (as working for all cases and optimal for most cases) and make optimization with anti-join later if we really need it.

modules/calcite/src/test/java/org/apache/ignite/testsuites/PlannerTestSuite.java
alex-plekhanov: https://issues.apache.org/jira/browse/IGNITE-14594

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/MinusNode.java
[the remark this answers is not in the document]
korlov42: > we still need to process all sources on the MAP phase

`select_1 EXCEPT select_2 EXCEPT select_3 EXCEPT select_n` should be evaluated as `((select_1 EXCEPT select_2) EXCEPT select_3) EXCEPT select_n`. Could you please explain why we need to process rows from `select_n` if results of `(...) EXCEPT select_3` is empty? It's unclear to me right now.


> I propose to start with the current implementation

works for me. It's better to benchmark both approaches and compare results to decide which one is better.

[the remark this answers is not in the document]
alex-plekhanov: > Could you please explain why we need to process rows from select_n on MAP phase if results of `(...) EXCEPT select_3` is empty?

For example, we have 'a' EXCEPT 'a' EXCEPT 'b' on node 1, and 'b' EXCEPT 'c' EXCEPT 'c' on node 2. In this case after the first step, we already have an empty set on node 1. If we stop processing other inputs on node 1, on reduce phase we will have the resulting set as 'b'. But this 'b' (from node 2) should be eliminated by the result from node 1.

alex-plekhanov:
@korlov42 I've fixed your comments, please have a look again.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/set/IgniteMapMinus.java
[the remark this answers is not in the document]
korlov42: Yes, you're right. The rule should be a little more complex:
```
        return ImmutableList.of(
            Pair.of(nodeTraits.replace(IgniteDistributions.random()), Commons.transform(inputTraits,
                t -> t.replace(t.getDistribution() == IgniteDistributions.single() ? IgniteDistributions.single() : IgniteDistributions.random())
            ))
        );
```
Perhaps, it worth to add additional planner's test to verifying a plan with bunch of table of all spectrum of distribution (affinity, single, broadcast)

The problem with current approach is that it will create unnecessary nodes. Assume we have N inputs with N different hash distributions. After deriving we will have N different variant for each distribution. This then creates N * (N - 1) converters (if possible) for every variant, whereas we could simple do all the job in single map-reduce phase.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/trait/DistributionTrait.java
```diff
@@ -142,7 +142,7 @@
                     && DistributionFunction.satisfy(function, other.function));
 
         if (other.getType() == RANDOM_DISTRIBUTED)
-            return getType() == HASH_DISTRIBUTED;
+            return getType() == HASH_DISTRIBUTED || getType() == SINGLETON;
```
korlov42, line 145: I don't sure this is a legit change. If SINGLETON will satisfy a RANDOM distribution, then there will be no additional exchange, which leads to singleton be a part of fragment with RANDOM distribution, but this should not be possible, since this fragment is allowed to be executed on several nodes, whereas SINGLETON is expected to be executed on only one

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/MinusNode.java
```diff
@@ -112,6 +112,14 @@ public void end(int idx) throws Exception {
 
         checkState();
 
+        if (type == AggregateType.SINGLE && grouping.isEmpty()) {
+            requested = 0;
+
+            downstream().end();
```
korlov42, line 118: Subsequent ending of downstream is unwanted because you have to be sure a caller side is waiting for next row. Currently it will be violated in case underlying source is empty, and caller invokes `request(0)`. This case is syntetic right now, yet still possible.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/trait/DistributionTrait.java
alex-plekhanov: This change was made in the middle of the development, looks like it's not required anymore. I will revert it.

alex-plekhanov: I do not quite understand. If `single` means that the plan node is allowed to execute only on one cluster node, then, for the MAP phase any input with `single` distribution should be prohibited. Since, if all inputs are single - `IgniteSingleMinus` node should be used, and if there is a mix of single and random plan nodes - random plan nodes should be converted to single with `IgniteExchange`. So, `IgniteMapMinus.deriveDistribution` should convert all inputs to `random`, or skip this plan if there is at least one `single` node. 

I was thinking previously that `single` trait means that data for this plan node is located only on one cluster node (nevertheless plan node can be executed on different cluster nodes, the same for `broadcast`, it shows only data distribution, but not force plan node to be executed on every cluster node). In this case change with `single().satisfies(random())` looks fine. And this change makes possible some plans like:
```
IgniteReduceMinus(all=[false])
  IgniteExchange(distribution=[single])
    IgniteMapMinus(all=[false])
      IgniteTableScan(table=[[PUBLIC, SINGLE_TBL1]])
      IgniteTableScan(table=[[PUBLIC, RANDOM_TBL1]])
```
Should we prohibit such plans?

korlov42: > single trait means that data for this plan node is located only on one cluster node 

This is correct. And the plan above is correct too. But the problem caused by limitation of current splitter implementation. When an exchange connects two fragments with SINGLE and RANDOM distribution, and the data flows from the first to the second, the IgniteReceiver nodes becomes a source of the data with a SINGLE distribution, but presented on every node of the fragment with RANDOM distribution, thereby creates N copies of the original data.
The same problem with BROADCAST to RANDOM transition.

korlov42: I have an ideas hot cope with this problem, but for now we should not allows such plans.

alex-plekhanov: Ok, reverted this change for a while.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/MinusNode.java
alex-plekhanov: Fixed.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/set/IgniteMapMinus.java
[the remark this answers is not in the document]
alex-plekhanov: Fixed.

korlov42 (approved):

alex-plekhanov:
Merged to sql-calcite branch

PR #9024, opened 2021-04-21
IGNITE-14607: Regex Based IP Filtering

atris:
This commit introduces the notion of regexes which can be registered and used as filters for incoming IPAddresses. A new API is introduced for the same and implemented in Azure Blob Storage IP Finder.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/TcpDiscoveryIpFinderAdapter.java
+++ b/modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/TcpDiscoveryIpFinderAdapter.java
@@ -18,7 +18,12 @@
 package org.apache.ignite.spi.discovery.tcp.ipfinder;
 
 import java.net.InetSocketAddress;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 import org.apache.ignite.Ignite;
 import org.apache.ignite.internal.util.tostring.GridToStringExclude;
 import org.apache.ignite.internal.util.typedef.internal.S;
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/TcpDiscoveryIpFinderAdapter.java
```diff
@@ -71,6 +79,24 @@
...
+        if (regexList == null) {
+            return true;
+        }
+
+        for (Pattern pattern : regexList) {
+            Matcher matcher = pattern.matcher(ipAddress);
+
+            if (matcher.matches()) {
+                return false;
+            }
```
AMashenkov, line 94:
```suggestion
            if (matcher.matches())
                return false;
```

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/TcpDiscoveryIpFinderAdapter.java
```diff
@@ -71,6 +79,24 @@
         return shared;
     }
 
+    /** {@inheritDoc} */
+    @Override public boolean isIpAddressAllowed(String ipAddress) {
+
+        if (regexList == null) {
+            return true;
+        }
```
AMashenkov, line 87:
```suggestion
        if (regexList == null)
            return true;
```

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/TcpDiscoveryIpFinderAdapter.java
```diff
@@ -85,6 +111,21 @@ public TcpDiscoveryIpFinderAdapter setShared(boolean shared) {
         return this;
     }
 
+    /**
+     * Adds a regex for filtering IPs
```
AMashenkov, line 115:
```suggestion
     * Adds a regex for filtering IPs.
     *
     * @param regex Regular expression.
     * @return {@code this} for chaining.
```

AMashenkov (changes requested):

atris:
Superseded by https://github.com/apache/ignite/pull/9048

PR #9050, opened 2021-04-27
IGNITE-14644 .NET: Log a warning about COMPlus_EnableAlternateStackCheck

ptupitsyn:
On Linux, Java overwrites SIGSEGV handler installed by .NET, so NullReferenceException causes a confusing "Stack smashing detected" error. This can be fixed by setting COMPlus_EnableAlternateStackCheck environment variable. Log a warning to raise user awareness.


isapego (approved):
Looks good to me.

PR #9058, opened 2021-04-28
IGNITE-14655 .NET: Improve DataStreamer API

ptupitsyn:
* Deprecate `AddData`, `RemoveData` methods: those methods return a `Task` that can't be awaited, which is confusing to the users. Replace with void `Add`, `Remove`.
* Deprecate `AutoFlushFrequency`, add `AutoFlushInterval` (uses `TimeSpan` instead of `long` - no confusion about time units)
* Add `FlushAsync`
* Add `GetCurrentBatchTask`
* Fix resource cleanup: release handles even if flush fails with an exception
* Fix async continuations: invoke on thread pool to avoid starving flusher thread
* Add benchmark - no difference with master


modules/platforms/dotnet/Apache.Ignite.Core/Datastream/IDataStreamer.cs
```diff
@@ -138,24 +134,44 @@ public interface IDataStreamer<TK, TV> : IDisposable
...
+        /// <li>Buffer is full.</li>
+        /// <li><see cref="Flush"/> or <see cref="TryFlush"/> is called.</li>
+        /// <li>Periodically when <see cref="AutoFlushInterval"/> is set.</li >
+        /// </ul>
         /// <para />
         /// If set to <c>0</c>, automatic flush is disabled.
         /// <para />
         /// Default is <c>0</c> (disabled).
         /// </summary>
+        [Obsolete("Use AutoFlushInterval.")]
```
isapego, line 157: Should not there be a justification for deprecation?

modules/platforms/dotnet/Apache.Ignite.Core/Datastream/IDataStreamer.cs
```diff
@@ -168,55 +184,90 @@ public interface IDataStreamer<TK, TV> : IDisposable
...
+        /// Loads any remaining buffered data, but doesn't close the streamer.
+        /// </summary>
+        Task FlushAsync();
+
         /// <summary>
         /// Closes this streamer optionally loading any remaining data.
         /// </summary>
         /// <param name="cancel">Whether to cancel ongoing loading operations. When set to <c>true</c>
-        /// there is not guarantees what data will be actually loaded to cache.</param>
+        /// there are no guarantees what data will be actually loaded to cache.</param>
```
isapego, line 270: Actually, "there is no guarantee" sounds more authentic to me.

isapego (approved):
Looks good overall. See my minor comments inline.

modules/platforms/dotnet/Apache.Ignite.Core/Datastream/IDataStreamer.cs
ptupitsyn: Fixed

modules/platforms/dotnet/Apache.Ignite.Core/Datastream/IDataStreamer.cs
ptupitsyn: The justification is that `TimeSpan` should be used for time interval purposes
* No ambiguity about units
* .NET-native type
* Consistent with other Ignite APIs and system APIs

However, I don't think we should provide those details in the deprecation message.

PR #9081, opened 2021-05-05
IGNITE-12291. Add MergeSort distributed cache query reducer.

timoninmaxim:
Thank you for submitting the pull request to the Apache Ignite.

In order to streamline the review of the contribution 
we ask you to ensure the following steps have been taken:

### The Contribution Checklist
- [ ] There is a single JIRA ticket related to the pull request. 
- [ ] The web-link to the pull request is attached to the JIRA ticket.
- [ ] The JIRA ticket has the _Patch Available_ state.
- [ ] The pull request body describes changes that have been made. 
The description explains _WHAT_ and _WHY_ was made instead of _HOW_.
- [ ] The pull request title is treated as the final commit message. 
The following pattern must be used: `IGNITE-XXXX Change summary` where `XXXX` - number of JIRA issue.
- [ ] A reviewer has been mentioned through the JIRA comments 
(see [the Maintainers list](https://cwiki.apache.org/confluence/display/IGNITE/How+to+Contribute#HowtoContribute-ReviewProcessandMaintainers)) 
- [ ] The pull request has been checked by the Teamcity Bot and 
the `green visa` attached to the JIRA ticket (see [TC.Bot: Check PR](https://mtcga.gridgain.com/prs.html))

### Notes
- [How to Contribute](https://cwiki.apache.org/confluence/display/IGNITE/How+to+Contribute)
- [Coding abbreviation rules](https://cwiki.apache.org/confluence/display/IGNITE/Abbreviation+Rules)
- [Coding Guidelines](https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines)
- [Apache Ignite Teamcity Bot](https://cwiki.apache.org/confluence/display/IGNITE/Apache+Ignite+Teamcity+Bot)

If you need any help, please email dev@ignite.apache.org or ask anу advice on http://asf.slack.com _#ignite_ channel.

PR #9086, opened 2021-05-06
IGNITE-12291 Refactoring of query limit to avoid dummy pages loading

timoninmaxim:
Refactoring limit handling for queries. Now it skips loading pages after limit exceeding, and cancel not-finished requests.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
@@ -28,7 +28,6 @@ import java.util.NoSuchElementException;
 import java.util.Queue;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import org.apache.ignite.IgniteCheckedException;
 import org.apache.ignite.IgniteLogger;
@@ -39,7 +38,6 @@ import org.apache.ignite.internal.processors.timeout.GridTimeoutObject;
 import org.apache.ignite.internal.util.future.GridFutureAdapter;
 import org.apache.ignite.internal.util.typedef.C1;
 import org.apache.ignite.internal.util.typedef.F;
-import org.apache.ignite.internal.util.typedef.X;
 import org.apache.ignite.internal.util.typedef.internal.CU;
 import org.apache.ignite.internal.util.typedef.internal.S;
 import org.apache.ignite.internal.util.typedef.internal.U;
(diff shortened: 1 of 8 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryManager.java
```diff
@@ -2858,7 +2859,7 @@ public int size() {
             null,
             false,
             keepBinary,
-            null).limit(limit);
+            null).limit(limit).pageSize(pageSize);
```
AMashenkov, line 2862:
```suggestion
            null)
            .limit(limit)
            .pageSize(pageSize);
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java
```diff
@@ -264,25 +266,7 @@ protected void onNodeLeft(UUID evtNodeId) {
...
-                cnt.addAndGet(col.size());
-            }
-            else if (capacity > 0) {
-                queue.add(new ArrayList<>((Collection<R>)col).subList(0, capacity));
-                capacity = 0;
-
-                cnt.addAndGet(capacity);
-            }
-        }
+        queue.add((Collection<R>)col);
```
AMashenkov, line 269: Why are you remove limit tracking from here? 
Seems, onDone() + cancel() methods could be called when the limit was reached.
onDone() will prevent requesting next pages (loadPage()), and cancel() will free resources on map nodes.

timoninmaxim: Hi! Sorry, could you please clarify which case do you mean? How does the cancel() invoke? 

If I understand you correctly, you told that empty collection (that will be inserted to the queue after limit exceed) will trigger cancel. But actually it's not true. In this case we just check that it's empty and ask for new page [1]. So nobody will send cancel().

You can run my test "GridCacheTextQueryPagesTest" from this PR on master code base - it will fail. It checks number of pages to be loaded during a query.

[1] https://github.com/apache/ignite/blob/7faa57c49118285b830fcf614a12abda65116f51/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryFutureAdapter.java#L213

timoninmaxim: Thanks for clarification, now understood your note. You're correct that it will work safe on current state of master. It's because we don't have any reduce of filter operations on received pages. But I did it by 2 reasons: 

1. In parallel activity I add MergeSort algorithm, it merging data from different pages. So there are cases when we get a data from single node, and it covers a specified limit. But actually we need another page from other nodes to compare. + Maybe there could be additional cases in future for other reduce or filtering operations. 

2. I don't like idea to make `enqueue()` responsible for finishing future, as there is already a working mechanism of `onDone()` and `cancel()` based on `hasNext()` result. Also now `enqueue()` creates a new collection from sublist in case of limit applying on input collection, there is not much overhead but it can be avoid by the same existing mechanism. I believe the less code the better. So I decide remove this duplicated mechanism.

timoninmaxim:
Close this PR. Limit handling moved to PR #9081 as it provides pluggable reducer functionality.

PR #9116, opened 2021-05-20
IGNITE-14753 Fix builds with maven 3.8.1+

solveme:
Update versions (migrate from http repositories):
  - apache parent
  - jotm (several tests broke after update, fixed)
  - spark-core_2.11

Fix broken configuration for jar plugin (due to versions upgrade in apache parent):
  - Remove 'useDefaultManifestFile' option from ignite parent pom (due to deprecation)
  - Add classifiers for artifacts in modules/extdata/uri

Remove duplicated dependencies definitions from several modules (maven complained to this during builds with warnings)

### DIFF
--- a/modules/extdata/uri/pom.xml
+++ b/modules/extdata/uri/pom.xml
@@ -64,7 +64,27 @@
         <module>modules/uri-dependency</module>
     </modules>
 
+    <properties>
+
+        <uri.fn>uri</uri.fn>
+        <uri.jar>${uri.fn}.jar</uri.jar>
+
+        <plain.fn>deployfile</plain.fn>
+        <plain.clr>plain</plain.clr>
+        <plain.jar>${plain.fn}-${plain.clr}.jar</plain.jar>
+
+        <well-signed.fn>deployfile</well-signed.fn>
+        <well-signed.clr>well-signed</well-signed.clr>
+        <well-signed.jar>${well-signed.fn}-${well-signed.clr}.jar</well-signed.jar>
+
+        <bad-signed.fn>deployfile</bad-signed.fn>
+        <bad-signed.clr>bad-signed</bad-signed.clr>
+        <bad-signed.jar>${bad-signed.fn}-${bad-signed.clr}.jar</bad-signed.jar>
+
+    </properties>
+
     <build>
+
         <resources>
             <resource>
                 <directory>src/main/java</directory>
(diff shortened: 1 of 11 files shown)


modules/extdata/uri/pom.xml
```diff
@@ -407,6 +430,7 @@
                     </execution>
                 </executions>
             </plugin>
+
```
vveider, line 433: This empty line is excess, closing tag should go one after another without empty lines. Only similar tags and sections are separated by new line in pom.xml

modules/extdata/uri/pom.xml
```diff
@@ -64,7 +64,27 @@
         <module>modules/uri-dependency</module>
     </modules>
 
+    <properties>
```
alamar, line 67: What is the reason for extdata-uri changes? Is it a simple refactoring?

modules/hibernate-4.2/src/test/java/org/apache/ignite/cache/hibernate/HibernateL2CacheTransactionalSelfTest.java
```diff
@@ -75,7 +76,7 @@
 
     /** {@inheritDoc} */
     @Override protected void beforeTestsStarted() throws Exception {
-        jotm = new Jotm(true, false);
+        jotm = new Jotm(true, false, new RmiLocalConfiguration());
```
alamar, line 79: What does it mean and why is it needed?

modules/urideploy/pom.xml
```diff
@@ -111,13 +111,6 @@
             <scope>test</scope>
         </dependency>
 
-        <dependency>
```
alamar, line 114: Removing ignite-tools may fail the suite since ignite-tools is needed to assist their execution. Why?

parent/pom.xml
```diff
@@ -293,9 +293,6 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-jar-plugin</artifactId>
-                    <configuration>
```
alamar, line 296: Why?

modules/hibernate-4.2/src/test/java/org/apache/ignite/cache/hibernate/HibernateL2CacheTransactionalSelfTest.java
solveme: 2-arg constructor in new Jotm version will call 3-arg constructor with null value for RMI configuration, but null value for RMI configuration will lead to NPE in other Jotm methods when you run tests.

modules/urideploy/pom.xml
solveme: I removed only duplicated dependency definition. I mean that there is a two line-to-line copies of 

        <dependency>
            <groupId>org.apache.ignite</groupId>
            <artifactId>ignite-tools</artifactId>
            <version>${project.version}</version>
            <scope>test</scope>
        </dependency>

one right after another. (Expand change view up for several lines)

parent/pom.xml
solveme: Jotm has third-party dependency that has another third-party dependency that is pulled from http repo. In 2.3.1-M1 version of Jotm this third party dependency was updated so build is not failing.

modules/extdata/uri/pom.xml
solveme: from http://apache-ignite-developers.2346864.n4.nabble.com/Building-with-maven-3-8-1-tp52689.html

Building ignite-exdata-uri with
latest jar plugin produces errors like:

     ignite-extdata-uri: You have to use a classifier to attach supplemental artifacts to the project instead of replacing them

Seems that jar plugin doesn't like when build produces multiple jars
even if they finalName's are different. Reworking build configuration
with classifiers fixed the problem.

So. Yep i reworked a little config since we will have to use classifiers

vveider (approved):

PR #9126, opened 2021-05-25
WIP -- IGNITE-12692

atris:


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteTableModify.java
```diff
@@ -96,4 +108,64 @@ public IgniteTableModify(RelInput input) {
...
+        return Pair.of(nodeTraits, ImmutableList.of(inTraits.get(0).replace(correlation)));
+    }
+
+    /**
+     *
+     */
+    @Override
+    public List<Pair<RelTraitSet, List<RelTraitSet>>> deriveCorrelation(RelTraitSet nodeTraits,
+                                                                        List<RelTraitSet> inTraits) {
+        Set<CorrelationId> corrIds = RexUtils.extractCorrelationIds(getSourceExpressionList());
```
korlov42, line 159: Currently IgniteTableModify doesn't introduce any correlations, so we should propagate its child's correlation only

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteTableModify.java
```diff
@@ -96,4 +108,64 @@ public IgniteTableModify(RelInput input) {
...
+                inTraits));
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<Pair<RelTraitSet, List<RelTraitSet>>> deriveDistribution(RelTraitSet nodeTraits,
+                                                                         List<RelTraitSet> inTraits) {
+        return ImmutableList.of(Pair.of(nodeTraits.replace(TraitUtils.distribution(inTraits.get(0))),
```
korlov42, line 128: we should propagate here only hash distribution of the table we going to modify

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/CalciteQueryProcessorTest.java
```diff
@@ -1110,7 +1110,8 @@ public void testInsertUpdateDeleteNonPrimitiveKey() throws Exception {
 
         assertEqualsCollections(F.asList(0, 0, "Igor", 1), row);
 
-        query = engine.query(null, "PUBLIC", "UPDATE DEVELOPER d SET name = 'Roman' WHERE id = ?", 0);
+        //query = engine.query(null, "PUBLIC", "UPDATE DEVELOPER d SET name = 'Roman' WHERE id = ?", 0);
```
zstan, line 1113: remove it.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteTableModify.java
```diff
@@ -96,4 +113,85 @@ public IgniteTableModify(RelInput input) {
...
+
+        IgniteDistribution distributionTrait = IgniteDistributions.affinity(0, "default",
+                getTraitSet().getTrait(DistributionTraitDef.INSTANCE));
+
+        if (distributionTrait.function().type() == RelDistribution.Type.HASH_DISTRIBUTED &&
+                distributionTrait.function().affinity()) {
+            networkCost = 0;
+        } else {
+            networkCost = totalBytes;
+        }
```
korlov42, line 193: this piece if code does't make any sense

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteTableModify.java
```diff
@@ -96,4 +113,85 @@ public IgniteTableModify(RelInput input) {
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override public List<Pair<RelTraitSet, List<RelTraitSet>>> deriveDistribution(RelTraitSet nodeTraits,
+                                                                         List<RelTraitSet> inTraits) {
+        IgniteDistribution distributionTrait = IgniteDistributions.affinity(0, "default",
+                nodeTraits.getTrait(DistributionTraitDef.INSTANCE));
+
+        IgniteDistribution targetDistributionTrait = IgniteDistributions.affinity(0, "default",
+                table.getDistribution());
```
korlov42, line 135: you don't need to create new distribution trait

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteTableModify.java
atris: The idea is to express that if the distribution is hash and the cache has affinity, then there is no network cost since there will be no need to move around data. Could you please let me know what needs to be corrected here?

PR #9154, opened 2021-06-04
IGNITE-14826 .NET: Support string and array keys in thin client with partition awareness

ptupitsyn:
* Add missing support for strings and arrays in `BinaryHashCodeUtils`
* Fix `byte` and `sbyte` hash codes: Java `byte` is signed, so the logic should be reversed


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Binary/BinaryHashCodeUtils.cs
```diff
@@ -45,11 +45,33 @@ public static unsafe int GetHashCode<T>(T val, Marshaller marsh, IDictionary<int
...
+                var val0 = TypeCaster<uint>.Cast(val);
+                return *(int*) &val0;
+            }
+
+            if (type == typeof(ulong))
+            {
+                var val0 = TypeCaster<ulong>.Cast(val);
+                return GetLongHashCode(*(long*) &val0);
+            }
+
```
isapego, line 65: What about all these types? They was not supported also?

isapego (approved):
Looks good overall

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Binary/BinaryHashCodeUtils.cs
ptupitsyn: They were supported - check the diff. I've reordered the checks, so that more likely use cases come early.

For example, I can imaging that `long`, `Guid` and `string` are among the most popular key type choices, while `double` would be very weird.

PR #9163, opened 2021-06-09
IGNITE-14856 Fix "Failed to resolve class name" on server when cache with indexed types is defined on client

ptupitsyn:
* Restore `registerMetadataForRegisteredCaches` call in `GridQueryProcessor#start` - this fixes Java-side behavior, but breaks registration for platform types, because `PlatformProcessor` is not available at this point
* Call `registerMetadataForRegisteredCaches` second time from `PlatformProcessorImpl.start` only for platform types

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
@@ -304,6 +304,8 @@ public class GridQueryProcessor extends GridProcessorAdapter {
         }, QRY_DETAIL_METRICS_EVICTION_FREQ, QRY_DETAIL_METRICS_EVICTION_FREQ);
 
         idxRebuildStateStorage.start();
+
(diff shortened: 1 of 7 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
```diff
@@ -1283,8 +1287,11 @@ private void registerTypeLocally(String clsName) throws BinaryObjectException {
 
             Class<?> cls = U.box(U.classForName(clsName, null, true));
 
-            if (cls != null)
-                binProc.binaryContext().registerClass(cls, true, false, true);
+            if (cls != null) {
+                if (!platformOnly) {
```
alamar, line 1291: For single-line if() statement, curly braces are not needed.

ptupitsyn: Fixed

isapego (approved):
Looks good to me.

PR #9195, opened 2021-06-24
IGNITE-14990 Incorrect values of cache, cache group, data region metrics after cluster re-activation

NSAmelchev:
Remove data region/cache/cache group metrics on deactivation without removing metrics configuration.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataRegionMetricsImpl.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataRegionMetricsImpl.java
@@ -726,6 +726,11 @@ public class DataRegionMetricsImpl implements DataRegionMetrics {
             metrics.reset();
     }
 
+    /** Removes all metric for data region. */
+    public void remove() {
+        kernalCtx.metric().remove(metricRegistry().name(), false);
+    }
+
     /** @param time Time to add to {@code totalThrottlingTime} metric in milliseconds. */
     public void addThrottlingTime(long time) {
         if (kernalCtx.performanceStatistics().enabled())
--- a/modules/core/src/test/java/org/apache/ignite/internal/metric/CacheMetricsAddRemoveTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/metric/CacheMetricsAddRemoveTest.java
@@ -114,7 +114,7 @@ public class CacheMetricsAddRemoveTest extends GridCommonAbstractTest {
 
         checkMetricsEmpty(cachePrefix);
 
-        createCache("persisted", "other-cache");
+        createCache("persisted", "other-cache", null);
 
         grid("client").cache("other-cache").put(1L, 1L);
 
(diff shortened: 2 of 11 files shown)


modules/core/src/test/java/org/apache/ignite/internal/metric/CacheMetricsAddRemoveTest.java
```diff
@@ -204,12 +253,14 @@ private void checkMetricsEmpty(String cachePrefix) {
```
ololo3000, line 253: `MetricRegistry mreg = mmgr.registry(cachePrefix);`

Can we just iterate over mmgr and check that there is no registry with specified name? Mentioned above call creates registry if it does not exist. I see no need for this.

modules/core/src/test/java/org/apache/ignite/internal/metric/CacheMetricsAddRemoveTest.java
```diff
@@ -204,12 +253,14 @@ private void checkMetricsEmpty(String cachePrefix) {
 
             assertNull(mreg.findMetric(metricName(cachePrefix, CACHE_GETS)));
             assertNull(mreg.findMetric(metricName(cachePrefix, CACHE_PUTS)));
+            assertNull(mreg.findMetric(metricName(cachePrefix, GET_TIME)));
```
ololo3000, line 256: Here and above, we must use mreg.findMetric(<metric_name>) without registry prefix. Otherwise, this check makes no sense and always true.

modules/core/src/test/java/org/apache/ignite/internal/metric/CacheMetricsAddRemoveTest.java
```diff
@@ -136,6 +136,44 @@ public void testCacheMetricsConfigurationNotRemovedOnStop() throws Exception {
         checkMetricsEmpty(cachePrefix);
     }
 
+    /** */
+    @Test
+    public void testCacheMetricsConfigurationNotRemovedOnStopTwoCachesPerGroup() throws Exception {
+        String cachePrefix = cacheMetricsRegistryName("other-cache", false);
```
ololo3000, line 142: Should we use nearEnabled flag instead of `false` here?

modules/core/src/test/java/org/apache/ignite/internal/metric/CacheMetricsAddRemoveTest.java
```diff
@@ -161,6 +200,9 @@ private void createCache(@Nullable String dataRegionName, @Nullable String cache
         if (cacheName != null)
             ccfg.setName(cacheName);
 
+        if (groupName != null)
+            ccfg.setGroupName(groupName);
+
```
ololo3000, line 205: I propose to set cache mode as well. It changes in parameters but does not actually apply to the cache.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataRegionMetricsImpl.java
```diff
@@ -726,6 +726,11 @@ public void clear() {
             metrics.reset();
     }
 
+    /** Removes all metric for data region. */
+    public void remove() {
+        kernalCtx.metric().remove(metricRegistry().name(), false);
```
ololo3000, line 731: Can we create dedicated variable for the registry name and calculate it once?

modules/core/src/test/java/org/apache/ignite/internal/metric/CacheMetricsAddRemoveTest.java
```diff
@@ -152,12 +200,25 @@ private void createCache(@Nullable String dataRegionName, @Nullable String cache
...
         if (nearEnabled)
             ccfg.setNearConfiguration(new NearCacheConfiguration());
 
         grid("client").createCache(ccfg);
 
         awaitPartitionMapExchange();
+
```
ololo3000, line 212: Can we just pass nearEnabled into cacheMetricsRegistryName instead of if statement?

modules/core/src/test/java/org/apache/ignite/internal/metric/CacheMetricsAddRemoveTest.java
NSAmelchev: I think this class should be refactored. This is not related to the issue and should be done in a separate issue.

modules/core/src/test/java/org/apache/ignite/internal/metric/CacheMetricsAddRemoveTest.java
NSAmelchev: Then we should fix `checkMetrics*` methods and other tests to provide the correct cache prefix. I think this is not related to the issue and should be done in a separate issue.

modules/core/src/test/java/org/apache/ignite/internal/metric/CacheMetricsAddRemoveTest.java
NSAmelchev: I think this is not related to the issue as well as other refactoring.

modules/core/src/test/java/org/apache/ignite/internal/metric/CacheMetricsAddRemoveTest.java
NSAmelchev: Thanks, fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataRegionMetricsImpl.java
NSAmelchev: I think code should not be refactored in this issue.

modules/core/src/test/java/org/apache/ignite/internal/metric/CacheMetricsAddRemoveTest.java
NSAmelchev: No, we can't. We should configure cache and near cache metrics.

nizhikov (approved):

PR #9230, opened 2021-07-06
Ignite-14999 : Support snapshot encryption with same master-key.

Vladsz83:
This PR suggests support of encrypted snapshots with same master key. 

Changes: 

1. Added encrypted file IO to the snapshot deltas to get ability to assemble encrypted snapshot data. 
2. Added general-purpose records to snapshot meta to transfer encryption keys (encrypted by master key). 
3. To keep consistency of the keys and encrypted data between snapshot and the persistent store, blocked re-encryption and master key change during snapshot operations and opposite.
4. Snapshot tests now have ‘encrypted’ option.


Vladsz83:
V2: https://github.com/apache/ignite/pull/9260

PR #9260, opened 2021-07-15
IGNITE-14999 : Support snapshot encryption with same master-key v2.

Vladsz83:
Second version of PR #9230 . Changed: another variant of absent group key processing.


Vladsz83:
Split into several tasks. First of them: #9269

PR #9271, opened 2021-07-22
IGNITE-15045 .NET: Fix NullPointerException in ContinuousQuery with security enabled

xtern:
Add null check to `SecurityAwareFilter`, because `PlatformContinuousQueryImpl` always sets itself as a remote filter, but serializes as null when the actual filter is not present.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/platform/cache/query/PlatformContinuousQueryImpl.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/platform/cache/query/PlatformContinuousQueryImpl.java
@@ -142,10 +142,8 @@ public class PlatformContinuousQueryImpl implements PlatformContinuousQuery {
 
                 qry.setLocalListener(this);
 
-                if (hasFilter || javaFilter != null) {
-                    //noinspection deprecation
-                    qry.setRemoteFilter(this); // Filter must be set always for correct resource release.
-                }
+                //noinspection deprecation
+                qry.setRemoteFilter(this); // Filter must be set always for correct resource release.
 
                 qry.setPageSize(bufSize);
                 qry.setTimeInterval(timeInterval);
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/platform/cache/query/PlatformContinuousQueryImpl.java
```diff
@@ -142,8 +142,10 @@ private static CacheEntryEventFilter getJavaFilter(Object filter, GridKernalCont
 
                 qry.setLocalListener(this);
 
-                //noinspection deprecation
-                qry.setRemoteFilter(this); // Filter must be set always for correct resource release.
+                if (hasFilter || javaFilter != null) {
+                    //noinspection deprecation
+                    qry.setRemoteFilter(this); // Filter must be set always for correct resource release.
```
ptupitsyn, line 147: Please note the comment: we can't make this call conditional. If we don't call `setRemoteFilter`, then `onQueryUnregister()` won't be called and platform resources won't be released.

I think we should fix `SecurityAwareFilter` instead to handle nulls correctly.

Alternatively, fix `writeReplace` to write a dummy filter instead of null, but this will affect performance.

ptupitsyn (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/cache/query/PlatformContinuousQueryImpl.java
xtern: Thanks for the clarification, I tried to prevent unnecessary serialization of the "security filter", but since the platform filter must always be set, it seems to me that the best solution would be to simply check for null in the "security filter".

ptupitsyn (approved):

PR #9315, opened 2021-08-10
IGNITE-15302 Add support of precision parameter for varbinary type

timoninmaxim:
1. Document QuerySqlField.precision for String (varchar).
2. Enable precision for byte[] (binary, varbinary).

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryTypeDescriptorImpl.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryTypeDescriptorImpl.java
@@ -647,8 +647,8 @@ public class QueryTypeDescriptorImpl implements GridQueryTypeDescriptor {
                 continue;
 
             if (String.class == propVal.getClass() || byte[].class == propVal.getClass()) {
-                int propValLen = propVal.getClass() == String.class ? ((String) propVal).length()
-                    : ((byte[]) propVal).length;
+                int propValLen = String.class == propVal.getClass() ? ((String)propVal).length()
+                    : ((byte[])propVal).length;
 
                 if (propValLen > prop.precision()) {
                     throw new IgniteSQLException("Value for a column '" + prop.name() + "' is too long. " +
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryTypeDescriptorImpl.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryTypeDescriptorImpl.java
@@ -646,11 +646,15 @@ public class QueryTypeDescriptorImpl implements GridQueryTypeDescriptor {
             if (propVal == null || prop.precision() == -1)
                 continue;
 
-            if (String.class == propVal.getClass() &&
-                ((String)propVal).length() > prop.precision()) {
-                throw new IgniteSQLException("Value for a column '" + prop.name() + "' is too long. " + 
-                    "Maximum length: " + prop.precision() + ", actual length: " + ((CharSequence)propVal).length(),
-                    isKey ? TOO_LONG_KEY : TOO_LONG_VALUE);
+            if (String.class == propVal.getClass() || byte[].class == propVal.getClass()) {
+                int propValLen = propVal.getClass() == String.class ? ((String) propVal).length()
+                    : ((byte[]) propVal).length;
+
+                if (propValLen > prop.precision()) {
+                    throw new IgniteSQLException("Value for a column '" + prop.name() + "' is too long. " +
+                        "Maximum length: " + prop.precision() + ", actual length: " + propValLen,
+                        isKey ? TOO_LONG_KEY : TOO_LONG_VALUE);
+                }
             }
             else if (BigDecimal.class == propVal.getClass()) {
                 BigDecimal dec = (BigDecimal)propVal;
--- a/modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/PrecisionTest.java
+++ b/modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/PrecisionTest.java
@@ -22,17 +22,10 @@ import java.util.Collections;
 import java.util.concurrent.Callable;
 import javax.cache.CacheException;
 import org.apache.ignite.IgniteCache;
-import org.apache.ignite.Ignition;
-import org.apache.ignite.binary.BinaryObject;
 import org.apache.ignite.cache.QueryEntity;
 import org.apache.ignite.cache.query.SqlFieldsQuery;
 import org.apache.ignite.cache.query.annotations.QuerySqlField;
-import org.apache.ignite.client.ClientCache;
-import org.apache.ignite.client.ClientCacheConfiguration;
-import org.apache.ignite.client.ClientException;
-import org.apache.ignite.client.IgniteClient;
 import org.apache.ignite.configuration.CacheConfiguration;
-import org.apache.ignite.configuration.ClientConfiguration;
 import org.apache.ignite.internal.IgniteEx;
 import org.apache.ignite.testframework.GridTestUtils;
 import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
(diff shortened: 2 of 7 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryTypeDescriptorImpl.java
```diff
@@ -646,11 +646,15 @@ else if (coCtx.kernalContext().cacheObjects().typeId(prop.type().getName()) !=
             if (propVal == null || prop.precision() == -1)
                 continue;
 
-            if (String.class == propVal.getClass() &&
-                ((String)propVal).length() > prop.precision()) {
-                throw new IgniteSQLException("Value for a column '" + prop.name() + "' is too long. " + 
-                    "Maximum length: " + prop.precision() + ", actual length: " + ((CharSequence)propVal).length(),
-                    isKey ? TOO_LONG_KEY : TOO_LONG_VALUE);
+            if (String.class == propVal.getClass() || byte[].class == propVal.getClass()) {
+                int propValLen = propVal.getClass() == String.class ? ((String) propVal).length()
```
ivandasch, line 650: According to codestyle it should be
`((String)propVal).length())`
The same is for casting to `byte[]`

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/PrecisionTest.java
```diff
@@ -0,0 +1,344 @@
...
+        assertNull(cache.get(0));
+    }
+
+    /** */
+    static class Person {
+        /** */
+        @QuerySqlField(precision = 5)
+        private final String name;
+
+        @QuerySqlField(precision = 5)
```
ivandasch, line 308: Missing empty javadoc

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/PrecisionTest.java
```diff
@@ -0,0 +1,344 @@
...
+            StringBuilder validStr = new StringBuilder("12345");
+
+            SqlFieldsQuery strQry = new SqlFieldsQuery("insert into PERSON(_KEY, name) values (?, ?)");
+            SqlFieldsQuery binQry = new SqlFieldsQuery("insert into PERSON(_KEY, arr) values (?, ?)");
+
+            for (int i = 1; i < CNT; i++) {
+                String v = validStr.append("0").toString();
+                int len = 5 + i;
+
+                assertClientPrecision(new Callable<Object>() {
```
ivandasch, line 255: I am not insisting, but may be use here just lambda?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/PrecisionTest.java
```diff
@@ -0,0 +1,344 @@
...
+            ") with \"CACHE_NAME=SQL,VALUE_TYPE=SQL_TYPE\""));
+
+        StringBuilder validStr = new StringBuilder("12345");
+
+        SqlFieldsQuery strQry = new SqlFieldsQuery("insert into TABLE(id, str) values (?, ?)");
+        SqlFieldsQuery binQry = new SqlFieldsQuery("insert into TABLE(id, bin) values (?, ?)");
+
+        IgniteCache tblCache = ignite(0).cache("SQL");
+
+        for (int i = 1; i < CNT; i++) {
```
ivandasch, line 93: I suppose that it is enough to test only for `CNT = 6`

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/PrecisionTest.java
timoninmaxim: Usage of pure lambda is broken for assertThrows. It converts lambda to RunnableX that rethrow any exception with IgniteException wrapper. So it's impossible to check exception class then.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/PrecisionTest.java
```diff
@@ -0,0 +1,344 @@
...
+            .setIndexedTypes(Integer.class, Person.class);
+
+        cfg.setCacheConfiguration(ccfg);
+
+        return cfg;
+    }
+
+    /** {@inheritDoc} */
+    @Override protected void beforeTest() throws Exception {
+        Ignite ignite = startGrids(2);
```
ivandasch, line 66: I'd suggests to start grid in one place --`beforeTestsStarted` , grid will be automatically stopped after tests finish.
You can just clear caches in `beforeTest` and `afterTest`. I'd suggest to create both caches in separate test methods.
Right now you have mixed approach and this is not consistent

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/PrecisionTest.java
```diff
@@ -0,0 +1,344 @@
...
+
+    /** */
+    @Test
+    public void testInsertTableVarColumns() {
+        cache.query(new SqlFieldsQuery("" +
+            "create table TABLE(" +
+            "   id int PRIMARY KEY," +
+            "   str varchar(5)," +
+            "   bin binary(5)" +
+            ") with \"CACHE_NAME=SQL,VALUE_TYPE=SQL_TYPE\""));
```
ivandasch, line 84: You can set VALUE_TYPE to `Person.class.getName()` and use the same assertions for both cases and eliminate copy-paste

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/PrecisionTest.java
timoninmaxim: actually for `CNT = 1`, as CNT shows number of additional symbols.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/PrecisionTest.java
```diff
@@ -0,0 +1,344 @@
...
+
+        StringBuilder validStr = new StringBuilder("12345");
+
+        SqlFieldsQuery strQry = new SqlFieldsQuery("insert into TABLE(id, str) values (?, ?)");
+        SqlFieldsQuery binQry = new SqlFieldsQuery("insert into TABLE(id, bin) values (?, ?)");
+
+        try (IgniteClient client = Ignition.startClient(new ClientConfiguration().setAddresses("127.0.0.1"))) {
```
ivandasch, line 193: I don't understand why you have decided to write tests for thin client. 
I can understand if you'd  create cache from IgniteClient using ClientCacheConfiguration, it's ok. 
But not in this case.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/PrecisionTest.java
```diff
@@ -0,0 +1,344 @@
...
+
+        /** */
+        Person(byte[] arr) {
+            this.name = null;
+            this.arr = arr;
+        }
+
+        /** {@inheritDoc} */
+        @Override public int hashCode() {
+            if (name != null)
```
ivandasch, line 325: Where do you use `equals` and `hashCode`?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/PrecisionTest.java
timoninmaxim: Works with lambda with return value. Thanks.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/PrecisionTest.java
timoninmaxim: Because it is a separate user interface for data. For example DataStreamer is broken, so I've decided to test ThinClient too. But also I agree that forgot to test it with cache created with thin client. Will add test for that.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/PrecisionTest.java
timoninmaxim: Will remove it.

ivandasch (approved):

PR #9336, opened 2021-08-17
removing MULTI_THREADED from H2 options

lsliwko:
The support for MULTI_THREADED has been removed in H2 v1.4.200:
https://github.com/h2database/h2database/releases/tag/version-1.4.200

If a H2 v1.4.200 is on classpath, it will crash Ignite:
Caused by: org.h2.jdbc.JdbcSQLNonTransientConnectionException: Unsupported connection setting "MULTI_THREADED" [90113-200]

It's due to hardcoded DB_OPTIONS:
https://github.com/andyglick/apache-ignite/blob/master/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java#L202

The fix is to either remove MULTI_THREADED argument or to add IGNORE_UNKNOWN_SETTINGS=TRUE


dkhjmu:
I have same problem... Thank you for sharing!

PR #9338, opened 2021-08-17
IGNITE-15263 Fix QuerySqlField.alias for affinity field

timoninmaxim:
Now affinity key correctly handles alias fields.


modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
```diff
@@ -557,6 +557,9 @@ public static QueryTypeCandidate typeForQueryEntity(GridKernalContext ctx, Strin
                         if (!F.isEmpty(qryEntity.getKeyFields()) && qryEntity.getKeyFields().contains(affField0)) {
                             affField = affField0;
 
+                            if (desc.aliases().containsKey(affField))
+                                affField = desc.aliases().get(affField);
```
AMashenkov, line 561:
```suggestion
                            affField = desc.aliases().getOrDefault(affField, affField);
```

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
```diff
@@ -578,6 +581,9 @@ public static QueryTypeCandidate typeForQueryEntity(GridKernalContext ctx, Strin
                     ((GridCacheDefaultAffinityKeyMapper)keyMapper).affinityKeyPropertyName(desc.keyClass());
 
                 if (affField != null) {
+                    if (desc.aliases().containsKey(affField))
+                        affField = desc.aliases().get(affField);
```
AMashenkov, line 585:
```suggestion
                    affField = desc.aliases().getOrDefault(affField, affField);
```

timoninmaxim: @AMashenkov , hi! Thanks for review. Applied both suggestions with an even additional improvement.

ivandasch (approved):
Looks good to me

PR #9366, opened 2021-09-01
IGNITE-15407 Fix PK flag ordering in SQL AST traverse

timoninmaxim:
This patch fixes a bug in the `SqlAstTraverser`. It incorrectly handled info about PK in cases of changing order of tables.

### DIFF
--- a/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/SqlAstTraverser.java
+++ b/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/SqlAstTraverser.java
@@ -198,7 +198,7 @@ class SqlAstTraverser {
         if (!joinIsValid) {
             log.warning(
                 String.format(
-                    "For join two partitioned tables join condition should be the equality operation of affinity keys." +
+                    "For join two partitioned tables join condition should contain the equality operation of affinity keys." +
                         " Left side: %s; right side: %s", left.getName(), right.getName())
             );
         }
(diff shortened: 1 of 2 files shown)


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/SqlAstTraverser.java
```diff
@@ -306,25 +306,32 @@ private void checkEqualityOperation(GridSqlOperation equalOp,
...
+        boolean actPkLeft = pkLeft;
+        boolean actPkRight = pkRight;
+
         if (leftTbl.equals(leftTblAls))
             actLeftCols = leftCols;
-        else if (leftTbl.equals(rightTblAls))
+        else if (leftTbl.equals(rightTblAls)) {
             actLeftCols = rightCols;
+            actPkLeft = pkRight;
+        }
```
AMashenkov, line 317: What's about leftTbl and rightTbl names? 
You've swap table columns, but not the names.

Seems, you can just swap column names instead.

modules/indexing/src/test/java/org/apache/ignite/sqltests/CheckWarnJoinPartitionedTables.java
```diff
@@ -46,41 +59,32 @@
     }
 
     /** {@inheritDoc} */
-    @Override protected void afterTest() throws Exception {
+    @Override protected void beforeTest() throws Exception {
```
AMashenkov, line 62: It is like a good practive to stop the grid after test.
Please, revert this or stop grid after all tests.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/SqlAstTraverser.java
timoninmaxim: @AMashenkov Thanks, nice catch. I've simplified this code even better after swap column names only.

AMashenkov (approved):

PR #9397, opened 2021-09-09
IGNITE-15484 Forbid duplicated field in CREATE INDEX clause

timoninmaxim:
The SQL clause "CREATE INDEX" allows to set duplicated field name. The last appearance will be applied silently. Now it fails while parse the query.


ivandasch (approved):
Looks good to me

PR #9419, opened 2021-09-17
IGNITE-15522 Add all() IndexQuery criterion.

timoninmaxim:
Introduces the `all()` IndexQuery criterion:
1. Allows to return full index range;
2. Useful for cases with multi-fielded index (A, B, C) with criteria for fields (A, C) but no for the the middle field B.


modules/core/src/main/java/org/apache/ignite/cache/query/IndexQueryCriteriaBuilder.java
```diff
@@ -24,6 +24,18 @@
  * Factory of {@link IndexQueryCriterion} for {@link IndexQuery}.
  */
 public class IndexQueryCriteriaBuilder {
+    /**
+     * All.
```
alex-plekhanov, line 28: Let's extend javadoc (describe why and when "all" is required)

PR #9424, opened 2021-09-20
IGNITE-15535 Add inclusive flags for the between() criterion

timoninmaxim:
between() supports exclusion of bounds to allow queries like (a > 10 && a <= 100).


timoninmaxim:
Choose different approach. See #9429

PR #9432, opened 2021-09-21
IGNITE-13631 Improve data storage metrics descriptions

dmekhanikov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataStorageMetricsImpl.java
```diff
@@ -187,25 +187,28 @@ public DataStorageMetricsImpl(
...
+            "WalFsyncDuration",
+            "Total time operations spent in the fsync phase writing WAL records to the disk " +
+                "since the start of the node in milliseconds.",
             rateTimeInterval,
             subInts);
 
         walFsyncTimeNum = mreg.hitRateMetric(
-            "WalFsyncTimeNum",
-            "Total count of fsync",
+            "WalFsyncNum",
```
nizhikov, line 210: Do we really need to change metric name?
Actually metric name is kind of public API and we don't want to change it without strong reason.

dmekhanikov: Well, the old name doesn't make much sense, so I'd rather rename it.
Personally I would rename the metric and specify the name of the new one in the release notes. WDYT?

nizhikov:
@dmekhanikov Overall looks good. Can you, also, change corresponding metric description in [new-metric.adoc](https://github.com/apache/ignite/tree/master/docs/_docs/monitoring-metrics/new-metrics.adoc)

PR #9436, opened 2021-09-23
IGNITE-14809 Add IndexQuery remote filter

timoninmaxim:
Add to IndexQuery functionality of remote filter:
- Consumes (K, V) as params
- Applies remotely on every node on query cursor.


ivandasch (approved):
Looks good to me!

PR #9462, opened 2021-10-01
IGNITE-12464 : Metrics for the Java Services V4

Vladsz83:
Implementation with the service context.


Vladsz83:
The final version is #9457.

PR #9518, opened 2021-10-22
IGNITE-15781 CacheGroupMetricsImpl#numberOfPartitionCopies uses existing set of partitions keys

timoninmaxim:
CacheGroupMetricsImpl#numberOfPartitionCopies created a lot of objects of `java.lang.Integer` for every call. Amount of objects linearly depends on cache partitions count. Replaced it with existing set objects.


nizhikov:
Hello. Can we modify the patch to use `GridPartitionStateMap#get(int)` and avoid any usage of Integer object here?

nizhikov:
It seems to me that almost all of the current usages of `GridDhtPartitionMap#get(Integer)` use int as a parameter and two stack below we unbox Integer to int anyway.

So we can simply refactor get to `GridDhtPartitionState#get` and made state method public.
```
    public GridDhtPartitionState get(int part) {
        return map.state(part);
    }
```

This will completely eliminate boxing -> unboxing during this piece of code.

nizhikov (approved):

PR #9526, opened 2021-10-25
IGNITE-15795 .NET: Add ServiceCallContext

xtern:
`ServiceCallContext` provides a way to pass additional data implicitly with all service method calls issued by the current service proxy.


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Services/ServiceContext.cs
```diff
@@ -18,7 +18,10 @@
 namespace Apache.Ignite.Core.Impl.Services
 {
     using System;
+    using System.Collections.Generic;
+    using System.Collections.ObjectModel;
```
ptupitsyn, line 22: Unused imports, please remove.

modules/platforms/dotnet/Apache.Ignite.Core/Services/IServiceCallContext.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// </code>
+    /// </summary>
+    public interface IServiceCallContext
+    {
+        /// <summary>
+        /// Gets the string attribute.
+        /// </summary>
+        /// <param name="name">Attribute name.</param>
+        /// <returns>String attribute value.</returns>
+        string Attribute(string name);
```
ptupitsyn, line 66:
```suggestion
        string GetAttribute(string name);
```

modules/platforms/dotnet/Apache.Ignite.Core/Services/IServiceCallContext.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// Gets the binary attribute.
+        /// </summary>
+        /// <param name="name">Attribute name.</param>
+        /// <returns>Binary attribute value.</returns>
+        byte[] BinaryAttribute(string name);
```
ptupitsyn, line 73:
```suggestion
        byte[] GetBinaryAttribute(string name);
```

modules/platforms/dotnet/Apache.Ignite.Core/Services/IServiceContext.cs
```diff
@@ -65,5 +66,13 @@ public interface IServiceContext
...
         /// </value>
         object AffinityKey { get; }
+
+        /// <summary>
+        /// Gets context of the current service call. 
+        /// </summary>
+        /// <returns>Context of the current service call.</returns>
+        /// <seealso cref="IServiceCallContext"/>
+        [IgniteExperimental]
+        IServiceCallContext CurrentCallContext();
```
ptupitsyn, line 76: Let's make it a property, more idiomatic:

```suggestion
        IServiceCallContext CurrentCallContext { get; }
```

modules/platforms/dotnet/Apache.Ignite.Core/Services/ServiceCallContextBuilder.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// </summary>
+        /// <param name="name">Attribute name.</param>
+        /// <param name="value">Attribute value.</param>
+        /// <returns>This for chaining.</returns>
+        public ServiceCallContextBuilder Put(string name, string value)
+        {
+            IgniteArgumentCheck.NotNullOrEmpty(name, "name");
+            IgniteArgumentCheck.NotNull(value, "value");
+
+            _attrs[name] = Encoding.UTF8.GetBytes(value);
```
ptupitsyn, line 46: String to byte conversion should go through `BinaryUtils.StringToUtf8Bytes` to be interoperable with Java because we have two different modes for that.

ptupitsyn: I think it'll be simpler to let the serializer deal with this, and use a `Dictionary<string, object>`.

modules/platforms/dotnet/Apache.Ignite.Core/Services/ServiceCallContextBuilder.cs
```diff
@@ -0,0 +1,79 @@
...
+        /** Context attributes. */
+        private readonly Dictionary<string, byte[]> _attrs = new Dictionary<string, byte[]>();
+        
+        /// <summary>
+        /// Put string attribute.
+        /// </summary>
+        /// <param name="name">Attribute name.</param>
+        /// <param name="value">Attribute value.</param>
+        /// <returns>This for chaining.</returns>
+        public ServiceCallContextBuilder Put(string name, string value)
```
ptupitsyn, line 41: To be consistent with `BinaryObjectBuilder`:

```suggestion
        public ServiceCallContextBuilder Set(string name, string value)
```

modules/platforms/dotnet/Apache.Ignite.Core/Services/ServiceCallContextBuilder.cs
```diff
@@ -0,0 +1,79 @@
...
+            return this;
+        }
+
+        /// <summary>
+        /// Put binary attribute.
+        /// </summary>
+        /// <param name="name">Attribute name.</param>
+        /// <param name="value">Attribute value.</param>
+        /// <returns>This for chaining.</returns>
+        public ServiceCallContextBuilder Put(string name, byte[] value)
```
ptupitsyn, line 57: Same as above:

```suggestion
        public ServiceCallContextBuilder Set(string name, byte[] value)
```

ptupitsyn (changes requested):

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Services/ServiceContext.cs
xtern: Done, thanks.

modules/platforms/dotnet/Apache.Ignite.Core/Services/ServiceCallContextBuilder.cs
xtern: Done.

modules/platforms/dotnet/Apache.Ignite.Core/Services/ServiceCallContextBuilder.cs
xtern: Done, the same change has been made to java.

modules/platforms/dotnet/Apache.Ignite.Core/Services/ServiceCallContextBuilder.cs
xtern: Done, the same change has been made to java.

modules/platforms/dotnet/Apache.Ignite.Core/Services/IServiceContext.cs
xtern: Thanks, the "getter" method has been changed.

But for the "setter" method I see 3 options:
1. Leave the static method SetCurrentCallContext(..) in the ServiceContext (current solution). 
2. Make it instance method (normal "setter").
3. Extract the threadlocal to another container-class with static methods to reading/storing the context (similar to current java implementation).

Personally, I prefer the second one, but accessing the ServiceContext instance while executing a service method requires a bit of reworking of the current service logic. 

Personally, I prefer the second one, but it requires a little rework of current service logic to access ServiceContext instance when executing the service method:
a. Store ServiceContext instead of a service in handleregistry.
b. Store reference to the Service in ServiceContext implementation.

ptupitsyn: Setter is internal, I think it is fine to keep it as is. No need to complicate.

modules/platforms/dotnet/Apache.Ignite.Core/Services/ServiceCallContextBuilder.cs
ptupitsyn: Not sure about Java, where `Put` is idiomatic for Maps.
There is no problem if some APIs are named differently in Java and C#, I think it is more important to use language-specific idioms so that the users are more comfortable.

modules/platforms/dotnet/Apache.Ignite.Core/Services/IServiceContext.cs
xtern: ok, thanks for the reply.

modules/platforms/dotnet/Apache.Ignite.Core/Services/ServiceCallContextBuilder.cs
xtern: ok, kept for java "put", thanks.

ptupitsyn (changes requested):

modules/platforms/dotnet/Apache.Ignite.Core/Services/ServiceCallContextBuilder.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// </summary>
+        /// <param name="name">Attribute name.</param>
+        /// <param name="value">Attribute value.</param>
+        /// <returns>This for chaining.</returns>
+        public ServiceCallContextBuilder Set(string name, byte[] value)
+        {
+            IgniteArgumentCheck.NotNullOrEmpty(name, "name");
+            IgniteArgumentCheck.NotNull(value, "value");
+
+            _attrs[name] = value.ToArray();
```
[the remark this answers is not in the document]
xtern: ok, added "Note: it is recommended to pass a copy of the array if the original can be changed later.". The same applies to Java code.

modules/platforms/dotnet/Apache.Ignite.Core/Services/IServiceCallContext.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// ...
+    ///
+    /// // Call this service with context.
+    /// IServiceCallContext callCtx = new ServiceCallContextBuilder().Set("user", "John").build();
+    /// HelloService helloSvc = ignite.GetServices().GetServiceProxy&lt;HelloService&gt;("hello-service", false, callCtx);
+    /// // Print "Hello John".
+    /// Console.WriteLine( helloSvc.call("Hello ") );
+    /// </code>
+    /// </summary>
+    public interface IServiceCallContext
```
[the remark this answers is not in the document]
xtern: Yes, thanks. I thought that this attribute is only applicable for methods/fields. 
I added `interface`-target and applied it to IServiceCallContext and ServiceCallContextBuilder.

ptupitsyn (approved):

PR #9564, opened 2021-11-15
IGNITE-15845 .NET: Fix TypeNameParser to ignore escaped characters in compiler-generated type names

ptupitsyn:
Compiler-generated type names use `\` to escape characters such as `[` and `,`, example: `MassTransit.Initializers.PropertyConverters.MessageDataPropertyConverter+<MassTransit-Initializers-IPropertyConverter<MassTransit-MessageData<System-Byte\[\]>\,MassTransit-MessageData<System-String>>-Convert>d__4`.

`TypeNameParser` was trying to parse those escaped characters as arrays, while it is just a part of the type name.


isapego (approved):
Looks good to me.

PR #9593, opened 2021-11-24
Ignite-15603 : Calcite. NPE on subquery returning multiple results.

Vladsz83:
Fixed correlated pass in table_scan and index_scan in order to put exchange before correlate

### DIFF
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/LogicalScanConverterRule.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/LogicalScanConverterRule.java
@@ -81,6 +81,12 @@ public abstract class LogicalScanConverterRule<T extends ProjectableFilterableTa
                     .replace(collation);
 
                 Set<CorrelationId> corrIds = RexUtils.extractCorrelationIds(rel.condition());
+
+                if (!F.isEmpty(rel.projects())) {
+                    corrIds = new HashSet<>(corrIds);
+                    corrIds.addAll(RexUtils.extractCorrelationIds(rel.projects()));
+                }
+
                 if (!corrIds.isEmpty())
                     traits = traits.replace(CorrelationTrait.correlations(corrIds));
 
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/LogicalScanConverterRule.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/LogicalScanConverterRule.java
@@ -18,6 +18,7 @@
 package org.apache.ignite.internal.processors.query.calcite.rule;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -46,6 +47,7 @@ import org.apache.ignite.internal.processors.query.calcite.schema.IgniteTable;
 import org.apache.ignite.internal.processors.query.calcite.trait.CorrelationTrait;
 import org.apache.ignite.internal.processors.query.calcite.trait.RewindabilityTrait;
 import org.apache.ignite.internal.processors.query.calcite.util.RexUtils;
+import org.apache.ignite.internal.util.typedef.F;
 
 /** */
 public abstract class LogicalScanConverterRule<T extends ProjectableFilterableTableScan> extends AbstractIgniteConverterRule<T> {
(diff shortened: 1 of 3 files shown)


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/LogicalScanConverterRule.java
```diff
@@ -81,6 +81,12 @@
                     .replace(collation);
 
                 Set<CorrelationId> corrIds = RexUtils.extractCorrelationIds(rel.condition());
+
+                if (!F.isEmpty(rel.projects())) {
```
alex-plekhanov, line 85: Why only for index scan? Table scan should be fixed too.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/CalciteQueryProcessorTest.java
```diff
@@ -1148,6 +1150,22 @@ public void testSelectWithOrdering() throws IgniteInterruptedCheckedException {
...
+    public void testSubqueryMultyValueError() throws IgniteInterruptedCheckedException {
+        sql("create table test_tbl(v INTEGER)", true);
+
+        sql("INSERT INTO test_tbl VALUES (1), (2), (3), (4), (5)", true);
+
+        Throwable err = GridTestUtils.assertThrows(log,
+            () -> sql("SELECT t0.v, (SELECT t0.v + t1.v FROM test_tbl t1) AS j FROM test_tbl t0"),
+            IgniteSQLException.class, null);
+
+        assertTrue(X.hasCause(err, "Subquery returned more than 1 value", IllegalArgumentException.class));
```
alex-plekhanov, line 1166: Can we write a query that is not throwing an exception after a fix, but throwing an exception without fix? For example by reducing the number of rows in test_tbl? If not, perhaps we should write planner test.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/CalciteQueryProcessorTest.java
```diff
@@ -1148,6 +1150,22 @@ public void testSelectWithOrdering() throws IgniteInterruptedCheckedException {
             .check();
     }
 
+    /**
+     * Checks exchnage-before-correlate and correct multy-value subquery error.
+     */
+    @Test
+    public void testSubqueryMultyValueError() throws IgniteInterruptedCheckedException {
```
alex-plekhanov, line 1157: The test name doesn't match javadoc. 
Actually, we should check correlated queries with correlates in projects (without correlates in filters), perhaps we should name the test (and fix the javadoc) according to this check. 
Also, fix typos (`multy` -> `multi`, `exchnage` -> `exchange`)

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/LogicalScanConverterRule.java
Vladsz83: Yep. IDE's Misscommit. Fixed.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/CalciteQueryProcessorTest.java
```diff
@@ -1148,6 +1150,18 @@ public void testSelectWithOrdering() throws IgniteInterruptedCheckedException {
...
+    /**
+     * Checks correlates are assigned before access.
+     */
+    @Test
+    public void testCorrelatesAssignedBeforeAccess() throws IgniteInterruptedCheckedException {
+        sql("create table test_tbl(v INTEGER)", true);
+
+        sql("INSERT INTO test_tbl VALUES (1)", true);
+
+        sql("SELECT t0.v, (SELECT t0.v + t1.v FROM test_tbl t1) AS j FROM test_tbl t0");
```
alex-plekhanov, line 1162: Let's assert result here

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/CalciteQueryProcessorTest.java
Vladsz83: Spelling fixed. Test renamed.

Filters? No filter is supposed. The best approach, I believe, would be the tree analysis. Besides, the tree is one of major goals to evolve and optimize. Probably, more effective solution here is to watch NPE/exception not to appear.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/CalciteQueryProcessorTest.java
Vladsz83: Done

alex-plekhanov (approved):

PR #9610, opened 2021-11-25
IGNITE-15921 Thin client: drop connection on invalid handshake without allocating buffer

ptupitsyn:
* Check first 3 bytes of the handshake message without allocating the buffer, drop connection in case of garbage data.
* Add 64 MiB handshake size limit.


modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/ClientMessage.java
```diff
@@ -35,6 +35,18 @@
     /** */
     private static final long serialVersionUID = -4609408156037304495L;
 
+    /** */
+    private static final int MAX_HANDSHAKE_SIZE = 1024 * 1024;
```
isapego, line 39: I'm not sure I like this solution as when we already implemented similia solution in the past and very soon got a complaint from a user that had really big password. Also, since that time we have introduced user attributes, which also can be pretty large and I can easily imagine handshake exceeding 1 MB. If we want to limit handshake size at least lets make it really big.

isapego (changes requested):
I propose to fix this issue in a little different way: 
1. Define `HANDSHAKE_MIN_SIZE` (size + version + client type at least),
2. Do not check or re-allocate buffer until we get message of size `HANDSHAKE_MIN_SIZE`
3. When we fully received message min header, check message type, major version for the client, client type, but do not check message size or at least set handshake max size limit larger value (e.g. first byte is zero looks like a safe check to me).

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/ClientMessage.java
ptupitsyn: We should pick a limit and stick to it. Current limit is essentially 2 GiB, but having a password or user attributes this large does not make sense, and may cause all sorts of performance issues.

If 1 MiB is too low, let's pick 10 MiB or 64 MiB, but no more than that.

isapego: 64MB seems OK to me, but let's see what @DirectXceriD and @slukyano think

DirectXceriD: 64Mb won't cause OOM (most likely), and it will cover case with large handshake payload, I'm good with that. @ptupitsyn @isapego thanks for letting me know, let's proceed.

ptupitsyn: Changed it to 64 MiB.

isapego (approved):

PR #9613, opened 2021-11-29
IGNITE-15515 Add exclusive find for BPlusTree

timoninmaxim:
Add opportunity to exclude boundaries for BPlusTree find. It helps avoid the filter step for IndexQuery for search with exclusive bounds for single criteria.


ibessonov (approved):
Can't say anything about index related code, but B+Tree changes look correct, thank you!

PR #9671, opened 2021-12-20
IGNITE-16111 Index rebuild handling

alex-plekhanov:

### Notes


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/logical/ExposeIndexRule.java
```diff
@@ -71,7 +71,10 @@ private static boolean preMatch(IgniteLogicalTableScan scan) {
             .map(idx -> idx.toRel(cluster, optTable, proj, condition, requiredCols))
             .collect(Collectors.toList());
 
-        assert !indexes.isEmpty();
+        indexes.removeIf(Objects::isNull);
```
korlov42, line 74: perhaps, it would be a bit better to filter out null with a stream API `.filter(Objects::nonNull)` before collecting them to a `indexes` collection

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteIndexScan.java
```diff
@@ -39,13 +41,18 @@
     /** */
     private final long sourceId;
 
+    /** */
+    private final RelCollation idxCollation;
```
korlov42, line 45: let's rename this to just `collation`

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteIndexScan.java
```diff
@@ -39,13 +41,18 @@
...
+
     /**
      * Constructor used for deserialization.
      *
      * @param input Serialized representation.
      */
     public IgniteIndexScan(RelInput input) {
-        super(changeTraits(input, IgniteConvention.INSTANCE));
+        super(changeTraits(input, IgniteConvention.INSTANCE, input.getCollation()));
```
korlov42, line 53: I don't sure it's a good idea to restore traits after deserialization. We either should do this for every node or don't do it at all.
Well, there is a few nodes (e.g. Exchange) which have an assertion inside their constructor, so `changeTraits` is used to overcome this problem. But for other cases I would prefer to avoid using it

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteIndexScan.java
```diff
@@ -126,7 +121,9 @@ private IgniteIndexScan(
     /** {@inheritDoc} */
     @Override protected RelWriter explainTerms0(RelWriter pw) {
         return super.explainTerms0(pw)
-            .itemIf("sourceId", sourceId, sourceId != -1);
+            .itemIf("sourceId", sourceId, sourceId != -1)
+            .item("collation", collation())
+            .item("idxCollation", idxCollation);
```
korlov42, line 126: why do we need both `collation` and `idxCollation`?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
```diff
@@ -298,13 +301,101 @@ public LogicalRelImplementor(
...
+                filterHasCorrelation ? null : filters,
+                projNodeRequired ? null : prj,
+                requiredColumns
+            );
 
-        Iterable<Row> rowsIter = idx.scan(ctx, group, filters, lower, upper, prj, requiredColumns);
+            Node<Row> node = new ScanNode<>(ctx, rowType, rowsIter);
 
-        return new ScanNode<>(ctx, rowType, rowsIter);
+            if (hasCollation || spoolNodeRequired) {
```
korlov42, line 333: Why do we create a SortNode for an empty collation?

korlov42: I think I got it... Let's add a comment here describing possible outcomes of this rewriting and the reasons of such outcomes

alex-plekhanov: Comment added

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteIndexScan.java
alex-plekhanov: `idxCollation` for index bounds, since `collation` an be trimmed.
I've removed `idxCollation` now, and add collation restoration by index bound.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteIndexScan.java
alex-plekhanov: Here we need information about collation on remote nodes, for other rels we don't need such information.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteIndexScan.java
alex-plekhanov: There is already `collation` (in traits), so it will be confusion. (but `idxCollation` already removed)

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/logical/ExposeIndexRule.java
alex-plekhanov: This block was rewritten, so there is no null check is required now.

korlov42 (approved):

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/logical/ExposeIndexRule.java
```diff
@@ -68,6 +68,9 @@ private static boolean preMatch(IgniteLogicalTableScan scan) {
         RexNode condition = scan.condition();
         ImmutableBitSet requiredCols = scan.requiredColumns();
 
+        if (igniteTable.isIndexRebuildInProgress())
```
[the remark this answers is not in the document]
alex-plekhanov: Calcite has some invariant and asserts preMatch after the rule is processed, so, in case of concurrent index rebuild flag change, there will be assertion error.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteIndexScan.java
alex-plekhanov: Sort relations don't have projections and "required columns", so sorting collation and trait collation are the same here. For index scans index collation and trait collation is different and there is confusion. Perhaps, it's better to store index collation in the "collation" field and restore output collation on remote nodes by index collation and projects. It's much easier than restoring index collation by bounds as currently implemented in this patch. WDYT?

alex-plekhanov: > Perhaps, it's better to store index collation in the "collation" field and restore output collation on remote nodes by index collation and projects. 

Hmm, but in this case sort node will be inserted even if it's not required (for example, when we just filter values by index), looks like it's not an optimal solution.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
```diff
@@ -298,13 +301,106 @@ public LogicalRelImplementor(
...
+                    requiredColumns,
+                    tbl.getRowType(typeFactory).getFieldCount()
+                ));
+            }
+
+            boolean sortNodeRequired = !collation.getFieldCollations().isEmpty();
+
+            if (sortNodeRequired) {
+                if (!spoolNodeRequired && projects != null)
+                    rowType = rel.getRowType();
```
korlov42, line 349: SortNode should have the same row type as its input

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/IndexRebuildPlannerTest.java
```diff
@@ -0,0 +1,367 @@
...
+        finally {
+            stop.set(true);
+        }
+
+        fut.get();
+    }
+
+    /** */
+    @Test
+    public void testIndexScanRewriter() throws Exception {
```
korlov42, line 136: I would prefer to move this test to a separate test class (LogicalRelImplementorSelfTest?)

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/IndexRebuildIntegrationTest.java
```diff
@@ -0,0 +1,316 @@
...
+            executeSql("INSERT INTO tbl2 VALUES (?, ?)", i, "val" + i);
+    }
+
+    /** */
+    @Test
+    public void testRebuildOnInitiatorNode() throws Exception {
+        String sql = "SELECT * FROM tbl WHERE id = 0 AND val='val0'";
+
+        QueryChecker validChecker = assertQuery(grid(0), sql)
+            .matches(QueryChecker.containsIndexScan("PUBLIC", "TBL", "IDX_ID_VAL"))
```
korlov42, line 104: We need to stop checking query plans in integration tests. There is already a test IndexRebuildPlannerTest which verifies a planner behaviour during index rebuilding. For integration tests the only thing that matters is the result set.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/IndexRebuildPlannerTest.java
alex-plekhanov: It still will be a subclass of AbstractPlannerTest since it requires planning methods. There will be code duplication with the current test (we need IndexRebuildAwareTable and table initialization). Is it ok?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/IndexRebuildIntegrationTest.java
alex-plekhanov: I disagree here. If the plan will be changed for some of these queries and there will be no index scan rewrite anymore, nobody will notice it, functionality became uncovered by tests and easily can be broken.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteIndexScan.java
```diff
@@ -87,9 +78,10 @@ public IgniteIndexScan(
         @Nullable List<RexNode> proj,
         @Nullable RexNode cond,
         @Nullable IndexConditions idxCond,
-        @Nullable ImmutableBitSet requiredCols
+        @Nullable ImmutableBitSet requiredCols,
+        @Nullable RelCollation collation
```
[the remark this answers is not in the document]
alex-plekhanov: Fixed

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
alex-plekhanov: Fixed

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/IndexRebuildPlannerTest.java
korlov42: If you need a template rel, you could use its serialized form and just deserialize it whenever you need it. As for code duplication, the table class could be moved to a separate class.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/IndexRebuildIntegrationTest.java
korlov42: This functionality is covered by the planner test. Besides, even with such a validation you still can't be sure you are verifying the right scenario, because for explain call the query will be optimized the second time, hence the plan could be different.

alex-plekhanov: Then we need to reuse the same queries for integration and planning tests. And we should ensure that queries in planning tests can't be changed without corresponding changes to integration tests. Perhaps, introduce some constants with queries. But I think it's not much convenient, so it's better to just recheck the required rel op in integration tests.

Without concurrent workload, the plan should be the same for explain and the query. It's much more likely that plan in planner test for the same query will be different than plan for integration test (and there are a lot of such cases currently).

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/IndexRebuildPlannerTest.java
alex-plekhanov: Fixed

modules/calcite/src/test/java/org/apache/ignite/testsuites/ExecutionTestSuite.java
```diff
@@ -53,6 +54,7 @@
     IntersectExecutionTest.class,
     RuntimeSortedIndexTest.class,
     LimitExecutionTest.class,
+    LogicalRelImplementorTest.class,
```
korlov42, line 57: Execution test suite is supposed to verify actual execution of particular node or group of nodes. Let's add this test to the IgniteCalciteTestSuite.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteIndexScan.java
```diff
@@ -111,11 +104,13 @@ private IgniteIndexScan(
...
```
korlov42, line 108: Did you forgot to remove annotation here?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/IndexRebuildIntegrationTest.java
korlov42: My point is you don't need to duplicate tests from planner test in integration test suite. And you don't need to duplicate test from LogicalRelImplementorTest too. Both planner part and rewriter part are covered.

Imo, for the integration part, we need to check if the cluster returns a valid result in the event that an index rebuild is triggered.

modules/calcite/src/test/java/org/apache/ignite/testsuites/ExecutionTestSuite.java
alex-plekhanov: Fixed

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteIndexScan.java
alex-plekhanov: Fixed

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/IndexRebuildIntegrationTest.java
alex-plekhanov: It's not a duplication, LogicalRelImplementorTest doesn't check execution node internals (and can't check it without reflection or without exposing these internals).

The integration part does not check the plan, it checks result, but ensures that this result was received using the correct code path.

korlov42 (approved):

alex-plekhanov:
Merged to sql-calcite branch.

PR #9722, opened 2022-01-11
IGNITE-16242 .NET: Fix platform cache behavior with persistent storage

ptupitsyn:
* Disable `readNoEntry` optimization when platform cache is enabled so that `GridCacheMapEntry` gets created and updates platform cache after node restart.
* Update platform cache when entry is initially unswapped from storage.
* Fix `cacheIdAndPartition` bitwise combination.
* Use `onDoneBeforeTopologyUnlock` instead of `onDoneAfterTopologyUnlock` to update platform cache topology version earlier and avoid null value on node start.
* Disable local partition scan optimization (IGNITE-12882) when persistence is enabled - there is no guarantee that all entries are in memory.


isapego (approved):
Looks good to me.

PR #9734, opened 2022-01-13
IGNITE-16294 .NET: Fix System.Enum serialization

Vladsz83:
.NET binary writer doesn't get `System.Enum` field as an enum. Happens because `System.Enum::IsEnum == false` and because we build user-type serialization based on declared field types (`FieldInfo::FieldType`), not on types of real assigned values.

Solution: Fixed the type caster so that it deals with `System.Enum`.


nizhikov (approved):

nizhikov:
LGTM 

@ptupitsyn Can you, please, take a look at this changes, also.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Common/TypeCaster.cs
```diff
@@ -87,9 +89,39 @@ private static class Casters<TFrom>
                 }
 
                 var paramExpr = Expression.Parameter(typeof(TFrom));
-                var convertExpr = Expression.Convert(paramExpr, typeof(T));
 
-                return Expression.Lambda<Func<TFrom, T>>(convertExpr, paramExpr).Compile();
+                if (typeof(TFrom) == typeof(Enum))
+                    return Expression.Lambda<Func<TFrom, T>>(TryConvertRawEnum(typeof(T), paramExpr),
```
ptupitsyn, line 94: What if `TryConvertRawEnum` returns `null`?

ptupitsyn (changes requested):

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Common/TypeCaster.cs
Vladsz83: Would fail with same conversion error.

ptupitsyn: `Expression.Lambda()` will throw `ArgumentNullException` if `TryConvertRawEnum` returns null. `ArgumentException` is always a bug. Let's check for null and throw `InvalidCastException` explaining that `Enum` can't be converted to the specified type.

ptupitsyn (approved):

ptupitsyn:
Looks good to me.

PR #9771, opened 2022-01-26
IGNITE-16400 : Deprecate direct references to local services. Fix service proxy.

Vladsz83:
As discussed in the threads related to the main ticket #IGNITE-12464, we deprecate references to local services. Also we will return service proxy every time even for local services by `serviceProxy()`

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/IgniteServices.java
+++ b/modules/core/src/main/java/org/apache/ignite/IgniteServices.java
@@ -567,7 +567,11 @@ public interface IgniteServices extends IgniteAsyncSupport {
      * @param name Service name.
      * @param <T> Service type
      * @return Deployed service with specified name.
+     * @see ServiceConfiguration#setStatisticsEnabled(boolean)
+     * @deprecated Use the proxies like {@link #serviceProxy(String, Class, boolean)}. References to local services
+     * corrupt the service statistics and bring no real performance optimization.
      */
+    @Deprecated
     public <T> T service(String name);
 
     /**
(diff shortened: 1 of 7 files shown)


modules/core/src/main/java/org/apache/ignite/IgniteServices.java
```diff
@@ -576,13 +580,15 @@ public void deployMultiple(String name, Service svc, int totalCnt, int maxPerNod
...
+     * corrupt the service statistics and bring no real performance optimization.
      */
+    @Deprecated
     public <T> Collection<T> services(String name);
 
     /**
-     * Gets a remote handle on the service. If service is available locally,
-     * then local instance is returned, otherwise, a remote proxy is dynamically
-     * created and provided for the specified service.
+     * Gets a handle on the service.
```
timoninmaxim, line 591: Let's keep a mention about remote services and proxy creation. 

For example, "Gets a handle on the remote or local service. Proxy is dynamically created and provided for the specified service".

modules/core/src/main/java/org/apache/ignite/internal/processors/service/IgniteServiceProcessor.java
```diff
@@ -1061,23 +1061,6 @@ public ServiceContextImpl serviceContext(String name) {
     ) throws IgniteException {
         ctx.security().authorize(name, SecurityPermission.SERVICE_INVOKE);
 
-        if (hasLocalNode(prj)) {
```
timoninmaxim, line 1064: Now `IgniteServiceProcessor#hasLocalNode()` is obsolete and can be safely removed.

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/IgniteServices.java
Vladsz83: Fixed

modules/core/src/main/java/org/apache/ignite/IgniteServices.java
```diff
@@ -613,8 +618,8 @@ public void deployMultiple(String name, Service svc, int totalCnt, int maxPerNod
         throws IgniteException;
 
     /**
-     * Gets a remote handle on the service with the specified caller context. The proxy
-     * is dynamically created and provided for the specified service.
+     * Gets a handle on remote or local service with the precified caller context. The proxy is dynamically created and
```
timoninmaxim, line 621: precified --> specified

timoninmaxim (approved):

PR #9775, opened 2022-01-27
IGNITE-16416 : Unreadable result on cluster deactivation via IgniteMXBean

Vladsz83:
IgniteMXBean can throw IgniteException on the cluster deactivation. It's is OK with the exception itself - cluster must not be deactivated without 'force' flag if that would issue data loss. Unfortunately, such exception is bound to the class path and might not be serialized by other remote size like jconsole.

Solution: exception is converted to basic RuntimeException in case of IgniteMXBean


modules/core/src/main/java/org/apache/ignite/internal/IgniteMXBeanImpl.java
```diff
@@ -330,8 +330,9 @@ public IgniteMXBeanImpl(IgniteKernal kernal) {
             ctx.state().changeGlobalState(newState, forceDeactivation, ctx.cluster().get()
                 .forServers().nodes(), false).get();
         }
-        catch (IgniteCheckedException e) {
-            throw U.convertException(e);
+        catch (IgniteException | IgniteCheckedException e) {
+            // Convert to deserializable one.
+            throw new RuntimeException("Unable to change cluster state to '" + state + "': " + e.getMessage());
```
timoninmaxim, line 335: Why don't you use `throw U.jmException(e);` as other beans?

Vladsz83: Fixed

timoninmaxim (approved):

PR #9788, opened 2022-02-02
IGNITE-16446: Fix IndexQuery incorrect work with indexes built for _VAL field

timoninmaxim:
IndexQuery handles indexes built for single _VAL (or alias) field correctly. Now it checks field type before sorting on an originating node.


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/reducer/IndexQueryReducer.java
```diff
@@ -138,12 +138,29 @@ public IndexQueryReducer(
...
+
+                o = prop.value(entry.getKey(), entry.getValue());
+            }
 
             return IndexKeyFactory.wrap(o, type, cctx.cacheObjectContext(), meta.keyTypeSettings());
         }
+
+        /** */
+        private boolean isKeyField(String fld) {
+            return fld.equals(typeDesc.keyFieldAlias()) || fld.equals(QueryUtils.KEY_FIELD_NAME);
```
alex-plekhanov, line 158: `equalsIgnoreCase`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/reducer/IndexQueryReducer.java
```diff
@@ -138,12 +138,29 @@ public IndexQueryReducer(
...
+        }
+
+        /** */
+        private boolean isValueField(String fld) {
+            return fld.equals(typeDesc.valueFieldAlias()) || fld.equals(QueryUtils.VAL_FIELD_NAME);
```
alex-plekhanov, line 163: `equalsIgnoreCase`?

modules/indexing/src/test/java/org/apache/ignite/cache/query/IndexQueryCacheKeyValueFieldsTest.java
```diff
@@ -0,0 +1,120 @@
...
+            new IndexQuery<String, Integer>(Integer.class)
+                .setCriteria(gt(fld, key ? key(pivot) : pivot))
+        );
+
+        assertEquals(CNT - pivot - 1, cursor.getAll().size());
+    }
+
+    /** */
+    private static String key(int val) {
+        return String.format("key_%1$" + 5 + "s", val).replace(' ', '0');
```
alex-plekhanov, line 118: `"key_%1$5s"` - for better readability.

modules/indexing/src/test/java/org/apache/ignite/cache/query/IndexQueryCacheKeyValueFieldsTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /** */
+    @Test
+    public void testKeyAliasField() {
+        check("f1", true);
+    }
+
+    /** */
+    @Test
+    public void testKeyField() {
```
alex-plekhanov, line 100: Let's also add tests with upper/lower case letters

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/reducer/IndexQueryReducer.java
timoninmaxim: `fld`, `keyFieldAlias` are already normalized here, so it is impossible to fail with getting non-normalized keys here.

But, also we should keep in mind the `CacheConfiguration#sqlEscapeAll()` flag. In case it set up, a user have to specify field in the same way as it was declared. For example, if field declared as "f1", then with `sqlEscapeAll=false` it will be normalized to "F1" and user can query it with both "f1" or "F1". But if `sqlEscapeAll=true`, then field will be "f1" and user has to run query with "f1" attribute only, and it's prohibited to use "F1" for querying. 

So `String#equals` works in both cases with or without escaping. I added tests for that.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/reducer/IndexQueryReducer.java
timoninmaxim: The same.

modules/indexing/src/test/java/org/apache/ignite/cache/query/IndexQueryCacheKeyValueFieldsTest.java
timoninmaxim: Added tests for cases and escaping.

modules/indexing/src/test/java/org/apache/ignite/cache/query/IndexQueryCacheKeyValueFieldsTest.java
```diff
@@ -0,0 +1,175 @@
...
+    protected boolean escape() {
+        return false;
+    }
+
+    /** */
+    @Parameterized.Parameters(name = "escape={0}")
+    public static Object[] params() {
+        return new Object[] { false, true };
+    }
+
```
alex-plekhanov, line 57: Test is not parametrized, looks like this is redundant

timoninmaxim: Removed.

alex-plekhanov (approved):

PR #9817, opened 2022-02-11
IGNITE-16462 .NET: Thin client: add heartbeats

ptupitsyn:
Implement keepalive: https://cwiki.apache.org/confluence/display/IGNITE/IEP-83+Thin+Client+Keepalive
* Add `OP_HEARTBEAT`, `OP_GET_IDLE_TIMEOUT` on server with `HEARTBEAT` feature flag.
* Implement heartbeats in .NET thin client:
  * `IgniteClientConfiguration.EnableHeartbeats`, default `false`.
  * `IgniteClientConfiguration.HeartbeatInterval`, default 30 seconds.
  * When heartbeats are enabled, get idle timeout from server, set effective heartbeat interval to `Math.Min(HeartbeatInterval, IdleTimeout / 3)`. Log warning when user-defined value is overridden.

### DIFF
--- a/modules/platforms/dotnet/Apache.Ignite.Core/Client/IgniteClientConfiguration.cs
+++ b/modules/platforms/dotnet/Apache.Ignite.Core/Client/IgniteClientConfiguration.cs
@@ -25,6 +25,7 @@ namespace Apache.Ignite.Core.Client
     using System.Xml;
     using Apache.Ignite.Core.Binary;
     using Apache.Ignite.Core.Client.Transactions;
+    using Apache.Ignite.Core.Configuration;
     using Apache.Ignite.Core.Impl.Binary;
     using Apache.Ignite.Core.Impl.Client;
     using Apache.Ignite.Core.Impl.Common;
@@ -274,6 +275,9 @@ namespace Apache.Ignite.Core.Client
         /// <para />
         /// When thin client connection is idle (no operations are performed), heartbeat messages are sent periodically
         /// to keep the connection alive and detect potential half-open state.
+        /// <para />
+        /// Can be used together with server-side idle timeout (<see cref="ClientConnectorConfiguration.IdleTimeout"/>)
+        /// to detect half-open connections on the server.
         /// </summary>
         public TimeSpan HeartbeatInterval { get; set; }
 
--- a/modules/platforms/dotnet/Apache.Ignite.Core/Client/IgniteClientConfiguration.cs
+++ b/modules/platforms/dotnet/Apache.Ignite.Core/Client/IgniteClientConfiguration.cs
@@ -65,6 +65,21 @@ namespace Apache.Ignite.Core.Client
         public static readonly TimeSpan DefaultSocketTimeout = TimeSpan.FromMilliseconds(5000);
 
         /// <summary>
+        /// Automatic heartbeat interval.
+        /// </summary>
+        public static readonly TimeSpan AutoHeartbeatInterval = TimeSpan.Zero;
+
+        /// <summary>
+        /// Disabled heartbeat interval.
+        /// </summary>
+        public static readonly TimeSpan DisabledHeartbeatInterval = TimeSpan.FromSeconds(-1);
+
+        /// <summary>
+        /// Default heartbeat interval.
+        /// </summary>
+        public static readonly TimeSpan DefaultHeartbeatInterval = AutoHeartbeatInterval;
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="IgniteClientConfiguration"/> class.
         /// </summary>
         public IgniteClientConfiguration()
@@ -267,7 +282,7 @@ namespace Apache.Ignite.Core.Client
         public int RetryLimit { get; set; }
 
         /// <summary>
-        /// TODO: Auto (0, default), Disabled (-1), manual (> 0).
+        /// TODO: Auto (0, default), Disabled (-1), manual (> 0). Explain the behavior.
         /// </summary>
         public TimeSpan HeartbeatInterval { get; set; }
 
--- a/modules/platforms/dotnet/Apache.Ignite.Core/Client/IgniteClientConfiguration.cs
+++ b/modules/platforms/dotnet/Apache.Ignite.Core/Client/IgniteClientConfiguration.cs
@@ -66,9 +66,9 @@ namespace Apache.Ignite.Core.Client
         public static readonly TimeSpan DefaultSocketTimeout = TimeSpan.FromMilliseconds(5000);
 
         /// <summary>
-        /// Default value for <see cref="DefaultHeartbeatInterval"/>.
+        /// Default value for <see cref="HeartbeatInterval"/>.
         /// </summary>
-        public static readonly TimeSpan DefaultDefaultHeartbeatInterval = TimeSpan.FromSeconds(30);
+        public static readonly TimeSpan DefaultHeartbeatInterval = TimeSpan.FromSeconds(30);
 
         /// <summary>
         /// Initializes a new instance of the <see cref="IgniteClientConfiguration"/> class.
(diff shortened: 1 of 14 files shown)


isapego (approved):
Looks good to me.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/ClientSocket.cs
```diff
@@ -170,6 +176,29 @@ internal sealed class ClientSocket : IDisposable
...
 
+            if (_features.HasFeature(ClientBitmaskFeature.Heartbeat) &&
+                clientConfiguration.HeartbeatInterval > TimeSpan.Zero)
+            {
+                _heartbeatInterval = clientConfiguration.HeartbeatInterval;
+
+                var serverIdleTimeout = TimeSpan.FromMilliseconds(
+                    DoOutInOp(ClientOp.GetIdleTimeout, null, r => r.Reader.ReadLong()));
+
+                if (_heartbeatInterval > serverIdleTimeout)
```
timoninmaxim, line 187: `ClientConnectorConfiguration#getIdleTimeout` can be 0 or negative, it means that the idleTimeout is disabled. We should skip the WARN for such cases. 

I don't see such check, but I'm not familiar with .Net so maybe I miss it.

ptupitsyn: Good catch, added a test and fixed the logic.

ivandasch: As for me, it is strange decision not to compute heartbeat interval based on server idle timeout.

ptupitsyn: @ivandasch I tried that and configuration gets too complicated. Maybe you have a better idea than this:
* `HeartbeatInterval` // `-1` is disabled (default), `0` is auto, `> 0` is manual (explicit)
* `DefaultAutoHeartbeatInterval` // when `HeartbeatInterval` is auto, but `idleTimeout` is `0`.

ivandasch: I suppose that HeartbetInterval should not be greater than one-third of IdleTimeout (this is approach from ZookeeperClient).
So logic is:
1. If idleTimeout is greater than 0, and heartbeat interval is not set or larger than one-third of idle timeout -- set it as one-third of idle timeout. If hearbtBeatInterval less that one-third of idleTimeout -- use heartbeatInterval as is
2. If idleTimeout is disabled, use heartbeat interval if it is provided.

ivandasch: > This will break potentially desirable behavior: don't enable heartbeats and actually drop thin clients that become idle.
I don't understand why, it would be great if you give an example of breakage

> We reset the timer on each write, see [ClientSocket.cs:854](https://github.com/apache/ignite/pull/9817/files#diff-81a8959836fbee45664e37261770ee4a215e2111c0eddd3535160ef7332232ecR854).
Great, I didn't catch it, sorry

ptupitsyn: As a user, I've set `idleTimeout` on server so that idle client connections are closed automatically and resources are not consumed. It is up to the clients to reconnect when needed.

Now, if we enable heartbeats automatically with one-third of `idleTimeout`, those idle connections will never be dropped, and server will consume resources keeping them open.

It is just a different use case. Some users want to keep client connections open forever, even when not doing anything. Other users may want to drop them.

Too many use cases and edge cases, that's why I suggest to keep it simple and let the user decide.

ivandasch: > As a user, I've set idleTimeout on server so that idle client connections are closed automatically and resources are not consumed. It is up to the clients to reconnect when needed.

Quite confusing desire as for me, cannot imagine that someone would expect this behavior. Idle client is the client that doesn't send anything, possibly stale, hang or partial dead. As for me, heartbeats is the good mechanism to notify that client is not idle, just on pause or probably waiting for some tasks.

ivandasch: As for me, it makes things more complicated for user. If we choose this approach, other clients should comply this decision. Let us discuss this on dev list

ptupitsyn: @ivandasch ok, let's continue on the dev list. Please describe your proposal there.

isapego: Agree, let's discuss on devlist and update IEP accordingly once everyone's agree with a solution.

ptupitsyn: As agreed on the dev list:
* `EnableHeartbeats` property added.
* `HeartbeatInterval` renamed to `DefaultHeartbeatInterval`.
* Logic and xmldocs updated.

ivandasch (approved):

modules/platforms/dotnet/Apache.Ignite.Core/Client/IgniteClientConfiguration.cs
```diff
@@ -64,6 +65,11 @@ public class IgniteClientConfiguration
         /// </summary>
         public static readonly TimeSpan DefaultSocketTimeout = TimeSpan.FromMilliseconds(5000);
 
+        /// <summary>
+        /// Default value for <see cref="DefaultHeartbeatInterval"/>.
+        /// </summary>
+        public static readonly TimeSpan DefaultDefaultHeartbeatInterval = TimeSpan.FromSeconds(30);
```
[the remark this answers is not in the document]
ptupitsyn: What would you suggest if we were doing this in Java or Python?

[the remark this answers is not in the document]
ivandasch: Just `HeartbeatInterval` and `DefaultHeartbeatInteval`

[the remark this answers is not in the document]
ptupitsyn: Agree, renamed.

timoninmaxim (approved):

timoninmaxim:
Hi, this patch LGTM. But I'm not sure, but maybe we should add some smoke tests that check connect a client to multiple nodes to be confident that the timer isn't shared between channels. We configure heartbeat on ClientConfiguration level, where also partition awareness configured. Then it looks meaningful and maybe it can be useful for regression testing, WDYT?

ptupitsyn:
@timoninmaxim good point, I've enabled `PartitionAwareness` in `ClientHeartbeatTest`, I think it should cover the regression matters.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/ClientSocket.cs
```diff
@@ -182,6 +203,43 @@ internal sealed class ClientSocket : IDisposable
...
+                        $"Server-side IdleTimeout is {serverIdleTimeoutMs}ms, " +
+                        $"using configured {nameof(IgniteClientConfiguration)}." +
+                        $"{nameof(IgniteClientConfiguration.HeartbeatInterval)}: " +
+                        clientConfiguration.HeartbeatInterval);
+
+                    return clientConfiguration.HeartbeatInterval;
+                }
+
+                _logger.Info($"Server-side IdleTimeout is {serverIdleTimeoutMs}ms, " +
+                             $"using 1/3 of it as heartbeat interval: {recommendedHeartbeatInterval}");
```
isapego, line 231: This should be a `Warn` in my opinion as we definitely have a problem with client/server configuration here and potentially unexpected behavior for the user.

ptupitsyn: Agree, fixed.

isapego (approved):

PR #9851, opened 2022-02-24
IGNITE-16549 .NET: Add IgniteClientConfiguration.SendServerExceptionStackTraceToClient

ptupitsyn:
* Propagate SendServerExceptionStackTraceToClient to .NET
* Java fix: Include stack trace for Compute errors
* Java fix: Include stack trace even when cause is null

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientRequestHandler.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientRequestHandler.java
@@ -125,8 +125,7 @@ public class ClientRequestHandler implements ClientListenerRequestHandler {
             msg = sqlState + ": " + msg;
         }
 
-        if ((X.getCause(e) != null || !X.getSuppressedList(e).isEmpty())
-                && ctx.kernalContext().sqlListener().sendServerExceptionStackTraceToClient())
+        if (ctx.kernalContext().sqlListener().sendServerExceptionStackTraceToClient())
             msg = msg + U.nl() + X.getFullStackTrace(e);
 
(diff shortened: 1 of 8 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientRequestHandler.java
```diff
@@ -125,8 +125,7 @@
             msg = sqlState + ": " + msg;
         }
 
-        if ((X.getCause(e) != null || !X.getSuppressedList(e).isEmpty())
-                && ctx.kernalContext().sqlListener().sendServerExceptionStackTraceToClient())
+        if (ctx.kernalContext().sqlListener().sendServerExceptionStackTraceToClient())
```
isapego, line 128: Should we preserve `!X.getSuppressedList(e).isEmpty()` logic here?

isapego (approved):
Looks good overall. See a single comment I have.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientRequestHandler.java
ptupitsyn: I'm not sure why. The option is called `sendServerExceptionStackTraceToClient`, so we just send the stack trace whatever it is, when enabled.

With those conditions in place I tried a few things like "disabled compute", "exception in compute task" and never got it to work.

isapego: Ok, your solution is reasonable then

PR #9873, opened 2022-03-05
IGNITE-16649 .NET: Fix missing binary schema update when field is removed

ptupitsyn:
When new binary object schema is a subset of existing schema, it was not sent to the cluster. Detect this case and send the schema.


ivandasch (approved):
Looks great!

PR #9887, opened 2022-03-14
Ignite-16685 : Expiration fails on topology version when cluster gets re-activated. V1 - Delayed expiration start.

Vladsz83:
Solutions:
Start expiration later, on confirmed topology.


Vladsz83:
Better version applied: https://github.com/apache/ignite/pull/9888

PR #9916, opened 2022-03-28
IGNITE-16749 .NET: Fix EntryPointNotFoundException on Alpine Linux

ptupitsyn:
On some Linux distros we can load `dlopen` and `pthread_*` symbols from `libcoreclr.so` to avoid dependency on `libc-dev`. However, this does not work on Alpine.

Add exception handler: if `libcoreclr` approach does not work, fall back to `libpthread.so` and `libdl.so`.

* `DllLoader` is called only once to load jvm.dll, it is fine to handle the exception inline.
* `UnmanagedThread` is called many times: perform the check once and set up delegates.

Tested on Alpine 3.15, Ubuntu 20.04, macOs Catalina 2019, Windows 10.


isapego (approved):
Looks good to me.

PR #9924, opened 2022-03-30
IGNITE-16582 Improve behavior of speed-based throttling when dirty pages ratio is low

rpuch:
Speed-based throttle contained 2 algorithms (it took max of the values they were returning):
    
* Simpler that throttled when current marking speed was higher than average CP speed
* More complex one that first computed max acceptable marking speed (taking into account clean pages ratio and average CP speed) and then throttled if current marking speed was higher than the acceptable speed AND the current dirty ratio was higher than the acceptable dirty ratio
    
There was a bug that prevented the second algorithm from ever kicking in.
    
The first one did not take into account the clean pages that we still had, so it throttled bursts even in situations when the buffering power of the clean pages was still enough to smooth the bursts. Also, the second algorithm seems to be more general than the first one, but more adaptive.
    
That's why this commit removes the first algorithm and enables the second one.

https://issues.apache.org/jira/browse/IGNITE-16582

### Notes

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedMemoryConsumptionThrottlingStrategy.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedMemoryConsumptionThrottlingStrategy.java
@@ -57,7 +57,7 @@ class SpeedBasedMemoryConsumptionThrottlingStrategy {
     /**
      * Total pages possible to store in page memory.
      */
-    private final long totalPages;
+    private volatile long pageMemTotalPages;
 
     /**
      * Last estimated speed for marking all clear pages as dirty till the end of checkpoint.
@@ -111,8 +111,6 @@ class SpeedBasedMemoryConsumptionThrottlingStrategy {
         this.pageMemory = pageMemory;
         this.cpProgress = cpProgress;
         this.markSpeedAndAvgParkTime = markSpeedAndAvgParkTime;
-
-        totalPages = pageMemory.totalPages();
     }
 
     /**
@@ -149,7 +147,7 @@ class SpeedBasedMemoryConsumptionThrottlingStrategy {
         final int cpWrittenPages = writtenPagesCounter.get();
         final long donePages = cpDonePagesEstimation(cpWrittenPages);
 
-        final long markDirtySpeed = markSpeedAndAvgParkTime.getSpeedOpsPerSec(curNanoTime);
+        final long instantaneousMarkDirtySpeed = markSpeedAndAvgParkTime.getSpeedOpsPerSec(curNanoTime);
         // NB: we update progress for speed calculation only in this (clean pages protection) scenario, because
         // we only use the computed speed in this same scenario and for reporting in logs (where it's not super
         // important to display an ideally accurate speed), but not in the CP Buffer protection scenario.
@@ -157,7 +155,8 @@ class SpeedBasedMemoryConsumptionThrottlingStrategy {
         // The progress is set to 0 at the beginning of a checkpoint, so we can be sure that the start time remembered
         // in cpWriteSpeed is pretty accurate even without writing to cpWriteSpeed from this method.
         cpWriteSpeed.setProgress(donePages, curNanoTime);
-        final long curCpWriteSpeed = cpWriteSpeed.getOpsPerSecond(curNanoTime);
+        // IDEA: use exponential moving average instead of regular moving average so that we react to changes faster?
+        final long avgCpWriteSpeed = cpWriteSpeed.getOpsPerSecond(curNanoTime);
 
         final int cpTotalPages = cpTotalPages();
 
@@ -166,11 +165,11 @@ class SpeedBasedMemoryConsumptionThrottlingStrategy {
             // CheckpointProgressImpl.clearCounters() is invoked at the end of a checkpoint (by falling through
             // between two volatile assignments). When we get here, we don't have any information about the total
             // number of pages in the current CP, so we calculate park time by only using information we have.
-            return parkTimeToThrottleByJustCPSpeed(markDirtySpeed, curCpWriteSpeed);
+            return parkTimeToThrottleByJustCPSpeed(instantaneousMarkDirtySpeed, avgCpWriteSpeed);
         }
         else {
-            return speedBasedParkTime(cpWrittenPages, donePages, markDirtySpeed,
-                    curCpWriteSpeed, cpTotalPages);
+            return speedBasedParkTime(cpWrittenPages, donePages, cpTotalPages, instantaneousMarkDirtySpeed,
+                    avgCpWriteSpeed);
         }
     }
 
@@ -183,6 +182,9 @@ class SpeedBasedMemoryConsumptionThrottlingStrategy {
      * @return estimation of work done (in pages)
      */
     private int cpDonePagesEstimation(int cpWrittenPages) {
+        // IDEA: this only works correctly if time-to-write a page is close to time-to-sync a page. In reality, this
+        // does not seem to hold, which produces wrong estimations. We could measure the real times in Checkpointer
+        // and make this estimation a lot more precise.
         return (cpWrittenPages + cpSyncedPages()) / 2;
     }
 
@@ -198,15 +200,15 @@ class SpeedBasedMemoryConsumptionThrottlingStrategy {
         boolean throttleByCpSpeed = curCpWriteSpeed > 0 && markDirtySpeed > curCpWriteSpeed;
 
         if (throttleByCpSpeed) {
-            return calcDelayTime(curCpWriteSpeed);
+            return nsPerOperation(curCpWriteSpeed);
         }
 
         return 0;
     }
 
     /***/
-    private long speedBasedParkTime(int cpWrittenPages, long donePages, long markDirtySpeed,
-                                    long curCpWriteSpeed, int cpTotalPages) {
+    private long speedBasedParkTime(int cpWrittenPages, long donePages, int cpTotalPages,
+                                    long instantaneousMarkDirtySpeed, long avgCpWriteSpeed) {
         final double dirtyPagesRatio = pageMemory.getDirtyPagesRatio();
 
         currDirtyRatio = dirtyPagesRatio;
@@ -220,8 +222,8 @@ class SpeedBasedMemoryConsumptionThrottlingStrategy {
                     donePages,
                     notEvictedPagesTotal(cpTotalPages),
                     threadIdsCount(),
-                    markDirtySpeed,
-                    curCpWriteSpeed);
+                    instantaneousMarkDirtySpeed,
+                    avgCpWriteSpeed);
         }
     }
 
@@ -237,8 +239,8 @@ class SpeedBasedMemoryConsumptionThrottlingStrategy {
      * @param donePages           roughly, written & fsynced pages count.
      * @param cpTotalPages        total checkpoint scope.
      * @param nThreads            number of threads providing data during current checkpoint.
-     * @param markDirtySpeed      registered mark dirty speed, pages/sec.
-     * @param curCpWriteSpeed     average checkpoint write speed, pages/sec.
+     * @param instantaneousMarkDirtySpeed registered (during approx last second) mark dirty speed, pages/sec.
+     * @param avgCpWriteSpeed     average checkpoint write speed, pages/sec.
      * @return time in nanoseconds to part or 0 if throttling is not required.
      */
     long getParkTime(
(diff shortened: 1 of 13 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedThrottleIntegrationTest.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override protected long getTestTimeout() {
+        return 3 * 60 * 1000;
+    }
+
+    /**
+     */
+    @Test
+    public void speedBasedThrottleShouldBeActivatedWhenNeeded() throws Exception {
+        AtomicBoolean throttled = new AtomicBoolean(false);
+        listeningLog.registerListener(message -> {
```
ibessonov, line 93: There are regexp-based or substring-based matchers for these loggers, you should probably use them. Just check out other usages

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteThrottleSandboxTest.java
```diff
@@ -172,14 +194,27 @@ public void testThrottle() throws Exception {
...
+                    else {
+                        ds.addData(ThreadLocalRandom.current().nextInt(keyCnt), new TestValue(ThreadLocalRandom.current().nextInt(),
+                            ThreadLocalRandom.current().nextInt()));
 
-                    putRate.increment();
+                        putRate.increment();
+                        putCount.incrementAndGet();
+                    }
+
+                    if (System.nanoTime() - startNanos > TimeUnit.MINUTES.toNanos(10))
```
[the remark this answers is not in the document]
rpuch: This test is only run manually when you want to see with your eyes how throttling happens. The test duration is 10 minutes. The duration is chosen as a sufficient, but not too long duration, and it's good to have it fixed because it becomes easier to compare the results of different runs.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteThrottleSandboxTest.java
```diff
@@ -247,4 +287,9 @@ private void deleteWorkFiles() throws Exception {
         cleanPersistenceDir();
         U.delete(U.resolveWorkDirectory(U.defaultWorkDirectory(), "snapshot", false));
     }
+
+    /** {@inheritDoc} */
+    @Override protected FailureHandler getFailureHandler(String igniteInstanceName) {
+        return new StopNodeOrHaltFailureHandler();
```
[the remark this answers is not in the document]
rpuch: This test never gets run on TC, it's only for running it manually on the developer's machine. And on the developer's machine, this failure handler makes the failure more 'visible' if it happens.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedThrottleIntegrationTest.java
rpuch: The existing library leads to a code that is not too comprehensible. Like ` if (logListener.check()) { break; }` - it would make a reader wonder what kind of a check is it? If the method would be named appropriately, like `expectationsSatisfied()`, it would make sense to use it. With the current state of the code, it seems that it's easier to understand the code I wrote than use the 'standard' tools.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedMemoryConsumptionThrottlingStrategy.java
```diff
@@ -149,15 +147,16 @@ private long computeParkTime(@NotNull AtomicInteger writtenPagesCounter, long cu
...
+        final long instantaneousMarkDirtySpeed = markSpeedAndAvgParkTime.getSpeedOpsPerSec(curNanoTime);
         // NB: we update progress for speed calculation only in this (clean pages protection) scenario, because
         // we only use the computed speed in this same scenario and for reporting in logs (where it's not super
         // important to display an ideally accurate speed), but not in the CP Buffer protection scenario.
         // This is to simplify code.
         // The progress is set to 0 at the beginning of a checkpoint, so we can be sure that the start time remembered
         // in cpWriteSpeed is pretty accurate even without writing to cpWriteSpeed from this method.
         cpWriteSpeed.setProgress(donePages, curNanoTime);
-        final long curCpWriteSpeed = cpWriteSpeed.getOpsPerSecond(curNanoTime);
+        // IDEA: use exponential moving average instead of regular moving average so that we react to changes faster?
```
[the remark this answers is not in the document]
rpuch: https://issues.apache.org/jira/browse/IGNITE-16878

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedMemoryConsumptionThrottlingStrategy.java
```diff
@@ -183,6 +182,9 @@ private long computeParkTime(@NotNull AtomicInteger writtenPagesCounter, long cu
      * @return estimation of work done (in pages)
      */
     private int cpDonePagesEstimation(int cpWrittenPages) {
+        // IDEA: this only works correctly if time-to-write a page is close to time-to-sync a page. In reality, this
```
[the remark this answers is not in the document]
rpuch: https://issues.apache.org/jira/browse/IGNITE-16879

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/IgniteThrottlingUnitTest.java
```diff
@@ -100,34 +100,51 @@ public class IgniteThrottlingUnitTest extends GridCommonAbstractTest {
      *
      */
     @Test
-    public void breakInCaseTooFast() {
+    public void shouldThrottleWhenWritingTooFast() {
```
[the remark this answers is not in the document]
rpuch: I renamed a test case to explain what should happen when. I cannot invent anything else that could be said about this test case, so writing the same thing in the comment would just create a duplication, I prefer avoiding it.

If you have an idea of what could be described in the javadoc, please let me know.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/IgniteThrottlingUnitTest.java
```diff
@@ -140,17 +157,14 @@ public void testCorrectTimeToPark() {
 
         int markDirtySpeed = 34422;
         int cpWriteSpeed = 19416;
-        long time = throttle.getCleanPagesProtectionParkTime(0.04,
+        long time = throttle.getCleanPagesProtectionParkTime(0.67,
```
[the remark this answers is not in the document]
rpuch: The old code was too pessimistic when calculating the delays, the new code does it differently, so the results are different, hence the numbers in the test are changed. Also, 0.04 is too small to trigger throttling with the new code, so I had to change it to 0.67 to have ability to verify the non-zero duration.

As for the case when we don't need to throttle at all, there is `shouldNotThrottleWhenWritingSlowly()` that verifies exactly this. It uses dirty ratio 0.47, not 0.04, but for 0.04 it would be the same.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteThrottleSandboxTest.java
```diff
@@ -172,14 +194,27 @@ public void testThrottle() throws Exception {
                 }
             }, "metrics-view");
 
+            final boolean sawForm = false;
```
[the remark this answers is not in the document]
rpuch: It is always `false`, because it is a manual sandbox, not a test for being run automatically. The user can turn it on and see how it goes in this mode :)

`saw form` is actually about the traffic having a form of a saw: for 10 seconds we add data at full speed, then for 10 seconds we don't do anything, then we repeat step 1.

I renamed the variable to `intermittentPutsMode`, does it look better, what do you think?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/IgniteThrottlingUnitTest.java
```diff
@@ -100,34 +100,51 @@ public class IgniteThrottlingUnitTest extends GridCommonAbstractTest {
      *
      */
     @Test
-    public void breakInCaseTooFast() {
+    public void shouldThrottleWhenWritingFasterThanTargetSpeedAndDirtyRatioIsAboveTargetRatio() {
```
SammyVimes, line 103: it's a bit too hard to read without spacing. Maybe leave the old name and add a clarifying comment?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/IgniteThrottlingUnitTest.java
```diff
@@ -100,34 +100,51 @@ public class IgniteThrottlingUnitTest extends GridCommonAbstractTest {
...
 
-        assertEquals(0, time);
+        assertEquals(0, parkTime);
+    }
+
+    /**
+     *
+     */
+    @Test
+    public void shouldNotThrottleWhenWritingFasterThanCPSpeedButThereAreManyCleanPages() {
```
SammyVimes, line 137: Same about long name

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteThrottleSandboxTest.java
```diff
@@ -142,25 +151,38 @@ public void testThrottle() throws Exception {
...
+                            }
                         }
                         catch (IgniteCheckedException e) {
                             e.printStackTrace();
+                            throw new RuntimeException("Something went wrong", e);
                         }
 
-                        System.out.println("@@@ putsPerSec=," + (putRate.value()) + ", getsPerSec=," + (getRate.value()) + ", dirtyPages=,"
-                            + dirtyPages + ", cpWrittenPages=," + cpWrittenPages + ", cpBufPages=," + cpBufPages);
+                        System.out.println("@@@ globalPutsPerSec="
```
SammyVimes, line 180: should be `log` instead of `System.out` probably

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedThrottleIntegrationTest.java
```diff
@@ -0,0 +1,114 @@
...
+        stopAllGrids();
+
+        super.afterTest();
+
+        cleanPersistenceDir();
+    }
+
+    /** {@inheritDoc} */
+    @Override protected long getTestTimeout() {
+        return 3 * 60 * 1000;
```
SammyVimes, line 85: 3 minutes for a test? Looks a bit odd

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedMemoryConsumptionThrottlingStrategy.java
```diff
@@ -367,11 +325,30 @@ private long calcSpeedToMarkAllSpaceTillEndOfCp(double dirtyPagesRatio, long don
...
+        double secondsTillCPEnd = 1.0 * (cpTotalPages - donePages) / avgCpWriteSpeed;
+
+        return (long)(remainedCleanPages / secondsTillCPEnd);
+    }
+
+    /** Returns total number of pages storable in page memory. */
+    private long pageMemTotalPages() {
+        long currentTotalPages = pageMemTotalPages;
+
+        if (currentTotalPages == 0) {
```
SammyVimes, line 344: When is it 0 I wonder?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedMemoryConsumptionThrottlingStrategy.java
```diff
@@ -183,6 +182,9 @@ private long computeParkTime(@NotNull AtomicInteger writtenPagesCounter, long cu
      * @return estimation of work done (in pages)
      */
     private int cpDonePagesEstimation(int cpWrittenPages) {
+        // TODO: IGNITE-16879 - this only works correctly if time-to-write a page is close to time-to-sync a page.
+        // In reality, this oes not seem to hold, which produces wrong estimations. We could measure the real times
```
SammyVimes, line 186:
```suggestion
        // In reality, this does not seem to hold, which produces wrong estimations. We could measure the real times
```

SammyVimes (changes requested):

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedThrottleIntegrationTest.java
rpuch: This is an ugly test, I admit. Locally it runs about 40 seconds. I just took a bit of room to avoid flakiness, 3 minutes seems to be not that terrible.

rpuch: What would you suggest?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedMemoryConsumptionThrottlingStrategy.java
rpuch: It is 0 in the very beginning, before it was written

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedMemoryConsumptionThrottlingStrategy.java
rpuch: Thanks, fixed this

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteThrottleSandboxTest.java
rpuch: Please note that this is not an automated test, it's a tool that is run manually. It looks like this way of writing to the console makes these lines a bit more visible for the monitoring human: they don't have that usual prefix added by the logging framework, plus they have that `@@@` as a visual anchor. If we change it to `log`, the logging prefix would mess this up, but it would not add too much value. Hence I prefer the vanilla prints.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/IgniteThrottlingUnitTest.java
rpuch: Shortened the name

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/IgniteThrottlingUnitTest.java
rpuch: Shortened the name

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteThrottleSandboxTest.java
SammyVimes: ok

SammyVimes (approved):
LGTM

PR #10024, opened 2022-05-16
IGNITE-16983 Java thin: Add AtomicLong partition awareness

ptupitsyn:
`AtomicLong` is just a cache entry underneath, where `GridCacheInternalKeyImpl` is the key. Use partition awareness to send requests to the primary node and improve performance.


isapego (approved):
Looks good to me.

PR #10045, opened 2022-05-27
IGNITE-17043 Shrink GridHandleTable on clear

SammyVimes:
https://issues.apache.org/jira/browse/IGNITE-17043

WHAT:
Automatically shrink the size of GridHandleTable on `clear` by the factor of two. Shrink only in case if element count is less than table size / 2, so we don't actually shrink table if we have a lot of big lists, for example.

WHY:
GridHandleTable grows during marshalling and never shrinks. Which leads to very long clear operations if, for example, we marshal a very big list and after that we marshal a very small list. In this case GridHandleTable will grow to the size of the very big list and subsequent clear operations will take a long time on every small list.

- [x] There is a single JIRA ticket related to the pull request. 
- [x] The web-link to the pull request is attached to the JIRA ticket.
- [x] The JIRA ticket has the _Patch Available_ state.
- [x] The pull request title is treated as the final commit message. 

### Notes

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/GridHandleTable.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/GridHandleTable.java
@@ -24,14 +24,17 @@ import java.util.Arrays;
  * assigned in ascending order.
  */
 public class GridHandleTable {
+    /** Initial size. */
+    private final int initCap;
+
+    /** Factor for computing size threshold. */
+    private final float loadFactor;
+
     /** Number of mappings in table/next available handle. */
     private int size;
 
     /** Size threshold determining when to expand hash spine. */
     private int threshold;
-
-    /** Factor for computing size threshold. */
-    private final float loadFactor;
 
     /** Maps hash value -> candidate handle value. */
     private int[] spine;
@@ -55,20 +58,30 @@ public class GridHandleTable {
      * @param loadFactor Load factor.
      */
     public GridHandleTable(int initCap, float loadFactor) {
+        this.initCap = initCap;
         this.loadFactor = loadFactor;
 
-        spine = new int[initCap];
-        next = new int[initCap];
-        objs = new Object[initCap];
-        spineEmpty = new int[initCap];
-        nextEmpty = new int[initCap];
+        init(initCap);
+
+        clear();
+    }
+
+    /**
+     * Initialize hash table.
+     *
+     * @param len Hash table length.
+     */
+    private void init(int len) {
+        spine = new int[len];
+        next = new int[len];
+        objs = new Object[len];
+        spineEmpty = new int[len];
+        nextEmpty = new int[len];
 
         Arrays.fill(spineEmpty, -1);
         Arrays.fill(nextEmpty, -1);
 
-        threshold = (int)(initCap * loadFactor);
-
-        clear();
+        threshold = (int)(len * loadFactor);
     }
 
     /**
@@ -107,6 +120,10 @@ public class GridHandleTable {
      * Resets table to its initial (empty) state.
      */
     public void clear() {
+        if (size < objs.length) {
+            shrink();
+        }
+
         System.arraycopy(spineEmpty, 0, spine, 0, spineEmpty.length);
         System.arraycopy(nextEmpty, 0, next, 0, nextEmpty.length);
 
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/internal/util/GridHandleTable.java
```diff
@@ -107,6 +120,10 @@ public int lookup(Object obj) {
      * Resets table to its initial (empty) state.
      */
     public void clear() {
+        if (size < objs.length) {
+            shrink();
```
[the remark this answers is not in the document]
SammyVimes: We still have to perform the rest of the "clear", because it's basically the same as in constructor + potentially we might not shrink at all

modules/core/src/main/java/org/apache/ignite/internal/util/GridHandleTable.java
```diff
@@ -107,6 +121,14 @@ public int lookup(Object obj) {
...
+        if (size < objs.length) {
+            if (shrink()) {
+                size = 0;
+
+                return;
+            }
+        }
+
         System.arraycopy(spineEmpty, 0, spine, 0, spineEmpty.length);
         System.arraycopy(nextEmpty, 0, next, 0, nextEmpty.length);
```
macrergate, line 133: How about clearing not all nextEmpty and objs array, but only first size elements?

macrergate, line 133:
```suggestion
        just some more thoughts: 
-         I guess that the array 'next' can be not cleaned at all as we  are resetting 'size' to zero and we never read the next array elements with index>=size

-       we could consider zero value as empty not -1, if we would change postitioning to 'int idx = (hash(obj) % spine.length) + 1; in this case  we wouldn't need to fill int arrays in init() method as they have zeroes by default.
        
```

modules/core/src/main/java/org/apache/ignite/internal/util/GridHandleTable.java
```diff
@@ -181,6 +203,32 @@ private void growEntries() {
...
+    /**
+     * Tries to gradually shrink hash table by factor of two when it's cleared.
+     *
+     * @return {@code true} if shrinked the table, {@code false} otherwise.
+     */
+    private boolean shrink() {
+        int newLen = objs.length;
+
+        if (newLen > initCap) {
+            int shrinked = (newLen - 1) / 2;
```
sashapolo, line 197: `shirnked` is incorrect, should be `shrank`

sashapolo, line 197: why `newLen - 1`?

sashapolo: I think this code can actually be written a little bit simpler:
```
int newLen = Math.max(objs.length / 2, initCap);

if (newLen >= size && newLen < objs.length) {
    init(newLen);

    return true;
}

return false;
```

SammyVimes: Because we grow it as "(size  << 1) + 1"

sashapolo: And why do we do that?

modules/core/src/main/java/org/apache/ignite/internal/util/GridHandleTable.java
SammyVimes: Sounds legit

SammyVimes: I like the first suggestion, but the second one will lead to unnecessary complication (because `size` will be affected by this change of indexing) and filling operation shouldn't be very frequent with the fix.

SammyVimes: On the other hand, objs should be cleared entirely to avoid referencing objects that are no longer needed

macrergate: I agree with you, let's ignore the second one, it's arguable.

modules/core/src/main/java/org/apache/ignite/internal/util/GridHandleTable.java
```diff
@@ -107,9 +110,15 @@ public int lookup(Object obj) {
...
+        if (size < objs.length) {
+            if (shrink()) {
+                size = 0;
+
+                return;
+            }
+        }
 
+        Arrays.fill(spine, -1);
         Arrays.fill(objs, null);
```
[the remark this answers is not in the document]
SammyVimes: You're right, fixed

modules/core/src/main/java/org/apache/ignite/internal/util/GridHandleTable.java
macrergate: 'Objs' array is filling from [0 to size), all the rest elements are always null.

SammyVimes: Yes, you're right, I had a little mixup)

ibessonov (approved):

PR #10105, opened 2022-06-23
IGNITE-17215: Write ClusterSnapshotRecord to WAL

timoninmaxim:
For PITR [1] process of recovering based on ClusterSnapshot + archived WALs.

It's required to have a point in WAL which splits whole WAL on 2 areas:
1. Before this point all data changes are contained within ClusterSnapshot, and no need to recover them from WAL archived files.
2. After this point all data need to be recovered from WAL archived files.

It's proposed to write ClusterSnapshotRecord while the checkpoint is running (cp#writeLock has acquired). ClusterSnapshot process guarantees:
1. there is no active transactions (or any data changes) in moment of running checkpoint.
2. ClusterSnapshot contains all data pages that will be persisted within this checkpoint process.

Then every logical record after begin CheckointRecord doesn't belong to ClusterSnapshot. Then it's safe to write ClusterSnapshotRecord within the checkpoint process.

[1] https://cwiki.apache.org/confluence/pages/editpage.action?pageId=211884314

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotFutureTask.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotFutureTask.java
@@ -343,8 +343,14 @@ class SnapshotFutureTask extends AbstractSnapshotFutureTask<Set<GroupPartitionId
             return;
 
         try {
-            if (cctx.wal() != null)
+            // Here we have the following warranties:
+            // 1. Checkpoint holds write acquire lock and Snapshot holds PME. Then there are not any concurrent updates.
+            // 2. This record is written before the related CheckpointRecord, and is flushed with CheckpointRecord or instead it.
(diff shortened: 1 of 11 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotFutureTask.java
```diff
@@ -342,6 +343,9 @@ private boolean stopping() {
             return;
 
         try {
+            if (cctx.wal() != null)
+                cctx.wal().log(new ClusterSnapshotRecord(snpName));
```
NSAmelchev, line 347: Is WAL will be flushed after checkpoint write lock?

timoninmaxim: Fixed.

NSAmelchev (approved):

PR #10117, opened 2022-06-28
IGNITE-17250 : Calcite. Make 'min()/max()' use first/last index value.

Vladsz83:


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/IndexMinMaxRule.java
```diff
@@ -0,0 +1,107 @@
...
+                aggr.getAggCallList().stream().filter(a -> a.getAggregation().getKind() == SqlKind.MIN
+                    || a.getAggregation().getKind() == SqlKind.MAX).count() != 1 ||
+                !idx.fields().get(0).equals(idxScan.getRowType().getFieldList().get(0).getName()))
+            return;
+
+        RexBuilder rexb = RexUtils.builder(idxScan.getCluster());
+        SqlAggFunction aggFun = aggr.getAggCallList().get(0).getAggregation();
+        boolean descending = idx.collation().getFieldCollations().get(0).getDirection().isDescending();
+
+        IndexConditions idxConditions = RexUtils.buildSortedIndexConditions(
```
alex-plekhanov, line 71: It's not a good idea to use IndexCondition as some label for index scan, lets create new relational operator instead.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/IndexMinMaxRule.java
```diff
@@ -0,0 +1,107 @@
...
+        IgniteIndex idx = table.getIndex(idxScan.indexName());
+
+        if (
+            table.isIndexRebuildInProgress() ||
+                idxScan.condition() != null ||
+                aggr.getGroupCount() > 0 ||
+                idx.collation().getFieldCollations().isEmpty() ||
+                aggr.getAggCallList().stream().filter(a -> a.getAggregation().getKind() == SqlKind.MIN
+                    || a.getAggregation().getKind() == SqlKind.MAX).count() != 1 ||
+                !idx.fields().get(0).equals(idxScan.getRowType().getFieldList().get(0).getName()))
```
alex-plekhanov, line 64: Strange and fragile condition. Let's use traits collation instead. idx.fields() method is also redundant.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/IndexScan.java
```diff
@@ -159,6 +162,47 @@ public IndexScan(
...
+    public List<Row> firstOrLast(boolean first, boolean notNull) {
+        reserve();
+
+        IndexingQueryFilter filter = new IndexingQueryFilterImpl(kctx, topVer, parts);
+        BPlusTree.TreeRowClosure<IndexRow, IndexRow> rowFilter = notNull ?
+            new BPlusTree.TreeRowClosure<IndexRow, IndexRow>() {
+                /** {@inheritDoc} */
+                @Override public boolean apply(BPlusTree<IndexRow, IndexRow> tree, BPlusIO<IndexRow> io, long pageAddr,
+                    int idx) throws IgniteCheckedException {
+                    return !F.isEmptyOrNulls(indexRow2Row(io.getLookupRow(tree, pageAddr, idx)));
```
alex-plekhanov, line 179: It's not correct to use some general isEmptyOrNulls, since Row it's an some abstract type, it can be anything (currently it's only array, but perhaps in some time in future there will be BinaryObject row available). Let's just check index row (indexRow.key(0).type = IndexKeyTypes.NULL). Also looks like notNull is always true, so this parameter is not required.

Vladsz83: Fixed

PR #10168, opened 2022-07-25
IGNITE-17408 Ensure buffer before read plain record type

timoninmaxim:
Ensure that buffer has enough remaining bytes for extracting plain record for encrypted records.

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotWalRecordTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotWalRecordTest.java
@@ -1,0 +1,192 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


NSAmelchev (approved):

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotWalRecordTest.java
```diff
@@ -0,0 +1,186 @@
...
+import org.apache.ignite.internal.processors.cache.persistence.wal.reader.IgniteWalIteratorFactory;
+import org.apache.ignite.internal.util.typedef.F;
+import org.apache.ignite.internal.util.typedef.T2;
+import org.apache.ignite.internal.util.typedef.internal.U;
+import org.apache.ignite.lang.IgniteBiTuple;
+import org.apache.ignite.testframework.GridTestUtils;
+import org.junit.Test;
+
+/** */
+public class IgniteClusterSnapshotWalRecordTest extends AbstractSnapshotSelfTest {
```
sashapolo, line 49: How are this test and other changes related to the bug described in the ticket?

timoninmaxim: @sashapolo hi! This test describes a scenario in that the bug is reproduced. It fails on reading WAL files. It's required for testing the ClusterSnapshotRecord, that should be written to WAL at the correct moment. The test-case `IgniteSnapshotManagerSelfTest.java#testSnapshotLocalPartitionMultiCpWithLoad` tests this, but it was very hard to debug why it was failing. Then I moved failure prone part of the test to the separate class to test it in more isolated case.

sashapolo: thank you for the explanation

sashapolo (approved):

PR #10195, opened 2022-08-11
IGNITE-14914 Add IndexQuery IN criterion

timoninmaxim:
This patch brings IN criterion to IndexQuery:

`IndexQuery.setCriteria(in("A", Arrays.asList(1, 2 ,3)));`

1. IN criterion accepts collection of values to find. This collections is transformed to `SortedSet(1, 2, 3)` because IndexQuery provides to user sorted result.
2. When IN applies on first indexed field - IN(A0, A1) for index (A, B) - it converts to multiple `eq` operations are joint with OR operation: `EQ(A0) or EQ(A1)`. 
3. Other range criteria for other fields are applied to every such separate operation: 
`IN(A0, A1) and GT(B)` converts to `(EQ(A0) and GT(B)) or (EQ(A1) and GT(B))`.
4. When IN applies to non-leading indexed field - IN(B0, B1) for index (A, B) - it works like a filter for prepared range:
`GTE(A) and IN(B0, B1)` converts to range `[[A, B0]; [A, B1]]` and every cache entry within this range is checked for being included to SortedSet(B0, B1).


modules/core/src/main/java/org/apache/ignite/internal/cache/query/index/IndexQueryProcessor.java
```diff
@@ -615,14 +503,22 @@ private boolean rowIsOutOfRange(
...
+                Set<IndexKey> inVals = qry.keyCond[keyIdx].inVals();
 
                 InlineIndexKeyType keyType = keyIdx < keyTypes.size() ? keyTypes.get(keyIdx) : null;
 
                 boolean descOrder = descOrderCache[keyIdx];
 
                 int maxSize = tree.inlineSize() - fieldOff;
 
+                if (inVals != null) {
+                    IndexRow row = io.getLookupRow(tree, pageAddr, idx);
```
tledkov, line 516: What do you think about using inline payload data if it is available for check IN condition?

modules/core/src/main/java/org/apache/ignite/internal/cache/query/index/IndexQueryProcessor.java
```diff
@@ -538,25 +442,12 @@ private boolean inclBoundary(IndexRangeQuery qry, boolean lower) {
...
-        if (val != null || isNull)
-            key = IndexKeyFactory.wrap(val, keyType, coctx, settings);
-
-        return key;
-    }
-
     /**
      * Checks index rows for matching to specified index criteria.
      */
     private static class IndexQueryCriteriaClosure implements BPlusTree.TreeRowClosure<IndexRow, IndexRow> {
```
tledkov, line 448: Maybe we should move inner classes to the top level?

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientCache.java
```diff
@@ -989,6 +990,17 @@ private QueryCursor<Cache.Entry<K, V>> indexQuery(IndexQuery<K, V> qry) {
                             serDes.writeObject(out, range.lower());
                             serDes.writeObject(out, range.upper());
                         }
+                        else if (c instanceof InIndexQueryCriterion) {
```
tledkov, line 993: Please add support backward compatibility for the thin client protocol:
see: `ClientBitmaskFeature`, `ClientProtocolContext`.
For new compatibility tests see: to `JavaThinCompatibilityTest`

tledkov (approved):
OK with me.

PR #10214, opened 2022-08-24
IGNITE-17299 .NET: Fix StringComparer serialization

ptupitsyn:
* Handle special case when deserialization constructor is missing.
* Add tests for all documented serializable system types.


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Binary/SerializableSerializer.cs
```diff
@@ -603,13 +597,33 @@ private static object ReadAsCustomType(Type customType, SerializationInfo serInf
...
+            var resObj = wrapper == null
                 ? customObj
                 : wrapper.GetRealObject(ctx);
+
+            // Special case: type is replaced, but there is no serialization ctor.
+            // Example: StringComparer.OrdinalIgnoreCase.
+            if (ctorFunc == null)
+            {
+                // Cached internally.
+                var members = FormatterServices.GetSerializableMembers(resObj.GetType());
```
isapego, line 615: What about non-serializable members? What if non-serializable method greatly affects object functionality?

isapego (approved):

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Binary/SerializableSerializer.cs
ptupitsyn: `SerializableSerializer` handles `ISerializable` types only. If a member of such type is marked with `[NonSerialized]`, it is skipped during serialization. This is what `GetSerializableMembers` does - returns all members that do not have `[NonSerialized]` attribute.

We are just following the same logic as the built-in binary serializer has.

isapego: I see now

PR #10262, opened 2022-09-21
IGNITE-17735 : Datastreamer may consume whole heap.

Vladsz83:
Limited max parallel batches for a node. The limitation now looks at cache persistence.


modules/core/src/main/java/org/apache/ignite/stream/StreamReceiver.java
```diff
@@ -41,4 +43,18 @@
...
+     * Sets max parallel batches (operations) per node if {@link IgniteDataStreamer#perNodeParallelOperations()} is not
+     * positive. In same way takes place only if positive.
+     *
+     * @param node       Node to send to.
+     * @param persistant {@code True} if the cache is persistent. {@code False} otherwise.
+     *
+     * @return Max parallel batches.
+     * @see IgniteDataStreamer#perNodeParallelOperations(int)
+     */
+    public default int perNodeParallelOperations(ClusterNode node, boolean persistant) {
```
ivandasch, line 57: persist*e*nt

Vladsz83: Fixed. In other places too.

Vladsz83:
Other, simplier impls: https://github.com/apache/ignite/pull/10343 or https://github.com/apache/ignite/pull/10351

PR #10299, opened 2022-10-10
IGNITE-17369 : v4, Check Datastremer registered at and before snapshot.

Vladsz83:


Vladsz83:
Discussed and decided not to ask and wait for the client nodes.

PR #10346, opened 2022-10-25
Ignite-17369 : v6, full snapshot check

Vladsz83:


Vladsz83:
Discussed and decided not to do full snapshot check each time.

PR #10360, opened 2022-11-02
IGNITE-17723 Added snapshot delta pages sort

NSAmelchev:
Added snapshot delta pages sort to almost sequential disk writes on apply


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -394,6 +403,13 @@ public class IgniteSnapshotManager extends GridCacheSharedManagerAdapter
     /** Snapshot transfer rate limit in bytes/sec. */
     private final DistributedLongProperty snapshotTransferRate = detachedLongProperty(SNAPSHOT_TRANSFER_RATE_DMS_KEY);
 
+    /**
+     * Snapshot delta sort batch size in pages count. If set then delta pages will be indexed by page index to almost
+     * sequential disk write on apply. This generates an extra delta read. If value is non-positive or not set then delta
+     * pages will be applied directly.
+     */
+    private final DistributedIntegerProperty deltaSortBatch = detachedIntegerProperty(SNAPSHOT_DELTA_SORT_BATCH_SIZE_KEY);
```
Mmuzaf, line 411: I'd rather avoid exposing some of the snapshot internal process things like "writing delta" or so on to a user.

Actually, I think it would be more friendly for a user to have a snapshot configuration property of a snapshot write mode `RANDOM` or `SEQUENTIAL`. The batch must be predefined by the system.

We don't even need to change this parameter at runtime.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -3254,19 +3271,73 @@ public LocalSnapshotSender(String snpName, @Nullable String snpPath) {
...
+
+                        @Override public int next() {
+                            return sortedIter.next();
+                        }
+
+                        private void advance() throws Exception {
+                            TreeMap<Integer, Integer> sorted = new TreeMap<>();
+
+                            while (idx < pagesCnt && sorted.size() < batchSize) {
+                                transferRateLimiter.acquire(pageSize);
```
Mmuzaf, line 3309: We are limiting only writes, why we should use the same for reads?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
```diff
@@ -3254,19 +3271,73 @@ public LocalSnapshotSender(String snpName, @Nullable String snpPath) {
...
+                                catch (Exception e) {
+                                    throw new IgniteException(e);
+                                }
+                            }
+
+                            return sortedIter.hasNext();
+                        }
+
+                        @Override public int next() {
+                            return sortedIter.next();
```
Mmuzaf, line 3302: It will produce NullPointerExcepction here if we call the `next` first time. It seems it should be IllegalStateException or something similar.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotDeltaTest.java
```diff
@@ -0,0 +1,217 @@
...
+            .setAffinity(new RendezvousAffinityFunction(false, partCnt));
+
+        String cacheDir = CACHE_DIR_PREFIX + DEFAULT_CACHE_NAME;
+
+        IgniteEx srv = startGridsWithCache(1, keys, (k) -> expPayload, ccfg);
+
+        if (sorted) {
+            setDeltaSortBatchSize(batchSize);
+
+            injectSequentialWriteCheck(srv, batchSize);
```
Mmuzaf, line 93: Why this factory can't be used as a JUnit test param?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
Mmuzaf, line 411: This will not work if encryption is enabled, right?

Mmuzaf (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
NSAmelchev: > We don't even need to change this parameter at runtime.

The snapshot path is set on each snapshot create process. I think we should allow users to set a mode per hardware. WDYT?

NSAmelchev: Encryption works by pages, so ordering will work.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteClusterSnapshotDeltaTest.java
NSAmelchev: Could you explain, please?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
NSAmelchev: Added check.

docs/_docs/snapshots/snapshots.adoc
```diff
@@ -95,6 +95,19 @@ increases the total amount of time for taking a snapshot. However, this keeps th
...
+=== Distributed properties
+
+The distributed properties listed in the table below allow you to configure snapshots at runtime:
+
+[cols="1,3,1",opts="header"]
+|===
+|Parameter | Description | Default Value
+|`snapshotDeltaSortBatchSize`| Snapshot delta sort batch size in pages count. If set then delta pages will be indexed
+by page index to almost sequential disk write on apply. This generates an extra delta read. If value is non-positive or
+not set then delta pages will be applied directly. | 0
```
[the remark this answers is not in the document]
NSAmelchev: I have refactored property to a sequential flag: `snapshotSequentialWrite`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
NSAmelchev: I have refactored property to a sequential flag: `snapshotSequentialWrite`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/IgniteSnapshotManager.java
NSAmelchev: I assume this is a total throttling limit to a disk to reduce snapshot process affect.

docs/_docs/snapshots/snapshots.adoc
```diff
@@ -95,6 +95,16 @@ increases the total amount of time for taking a snapshot. However, this keeps th
 
 See the link:perf-and-troubleshooting/thread-pools-tuning[Ignite Snapshot Execution Pool,window=_blank] page for more details.
 
+=== Distributed properties
+
+The distributed properties listed in the table below allow you to configure snapshots at runtime:
```
Mmuzaf, line 100: Please, reflect in this documentation changes made in the source code.

NSAmelchev: Done.

Mmuzaf (approved):

PR #10387, opened 2022-11-18
IGNITE-18191 .NET: Fix startup on Java 15+

ptupitsyn:
* Add JVM arguments for Java 15+
* Fix `ClientServerCompatibilityTest` 
  * Skip old Ignite versions when new JDK is detected.
  * Fix error handling in `PlatformProcessUtils.startProcess`

NOTE: The following warning may appear on startup with JDK 17+: `OpenJDK 64-Bit Server VM warning: Ignoring option --illegal-access=permit; support was removed in 17.0`. It is harmless.

Unfortunately, with JNI there is no way to tell apart Java versions above 10: https://docs.oracle.com/en/java/javase/17/docs/specs/jni/functions.html#getversion. So we have to use a universal set of JVM args for everything Java 9+, which may produce warnings.

### DIFF
--- a/modules/platforms/dotnet/Apache.Ignite.Core.Tests/TestUtils.cs
+++ b/modules/platforms/dotnet/Apache.Ignite.Core.Tests/TestUtils.cs
@@ -625,7 +625,7 @@ namespace Apache.Ignite.Core.Tests
                 Logger = noLogger ? null : new TestContextLogger(),
 
                 // TODO: Revert me
-                // JvmDllPath = "/usr/lib/jvm/java-11-openjdk-amd64/lib/server/libjvm.so"
+                JvmDllPath = "/usr/lib/jvm/java-11-openjdk-amd64/lib/server/libjvm.so"
             };
         }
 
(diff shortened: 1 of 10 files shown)


modules/platforms/dotnet/Apache.Ignite.Core.Tests/TestUtils.cs
```diff
@@ -77,8 +77,8 @@ public static class TestUtils
             ? new List<string>
             {
                 "-XX:+HeapDumpOnOutOfMemoryError",
-                "-Xms2g",
-                "-Xmx4g",
+                "-Xms4g",
+                "-Xmx7g",
```
isapego, line 81: Why do we increase default memory limits in test?

isapego (approved):

modules/platforms/dotnet/Apache.Ignite.Core.Tests/TestUtils.cs
ptupitsyn: I've seen some intermittent failures caused by OOM on Java side while running tests. This is an attempt to improve stability.  If not, we should investigate increased memory usage. I'm keeping an eye on it.

PR #10410, opened 2022-11-30
IGNITE-15737 IoStatisticsSelfTest#testPersistentIOGlobalStat failure.

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-15737

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/metric/IoStatisticsSelfTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/metric/IoStatisticsSelfTest.java
@@ -30,9 +30,7 @@ import org.apache.ignite.internal.processors.metric.GridMetricManager;
 import org.apache.ignite.internal.processors.metric.MetricRegistry;
 import org.apache.ignite.spi.metric.LongMetric;
 import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
-import org.hamcrest.Matchers;
 import org.jetbrains.annotations.Nullable;
-import org.junit.Assert;
 import org.junit.Test;
 
 import static org.apache.ignite.internal.metric.IoStatisticsCacheSelfTest.logicalReads;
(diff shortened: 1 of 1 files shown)


modules/core/src/test/java/org/apache/ignite/internal/metric/IoStatisticsSelfTest.java
```diff
@@ -145,28 +151,26 @@ private void ioStatGlobalPageTrackTest(boolean isPersistent) throws Exception {
         long physicalReadsCnt = physicalReads(mmgr, CACHE_GROUP, DEFAULT_CACHE_NAME, null);
 
         if (isPersistent)
-            Assert.assertTrue(physicalReadsCnt > 0);
+            Assert.assertThat(physicalReadsCnt, Matchers.greaterThan(0L));
```
rpuch, line 154: If we import `assertThat` and `greaterThan` statically, then the assertion reads as a plain text (which is probably good for readability): 'assert that phisicalReadsCnt greater than 0`

modules/core/src/test/java/org/apache/ignite/internal/metric/IoStatisticsSelfTest.java
```diff
@@ -190,9 +194,9 @@ private IgniteConfiguration getConfiguration(boolean isPersist) throws Exception
             DataStorageConfiguration dsCfg = new DataStorageConfiguration()
                 .setDefaultDataRegionConfiguration(
                     new DataRegionConfiguration()
-                        .setMaxSize(30L * 1024 * 1024)
-                        .setPersistenceEnabled(true))
-                .setWalMode(WALMode.LOG_ONLY);
+                        .setMaxSize(25 * MB)
```
rpuch, line 197: Looks like this is THE change that fixes test flakiness. Looks like 30Mb produces a flaky test, but 25Mb does not. Could you please add a comment explaining why this value of 25Mb was chosen (is there a calculation that demonstrates that a test will be stable with 25Mb, or it's empirically chosen)?

modules/core/src/test/java/org/apache/ignite/internal/metric/IoStatisticsSelfTest.java
tkalkirill: Fix it

modules/core/src/test/java/org/apache/ignite/internal/metric/IoStatisticsSelfTest.java
tkalkirill: Tried to fix it

rpuch (approved):

PR #10623, opened 2023-04-05
IGNITE-19221 Java thin: Add ClientConfiguration.clusterDiscoveryEnabled

ptupitsyn:
Cluster discovery is always enabled currently, which can be undesirable in some use cases. Add a config property to disable discovery.


alex-plekhanov:
Let's fix also AbstractThinClientTest.getClientConfiguration(). Here static address finder is used to disable discovery, now we can use property (also, perhaps ReliabilityTest should be fixed too, where StaticAddressFinder is used and StaticAddressFinder should be removed).

alex-plekhanov (approved):

PR #10858, opened 2023-07-27
IGNITE-20061 Add SystemProperty for H2Connection#STATEMENT_CACHE_SIZE 

asdfgh19:
JIRA: [IGNITE-20061](https://issues.apache.org/jira/browse/IGNITE-20061)

When the memory is sufficient, we hope to increase the size of H2StatementCache to improve performance;

When the memory is limited, we hope to reduce the size of H2StatementCache to avoid memory overflow;

So I propose to add a system property IGNITE_H2_INDEXING_CACHE_STATEMENT_CACHE_SIZE to configure the H2StatementCache size.


asdfgh19:
@ptupitsyn Can you help to review this pr? Thanks

PR #10891, opened 2023-08-13
IGNITE-20135 : NO_INDEX hint for calcite engine v2.

Vladsz83:


Vladsz83:
Implemented in #10902

PR #10955, opened 2023-09-23
IGNITE-20481 Java client: Balance requests across connections

weetyre:

JIRA ticket:
https://issues.apache.org/jira/browse/IGNITE-20481

Description:

The ignite 2.15.0 version does not change the default channel for processing when the client is connected normally, resulting in load imbalance performance problems between ignite nodes. This modification adds a round-robin method to achieve the purpose of using a new available channel for each request to achieve complex balancing.


modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
```diff
@@ -778,29 +781,27 @@ private <T> T applyOnDefaultChannel(
                 if (closed)
                     throw new ClientException("Channel is closed");
 
+
```
Shoothzj, line 784: unnecessary change

weetyre: good catch! I have fixed it!

weetyre:
@ptupitsyn @petrov-mg @ivan-fedorenkov @sk0x50  thanks in advance. :)

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
```diff
@@ -781,26 +784,24 @@ private <T> T applyOnDefaultChannel(
...
+                            break;
                         }
-                        while (hld.ch == null && idx != idx0);
                     }
                 }
                 finally {
                     curChannelsGuard.readLock().unlock();
                 }
 
+                curChannelsGuard.readLock().lock();
```
ptupitsyn, line 803: 1. Do we really need this?
2. Corresponding unlock seems to be missing.

ptupitsyn (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
```diff
@@ -70,8 +70,11 @@ final class ReliableChannel implements AutoCloseable {
     /** Limit of attempts to execute each service. */
     private volatile int attemptsLimit;
 
-    /** Index of the current channel. */
-    private volatile int curChIdx = -1;
+    /** Index of the current channel (for round-robin balancing). */
+    private final AtomicInteger curChIdx = new AtomicInteger();
+
+    /** Index of the default channel. */
+    private volatile int defaultChIdx = -1;
```
Shoothzj, line 77: If this variable is changed every time a request is made, the name `defaultChIdx` is not appropriate.

weetyre:
@ptupitsyn @petrov-mg @ivan-fedorenkov @sk0x50 thanks in advance again. :)

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ReliableChannel.java
```diff
@@ -487,10 +490,10 @@ private void rollCurrentChannel(ClientChannelHolder hld) {
      * On current channel failure.
      */
     private void onChannelFailure(ClientChannel ch, Throwable t, @Nullable List<ClientConnectionException> failures) {
-        // There is nothing wrong if curChIdx was concurrently changed, since channel was closed by another thread
+        // There is nothing wrong if defaultChIdx was concurrently changed, since channel was closed by another thread
         // when current index was changed and no other wrong channel will be closed by current thread because
         // onChannelFailure checks channel binded to the holder before closing it.
-        onChannelFailure(channels.get(curChIdx), ch, t, failures);
+        onChannelFailure(channels.get(lastDefaultChIdx), ch, t, failures);
```
ptupitsyn, line 496: This won't work properly anymore, since we use a new index for every operation.

`channels.get(lastDefaultChIdx)` won't match `ch` in most cases, so `hld.closeChannel();` won't be called inside `onChannelFailure`.

I think we should rework failure scenario a bit:
* Get rid of `lastDefaultChIdx`, new `curChIdx` is enough
* Remove `ClientChannel ch` parameter from `onChannelFailure`. Use `nodeChannels` to resolve `ClientChannelHolder` from `ClientChannel`.

ptupitsyn (changes requested):
Please add a test to verify new functionality.

alex-plekhanov:
@weetyre, what problem are you trying to solve with this patch? 
When partition awareness is enabled connections are already balanced. Only reason to have partition awareness disabled is to have only one connection per client, so no balancing required here. Your patch breaks this behaviour (perhaps additionaly breaks other minor things).

PR #11016, opened 2023-10-26
IGNITE-20740 Fix SocketChannel leak when connection fails

shoothzj:
https://issues.apache.org/jira/browse/IGNITE-20740

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/client/thin/io/gridnioserver/GridNioClientConnectionMultiplexer.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/client/thin/io/gridnioserver/GridNioClientConnectionMultiplexer.java
@@ -146,8 +146,30 @@ public class GridNioClientConnectionMultiplexer implements ClientConnectionMulti
         rwLock.readLock().lock();
 
         try {
-            SocketChannel ch = SocketChannel.open();
-            ch.socket().connect(new InetSocketAddress(addr.getHostName(), addr.getPort()), connTimeout);
+            SocketChannel ch = null;
+            try {
+                ch = SocketChannel.open();
+                ch.socket().connect(new InetSocketAddress(addr.getHostName(), addr.getPort()), connTimeout);
+            }
+            catch (Exception e) {
+                if (ch != null) {
+                    if (ch.socket() != null) {
+                        try {
+                            ch.socket().close();
+                        }
+                        catch (Exception ex) {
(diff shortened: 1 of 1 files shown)


shoothzj:
@ptupitsyn @petrov-mg @ivan-fedorenkov @sk0x50  PTAL, thanks

modules/core/src/main/java/org/apache/ignite/internal/client/thin/io/gridnioserver/GridNioClientConnectionMultiplexer.java
```diff
@@ -146,8 +146,30 @@ public GridNioClientConnectionMultiplexer(ClientConfiguration cfg) {
...
+                            throw new ClientConnectionException(ex.getMessage(), ex);
+                        }
+                    }
+                    try {
+                        ch.close();
+                    }
+                    catch (Exception ex) {
+                        throw new ClientConnectionException(ex.getMessage(), ex);
+                    }
+                }
```
ptupitsyn, line 170:
```suggestion
                if (ch != null) {
                    if (ch.socket() != null) {
                        try {
                            ch.socket().close();
                        }
                        catch (Exception ignored) {
                        }
                    }
                    
                    try {
                        ch.close();
                    }
                    catch (Exception ignored) {
                    }
                }
```

The idea is to throw the original exception to the user. We don't care about exceptions that happen int `close()`.

ptupitsyn (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/client/thin/io/gridnioserver/GridNioClientConnectionMultiplexer.java
shoothzj: I agree. Can we do some log here?

shoothzj:
@ptupitsyn Thanks for your review, Could you please review again? Thanks.

ptupitsyn (approved):

ptupitsyn:
TC build triggered: https://ci.ignite.apache.org/viewLog.html?buildId=7591435&

shoothzj:
@ptupitsyn Can this pr being merged? Thanks :)

ptupitsyn:
> @ptupitsyn Can this pr being merged? Thanks :)

Yes, once we have a good TeamCity run. I'll deal with this, don't worry.

PR #11060, opened 2023-11-21
IGNITE-20904 Fix Messaging SystemView code.

CyrilFeng:
GridMessageListenHandler.cacheName() caused an IllegalStateException when querying the SystemView with SQL. I made it return null, and I adjusted the GridMessageListenHandler.orderedTopic() method so that the SystemView can return topicName.


nizhikov:
Hello @CyrilFeng 

Thanks for PR.


Can you, please, add tests to `SystemViewSelfTest` to test fixed issue?
Looks like you can extend `testContinuousQuery` or take it as a pattern for the new one.

PR #11107, opened 2023-12-18
IGNITE-20949 .NET: Thin client: Add service deployment awareness

Vladsz83:
To improve service invocation performance, send client requests directly to the nodes where the service is deployed.


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Services/ServicesClient.cs
```diff
@@ -35,6 +39,9 @@ namespace Apache.Ignite.Core.Impl.Client.Services
     /// </summary>
     internal class ServicesClient : IServicesClient
     {
+        /** Max service topology update period in mills. */
```
ptupitsyn, line 42:
```suggestion
        /** Max service topology update period in millis. */
```

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Services/ServicesClient.cs
```diff
@@ -202,7 +215,123 @@ public IServicesClient WithServerKeepBinary()
...
+
+                        return filteredTopology;
+                    }, (status, err) =>
+                    {
+                        Interlocked.Exchange(ref _updateInProgress, 0);
+
+                        log.Error("Failed to update topology of the service '" + _srvcName + "'.", err);
+
+                        return _nodes;
+                    }).Result;
```
ptupitsyn, line 310: Don't use `.Result` - it blocks the thread to wait for the Task. Prefer `await`.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Services/ServicesClient.cs
```diff
@@ -202,7 +215,123 @@ public IServicesClient WithServerKeepBinary()
...
+            /// Provides last known service topology and asynchronously updates it if required.
+            /// </summary>
+            internal IList<Guid> GetAndUpdate()
+            {
+                long lastKnownAff = Interlocked.Read(ref _lastAffTop);
+                long curAff = _srvc._ignite.Socket.GetTopologyVersion();
+                long lastUpdateTime = Interlocked.Read(ref _lastUpdateRequestTime);
+                
+                if(lastKnownAff == 0 || curAff > lastKnownAff || DateTime.Now.Ticks - lastUpdateTime >= SrvTopUpdatePeriod)
+                    UpdateTopologyAsync();
```
ptupitsyn, line 331: What's the idea - do we want to update in background and return current nodes meanwhile?

ptupitsyn (changes requested):

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Services/ServicesClient.cs
Vladsz83: Yes. Just like in Java Thin. The same.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Services/ServicesClient.cs
Vladsz83: Yes, forgot the async

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Services/ServicesClient.cs
```diff
@@ -202,7 +216,123 @@ public IServicesClient WithServerKeepBinary()
...
+            internal ServiceTopology(string name, ServicesClient svcClient)
+            {
+                _svcName = name;
+                _svcClient = svcClient;
+            }
+
+            /// <summary>
+            /// Asynchronously updates the service topology.
+            /// </summary>
+            private async void UpdateTopologyAsync()
```
ptupitsyn, line 268:
```suggestion
            private async Task UpdateTopologyAsync()
```

`async void` is not recommended:
* https://learn.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming
* https://learn.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Services/ServicesClient.cs
```diff
@@ -202,7 +216,123 @@ public IServicesClient WithServerKeepBinary()
...
+                        Interlocked.Exchange(ref _lastUpdateRequestTime, DateTime.Now.Ticks);
+                        Interlocked.Exchange(ref _lastAffTop, topVer);
+                        Interlocked.Exchange(ref _updateInProgress, 0);
+                        
+                        var filteredTopology = FilterTopology(res, groupNodes?.Select(n => n.Id).ToList());
+                        
+                        log.Debug("Topology of service '" + _svcName + "' has been updated. The " +
+                                  "service instance nodes: " + string.Join(", ", res.Select(gid=>gid.ToString())) +
+                                  ". Effective topology with the cluster group is: " + 
+                                  string.Join(", ", filteredTopology.Select(gid=>gid.ToString())) + '.');
```
ptupitsyn, line 301: At this point we haven't yet updated `_nodes`. Let's move those lines to the end of `UpdateTopologyAsync` method, since we have `await` now.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Services/ServicesClient.cs
```diff
@@ -202,7 +217,127 @@ public IServicesClient WithServerKeepBinary()
...
+
+                        return _nodes;
+                    }).ConfigureAwait(false);
+                
+                _nodes = FilterTopology(top, groupNodes?.Select(n => n.Id).ToList());
+                
+                Interlocked.Exchange(ref _lastUpdateRequestTime, DateTime.Now.Ticks);
+                Interlocked.Exchange(ref _lastAffTop, topVer);
+
+                log.Debug("Topology of service '" + _svcName + "' has been updated. The " +
```
ptupitsyn, line 308:
```suggestion
                if (log.IsEnabled(LogLevel.Debug))
                    log.Debug("Topology of service '" + _svcName + "' has been updated. The " +
```

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Services/ServicesClient.cs
```diff
@@ -202,7 +217,127 @@ public IServicesClient WithServerKeepBinary()
...
+
+                        return res;
+                    }, (status, err) =>
+                    {
+                        log.Error("Failed to update topology of the service '" + _svcName + "'.", err);
+
+                        return _nodes;
+                    }).ConfigureAwait(false);
+                
+                _nodes = FilterTopology(top, groupNodes?.Select(n => n.Id).ToList());
```
ptupitsyn, line 303: `ToList` is unnecessary, `Intersect` allows `IEnumerable`.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Services/ServicesClient.cs
```diff
@@ -202,7 +217,127 @@ public IServicesClient WithServerKeepBinary()
...
+                
+                Interlocked.Exchange(ref _lastUpdateRequestTime, DateTime.Now.Ticks);
+                Interlocked.Exchange(ref _lastAffTop, topVer);
+
+                log.Debug("Topology of service '" + _svcName + "' has been updated. The " +
+                          "service instance nodes: " + string.Join(", ", top.Select(gid=>gid.ToString())) +
+                          ". Effective topology with the cluster group is: " + 
+                          string.Join(", ", _nodes.Select(gid=>gid.ToString())) + '.');
+
+                _updateInProgress = 0;
```
ptupitsyn, line 313: Mixing `volatile` and `Interlocked` makes the code harder to reason about. It is considered a good practice to stick to one synchronization mechanism for a given field. Let's remove `volatile` and use `Interlocked.CompareExchange(ref _updateInProgress, 0, 0)` to read the value.

See https://stackoverflow.com/questions/24808291/reading-an-int-thats-updated-by-interlocked-on-other-threads

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Services/ServicesClient.cs
```diff
@@ -202,7 +217,127 @@ public IServicesClient WithServerKeepBinary()
...
+            {
+                if (_updateInProgress != 0)
+                    return _nodes;
+
+                var curAff = _svcClient._ignite.Socket.GetTopologyVersion();
+                var lastKnownAff = Interlocked.Read(ref _lastAffTop);
+                var sinceLastUpdate = DateTime.Now.Ticks - Interlocked.Read(ref _lastUpdateRequestTime);
+
+                if (curAff > lastKnownAff || sinceLastUpdate > SrvTopUpdatePeriod)
+                    UpdateTopologyAsync().ConfigureAwait(false);
```
ptupitsyn, line 337:
```suggestion
                    _ = UpdateTopologyAsync().ConfigureAwait(false);
```

Discard the task explicitly to show that we did not skip `await` by mistake.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Client/Services/ServiceAwarenessTest.cs
```diff
@@ -0,0 +1,518 @@
...
+    using Apache.Ignite.Core.Log;
+    using Apache.Ignite.Core.Services;
+    using Apache.Ignite.Core.Tests.Client.Cache;
+    using Apache.Ignite.Core.Tests.Services;
+    using NUnit.Framework;
+
+    /// <summary>
+    /// Service awareness tests.
+    /// </summary>
+    public class ServicesAwarenessTest
```
ptupitsyn, line 38: This class reimplements a lot of logic from `ClientTestBase` - is that intentional?

ptupitsyn (changes requested):

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Client/Services/ServiceAwarenessTest.cs
Vladsz83: Yes. I saw it. Not so much shared logic. Just cluster and logs management. I would need to fix the base classs too. But ok, implemented relying on ClientTestBase

ptupitsyn:
I've pushed some changes to your branch - please review them.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Services/ServicesClient.cs
```diff
@@ -35,6 +40,9 @@ namespace Apache.Ignite.Core.Impl.Client.Services
     /// </summary>
     internal class ServicesClient : IServicesClient
     {
+        /** Max service topology update period in millis. */
+        internal static readonly TimeSpan SrvTopUpdatePeriod = TimeSpan.FromSeconds(10);
```
ptupitsyn, line 44: 10 seconds is a bit too short. How often do we expect the topology to change?

Vladsz83: Java thin client has this period. But we can extend them both to one minute probably. WDYT?

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Client/Services/ServiceAwarenessTests.cs
```diff
@@ -337,14 +337,19 @@ private static void CallService(IJavaService service, bool failureExpected)
...
         {
-            var logEntry = log.Entries
+            var allLogEntries = log.Entries
                 .Where(e => e.Message.Contains(stringToSearch))
-                .Select(e => e.Message)
-                .Single();
+                .Select(e => e.Message);
+            
+            // Check service topology is updated strictly once.
+            Assert.AreEqual(1, allLogEntries.Count());
```
ptupitsyn, line 350: `Single()` ensures 1 and only 1 element.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Services/ServicesClient.cs
ptupitsyn: Let's extend to a minute and maybe file tickets to make those configurable?

Vladsz83: 1 minute is ok to me. But I dont think anyone is going to configure there tiemouts. Do we need any ticket? A minor improvemets?

Vladsz83: Btw. Service topology is not updated if client doesn't call the service. This timeout works only with the traffic.

ptupitsyn: My point is, we should optimize for a common use case, which is a stable topology. Requesting new assignment every 10 seconds seems wasteful. One minute is reasonable.

> I dont think anyone is going to configure these tiemouts.
Ok, I agree. Let's keep it hardcoded, we can always make it configurable if users ask for that.

ptupitsyn:
@Vladsz83 The PR looks good to me, thank you! Please get a green TC visa and we can merge it.

Vladsz83:
> @Vladsz83 The PR looks good to me, thank you! Please get a green TC visa and we can merge it.

Ok, good. Thanks. I'm gonna push one more commit. A test timeout issue arised on the Java test wit new topology poll period.

PR #11223, opened 2024-02-05
IGNITE-21156 : [IEP-114] Custom metrics introduction - IgniteMetricRegistry

Vladsz83:

### Notes


NSAmelchev (approved):

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetricRegistry.java
```diff
@@ -0,0 +1,103 @@
...
+public interface IgniteMetricRegistry extends ReadOnlyMetricRegistry {
+    /**
+     * Registers an int metric which value will be queried from the specified supplier.
+     *
+     * @param name Metric name.
+     * @param supplier Metric value supplier.
+     * @param desc Metric description.
+     * @return New or previously registered {@link IntMetric}.
+     */
+    IntMetric register(String name, IntSupplier supplier, @Nullable String desc);
```
nizhikov, line 50: Do we really need to return `IntMetric` here and in other methos?
After registration metrics exposed by exporter.
What are scenarios when user need to store `IntMetric` instance?

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetricRegistry.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Removes metrics with the {@code name}.
+     *
+     * @param name Metric name..
+     */
+    void remove(String name);
+
+    /** Resets all metrics of this metric registry. */
+    void reset();
```
nizhikov, line 102: Can you, please, clarify what "reset metric" means exactly?
All metrics registered with public API have suppliers to provide current value of the metric.
It's not clear what this method will do.

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetrics.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+@IgniteExperimental
+public interface IgniteMetrics extends Iterable<ReadOnlyMetricRegistry> {
+    /**
+     * Gets or creates custom metric registry named "custom." + {@code registryName}.
+     *
+     * @param registryName name part to add to the prefix "custom.".
+     * @return {@link IgniteMetricRegistry} registry.
+     */
+    IgniteMetricRegistry customRegistry(String registryName);
```
nizhikov, line 50: Why we have two methods `customRegistry` and `findRegistry`?
Is it possible to use only one? Like:

```
IgniteMetricRegistry registry(String name);
```

Vladsz83: `registry()` or `customRegistry()` always create a metric registry. `findRegistry()` doesn't. Hence, is a way just to take metrics if they exist.
`custom` is to emphasize that only custom metrics are affected. In the first ticket implementations we allowed read-only access to the system metrics as suggested in IGNITE-12553. But decided to simplify the first commit. In future we plan to expose read-only system metrics with `findRegistry()` and implement IGNITE-12553. WDYT?

nizhikov: > Hence, this is the way just to take metrics if they exist.

But we can iterate over all registry and find required. Do we really need one more method for that?
What is use-case for obtaining just one registry without creation?

> WDYT?

Sorry, I don't get it.
What is the API we want to create?

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetricRegistry.java
Vladsz83: Changed the javadoc to 'Calls {@link Metric#reset()} for all the registered metrics.'. Is that enough?

nizhikov: Why do we need this method, in the first place?

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetricRegistry.java
Vladsz83: It is not necessary but looks handy. User may read metric in its own way, not only by an exporter. We do similar, for instance, with `CacheMetricImpl#offHeapEntriesCnt, offHeapPrimaryEntriesCnt, heapEntriesCnt, cacheSize`. Using same API. It is simple to return additional getters.

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetricRegistry.java
Vladsz83: What do you mean 'in the first place'? Dropping, resetting, of a user metric looks intrinsic. Why not?

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetrics.java
Vladsz83: We want give a user the abilities to create, remove and get its own metrics and metric registries. We want to prevent affecting internal metrics by a user. User should cleary see it can change only custom metrics. In future, we plan to provide an ability to get read-only internal metrics as described in IGNITE-12553. That we have 'findRegistry()'. It will expose internal metric too, but not change them.

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetricRegistry.java
nizhikov: Because, all metrics we can create with the current API are those that works with `Supplier`.
And `reset` method intention is to assign 0 or some default value to the counter or something similar.
Please, take a look at `IntGauge` and other similar implementions.

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetrics.java
nizhikov: `findRegistry` will be used only in the future improvements?

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetricRegistry.java
Vladsz83: Yes, but we provide `reset()` even for the gauges.

nizhikov: `reset` for gauges are no-op.

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetrics.java
Vladsz83: > `findRegistry` will be used only in the future improvements?

No. It is supposed that user should be able try to get own metric registry witout creating it.

nizhikov: I don't think we should provide such method.

Seems like other API's don't provide it, also.
Please, take a look at dropwizard implementation:

https://github.com/dropwizard/metrics/blob/release/4.2.x/metrics-core/src/main/java/com/codahale/metrics/SharedMetricRegistries.java

nizhikov: Let's simplify `customRegistry` to well-known name `getOrCreate`.

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetrics.java
```diff
@@ -0,0 +1,71 @@
...
+    @Nullable ReadOnlyMetricRegistry findRegistry(String registryName);
+
+    /**
+     * Removes custom metric registry.
+     * <p>
+     * Note: The registry name have to follow the name convention as described at {@link IgniteMetrics}.
+     *
+     * @param registryName Registry name starting with 'custom.'.
+     */
+    void removeCustomRegistry(String registryName);
```
nizhikov, line 70: Let's rename this to well-known `remove`.

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetricRegistry.java
nizhikov: User can re-use supplier to read metric value. There is no need to return `IntMetric` here and in other methods.
Let's return `void` here.

nizhikov:
Maybe we should use "user provided metrics" or "user metrics" instead of "custom"?

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetricRegistry.java
```diff
@@ -0,0 +1,90 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Metric registry. Allows to get, add or remove metrics.
+ *
+ * @see IgniteMetrics
+ * @see ReadOnlyMetricRegistry
+ */
+@IgniteExperimental
+public interface IgniteMetricRegistry extends ReadOnlyMetricRegistry {
```
nizhikov, line 36: I doubt about naming here - `IgniteMetricRegistry` is something that belongs to Ignite.
But, here it used for registry that will be managed and fullfillled by the user.

maybe we can use just `MetricRegistry` here.
And rename internal implementation to `MetricRegistryImpl`?

nizhikov:
@Vladsz83 Thanks for the updates. Overall looks good. Let's agree on naming - I left some suggestions in PR.

Vladsz83:
> Maybe we should use "user provided metrics" or "user metrics" instead of "custom"?

@nizhikov where exactly? Other famous databases operate with 'custom metrics'.

Vladsz83:
> > where exactly?
> 
> In comments, method naming, etc.
> 
> > Other famous databases operate with 'custom metrics'.
> 
> Can you, please, provide an example?

You can take the examples from the IEP within 'Motivation' section: https://cwiki.apache.org/confluence/display/IGNITE/IEP-114+Custom+metrics#IEP114Custommetrics-Motivation

nizhikov:
I see - so "custom metric" is commonly used naming. OK, let's keep it.

modules/core/src/main/java/org/apache/ignite/metric/IgniteMetricRegistry.java
nizhikov: @Vladsz83 Let's decide on this suggestion before proceed with the merge.

Vladsz83: We've discussed remaning internal MetricRegistry -> MetricRegistryImpl here : https://github.com/apache/ignite/pull/11216/files . Also, we tried temporary keep two internal and external interfaces MetricRegistry to avoid related tons of code cnahges :  https://github.com/apache/ignite/pull/11221/files . The decision was that twe need more exposed, available for internal usage metric interfaces, internal code remenings and code cleanup before using 'MetricRegistry'. Since we declare the interface @IgniteExperimental, we could rename it later.

Vladsz83: Also, we plan to expose internal read-only metrics as noted before using the same API, ` findRegistry(String name)`. They belong to Ignite.

nizhikov: > Also, we plan to expose internal read-only metrics as noted before using the same API

Then method must return `ReadOnlyMetricRegistry`.

> to avoid related tons of code cnahges

I don't think "tons of changes" is the right reason to use bad naming.
Let's make API more clear by using clear naming.

Vladsz83: >> Then method must return ReadOnlyMetricRegistry.

Yes, but not IgniteReadOnlyMetricRegistry. Ok, I'll try to rename and make use MetricRegistry (not MetricRegistryImpl) as much as possible

Vladsz83: Ok, I prepared another PR with the renaming internal `class MetricRegistry` -> `class MetricRegistryImpl` and new `interface IgniteMetricRegistry` -> `interface MetricRegistry` : https://github.com/apache/ignite/pull/11293/files . 
Used `MetricRegistry` instead of `MetricRegistryImpl` where is possible. We planned to add methods like `longMetric` to the new registry interface in a next PR. Then, we could refactor to more internal usage of `interface MetricRegistry`.

I recalled one problem that we discussed: `CdcConsumer`. It is `@IgniteExperimental` but also a public API. And this feature look crucial. It uses internal `org.apache.ignite.internal.processors.metric.MetricRegistry`. Renaming to `MetricRegistryImpl` will break at least the extensions. And any external code which uses `CdcConsumer` and probably related `org.apache.ignite.internal.processors.metric.MetricRegistry` (`CdcConsumer#start(MetricRegistry mreg)`). IMHO, keeping `@IgniteExperimental IgniteMetricRegistry` for a while and possible later renaming to `@IgniteExperimental MetricRegistry` is a less evil. WDYT?

PR #11264, opened 2024-03-04
IGNITE-21660 ClientQueryCursor is not closed in getAll

shoothzj:
https://issues.apache.org/jira/projects/IGNITE/issues/IGNITE-21660

ClientQueryCursor is not closed in `getAll()` method. According to the annotation, it should be closed.

Annotation:
Since all the results will be fetched, all the resources will be closed automatically after this call, e.g. there is no need to call {@link #close()} method in this case.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientQueryCursor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientQueryCursor.java
@@ -40,19 +40,24 @@ class ClientQueryCursor<T> implements QueryCursor<T> {
 
     /** {@inheritDoc} */
     @Override public List<T> getAll() {
-        while (true) {
-            try {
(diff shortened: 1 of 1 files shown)


shoothzj:
@ptupitsyn cloud you please take a look? thanks :)

ptupitsyn:
@shoothzj there is no bug. We don't need to call `close` in `getAll`, because the server cleans up all resources automatically when it returns the last page of the query.
See `ClientCacheQueryCursor#writePage`: https://github.com/apache/ignite/blob/2fbbb676386515ea881e4e61f08864d6bc93225a/modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cache/ClientCacheQueryCursor.java#L88-L89

ptupitsyn:
I've closed the PR and the JIRA ticket. Let me know if you think there is still something missing.

PR #11298, opened 2024-04-03
IGNITE-21656 :  Cache dump fails on a cache with a node filter. V3

Vladsz83:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotPartitionsVerifyHandler.java
```diff
@@ -362,6 +362,9 @@ private Map<PartitionKeyV2, PartitionHashRecordV2> checkDumpFiles(
         SnapshotHandlerContext opCtx,
         Set<File> partFiles
     ) {
+        if (F.isEmpty(partFiles))
```
timoninmaxim, line 365: It looks like this condition should be outside of the method. If param is empty then no need to check them?

Vladsz83: Fixed but works wierd. Snapshot contains meta. It is a partition but might be ignored too

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/dump/Dump.java
```diff
@@ -163,10 +163,12 @@ public Dump(
      * @return Standalone kernal context.
      */
     private GridKernalContext standaloneKernalContext(File dumpDir, IgniteLogger log) {
-        File binaryMeta = CacheObjectBinaryProcessorImpl.binaryWorkDir(dumpDir.getAbsolutePath(), F.first(metadata).folderName());
+        SnapshotMetadata meta = metadata.stream().filter(m -> !m.partitions().isEmpty()).findFirst().orElse(F.first(metadata));
```
timoninmaxim, line 166: Looks overcomplicated.

timoninmaxim: Use F.view() instead

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/dump/Dump.java
```diff
@@ -163,10 +163,12 @@ public Dump(
...
+
+        File binaryMeta = CacheObjectBinaryProcessorImpl.binaryWorkDir(dumpDir.getAbsolutePath(), meta.folderName());
         File marshaller = new File(dumpDir, DFLT_MARSHALLER_PATH);
 
-        A.ensure(binaryMeta.exists(), "binary metadata directory not exists");
+        A.ensure(meta.partitions().isEmpty() || binaryMeta.exists(), "binary metadata directory not exists");
```
timoninmaxim, line 171: If meta.partitions() is empty, it means dump doesn't have partitions at all - it looks like a bug.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotPartitionsVerifyHandler.java
```diff
@@ -362,6 +362,13 @@ private Map<PartitionKeyV2, PartitionHashRecordV2> checkDumpFiles(
         SnapshotHandlerContext opCtx,
         Set<File> partFiles
     ) {
+        if (F.isEmpty(partFiles)) {
+            log.info("Dump data integrity check skipped [snpName=" + opCtx.metadata().snapshotName()
```
timoninmaxim, line 366: snpName -> dumpName, let's also add param for "this node" - consistentId?.

Vladsz83: Thought about it. All the other code uses 'snp' and 'snapshot' for dumps. But ok

PR #11373, opened 2024-06-04
IGNITE-22349 .NET: Add compute task session support

kukushal:
Add compute task session support to .NET. This allows the user to change the priority of a compute job via `grid.task.priority` attribute (see [docs](https://ignite.apache.org/docs/latest/distributed-computing/job-scheduling#priority-ordering)).

The new injectable .NET `IComputeTaskSession` has existing Java counterpart [ComputeTaskSession](https://github.com/apache/ignite/blob/2.16.0/modules/core/src/main/java/org/apache/ignite/compute/ComputeTaskSession.java) and supports only the `SetAttributes` method. It does not support the rest of the Java API since only
the `SetAttributes` is required to address IGNITE-22349. The `GetAttribute` method was added to allow creating 
automated tests.

Changes:
* The new `IComputeTaskSession` interface supports the `SetAttributes` and `GetAttribute` operations with the same
signature as their Java `ComputeTaskSession` counterparts.
  * The `ComputeTaskSession` implementation delegates operations to the Java side via `PlatformJniTarget`, which
  receives the unmanaged session pointer in the JNI `ComputeTaskMap` and `ComputeJobExecute` callbacks.
  * On the Java side the new `PlatformComputeTaskSession` class receives the JNI calls and delegates operations to 
  wrapped `ComputeTaskSession`.
  * The `ComputeTaskSession` on the Java side in injected into `PlatformFullJob` or `PlatformFullTask`. At first glance
  the injection might seem slow. However injectors are cached so subsequent injections would be as fast as setting the
  field directly. The alternative approach to pass `ComputeTaskSession` directly to the `PlatformFullJob` and
  `PlatformFullTask` constructors seems to required lots of code changes. 
* The new .NET `ComputeTaskSessionFullSupportAttribute` explicitly enables the compute session attributes API in the 
same way the Java's [ComputeTaskSessionFullSupport](https://github.com/apache/ignite/blob/2.16.0/modules/core/src/main/java/org/apache/ignite/compute/ComputeTaskSessionFullSupport.java) annotation does. This flag is passed to Java as the `Execute` JNI operation parameter and stored in the `PlatformFullTask`.
* The new .NET `TaskSessionResourceAttribute` injects the `IComputeTaskSession` into implementations of `IComputeTask` and `IComputeJob`.
* Added automated tests for distributing session attributes on the same and different nodes.
* Documentation: added new "C#/.NET" tab in the `MyUrgentTask` code snippet in the [Priority Ordering](https://ignite.apache.org/docs/latest/distributed-computing/job-scheduling#priority-ordering) section.


modules/platforms/dotnet/Apache.Ignite.Core/Compute/IComputeTaskSession.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// Stores custom compute task attributes. Specific compute task implementations must be annotated with the
+    /// <see cref="ComputeTaskSessionFullSupportAttribute"/> to enable distributing the task attributes to the compute
+    /// jobs that the task creates.
+    /// </summary>
+    public interface IComputeTaskSession
+    {
+        /// <summary>
+        /// Gets the value of the given key or <c>null</c> if the key does not exist.
+        /// </summary>
+        TV GetAttribute<TV>(object key);
```
ptupitsyn, line 32:
```suggestion
        TV GetAttribute<TK, TV>(TK key);
```

modules/platforms/dotnet/Apache.Ignite.Core/Compute/IComputeTaskSession.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        /// <summary>
+        /// Stores the collection of attributes.
+        /// </summary>
+        void SetAttributes<TK, TV>(IEnumerable<KeyValuePair<TK, TV>> attrs);
```
ptupitsyn, line 37:
```suggestion
        void SetAttributes<TK, TV>(params KeyValuePair<TK, TV>[] attrs);
```

This way we don't need an extension method below.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Compute/ComputeTaskSessionFullSupportAttribute.cs
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Apache.Ignite.Core.Compute
+{
+    using System;
+    using static System.AttributeTargets;
+
+    /// <summary>
```
ptupitsyn, line 23: Let's add a note about performance - Java annotation javadoc has this:
```
By default, attributes and checkpoints are disabled for performance reasons.
```

ptupitsyn (changes requested):

kukushal:
@ptupitsyn , @shishkovilja , I addressed your code review comments and also fixed some C++ code that I forgot to initially change and that caused some C++ test failures. Could you please take another look and approve the PR if you have no other comments. Thank you!

modules/platforms/dotnet/Apache.Ignite.Core/Compute/IComputeTaskSession.cs
kukushal: Done

modules/platforms/dotnet/Apache.Ignite.Core/Compute/IComputeTaskSession.cs
kukushal: Done

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Compute/ComputeTaskSessionFullSupportAttribute.cs
kukushal: Done

ptupitsyn (approved):

shishkovilja (approved):

kukushal:
@ptupitsyn , thank you for the veview. Would you be willing to merge the changes please?

shishkovilja:
@kukushal , @ptupitsyn , guys, it seems, that build was broken by this PR. I have failed a ticket in order to fix poblem: https://issues.apache.org/jira/browse/IGNITE-22586

PR #11391, opened 2024-06-13
IGNITE-22662 : Snapshot check as distributed process

Vladsz83:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckDistributedProcess.java
```diff
@@ -0,0 +1,549 @@
...
+    private static final IgniteInternalFuture FINISHED_FUT = new GridFinishedFuture<>();
+
+    /** */
+    private final IgniteLogger log;
+
+    /** */
+    private final GridKernalContext kctx;
+
+    /** Snapshot check requests per snapshot on every node. */
+    final Map<String, SnapshotCheckProcessRequest> requests = new ConcurrentHashMap<>();
```
timoninmaxim, line 70: Let's make the field private and provide a method for getting all current requests.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckDistributedProcess.java
```diff
@@ -0,0 +1,549 @@
...
+    }
+
+    /**
+     * Stops the process with the passes exception.
+     *
+     * @param th The interrupt reason.
+     * @param rqFilter If not {@code null}, used to filter which requests/process to stop. If {@code null}, stops all the validations.
+     */
+    public void interrupt(Throwable th, @Nullable Function<SnapshotCheckProcessRequest, Boolean> rqFilter) {
+        if (requests.isEmpty())
```
timoninmaxim, line 103: useless check

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckDistributedProcess.java
```diff
@@ -0,0 +1,549 @@
...
+            .filter(e -> requiredNodes.contains(e.getKey()) && e.getValue() != null)
+            .collect(Collectors.toMap(e -> kctx.cluster().get().node(e.getKey()), Map.Entry::getValue));
+    }
+
+    /**
+     * @param snpName Snapshot name of the validation process. If {@coe null}, ignored.
+     * @param procId  If {@code snpName} is {@code null}, is used to find the operation request.
+     * @return Current snapshot checking request by {@code snpName} or {@code procId}.
+     */
+    private @Nullable SnapshotCheckProcessRequest curRequest(@Nullable String snpName, UUID procId) {
```
timoninmaxim, line 306: AFAIK, we don't use abbreviations in method names. `currentRequest`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckDistributedProcess.java
```diff
@@ -0,0 +1,549 @@
...
+        Map<UUID, Throwable> errors
+    ) {
+        clean(procId, null, results, errors);
+
+        return FINISHED_FUT;
+    }
+
+    /** Phase 2 beginning.  */
+    private IgniteInternalFuture<HashMap<PartitionKeyV2, PartitionHashRecordV2>> validateParts(
+        SnapshotCheckProcessRequest incReq) {
```
timoninmaxim, line 139: What does `incReq` mean?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckDistributedProcess.java
```diff
@@ -0,0 +1,549 @@
...
+        SnapshotCheckProcessRequest locReq;
+
+        if (stopAndCleanOnError(incReq, null) || (locReq = requests.get(incReq.snapshotName())) == null)
+            return FINISHED_FUT;
+
+        assert locReq.equals(incReq);
+
+        // Store metas to collect cluster operation result laster.
```
timoninmaxim, line 147: laster?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckDistributedProcess.java
```diff
@@ -0,0 +1,549 @@
...
+     */
+    private static void stopFutureOnAnyFailure(@Nullable GridFutureAdapter<?> fut, Runnable action) {
+        if (fut == null) {
+            action.run();
+
+            return;
+        }
+
+        try {
+            action.run();
```
timoninmaxim, line 521: Code duplication.

```
try {
    action.run();
}
catch (Throwable th) {
    if (fut != null)
        fut.onDone(th);
}
```

timoninmaxim: Also, it looks like that the `fut` is never `null`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckDistributedProcess.java
```diff
@@ -0,0 +1,549 @@
...
+        Map<UUID, HashMap<PartitionKeyV2, PartitionHashRecordV2>> results,
+        Map<UUID, Throwable> errors
+    ) {
+        clean(procId, null, results, errors);
+
+        return FINISHED_FUT;
+    }
+
+    /** Phase 2 beginning.  */
+    private IgniteInternalFuture<HashMap<PartitionKeyV2, PartitionHashRecordV2>> validateParts(
```
timoninmaxim, line 138: Why does it actually `HashMap` and not just a `Map`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckDistributedProcess.java
```diff
@@ -0,0 +1,549 @@
...
+                    .checkPartitions(locReq.meta(), snpDir, locReq.groups(), false, true, false);
+
+                locPartsChkFut.onDone(res instanceof HashMap ? (HashMap<PartitionKeyV2, PartitionHashRecordV2>)res
+                    : new HashMap<>(res));
+            }
+            catch (IgniteCheckedException e) {
+                throw new IgniteException("Failed to calculate snapshot partition hashes, req: " + incReq, e);
+            }
+
+            // No need to wait to the cealup if current node is just a worker.
```
timoninmaxim, line 184: cealup?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckDistributedProcess.java
```diff
@@ -0,0 +1,549 @@
...
+     * If required, stops and clean related validation if errors occured
+     *
+     * @return {@code True} if the validation stopped and cleaned. {@code False} otherwise.
+     */
+    private boolean stopAndCleanOnError(SnapshotCheckProcessRequest req,
+        @Nullable Map<UUID, Throwable> occuredErrors) {
+        assert req != null;
+
+        if (!F.isEmpty(occuredErrors)) {
+            occuredErrors = occuredErrors.entrySet().stream()
```
timoninmaxim, line 202: I suppose the errors can appear only from nodes that are part of the request? Because these errors are passed in reduce with the request

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckDistributedProcess.java
```diff
@@ -0,0 +1,549 @@
...
+    private Map<ClusterNode, Exception> collectErrors(Map<UUID, Throwable> errors, Set<UUID> requiredNodes) {
+        if (errors.isEmpty())
+            return Collections.emptyMap();
+
+        return errors.entrySet().stream().filter(e -> requiredNodes.contains(e.getKey()) && e.getValue() != null)
+            .collect(Collectors.toMap(e -> kctx.cluster().get().node(e.getKey()), e -> asException(e.getValue())));
+    }
+
+    /** */
+    private Map<ClusterNode, Map<PartitionKeyV2, PartitionHashRecordV2>> collecPartsHashes(
```
timoninmaxim, line 289: collec?

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckDistributedProcess.java
Vladsz83: Well, it should. But the process works on any node. Theoretically, any unrelated, unexpected exception might occur on any node. Let's keep an assert.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckDistributedProcess.java
Vladsz83: Because `DistributedProcess` requires a `Serializable` result:
`DistributedProcess<I extends Serializable, R extends Serializable>`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckDistributedProcess.java
Vladsz83: Irrelevant for now. The code was changed. The null check is above. The cluster opearation result future can be null in `clean(...) for example.`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotFullCheckDistributedProcess.java
```diff
@@ -0,0 +1,549 @@
...
+
+        log = kctx.log(getClass());
+
+        phase1CheckMetas = new DistributedProcess<>(kctx, SNAPSHOT_CHECK_METAS, this::prepareAndCheckMetas,
+            this::reducePreparationAndMetasCheck);
+
+        phase2PartsHashes = new DistributedProcess<>(kctx, SNAPSHOT_VALIDATE_PARTS, this::validateParts,
+            this::reduceValidatePartsAndFinish);
+
+        kctx.event().addLocalEventListener((evt) -> nodeLeft(((DiscoveryEvent)evt).eventNode()), EVT_NODE_FAILED, EVT_NODE_LEFT);
```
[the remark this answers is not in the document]
Vladsz83: Yes. But I wanted to make the code more independent. The process is able to handle most of the events itself. WDYT?

modules/core/src/main/java/org/apache/ignite/internal/management/cache/VerifyBackupPartitionsTaskV2.java
```diff
@@ -110,7 +111,7 @@ public class VerifyBackupPartitionsTaskV2 extends ComputeTaskAdapter<CacheIdleVe
 
     /** {@inheritDoc} */
     @Nullable @Override public IdleVerifyResultV2 reduce(List<ComputeJobResult> results) throws IgniteException {
-        return reduce0(results);
```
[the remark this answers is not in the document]
Vladsz83: Irrelevant. The code was already chenged

modules/core/src/main/java/org/apache/ignite/internal/management/cache/VerifyBackupPartitionsTaskV2.java
```diff
@@ -138,34 +139,43 @@ public class VerifyBackupPartitionsTaskV2 extends ComputeTaskAdapter<CacheIdleVe
...
-        }
+        });
 
-        if (results.size() != ex.size())
-            return new IdleVerifyResultV2(clusterHashes, ex);
+        if (results.size() != errors.size())
+            return new IdleVerifyResultV2(clusterHashes, errors);
         else
-            return new IdleVerifyResultV2(ex);
+            return new IdleVerifyResultV2(errors);
```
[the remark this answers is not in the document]
Vladsz83: Irrelevant. The code was already changed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,543 @@
...
+            if (locReq.error() != null)
+                locPartsChkFut.onDone(locReq.error());
+
+            if (locPartsChkFut.isDone())
+                return;
+
+            File snpDir = kctx.cache().context().snapshotMgr().snapshotLocalDir(locReq.snapshotName(), locReq.snapshotPath());
+
+            try {
+                Map<PartitionKeyV2, PartitionHashRecordV2> res = kctx.cache().context().snapshotMgr().checker()
```
timoninmaxim, line 172: Looks weird that `SnapshotCheckProcess` doesn't have own instance of `SnapshotChecker`. `IgniteSnapshotManager` checks nothing but uses only `read*` methods from the `SnapshotChecker` class. Consider refactoring here.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckDistributedProcess.java
timoninmaxim: Let's use DTO classes, like ValidatePartitionsResponse. It's very weird to see smth like. 

```
            if (!(locMetas instanceof ArrayList))
                locMetas = new ArrayList<>(locMetas);
...
            locMetasChkFut.onDone((ArrayList<SnapshotMetadata>)locMetas);
```

Or just use an assert `assert locMetas instanceof ArrayList` to get an exception during tests. Recreate collection isn't a right way

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,543 @@
...
+                locReq.meta(locMetas.get(0));
+
+            if (!(locMetas instanceof ArrayList))
+                locMetas = new ArrayList<>(locMetas);
+
+            // A node might have already gone before this. No need to proceed.
+            Throwable locRqErr = locReq.error();
+
+            if (locRqErr != null)
+                locMetasChkFut.onDone(locRqErr);
```
timoninmaxim, line 365: You use both `locReq` and `locRqErr`. Let's use the same abbreviation for requests.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,543 @@
...
+
+        GridFutureAdapter<ArrayList<SnapshotMetadata>> locMetasChkFut = new GridFutureAdapter<>();
+
+        assert locReq.fut() == null;
+
+        locReq.fut(locMetasChkFut);
+
+        IgniteSnapshotManager snpMgr = kctx.cache().context().snapshotMgr();
+
+        snpMgr.snapshotExecutorService().submit(() -> stopFutureOnAnyFailure(locMetasChkFut, () -> {
```
timoninmaxim, line 336: As I can see in the code of `SnapshotChecker` it already uses the executor service for some checks. Let's delegate all work with executor services to the checker.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,543 @@
...
+            if (!F.isEmpty(locMetas))
+                locReq.meta(locMetas.get(0));
+
+            if (!(locMetas instanceof ArrayList))
+                locMetas = new ArrayList<>(locMetas);
+
+            // A node might have already gone before this. No need to proceed.
+            Throwable locRqErr = locReq.error();
+
+            if (locRqErr != null)
```
timoninmaxim, line 364: You already check the error in this thread on L344. I suppose if we check it once and then run the check then we should respond with result of the check

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
Vladsz83: I thought about it. The checker is shared. The tasks use it too. Once we remove the tasks, we could move the checker to the process. WDYT?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotFullCheckDistributedProcess.java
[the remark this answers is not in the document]
Vladsz83: I don't think that a code duplication definitely means a problem. IMHO, much worse is a violation of the principles like the encapsulation. SnapshotManager ignores clients:

if (ctx.clientNode())
    return;
            
It won't deliver events to the check process in some cases. Why? Why it should decide what to do with an event instead of the dedicated process? Client can start snapshot checking too. It keeps the launch future, results and partial results. It requires cleaning too. Several tests fail on it. Client must know about a node leave. But IgniteSnasphot manager doesn't allow. As I remember, I fixed a couple of similar bugs in the snapshots: we ignored client and non-baseline node.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,549 @@
...
+
+            Collection<Integer> grpIds = F.isEmpty(locReq.groups()) ? null : F.viewReadOnly(locReq.groups(), CU::cacheId);
+
+            // An error can occur when the local future is not initialized yet.
+            Throwable locReqErr = locReq.error();
+
+            if (locReqErr != null)
+                locMetasChkFut.onDone(locReqErr);
+
+            if (!locMetasChkFut.isDone()) {
```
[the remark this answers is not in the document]
timoninmaxim: Actually, do we really need this check at all? The future inside `checkLocalMetas` isn't interruptible, then in case of error happened a moment later after the check, it unconditionally checks all metas. I think we can skip this `if` at all in favor of simplicity.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,549 @@
...
+     * @return Current snapshot checking request by {@code snpName} or {@code procId}.
+     */
+    private @Nullable SnapshotCheckProcessRequest currentRequest(@Nullable String snpName, UUID procId) {
+        return snpName == null
+            ? requests.values().stream().filter(rq -> rq.requestId().equals(procId)).findFirst().orElse(null)
+            : requests.get(snpName);
+    }
+
+    /** Phase 1 beginning: prepare, collect and check local metas. */
+    private IgniteInternalFuture<CheckResultDTO> prepareAndCheckMetas(SnapshotCheckProcessRequest extReq) {
```
timoninmaxim, line 299: Let's add javadocs for `extReq`. I suppose it means `external`. I'm not sure, why it isn't just a `req`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,549 @@
...
+    private void clean(
+        UUID rqId,
+        @Nullable Throwable opErr,
+        @Nullable Map<UUID, CheckResultDTO> results,
+        @Nullable Map<UUID, Throwable> errors
+    ) {
+        GridFutureAdapter<SnapshotPartitionsVerifyTaskResult> clusterOpFut = clusterOpFuts.remove(rqId);
+
+        stopFutureOnAnyFailure(clusterOpFut, () -> {
+            SnapshotCheckProcessRequest locRq = currentRequest(null, rqId);
```
timoninmaxim, line 208: you still have in the same class `locReq` and `locRq`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,549 @@
...
+            log.debug("Started partitions validation as part of the snapshot checking. Request=" + locReq + '.');
+    }
+
+    /** Finds current snapshot name from the metas. */
+    private @Nullable String snpName(@Nullable Map<UUID, CheckResultDTO> results) {
+        if (F.isEmpty(results))
+            return null;
+
+        // Ensure the same snapshot name or empty.
+        assert F.flatCollections(results.values().stream().filter(r -> r != null && r.metas != null).map(res -> res.metas)
```
timoninmaxim, line 432: It looks complicated. I suppose it can be check during map phase. Or, maybe, while constructing `CheckResultDTO`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,549 @@
...
+        if (locReq == null || stopAndCleanOnError(locReq, errors) || !kctx.localNodeId().equals(locReq.opCoordId))
+            return;
+
+        Throwable stopClusterProcErr = null;
+
+        try {
+            assert !F.isEmpty(results) || !F.isEmpty(errors) || locReq.error() != null;
+
+            if (locReq.error() != null)
+                throw locReq.error();
```
timoninmaxim, line 380: I suppose it's already handled in 371L, doesn't it? There is already collection of all results from all nodes, then concurrent errors don't affect the phase.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,549 @@
...
+
+            Map<ClusterNode, Exception> errs = new HashMap<>();
+
+            results.forEach((nodeId, nodeRes) -> {
+                // A node might be non-baseline (not required).
+                if (locReq.nodes.contains(nodeId)) {
+                    assert nodeRes != null && nodeRes.partsHashes == null;
+                    assert kctx.cluster().get().node(nodeId) != null;
+
+                    locReq.metas.put(kctx.cluster().get().node(nodeId), nodeRes.metas);
```
timoninmaxim, line 392: `locReq.metas` are volatile and it's not a concurrent collection. Then I suppose it protects the reference only. Then common practice is to fully prepare collection first and after that set the reference.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,549 @@
...
+                }
+            });
+
+            errors.forEach((nodeId, nodeErr) -> {
```
timoninmaxim, line 396: If `errors` are not empty this code is never executed. It is handled within `stopAndCleanOnError`

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,549 @@
...
+
+    /** Phase 1 beginning: prepare, collect and check local metas. */
+    private IgniteInternalFuture<CheckResultDTO> prepareAndCheckMetas(SnapshotCheckProcessRequest extReq) {
+        SnapshotCheckProcessRequest locReq = requests.computeIfAbsent(extReq.snapshotName(), snpName -> extReq);
+
+        if (!locReq.equals(extReq)) {
+            Throwable err = new IllegalStateException("Validation of snapshot '" + extReq.snapshotName()
+                + "' has already started. Request=" + locReq + '.');
+
+            clean(extReq.requestId(), err, null, null);
```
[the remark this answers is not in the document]
Vladsz83: Yes, we could. I failed to do it quickly. Error processing is one of the huges parts of code and tests. Also It doesn't become simplier. Why changing? Repeating other process as much as possible was my goal at all. We do lesser serialization here, exist asap.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,549 @@
...
+                log.debug("Skipping snapshot local metadatas collecting for snapshot validation, request=" + extReq
+                    + ". Current node is not required.");
+            }
+
+            return new GridFinishedFuture<>();
+        }
+
+        GridFutureAdapter<CheckResultDTO> locMetasChkFut = new GridFutureAdapter<>();
+
+        stopFutureOnAnyFailure(locMetasChkFut, () -> {
```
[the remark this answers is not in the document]
Vladsz83: Same. This requires a lot of logic change. What is wrong now? We stop and clean asap, do fewer serializations.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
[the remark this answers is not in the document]
Vladsz83: I think we should not start any working if we cleary know that an error already arised. Why not checking it before starting?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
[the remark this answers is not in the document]
Vladsz83: No, we do `return new GridFinishedFuture<>(err);` in the restore process. Throwing exceptiong leads to `            U.error(log, "Failed to handle InitMessage [id=" + p.id + ']', err);` in DistributedProcess.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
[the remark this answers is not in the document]
Vladsz83: We do `return new GridFinishedFuture<>(err);` in the restore process. Throwing exceptiong leads to `            U.error(log, "Failed to handle InitMessage [id=" + p.id + ']', err);` in DistributedProcess.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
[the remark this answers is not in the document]
Vladsz83: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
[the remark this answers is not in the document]
timoninmaxim: > We do return new GridFinishedFuture<>(err); in the restore process

I see that both options are equal:

1. Throwing exception leads to setting the future to `p.resFut.onDone(err);`.
2. Returning error leads to `p.resFut.onDone(fut.error());`

I think throwing an exception makes code simplier.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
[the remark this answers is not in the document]
timoninmaxim: > Why not checking it before starting?

Because you don't check it after start, for example for every iteration of underlying loop inside `checkLocalMetas`. You should check it everywhere to reduce amount of work to be done. Currently it looks weird and only complicates the code.

This exception is throwed by very rare condition and amount of job to be skipped is actually very small. Let's skip this handling of this case.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,503 @@
...
+                return;
+
+            SnapshotMetadataVerificationTaskResult metasRes = new SnapshotMetadataVerificationTaskResult(
+                locReq.metas,
+                SnapshotChecker.reduceMetasResults(locReq.snapshotName(), locReq.snapshotPath(), locReq.metas, null,
+                    kctx.cluster().get().localNode().consistentId())
+            );
+
+            if (!F.isEmpty(metasRes.exceptions()))
+                stopClusterProcErr = new IgniteSnapshotVerifyException(metasRes.exceptions());
```
timoninmaxim, line 370: `throw new IgniteSnapshotVerifyException(metasRes.exceptions());`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,503 @@
...
+        // Local meta might be null if current node started after the snapshot creation or placement.
+        if (!req.nodes.contains(kctx.localNodeId()) || locReq.meta() == null)
+            return new GridFinishedFuture<>();
+
+        GridFutureAdapter<CheckResultDTO> locPartsChkFut = new GridFutureAdapter<>();
+
+        stopFutureOnAnyFailure(locPartsChkFut, () -> {
+            locReq.fut(locPartsChkFut);
+
+            if (!locPartsChkFut.isDone()) {
```
timoninmaxim, line 159: The same, but `checkPartitions` looks very heavy workload. How do you see if it possible to interrupt the processing in case of error?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,503 @@
...
+            else
+                finished = finishClusterFutureWithErr(clusterOpFut, err, errors0);
+
+            if (finished && log.isInfoEnabled())
+                log.info("Snapshot validation process finished, reqId: " + reqId + '.');
+        });
+    }
+
+    /** */
+    private Throwable stopAndCleanLocRequest(SnapshotCheckProcessRequest locRq, @Nullable Throwable err) {
```
timoninmaxim, line 219: Let's inline this method

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,503 @@
...
+        if (err == null && locRq.error() != null)
```
timoninmaxim, line 220: It's still `locReq` and `locRq` in this class

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,503 @@
...
+     *
+     * @param th The interrupt reason.
+     * @param rqFilter If not {@code null}, used to filter which requests/process to stop. If {@code null}, stops all the validations.
+     */
+    void interrupt(Throwable th, @Nullable Function<SnapshotCheckProcessRequest, Boolean> rqFilter) {
+        requests.values().forEach(rq -> {
+            if (rqFilter == null || rqFilter.apply(rq)) {
+                rq.error(th);
+
+                clean(rq.requestId(), th, null, null);
```
timoninmaxim, line 107: Is there case when it's not cleaned during `reduce` phase?

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
Vladsz83: Yes, we should be able to interrupt here. But I see it as a new ticket like "Snapshot check cancelation." Currently, there is no normal snapshot check cancelation.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
timoninmaxim: > When the reduce phase is not called at all.

Could you please describe the case? AFAIU, `requests` are filled only on the `map` phase. By code of `DistributedProcess` I see if `map` phase begins it means that it's guaranteed that `reduce` phase will call (even if coordinator fails).

The only case it might not be call - stopping local node. But it looks like we needn't clean anything in this case.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,496 @@
...
+            inclCstHndlrs,
+            true
+        );
+
+        GridFutureAdapter<SnapshotPartitionsVerifyTaskResult> clusterOpFut = new GridFutureAdapter<>();
+
+        clusterOpFut.listen(fut -> clusterOpFuts.remove(reqId));
+
+        stopFutureOnAnyFailure(clusterOpFut, () -> {
+            clusterOpFuts.put(reqId, clusterOpFut);
```
timoninmaxim, line 427: This code never fails. No need to wrap it in the method

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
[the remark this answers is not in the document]
timoninmaxim: Still not fixed

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,496 @@
...
+                if (locWorkingFut != null)
+                    finished = err == null ? locWorkingFut.onDone() : locWorkingFut.onDone(err);
+
+                requests.remove(locReq.snapshotName());
+
+                if (finished && log.isInfoEnabled())
+                    log.info("Finished snapshot local validation, req: " + locReq + '.');
+            }
+
+            if (clusterOpFut == null || clusterOpFut.isDone())
```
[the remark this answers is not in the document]
Vladsz83: Every iteration is supposed to clean certain request/process and possible cluster-wide operation future. Consider you launch several snapshot checks from the same node. This node keeps several requests and several cluster-wide operation futures. Every interrupt-iteration cleans each local operation request and each operation future.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,496 @@
...
+        @Nullable Map<UUID, Throwable> errors
+    ) {
+        GridFutureAdapter<SnapshotPartitionsVerifyTaskResult> clusterOpFut = clusterOpFuts.remove(reqId);
+
+        stopFutureOnAnyFailure(clusterOpFut, () -> {
+            SnapshotCheckProcessRequest locReq = currentRequest(null, reqId);
+
+            Throwable err = opErr;
+
+            if (locReq != null) {
```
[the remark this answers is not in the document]
Vladsz83: No, we cannot. In the case when you launch full checking of the same snapshot, you have no request but have the future on the originator node. Tests like `testConcurrentTheSameSnpFullRestorationWhenFullyCheckingDeclined()` or `testConcurrentTheSameSnpFullChecksDeclined()` cover this case.

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,466 @@
...
+        SnapshotCheckProcessRequest locReq = currentRequest(null, req.requestId());
+
+        if (locReq == null)
+            return new GridFinishedFuture<>();
+
+        assert locReq.equals(req);
+
+        // Store metas on the originator to redulte relusts later.
+        if (req.operationalNodeId().equals(kctx.localNodeId()))
+            locReq.metas = req.metas;
```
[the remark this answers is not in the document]
Vladsz83: Not for the operation initiator, only for the operation coordinator. The coordinator processes prossible errors and may stop the process. Here we are not sure that metas are needed. Also, first phase's results are reduces and processed by the operation coordinator. Other node should not work here. But the initiator needs the metas too. It forms process/future result (`SnapshotPartitionsVerifyTaskResult` in `reduceValidatePartsAndFinish()`/`clean()`) in the future. If first phase was OK, the initiator takes properly collected metas on the validation preparation.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,466 @@
...
+
+        SnapshotCheckProcessRequest locReq = currentRequest(null, req.requestId());
+
+        if (locReq == null)
+            return new GridFinishedFuture<>();
+
+        assert locReq.equals(req);
+
+        // Store metas on the originator to redulte relusts later.
+        if (req.operationalNodeId().equals(kctx.localNodeId()))
```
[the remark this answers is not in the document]
Vladsz83: Operation initiator/originator is a node who initiated the process. Probably a client node. We should not manage the process from a client. Operation coordinator is a node which controls the process, reduces first phase, decides if second phase is required if no error appears. Currently, the operation coordinator is one of the required server nodes (randomly selected).

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,466 @@
...
+     *
+     * @param th The interrupt reason.
+     * @param rqFilter If not {@code null}, used to filter which requests/process to stop. If {@code null}, stops all the validations.
+     */
+    void interrupt(Throwable th, @Nullable Function<SnapshotCheckProcessRequest, Boolean> rqFilter) {
+        requests.values().forEach(req -> {
+            if (rqFilter == null || rqFilter.apply(req)) {
+                req.error(th);
+
+                clean(req.requestId(), th, null, null);
```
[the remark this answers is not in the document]
Vladsz83: Why not stopping pontentially long futures like partition validations? Yes, we can't do it right now. But we should. `clean()` assigns the error the the current futures. WDYT? I've put some comments and renemed the method.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,466 @@
...
+        return snpName == null
+            ? requests.values().stream().filter(req -> req.requestId().equals(procId)).findFirst().orElse(null)
+            : requests.get(snpName);
+    }
+
+    /** Phase 1 beginning: prepare, collect and check local metas. */
+    private IgniteInternalFuture<CheckResultDTO> prepareAndCheckMetas(SnapshotCheckProcessRequest req) {
+        SnapshotCheckProcessRequest locReq = requests.computeIfAbsent(req.snapshotName(), snpName -> req);
+
+        if (!locReq.equals(req)) {
```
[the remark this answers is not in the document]
Vladsz83: It also calls `super.equals(other)`. Doesn't it?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,462 @@
...
+        kctx.event().addLocalEventListener((evt) -> {
+            if (requests.isEmpty())
+                return;
+
+            DiscoveryEvent devt = (DiscoveryEvent)evt;
+
+            Throwable err = new ClusterTopologyCheckedException("Snapshot checking stopped. " +
+                "A required node or the initiator node left the cluster: " + devt.eventNode() + '.');
+
+            interrupt(err, req -> req.nodes().contains(devt.eventNode().id())
```
timoninmaxim, line 96: Looks like we can explicitly use filter here instead of provide it to the `interrupt` method.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,475 @@
...
+        String snpName,
+        @Nullable String snpPath,
+        @Nullable Collection<String> grpNames,
+        boolean inclCstHndlrs
+    ) {
+        assert !F.isEmpty(snpName);
+
+        UUID reqId = UUID.randomUUID();
+
+        List<UUID> requiredNodes = new ArrayList<>(F.viewReadOnly(kctx.discovery().discoCache().aliveBaselineNodes(), F.node2id()));
```
timoninmaxim, line 368: Why not Set?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,475 @@
...
+        private final SnapshotCheckProcessRequest req;
+
+        /** Working future. */
+        private final GridFutureAdapter<CheckResultDTO> fut = new GridFutureAdapter<>();
+
+        /** Collected cluster metas. */
+        @Nullable private Map<ClusterNode, List<SnapshotMetadata>> clusterMetas;
+        
+        /** Snapshot local metadata. */
+        @Nullable private SnapshotMetadata locMeta;
```
timoninmaxim, line 436: Can we calculate it from `clusterMetas?`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,475 @@
...
+        /** An error occured. */
+        @Nullable private volatile Throwable err;
+
+        /** Creates operation context. */
+        private SnapshotCheckContext(SnapshotCheckProcessRequest req) {
+            this.req = req;
+        }
+
+        /** */
+        boolean isNodeRequired(UUID nodeId) {
```
timoninmaxim, line 447: prefix `is` isn't used for internal classes.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,475 @@
...
+    /** Phase 1 beginning: prepare, collect and check local metas. */
+    private IgniteInternalFuture<CheckResultDTO> prepareAndCheckMetas(SnapshotCheckProcessRequest req) {
+        SnapshotCheckContext ctx = contexts.computeIfAbsent(req.snapshotName(), snpName -> new SnapshotCheckContext(req));
+
+        if (!ctx.req.equals(req)) {
+            return new GridFinishedFuture<>(new IllegalStateException("Validation of snapshot '" + req.snapshotName()
+                + "' has already started. Request=" + ctx + '.'));
+        }
+
+        if (!req.nodes.contains(kctx.localNodeId())) {
```
timoninmaxim, line 237: Can we check it before creating the context? If node should be skipped then we can just return `new GridFinishedFuture()`.

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
Vladsz83: > Why not Set?

AbstractSnapshotOperationRequest creates set: `this.nodes = new HashSet<>(nodes);`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,485 @@
...
+
+        clusterOpFuts.put(reqId, clusterOpFut);
+
+        phase1CheckMetas.start(req.requestId(), req);
+
+        return clusterOpFut;
+    }
+
+    /** Properly sets errror to the cluster operation future. */
+    static boolean finishClusterFutureWithErr(
```
timoninmaxim, line 416: This method invokes the only branch, as `propogatedError == null`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,485 @@
...
+
+            GridFutureAdapter<?> fut = clusterOpFuts.get(ctx.req.reqId);
+
+            if (fut != null)
+                fut.onDone(err);
+        }
+    }
+
+    /** */
+    Map<String, SnapshotCheckContext> requests() {
```
timoninmaxim, line 116: Wrong name

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,485 @@
...
+        @Nullable Map<ClusterNode, List<SnapshotMetadata>> clusterMetas;
+
+        /** Creates operation context. */
+        private SnapshotCheckContext(SnapshotCheckProcessRequest req) {
+            this.req = req;
+        }
+    }
+
+    /** A DTO used to transfer nodes' results for the both phases. */
+    private static final class CheckResultDTO implements Serializable {
```
timoninmaxim, line 463: `SnapshotOperationResponse` by analogue with `IgniteSnapshotManager#SnapshotOperationResponse`

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,485 @@
...
+    /** Phase 2 beginning.  */
+    private IgniteInternalFuture<CheckResultDTO> validateParts(SnapshotCheckProcessRequest req) {
+        GridFutureAdapter<?> clusterOpFut = clusterOpFuts.get(req.requestId());
+
+        if (clusterOpFut == null && !req.nodes().contains(kctx.localNodeId()))
+            return new GridFinishedFuture<>();
+
+        SnapshotCheckContext ctx = context(null, req.requestId());
+
+        assert ctx != null;
```
timoninmaxim, line 180: May be null, if onNodeLeft invoked before this phase started.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,485 @@
...
+    private void reducePreparationAndMetasCheck(
+        UUID reqId,
+        Map<UUID, CheckResultDTO> results,
+        Map<UUID, Throwable> errors
+    ) {
+        String snpName = snpName(results);
+
+        SnapshotCheckContext ctx = context(snpName, reqId);
+
+        // The context is not stored in the case of concurrent check of the same snapshot but the operation future is registered.
```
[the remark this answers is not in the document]
Vladsz83: The second phase cannot start if an error occurs on the first phase. There cannot be case when clusterOpFut exists but no context registered.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,485 @@
...
+        assert ctx != null;
+        assert ctx.req.reqId.equals(req.reqId);
+
+        ctx.fut = new GridFutureAdapter<>();
+
+        // Excludes non-baseline initiator.
+        if (!baseline(kctx.localNodeId()))
+            return new GridFinishedFuture<>();
+
+        // Local meta might be null if current node started after the snapshot creation or placement.
```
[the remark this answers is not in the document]
Vladsz83: For example `testRestoreFromAnEmptyNode()` covers this case. Or start node1, node 2, create snapshot, start node 3, change baseline, launch snapshot check. Node 3 has no any the snapshot data.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
Vladsz83: No. In `SnapshotManager` the error might appear.

timoninmaxim: This method defined in this class, and it's used in single pattern here. It looks like `IgniteSnapshotManager` needs own method for its specific case. Also holding in mind that `IgniteSnapshotManager` will not check anything in the future and delegate it to `SnapshotCheckProcess`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
timoninmaxim: You add an element to the `requiredNodes` which is likely will be duplicated. It's quite weird to see `List`, especially when `SnapshotCheckProcessRequest` requires `Collection`. Let's explicitly use `Set`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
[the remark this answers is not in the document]
timoninmaxim: Can we use this check instead checking the baseline? It looks like that this check already covers the check above

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,479 @@
...
+        if (ctx.locMeta == null)
+            ctx.fut.onDone();
+        else {
+            File snpDir = kctx.cache().context().snapshotMgr().snapshotLocalDir(req.snapshotName(), req.snapshotPath());
```
timoninmaxim, line 189: `snpDir = ctx.locMeta.folderName()`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,479 @@
...
+        Iterator<Map.Entry<String, SnapshotCheckContext>> it = contexts.entrySet().iterator();
+
+        while (it.hasNext()) {
+            SnapshotCheckContext ctx = it.next().getValue();
+
+            if (!ctx.req.nodes().contains(nodeId))
+                continue;
+
+            if (ctx.fut != null)
+                ctx.fut.onDone(err);
```
timoninmaxim, line 104: You have a case when phase 1 ends, phase 2 starts (get context but not set future), node fail. In this case `clusterOpFut` will return fast but phase 2 still runs in background.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,479 @@
...
+            // Operation node is the initiator collecting the final check operation result.
+            if (clusterOpFut != null) {
+                Map<ClusterNode, Exception> errors0 = mapErrors(errors);
+
+                if (!F.isEmpty(results)) {
+                    assert results.values().stream().noneMatch(res -> res != null && res.metas != null);
+
+                    Map<ClusterNode, Map<PartitionKeyV2, PartitionHashRecordV2>> results0 = mapPartsHashes(results, ctx.req.nodes());
+
+                    IdleVerifyResultV2 chkRes = SnapshotChecker.reduceHashesResults(results0, errors0);
```
timoninmaxim, line 159: It might throw error and then `clusterOpFut` will not finish.

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,479 @@
...
+                // which must be validated.
+                if (ctx.req.nodes().contains(nodeId) && baseline(nodeId)) {
+                    assert nodeRes != null && nodeRes.partsHashes == null;
+
+                    metas.put(kctx.cluster().get().node(nodeId), nodeRes.metas);
+                }
+            });
+
+            Map<ClusterNode, Exception> metasCheck = SnapshotChecker.reduceMetasResults(ctx.req.snapshotName(), ctx.req.snapshotPath(),
+                metas, null, kctx.cluster().get().localNode().consistentId());
```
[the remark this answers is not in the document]
Vladsz83: It doesn't know which node is used to check. Probably not a node at all, but an utility application call.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
Vladsz83: It is not a full path to read.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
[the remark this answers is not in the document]
Vladsz83: > What does `placement` means?

An action which puts snapshot files before validation. Like copying from another cluster

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
Vladsz83: It doesn't declare an exception and doesn't throw it anywhere inside.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
Vladsz83: But in this case phase 2 won't work. And new checking of the same snapshot is even available. I think we can stop the future.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,596 @@
...
+        boolean check
+    ) {
+        GridFutureAdapter<SnapshotCheckResponse> fut = new GridFutureAdapter<>();
+
+        CompletableFuture<? extends Map<?, ?>> workingFut;
+
+        IgniteSnapshotManager snpMgr = kctx.cache().context().snapshotMgr();
+
+        if (allRestoreHandlers) {
+            workingFut = CompletableFuture.supplyAsync(() -> {
```
timoninmaxim, line 227: We should specify the snapshot executor for the async job

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,596 @@
...
+        return fut;
+    }
+
+    /**
+     * Call snapshot all the registered validaton handlers.
+     *
+     * @see IgniteSnapshotManager#handlers()
+     * @see #checkCustomHandlersResults(String, Map)
+     */
+    GridFutureAdapter<SnapshotCheckResponse> invokeCustomHandlers(
```
timoninmaxim, line 256: This method isn't used in this class. It looks like `SnapshotHandlerRestoreJob` should call method from the `SnapshotChecker` instead

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,596 @@
...
+        return clusterOpFut;
+    }
+
+    /** @return {@code True} if the provided node id is id of a baseline node. */
+    private boolean baseline(UUID nodeId) {
+        return CU.baselineNode(kctx.cluster().get().node(nodeId), kctx.state().clusterState());
+    }
+
+    /** Converts failure to an exception if it is not. */
+    private static Exception asException(Throwable th) {
```
timoninmaxim, line 539: Can be inlined

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,596 @@
...
+
+    /**
+     * Stops all the processes with the passed exception.
+     *
+     * @param err The interrupt reason.
+     */
+    void interrupt(Throwable err) {
+        contexts.forEach((snpName, ctx) -> {
+            if (ctx.fut != null)
+                ctx.fut.onDone(err);
```
timoninmaxim, line 133: It is possible `interrupt` will not interrupt anything. You still have a period of time:
1. First phase don't start and futures are nulls.
2. First phase finished, second phase still not started and futures are nulls.

I again recommend you create a future that lives on every node during all process and phase futures listen it.

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
Vladsz83: I'm afraid not if we use the same executor somewhere inside the task. I tried. The executor might be confgured with single thred. This thread is blocked with the waiting task. No thread left for the workers. Tests like `testChangeSnapshotTransferRateInRuntime()` hangs.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
Vladsz83: This call in in the same thread and discovery managed. We either start or reduce any phase or process this interruption. Absence of the futures during the interuption doesn't look like problem. It means only that we have nothing to stop this time. Isn't it?

Vladsz83: `interrupt()` is a node stopping. SnapshotManager stops, its thread pool stops. Discovery should not work, process phases should start and be able to work because SnapshotManager stops. No problem is expected. Also, keeping single future requires resseting it. The same race. Canceling done future won't do anything. And the `reset()` would revive and restart it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,510 @@
...
+
+        return new GridFinishedFuture<>();
+    }
+
+    /** Phase 2 beginning.  */
+    private IgniteInternalFuture<SnapshotCheckResponse> validateParts(SnapshotCheckProcessRequest req) {
+        SnapshotCheckContext ctx;
+
+        // The context can be null, if a required node leaves before this phase.
+        if (!req.nodes().contains(kctx.localNodeId()) || (ctx = context(null, req.requestId())) == null || ctx.locMeta == null)
```
timoninmaxim, line 200: does `ctx.locMeta == null` cover case `!req.nodes().contains(kctx.localNodeId())`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,510 @@
...
+                    if (err != null)
+                        ctx.fut.onDone(err);
+                    else
+                        ctx.fut.onDone(new SnapshotCheckResponse(res));
+                });
+        }
+        else {
+            File snpDir = snpMgr.snapshotLocalDir(req.snapshotName(), req.snapshotPath());
+
+            GridFutureAdapter<SnapshotCheckResponse> fut = new GridFutureAdapter<>();
```
timoninmaxim, line 219: ctx.fut?

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,510 @@
...
+        while (it.hasNext()) {
+            SnapshotCheckContext ctx = it.next().getValue();
+
+            if (!ctx.req.nodes().contains(nodeId))
+                continue;
+
+            if (ctx.fut != null)
+                ctx.fut.onDone(err);
+
+            it.remove();
```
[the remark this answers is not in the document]
Vladsz83: Because we do not store error any more. We cannot recognize that we should not work at the reduce. The phase won't work if there is no context. If it is, the phase works. Also because stop future here.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
Vladsz83: No. We've added client-initiator to the required nodes. It has no meta. Previously, the required nodes were only data nodes. Also snapshot might be placed in any manner to the baseline nodes and/or be restored from another cluster. Any node may miss snapshot meta. Test `testRestoreFromAnEmptyNode()` show case like that.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/AbstractSnapshotOperationRequest.java
```diff
@@ -0,0 +1,120 @@
...
+     * @param reqId Request ID.
+     * @param snpName Snapshot name.
+     * @param snpPath Snapshot directory path.
+     * @param grps List of cache group names.
+     * @param incIdx Incremental snapshot index.
+     * @param nodes IDs of the nodes that must be alive to complete the operation.
+     */
+    protected AbstractSnapshotOperationRequest(
+        UUID reqId,
+        @Nullable UUID opNodeId,
```
timoninmaxim, line 71: Useless param

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/AbstractSnapshotOperationRequest.java
```diff
@@ -0,0 +1,120 @@
...
+    /** Serial version uid. */
+    private static final long serialVersionUID = 0L;
+
+    /** Request ID. */
+    @GridToStringInclude
+    protected final UUID reqId;
+
+    /** Snapshot name. */
+    @GridToStringInclude
+    protected final String snpName;
```
timoninmaxim, line 43: Let's use either protected fields, either getters with private fields. Same for descending classes

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotChecker.java
```diff
@@ -536,19 +628,138 @@ private boolean hasExpiringEntries(
...
+            if (meta.dump())
+                return checkDumpFiles(snpDir, meta, grps, locNode.consistentId(), checkParts, skipPartsHashes);
+
+            try {
+                return checkSnapshotFiles(snpDir, grps, meta, forCreation, checkParts, skipPartsHashes);
+            }
+            catch (IgniteCheckedException e) {
+                throw new IgniteException("Failed to check partitions of snapshot '" + meta.snapshotName() + "'.", e);
+            }
+        });
```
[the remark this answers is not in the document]
Vladsz83: Nope. The check funtions uses it, the same executor. If it configured with just 1 thread (`setSnapshotThreadPoolSize(1)`), we'll freeze here. Tests like `testChangeSnapshotTransferRateInRuntime()` would hang.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotChecker.java
```diff
@@ -218,73 +221,109 @@ public <T> T readFromFile(File smf)
         }
     }
 
-    /** */
-    public List<SnapshotMetadata> checkLocalMetas(File snpFullPath, @Nullable Collection<Integer> cacheGrpIds,
-        @Nullable Object locNodeConsistId) {
-        List<SnapshotMetadata> snpMetas = readSnapshotMetadatas(snpFullPath, locNodeConsistId);
+    /** Launches local metas checking and waits for the result, handles execution exceptions. */
+    public List<SnapshotMetadata> checkLocalMetasResult(File snpPath, @Nullable Collection<Integer> grpIds, @Nullable Object locNodeCstId) {
```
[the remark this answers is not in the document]
Vladsz83: > Do we actually need this method, I suppose invoker can call `.get()` itself

Fixed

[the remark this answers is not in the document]
Vladsz83: > Also use a timeout for the `get()`.

I'n afraid, user cannot pass any timeout on snapshot checking. Even impossible to cancel it yet.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,508 @@
...
+    private Map<ClusterNode, Map<PartitionKeyV2, PartitionHashRecordV2>> mapPartsHashes(
+        @Nullable Map<UUID, SnapshotCheckResponse> results,
+        Collection<UUID> requiredNodes
+    ) {
+        if (F.isEmpty(results))
+            return Collections.emptyMap();
+
+        // A not required node can leave the cluster and its result can be null.
+        return results.entrySet().stream()
+            .filter(e -> requiredNodes.contains(e.getKey()) && e.getValue() != null)
```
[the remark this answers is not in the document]
Vladsz83: No. Results can contain nulls, come from not required, not a data nodes. NPEs will arise.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,508 @@
...
+    Map<String, SnapshotCheckContext> contexts() {
+        return Collections.unmodifiableMap(contexts);
+    }
+
+    /**
+     * Stops all the processes with the passed exception.
+     *
+     * @param err The interrupt reason.
+     */
+    void interrupt(Throwable err) {
```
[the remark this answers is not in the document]
Vladsz83: This method shoud be be renamed to 'onStop()'. Everything is stopping. Snapshot manager is stopping, thread pools is stopping. Discovery should not accept or process messages. No another check process must be able to start. Even it starts, it must not be able to work due to stopping thread pools. No problems are excpected.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
```diff
@@ -0,0 +1,501 @@
...
+        if (!req.nodes().contains(kctx.localNodeId()))
+            return new GridFinishedFuture<>();
+
+        if (kctx.isStopping())
+            return new GridFinishedFuture<>(new NodeStoppingException("The node is stopping: " + kctx.localNodeId()));
+
+        SnapshotCheckContext ctx = contexts.computeIfAbsent(req.snapshotName(), snpName -> new SnapshotCheckContext(req));
+
+        if (!ctx.req.requestId().equals(req.requestId())) {
+            return new GridFinishedFuture<>(new IllegalStateException("Validation of snapshot '" + req.snapshotName()
```
timoninmaxim, line 269: Just throw exception here

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotChecker.java
```diff
@@ -536,24 +618,145 @@ private boolean hasExpiringEntries(
...
+    public Map<PartitionKeyV2, PartitionHashRecordV2> checkPartitionsResult(
+        SnapshotMetadata meta,
+        File snpDir,
+        @Nullable Collection<String> groups,
+        boolean forCreation,
+        boolean checkParts,
+        boolean skipPartsHashes
+    ) {
+        try {
+            return checkPartitions(meta, snpDir, groups, forCreation, checkParts, skipPartsHashes).get();
```
timoninmaxim, line 631: Delegate `get()` call to an invoker. Checker should have only one method that checks partitions.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotCheckProcess.java
Vladsz83: We've already discussed that. We do not throws exceptions on another DPs. We return futures. Additional error messages are posted in the log in case of direct exception in Distributed process:

` catch (Throwable err) {
                U.error(log, "Failed to handle InitMessage [id=" + p.id + ']', err);
`

PR #11423, opened 2024-07-08
Dedicated Snapshot Checker

Vladsz83:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotChecker.java
```diff
@@ -0,0 +1,698 @@
...
+        if (!F.isEmpty(baselineNodes) && F.isEmpty(knownExceptions)) {
+            throw new IgniteException("No snapshot metadatas found for the baseline nodes " +
+                "with consistent ids: " + String.join(", ", baselineNodes));
+        }
+
+        return resultExceptions;
+    }
+
+    /** */
+    protected IgniteBiTuple<Map<Integer, File>, Set<File>> preparePartitions(SnapshotMetadata meta, Collection<Integer> grps, File snpDir) {
```
timoninmaxim, line 330: Why is it protected?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotChecker.java
```diff
@@ -0,0 +1,698 @@
...
+        boolean skipHash,
+        boolean punchHoleEnabled
+    ) throws IgniteCheckedException {
+        IgniteBiTuple<Map<Integer, File>, Set<File>> grpAndPartFiles = preparePartitions(meta, grpIds, snpDir);
+
+        Map<PartitionKeyV2, PartitionHashRecordV2> res = new ConcurrentHashMap<>();
+        ThreadLocal<ByteBuffer> buff = ThreadLocal.withInitial(() -> ByteBuffer.allocateDirect(meta.pageSize())
+            .order(ByteOrder.nativeOrder()));
+
+        GridKernalContext snpCtx = IgniteSnapshotManager.createStandaloneKernalContext(log, compression, snpDir, meta.folderName());
```
timoninmaxim, line 389: There is already GridKernalContext in the class. I suppose user of the class is responsible for providing the context

modules/core/src/main/java/org/apache/ignite/internal/management/snapshot/SnapshotCreateCommandArg.java
```diff
@@ -67,6 +72,17 @@ public class SnapshotCreateCommandArg extends IgniteDataTransferObject {
         dest = U.readString(in);
         sync = in.readBoolean();
         incremental = in.readBoolean();
+        onlyPrimary = in.readBoolean();
+    }
+
+    /** */
+    public boolean onlyPrimary() {
```
timoninmaxim, line 79: There is not only the test, it also changes this class, let's move this test to a separate ticket

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotChecker.java
```diff
@@ -0,0 +1,698 @@
...
+    /** */
+    public SnapshotChecker(
+        GridKernalContext kctx,
+        Marshaller marshaller,
+        ExecutorService executorSrvc,
+        @Nullable ClassLoader marshallerClsLdr
+    ) {
+        this.kctx = kctx;
+
+        this.marshaller = marshaller;
```
timoninmaxim, line 129: It looks like `marshaller` and `marshallerClsLdr` is used only in single method. Should we move then this params from constructor to those methods?

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/snapshot/SnapshotChecker.java
Vladsz83: Not sure. This single method is used in other `read-smth` methods like `readSnapshotMetadata()` where the params would be required too. Looks unhandy.

PR #11448, opened 2024-07-22
IGNITE-22776 : SQL table statistic is updated only once.

Vladsz83:


modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/stat/StatisticsObsolescenceTest.java
```diff
@@ -18,22 +18,154 @@
...
+
+                while (!stop.get()) {
+                    op.accept(key.getAndIncrement());
+
+                    if (++opCnt == workingRowsNum / 3) {
+                        opCnt = 0;
+
+                        enoughDataChanged.set(true);
+
+                        assertTrue(waitForCondition(() -> stop.get() || !enoughDataChanged.get(), getTestTimeout()));
```
alex-plekhanov, line 114: I think `CyclicBarrier` is more straitforward way to synchronize two threads here. 
For example we can replace modifiing thread as:
```
                while (!barrier.isBroken()) {
                    op.accept(key.getAndIncrement());

                    if (++opCnt == workingRowsNum / 3) {
                        opCnt = 0;

                        barrier.await();
                        barrier.await();
                    }
                }
```
And remove `enoughDataChanged` and `stop` variables, and several `waitForCondition` calls.
In checking thread we can place 4 `barrier.await()` methods and `barrier.reset()` in the `finally` block.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/stat/StatisticsObsolescenceTest.java
```diff
@@ -18,22 +18,154 @@
...
+                    }
+                }
+            });
+
+            assertTrue(waitForCondition(enoughDataChanged::get, getTestTimeout()));
+
+            waitForStatsUpdates(initStat1, osbInterval);
```
alex-plekhanov, line 121: I'm afraid in some cases `osbInterval` will be not enough. For example, you start to wait for statistics update, but in this time check task is executed and there are less than 15% of rows changed. Next check task will be executed after `osbInterval`, and  `waitForStatsUpdates` will fail.
How about waiting for `osbInterval * 2`?

alex-plekhanov (approved):

PR #11518, opened 2024-09-08
IGNITE-23114 JDBC thin: add support for stream APIs for BLOB

skorotkov:

### Notes


modules/core/src/main/java/org/apache/ignite/internal/jdbc2/lob/JdbcBlobBuffer.java
```diff
@@ -0,0 +1,354 @@
...
+     * Provide OutputStream through which the data can be written to buffer starting from
+     * the (zero-based) position.
+     *
+     * @param pos The zero-based offset to the first byte to be written. Must not be negative
+     *            or greater than total count of bytes in buffer.
+     *
+     * @return OutputStream instance.
+     */
+    public OutputStream getOutputStream(long pos) {
+        long totalCnt = storage.totalCnt();
```
nizhikov, line 107: Looks like sanity checks double. Here and in

`JdbcBlob#setBytes`, `JdbcBlob#setBinaryStream` where it's used.
Can we keep only one copy of sanity checks?

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/lob/JdbcBlobBuffer.java
```diff
@@ -0,0 +1,354 @@
...
+     * Provide InputStream through which the no more than {@code len} bytes can be read
+     * from buffer starting from the specified zero-based position {@code pos}.
+     *
+     * @param pos The zero-based offset to the first byte to be retrieved. Must not be negative
+     *            or greater than total count of bytes in buffer.
+     * @param len The length in bytes of the data to be retrieved. Must not be negative.
+     * @return InputStream instance.
+     */
+    public InputStream getInputStream(long pos, long len) {
+        if (len < 0)
```
nizhikov, line 139: Looks like sanity checks double. Here and in

`JdbcBlob#getBytes`, `JdbcBlob#getBinaryStream` where it's used.
Can we keep only one copy of sanity checks?

modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java
```diff
@@ -1880,6 +1892,72 @@ public void writeFieldIdNoSchemaUpdate(int fieldId) {
...
+    private int doWriteByteArrayFromInputStream(InputStream in, int limit) throws IOException {
+        out.unsafeWriteByte(GridBinaryMarshaller.BYTE_ARR);
+
+        int lengthPos = out.position();
+        out.position(lengthPos + 4);
+
+        int readLen;
+        int writtenLen = 0;
+
+        byte[] buf = new byte[DEFAULT_BUFFER_SIZE];
```
nizhikov, line 1942: byte[] buf = new byte[Math.min(limit, DEFAULT_BUFFER_SIZE)];

modules/core/src/main/java/org/apache/ignite/internal/binary/streams/BinaryAbstractOutputStream.java
```diff
@@ -29,14 +29,17 @@ public abstract class BinaryAbstractOutputStream extends BinaryAbstractStream
     /** Minimal capacity when it is reasonable to start doubling resize. */
     private static final int MIN_CAP = 256;
 
+    /** Max capacity when it is still reasonable doubling resize. */
+    protected static final int MAX_CAP = 32 * 1024 * 1024;
```
nizhikov, line 33: This changes looks like overcomplication without clear performance benefits.
Do we really need them? 
Can we revert it here and move to the separate PR only if there are clear real-worls benchmarks that show some advantage of new restriction

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
```diff
@@ -246,12 +278,73 @@ else if (cls == Timestamp[].class)
             writer.writeTimestampArray((Timestamp[])obj);
         else if (cls == java.util.Date[].class || cls == java.sql.Date[].class)
             writer.writeDateArray((java.util.Date[])obj);
+        else if (obj instanceof SqlInputStreamWrapper) {
+            try {
```
nizhikov, line 282: Let's move exception handling inside `writeInputStreamAsByteArray` method.

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
```diff
@@ -246,12 +278,73 @@ else if (cls == Timestamp[].class)
...
+                writeInputStreamAsByteArray(writer, (SqlInputStreamWrapper)obj);
+            }
+            catch (IOException e) {
+                throw new BinaryObjectException(e);
+            }
+        }
+        else if (obj instanceof Blob) {
+            try {
+                writeBlobAsByteArray(writer, (Blob)obj);
```
nizhikov, line 291: Let's move exception handling inside `writeBlobAsByteArray` method.

modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java
```diff
@@ -1880,6 +1892,72 @@ public void writeFieldIdNoSchemaUpdate(int fieldId) {
...
+        }
+
+        int savedPos = out.position();
+
+        out.position(lengthPos);
+        out.unsafeWriteInt(writtenLen);
+
+        out.position(savedPos);
+
+        return writtenLen;
```
nizhikov, line 1958: Looks like all usages of `doWriteByteArrayFromInputStream` check that `limit` equal to `writtenLen` so we can move check inside the method.

modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java
```diff
@@ -1880,6 +1892,72 @@ public void writeFieldIdNoSchemaUpdate(int fieldId) {
...
+     *
+     * @param in InputStream.
+     * @return Number of bytes written or -1 if stream contains more than {@code MAX_ARRAY_SIZE} bytes.
+     */
+    public int writeByteArrayFromInputStream(InputStream in) throws IOException {
+        out.unsafeEnsure(1 + 4);
+
+        int writtenLen = doWriteByteArrayFromInputStream(in, MAX_ARRAY_SIZE);
+
+        if (writtenLen == MAX_ARRAY_SIZE && in.read() != -1)
```
nizhikov, line 1919: This method writes `InputStream` from the user and can block forever on `in.read()` invocation.

modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java
```diff
@@ -1880,6 +1892,72 @@ public void writeFieldIdNoSchemaUpdate(int fieldId) {
...
+
+        int lengthPos = out.position();
+        out.position(lengthPos + 4);
+
+        int readLen;
+        int writtenLen = 0;
+
+        byte[] buf = new byte[DEFAULT_BUFFER_SIZE];
+
+        while (-1 != (readLen = in.read(buf, 0, Math.min(buf.length, limit - writtenLen)))
```
nizhikov, line 1944: we need to exchange order of checks.
First, check `writtenLen < limit` then read and check for success.
Looks like, now, you will try to read one more time (zero byte) when `writtenLen < limit == false`.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/lob/JdbcBlobStorageContext.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.jdbc2.lob;
+
+/**
+ * Additional storage-specific data to be saved in the {@link JdbcBlobBufferPointer}.
+ */
+interface JdbcBlobStorageContext {
```
nizhikov, line 23: We can use implementation everywheren as interface has only one.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/lob/JdbcBlobBufferPointer.java
```diff
@@ -0,0 +1,82 @@
...
+
+        return this;
+    }
+
+    /**
+     * Set context.
+     *
+     * @param context New context.
+     */
+    JdbcBlobBufferPointer setContext(JdbcBlobStorageContext context) {
```
nizhikov, line 67: There are two usage of method both with zeroed context.
Let's simplify this code and move InMemContext#idx, InMemContext#inBufPos inside JdbcBlobBuffer


```
JdbcBlobMemoryStorage {
...
    /** {@inheritDoc} */
    @Override public JdbcBlobBufferPointer createPointer() {
        return new JdbcBlobBufferPointer().setContext(new InMemContext(0, 0));
    }

...
    private InMemContext recoverContext(JdbcBlobBufferPointer pointer) {
        long pos = pointer.getPos();

        pointer.setContext(new InMemContext(0, 0));
        pointer.setPos(0);

        advance(pointer, pos);

        return (InMemContext)pointer.getContext();
    }
```
}
```

Comment (skorotkov) [reply modules/core/src/main/java/org/apache/ignite/internal/jdbc2/lob/JdbcBlobBufferPointer.java:67]: InMemContext#idx, InMemContext#inBufPos can not be moved inside the JdbcBlobBuffer.  

The Context is stored in some JdbcBlobBufferPointer.  JdbcBlobBufferPointer keeps some particular position in the JdbcBlobBuffer and transparently contains additional storage-specific context for fast access to data. Additional context is needed for RW storage (idx + inBufPos) but is not needed for the RO one. That is why there is a single implementation of context.

There may be **many** pointers pointing to **different** positions within the single JdbcBlobBuffer.  This is because several independent InputStreams and OutputStreams may be created for Blob each containing the its own JdbcBlobBufferPointer.

Keeping in mind the copy-on-write behaviour of JdbcBlobBuffer the underlining storage may be changed from RO to RW after the stream (and therefore the pointer) was created.  Streams should remain valid and be transparent to the RO-2-RW switch, so the existing pointers should be able to be "upgraded" to work with the RW storage. The recoverContext() function does that.

***
So, the only simplification comes to mind is to move the idx and inBufPos directly to JdbcBlobBufferPointer.  That way we remove the JdbcBlobStorageContext interface and InMemContext implementation.  But broke a little bit the OOD priciples  - now the JdbcBlobBufferPointer  will "know" about two specific implementations of storage interface.

Doing so we may also replace the JDbcBlobStorage interface with just the abstract base class for two remaining in-memory storage implementations.

***

What do you think?

Comment (nizhikov) [reply modules/core/src/main/java/org/apache/ignite/internal/jdbc2/lob/JdbcBlobBufferPointer.java:67]: > So, the only simplification comes to mind is to move the idx and inBufPos directly to JdbcBlobBufferPointer.

Yes, you are correct - I meant moving two variables to `JdbcBlobBufferPointer`.

File: modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java (hunk 8 above)
Comment (skorotkov) [reply modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java:1912]: Generally any call to InputStream in all other new stream-related methods can block as well.

I think it's not a problem if method is called from the Thin JDBC Driver only (which is true for now). Since it's a user application thread would block.

On the side if (in future) some server code would call these methods it may be a problem.

***

Reading asyncroniously with timeout looks an overkill for me for the Thin JDBC driver case.  May be just take these methods outside of the BinaryWriterExImpl to JDBC-specific part (to  SqlListenerUtils).

Initially I did so but decided that it's better to have the serialization code at the same file.

***
What is your preference for this?

Comment (nizhikov) [reply modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java:1912]: > Generally any call to InputStream in all other new stream-related methods can block as well.

Why do we need to read extra byte here, in the first place?
My point is that `in.read() != -1` are useless here.
I think we must  handle "too much data" case differently - throw from the `doWriteByteArrayFromInputStream` maybe?

Comment (skorotkov) [reply modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java:1912]: This is a case when user didn't provide the length of data in the stream vai the API.

So we accept any actual length <= `MAX_ARRAY_SIZE` and thus need to check that there is no more data in stream if `MAX_ARRAY_SIZE` was achieved.

In any case whether here or in `doWriteByteArrayFromInputStream` we still need to do extra `in.read()` to be sure that no more data present in stream.

Moreove in the `doWriteByteArrayFromInputStream` it's supposed that we are not allowed to read more bytes (and move stream pointer) to hadle case what used does pass the length of data.

Or may be I just do not see the obvious solution ?

Comment (skorotkov) [reply modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java:1912]: 1. To throw exception to user saying that streams longer then `MAX_ARRAY_SIZE` are not supported. Instead of silently strip the tail.
2. Hm, not sure how to answer - I see no other way

Comment (nizhikov) [reply modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java:1912]: > To throw exception to user saying that streams longer then MAX_ARRAY_SIZE are not supported. Instead of silently strip the tail.

We can do it when we read more then `MAX_ARRAY_SIZE` bytes from stream while writing data inside `doWriteByteArrayFromInputStream`.

Comment (skorotkov) [reply modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java:1912]: Either read 1 byte and expect -1, or read array and expect zero. In any case extra call.

Comment (skorotkov) [reply modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java:1912]: I think we can not.  

Since `doWriteByteArrayFromInputStream` is supposed to not read more than limit. No single byte.

I can create another function which would be almost the same but allowed to (try) read more data then limit. In this function the extra read call will only be performed if actual data length is multiple of DEFAULT_BUFFER_SIZE.

 `doWriteByteArrayFromInputStream`  will be used for case for known (declared) length

New  function will be used for case if length is not provided.

Comment (nizhikov) [reply modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java:1912]: > Since doWriteByteArrayFromInputStream is supposed to not read more than limit. No single byte.

But you will read extra byte `if (writtenLen == MAX_ARRAY_SIZE && in.read() != -1)` here.

Comment (skorotkov) [reply modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java:1912]: only in case if user didn't provide stream length via API call
so we are free to read as we want

Comment (nizhikov) [reply modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java:1912]: Looks like with this version we can still have one method and fail if read more then `MAX_ARRAY_SIZE` from InputStream. Logic relies on `-1` as unlimited read flag.

```
    /**
     * Write byte array from the InputStream with uknown length.
     *
     * @param in InputStream.
     * @return Number of bytes written or -1 if stream contains more than {@code MAX_ARRAY_SIZE} bytes.
     */
    public int writeByteArrayFromInputStream(InputStream in) throws IOException {
        out.unsafeEnsure(1 + 4);

        return doWriteByteArrayFromInputStream(in, -1);
    }

    /**
     * Write byte array from the InputStream. No more than {@code limit} bytes will be written.
     *
     * @param in InputStream.
     * @param limit Max length of data to be read from the stream.
     * @return Number of bytes written.
     * @throws IOException If an I/O error occurs.
     */
    private int doWriteByteArrayFromInputStream(InputStream in, int limit) throws IOException {
        out.unsafeWriteByte(GridBinaryMarshaller.BYTE_ARR);

        int lengthPos = out.position();

        out.position(lengthPos + 4);

        int written = 0;

        byte[] buf = new byte[Math.min(limit, DEFAULT_BUFFER_SIZE)];

        while (true) {
            int read = limit > 0
                ? in.read(buf, 0, Math.min(buf.length, limit - written))
                : in.read(buf, 0, buf.length);

            if (read == -1)
                break;

            if (read + written > MAX_ARRAY_SIZE)
                throw new IgniteException("Too much data. Can't write more then " + MAX_ARRAY_SIZE + " bytes from stream");

            out.writeByteArray(buf, 0, read);

            written += read;

            if (written >= limit)
                break;
        }

        out.position(lengthPos);
        out.unsafeWriteInt(written);
        out.position(out.position() + written);

        return written;
    }
```

Comment (skorotkov) [reply modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryWriterExImpl.java:1912]: Okay

File: modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinResultSet.java [hunk 12]
```diff
@@ -699,9 +719,9 @@ else if (cls == String.class) {
 
     /** {@inheritDoc} */
     @Override public InputStream getBinaryStream(String colLb) throws SQLException {
-        ensureNotClosed();
+        int colIdx = findColumn(colLb);
```
nizhikov, line 722: Variable can be inlined

`return getBinaryStream(findColumn(colLb)`

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinResultSet.java
```diff
@@ -1298,12 +1318,26 @@ else if (cls == String.class || cls == Character.class) {
 
     /** {@inheritDoc} */
     @Override public Blob getBlob(int colIdx) throws SQLException {
-        return new JdbcBlob(getBytes(colIdx));
+        Object val = getValue(colIdx);
+
+        if (val == null)
+            return null;
+
+        if (val instanceof JdbcBlobBuffer) {
```
nizhikov, line 1326: Let's simplify a bit here:

```
     if (!(val instanceof JdbcBlobBuffer)) 
            throw new SQLException("Cannot convert to Blob [colIdx=" + colIdx + "]");

        return new JdbcBlob(JdbcBlobBuffer.shallowCopy((JdbcBlobBuffer)val));
```

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinResultSet.java
```diff
@@ -1327,12 +1361,16 @@ else if (cls == String.class || cls == Character.class) {
 
     /** {@inheritDoc} */
     @Override public Blob getBlob(String colLb) throws SQLException {
-        return new JdbcBlob(getBytes(colLb));
+        int colIdx = findColumn(colLb);
```
nizhikov, line 1364: colIdx can be inlined.

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinPreparedStatement.java
```diff
@@ -446,7 +446,17 @@ private void executeWithArguments(JdbcStatementType stmtType) throws SQLExceptio
     @Override public void setBinaryStream(int paramIdx, InputStream x, long length) throws SQLException {
         ensureNotClosed();
 
-        throw new SQLFeatureNotSupportedException("Streams are not supported.");
+        if (length < 0)
+            throw new SQLException("Invalid argument. Length should be greater than 0.");
+
+        if (length > MAX_ARRAY_SIZE)
+            throw new SQLFeatureNotSupportedException("Invalid argument. InputStreams with length > " + MAX_ARRAY_SIZE +
```
nizhikov, line 453: with length greater then ...

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinResultSet.java
```diff
@@ -587,9 +598,18 @@ else if (cls == String.class) {
...
-        ensureNotClosed();
+        Object val = getValue(colIdx);
+
+        if (val == null)
+            return null;
+
+        if (val instanceof JdbcBlobBuffer)
+            return ((JdbcBlobBuffer)val).getInputStream();
+        else {
+            byte[] bytes = getBytes(colIdx);
```
nizhikov, line 609: `bytes` can be inlinded.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/lob/JdbcBlobStorageContext.java
nizhikov: This intefrace should be removed, or you can move logic from a single implementation here and use it everywhere.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/lob/JdbcBlobReadOnlyMemoryStorage.java
```diff
@@ -0,0 +1,115 @@
...
+
+import java.io.IOException;
+import org.apache.ignite.internal.util.typedef.internal.U;
+
+/**
+ * Heap memory based read-only implementation of the {@link JdbcBlobStorage} which wraps
+ * a part of an externally provided byte array which can not be modified.
+ *
+ * <p>Says, allows direct read-only access to binary data stored in the incoming JDBC
+ * responce message without copying.
```
nizhikov, line 28: responce -> response

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/lob/JdbcBlobMemoryStorage.java
```diff
@@ -0,0 +1,323 @@
...
+ */
+class JdbcBlobMemoryStorage implements JdbcBlobStorage {
+    /** Max capacity when it is still reasonable to double size for new buffer. */
+    private static final int MAX_CAP = 32 * 1024 * 1024;
+
+    /** The list of buffers. */
+    private List<byte[]> buffers = new ArrayList<>();
+
+    /** The total number of bytes in all buffers. */
+    private long totalCnt;
```
nizhikov, line 39: We know for sure that `totalCnt <= MAX_ARRAY_SIZE`, isn't it?
So, we want to use `int` variables to track position and data length everywhere internally.

Let's use `int` here and in other places, it should simplify code.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/lob/JdbcBlobReadWriteStorage.java
```diff
@@ -0,0 +1,275 @@
...
+        }
+    }
+
+    /** {@inheritDoc} */
+    @Override public long totalCnt() {
+        return totalCnt;
+    }
+
+    /** {@inheritDoc} */
+    @Override public JdbcBlobBufferPointer createPointer() {
```
nizhikov, line 64: Method body equal to the parent. Can be removed.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/lob/JdbcBlobBufferPointer.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+package org.apache.ignite.internal.jdbc2.lob;
+
+/**
+ * Keeps a pointer to some position in a {@link JdbcBlobBuffer}.
+ */
+class JdbcBlobBufferPointer {
+    /** Current buffer position. */
+    private long pos;
```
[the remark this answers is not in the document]
nizhikov: Can we migrate `pos` to the int ?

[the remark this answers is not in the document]
skorotkov: Yes, I migrating to `int` now as you suggested in other comment.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/lob/JdbcBlobBuffer.java
```diff
@@ -0,0 +1,333 @@
...
+     * @param storage Storage instance.
+     */
+    private JdbcBlobBuffer(JdbcBlobStorage storage) {
+        this.storage = storage;
+    }
+
+    /**
+     * @return Total number of bytes in the buffer.
+     */
+    public long totalCnt() {
```
nizhikov, line 93: Le't rename this and all related variables to `length`, 'len` or `size` as this naming commonly used in API we working on.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/lob/JdbcBlobBuffer.java
```diff
@@ -0,0 +1,333 @@
...
+
+    /**
+     * Truncate the buffer to the specified length.
+     *
+     * @param len The length in bytes to be truncated. Must not be negative
+     *            or greater than total count of bytes in buffer.
+     */
+    public void truncate(long len) throws IOException {
+        try {
+            storage.truncate(len);
```
nizhikov, line 144: Logic based on exception must be omited.
Let's rework this as following:

```
if (storage.readOnly())
    switchToReadWriteMemoryStorage();

storage.truncate(len);
```

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/lob/JdbcBlobReadWriteStorage.java
```diff
@@ -0,0 +1,275 @@
...
+    /** {@inheritDoc} */
+    @Override public void write(JdbcBlobBufferPointer pos, int b) throws IOException {
+        if (pos.getPos() >= totalCnt + 1)
+            throw new IOException("Writting beyond end of Blob, it probably was truncated after OutputStream was created " +
+                    "[pos=" + pos.getPos() + ", totalCnt=" + totalCnt + "]");
+
+        byte[] buf = getBuf(pos);
+
+        if (buf == null)
+            buf = addNewBuffer(1);
```
nizhikov, line 117: We don't want to create byte[1] to write data.
Let's create new array of reasonable size here.

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinPreparedStatement.java
```diff
@@ -528,7 +539,8 @@ private void executeWithArguments(JdbcStatementType stmtType) throws SQLExceptio
     private void setArgument(int paramIdx, Object val) throws SQLException {
         ensureNotClosed();
 
-        if (val != null && !SqlListenerUtils.isPlainType(val.getClass()))
+        if (val != null && !SqlListenerUtils.isPlainType(val.getClass())
+                && !(val instanceof SqlInputStreamWrapper) && !(val instanceof Blob))
```
nizhikov, line 543: Looks like additional checks must be inside `isPlainType` method.

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlInputStreamWrapper.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+package org.apache.ignite.internal.processors.odbc;
+
+import java.io.InputStream;
+
+/**
+ * InputStream wrapper used to pass it as argument to PreparedStatement.
+ */
+public class SqlInputStreamWrapper {
```
nizhikov, line 25: Used only from `JdbcThinPreparedStatement`.
Must be moved there and made package private.

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlInputStreamWrapper.java
```diff
@@ -0,0 +1,64 @@
...
+        stream = inputStream;
+        len = null;
+    }
+
+    /**
+     * Returns input stream for the enclosed data.
+     *
+     * @return Input stream.
+     */
+    public InputStream getInputStream() {
```
nizhikov, line 54: get'ers for internal classes must have `inputStream` and `length` names.
See - https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines#CodingGuidelines-Gettersandsetters

> Getters and setters should be implemented without is/get/set prefixes in internal APIs.

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlInputStreamWrapper.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * InputStream wrapper used to pass it as argument to PreparedStatement.
+ */
+public class SqlInputStreamWrapper {
+    /** Input stream wrapped. */
+    private final InputStream stream;
+
+    /** Length of data in the input stream. May be null if unknown. */
+    private final Integer len;
```
nizhikov, line 30: Let's use `int` type and `-1` for unknown length.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,463 @@
...
+ */
+public class JdbcBinaryBuffer {
+    /** Byte array storing data. */
+    private byte[] arr;
+
+    /** Offset the data starts in the array. */
+    private int off;
+
+    /** The length of data. */
+    private int length;
```
nizhikov, line 51: length -> len - let's use consistent naming across all codebase.

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
```diff
@@ -246,12 +274,61 @@ else if (cls == Timestamp[].class)
...
             throw new BinaryObjectException("Custom objects are not supported");
     }
 
+    /**
+     * Write byte array from the InputStream enclosed in the stream wrapper.
+     *
+     * @param writer Writer.
+     * @param wrapper stream wrapper
+     */
+    private static void writeInputStreamAsByteArray(BinaryWriterExImpl writer, SqlInputStreamWrapper wrapper)
```
nizhikov, line 293: writeInputStreamAsByteArray -> writeInputStream

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
```diff
@@ -246,12 +274,61 @@ else if (cls == Timestamp[].class)
...
+            throws BinaryObjectException {
+        InputStream in = wrapper.getInputStream();
+        Integer len = wrapper.getLength();
+
+        if (len == null)
+            writer.writeByteArrayFromInputStream(in);
```
nizhikov, line 299: writeByteArrayFromInputStream -> `writeByteArray` or just `write`

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
```diff
@@ -246,12 +274,61 @@ else if (cls == Timestamp[].class)
...
+        else {
+            int written = writer.writeByteArrayFromInputStream(in, len);
```
nizhikov, line 301: writeByteArrayFromInputStream -> writeByteArray or just write

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
```diff
@@ -246,12 +274,61 @@ else if (cls == Timestamp[].class)
...
+     */
+    private static void writeBlobAsByteArray(BinaryWriterExImpl writer, Blob blob)
+            throws BinaryObjectException {
+        try {
+            int len = (int)blob.length();
+            InputStream in = blob.getBinaryStream(1, len);
+
+            int written = writer.writeByteArrayFromInputStream(in, len);
+
+            if (len != written)
```
nizhikov, line 323: parentess for multiline `if`.

if (len != written) {
throw new BinaryObjectException("Blob length mismatch. [declaredLength=" + len + ", " +
                        "actualLength=" + written + "]");
}

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
```diff
@@ -246,12 +274,61 @@ else if (cls == Timestamp[].class)
...
+            throws BinaryObjectException {
+        InputStream in = wrapper.getInputStream();
+        Integer len = wrapper.getLength();
+
+        if (len == null)
+            writer.writeByteArrayFromInputStream(in);
+        else {
+            int written = writer.writeByteArrayFromInputStream(in, len);
+
+            if (len != written)
```
nizhikov, line 303: parentess for multiline if.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBlob.java
```diff
@@ -134,57 +160,113 @@ public JdbcBlob(byte[] arr) {
...
     @Override public void truncate(long len) throws SQLException {
         ensureNotClosed();
 
-        if (len < 0 || len > arr.length)
-            throw new SQLException("Invalid argument. Length can't be " +
-                "less than zero or greater than Blob length [len=" + len + ']');
+        int blobLen = buf.length();
 
-        arr = Arrays.copyOf(arr, (int)len);
+        if (len < 0 || len > blobLen)
```
nizhikov, line 198: parentess for multiline if.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBlob.java
```diff
@@ -134,57 +160,113 @@ public JdbcBlob(byte[] arr) {
...
     }
 
     /** {@inheritDoc} */
     @Override public OutputStream setBinaryStream(long pos) throws SQLException {
-        throw new SQLFeatureNotSupportedException();
+        ensureNotClosed();
+
+        int blobLen = buf.length();
+
+        if (pos < 1 || pos - 1 > blobLen)
```
nizhikov, line 185: parentess for multiline if.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBlob.java
```diff
@@ -134,57 +160,113 @@ public JdbcBlob(byte[] arr) {
...
-            throw new ArrayIndexOutOfBoundsException();
+        int blobLen = buf.length();
 
-        byte[] dst = arr;
-
-        if (idx + len > arr.length) {
-            dst = new byte[arr.length + (len - (arr.length - idx))];
-
-            U.arrayCopy(arr, 0, dst, 0, idx);
+        if (pos < 1 || pos - 1 > blobLen)
```
nizhikov, line 165: parentess for multiline if.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBlob.java
```diff
@@ -18,54 +18,80 @@
...
     }
 
     /** {@inheritDoc} */
     @Override public byte[] getBytes(long pos, int len) throws SQLException {
         ensureNotClosed();
 
-        if (pos < 1 || (arr.length - pos < 0 && arr.length > 0) || len < 0)
+        int blobLen = buf.length();
+
+        if (pos < 1 || (pos > blobLen && blobLen > 0) || len < 0)
```
nizhikov, line 80: parentess for multiline if.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,463 @@
...
+            if (pos > length)
+                throw new IOException("Writting beyond end of Blob, it probably was truncated after OutputStream was created " +
+                        "[pos=" + pos + ", blobLength=" + length + "]");
+
+            JdbcBinaryBuffer.this.write(pos++, b);
+        }
+
+        /** {@inheritDoc} */
+        @Override public void write(byte[] b, int off, int len) throws IOException {
+            if (pos > length)
```
nizhikov, line 454: parentess for multiline if.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,463 @@
...
+         *
+         * @param pos Starting position (zero-based).
+         */
+        private BufferOutputStream(int pos) {
+            this.pos = pos;
+        }
+
+        /** {@inheritDoc} */
+        @Override public void write(int b) throws IOException {
+            if (pos > length)
```
nizhikov, line 445: parentess for multiline if.

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
```diff
@@ -246,12 +274,61 @@ else if (cls == Timestamp[].class)
...
     }
 
+    /**
+     * Write byte array from the InputStream enclosed in the stream wrapper.
+     *
+     * @param writer Writer.
+     * @param wrapper stream wrapper
+     */
+    private static void writeInputStreamAsByteArray(BinaryWriterExImpl writer, SqlInputStreamWrapper wrapper)
+            throws BinaryObjectException {
```
nizhikov, line 294: no reason for new line for throws

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
```diff
@@ -246,12 +274,61 @@ else if (cls == Timestamp[].class)
...
+    }
+
+    /**
+     * Write byte array from the Blob instance.
+     *
+     * @param writer Writer.
+     * @param blob Blob.
+     */
+    private static void writeBlobAsByteArray(BinaryWriterExImpl writer, Blob blob)
+            throws BinaryObjectException {
```
nizhikov, line 316: no reason for new line for throws

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
```diff
@@ -246,12 +274,61 @@ else if (cls == Timestamp[].class)
...
+     *
+     * @param writer Writer.
+     * @param wrapper stream wrapper
+     */
+    private static void writeInputStreamAsByteArray(BinaryWriterExImpl writer, SqlInputStreamWrapper wrapper)
+            throws BinaryObjectException {
+        InputStream in = wrapper.getInputStream();
+        Integer len = wrapper.getLength();
+
+        if (len == null)
```
nizhikov, line 298: no reason to check if we use `-1` as "unknown length" as we did inside `BinaryWriterExImpl#doWriteByteArrayFromInputStream`, already.

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlInputStreamWrapper.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * @param inputStream Input stream.
+     * @param len Length of data in the input stream. May be null if unknown.
+     */
+    public SqlInputStreamWrapper(InputStream inputStream, Integer len) {
```
nizhikov, line 36: `inputStream` -> `stream`

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlInputStreamWrapper.java
```diff
@@ -0,0 +1,64 @@
...
+        stream = inputStream;
+        this.len = len;
+    }
+
+    /**
+     * @param inputStream Input stream.
+     */
+    public SqlInputStreamWrapper(InputStream inputStream) {
```
nizhikov, line 44: `inputStream` -> `stream`

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlInputStreamWrapper.java
```diff
@@ -0,0 +1,64 @@
...
+        stream = inputStream;
```
nizhikov, line 45: `this(stream, -1);`

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
```diff
@@ -246,12 +274,61 @@ else if (cls == Timestamp[].class)
...
+     * Write byte array from the Blob instance.
+     *
+     * @param writer Writer.
+     * @param blob Blob.
+     */
+    private static void writeBlobAsByteArray(BinaryWriterExImpl writer, Blob blob)
+            throws BinaryObjectException {
+        try {
+            int len = (int)blob.length();
+            InputStream in = blob.getBinaryStream(1, len);
```
nizhikov, line 319: `in` variable can be inlined.
Looks like we can simplify to: 
```
int written = written.writeByteArrayFromInputStream(blob.getBinaryStream(), len);
```

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,463 @@
...
+     * Truncate data in this buffer to specified length.
+     *
+     * @param len New length.
+     */
+    void truncate(int len) {
+        byte[] newArr = new byte[Math.max(MIN_CAP, len)];
+
+        U.arrayCopy(arr, off, newArr, 0, len);
+
+        arr = newArr;
```
nizhikov, line 171: Let's create a single method that copy data into new array and set `isReadOnly = false` to make sure we do it consistent.

All methods that writes to the `arr` must check `isReadOnly` flag, also.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,463 @@
...
+     * Calculate new capacity.
+     *
+     * @param curCap Current capacity.
+     * @param reqLen Required new data length.
+     * @return New capacity.
+     */
+    protected static int capacity(int curCap, int reqLen) {
+        int newCap;
+
+        if (reqLen < MIN_CAP)
```
nizhikov, line 290: reqLen <= MIN_CAP ?

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,463 @@
...
+    }
+
+    /**
+     * Calculate new capacity.
+     *
+     * @param curCap Current capacity.
+     * @param reqLen Required new data length.
+     * @return New capacity.
+     */
+    protected static int capacity(int curCap, int reqLen) {
```
nizhikov, line 287: Should we add 
```
assert reqLen > curCap
```

nizhikov: curCap -> `cap`
reqLen -> `newCap`.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
nizhikov:
```
if (reqLen <= MIN_CAP)
    return MIN_CAP;

int resCap = Math.max(cap, MIN_CAP);

while (resCap < reqCap) {
    resCap <<= 1;

    if (resCap < 0)
        return MAX_ARRAY_SIZE;
}
```

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,463 @@
...
+        int newCap;
+
+        if (reqLen < MIN_CAP)
+            newCap = MIN_CAP;
+        else {
+            newCap = Math.max(curCap, MIN_CAP);
+
+            while (newCap < reqLen) {
```
nizhikov, line 295: Infinite loop on `reqLen == MAX_ARRAY_SIZE`?

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
skorotkov: `writeByteArray` already exists, so will change to `writeInputStream`

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
skorotkov: No.  

There is a unit test for that case

https://github.com/apache/ignite/pull/11518/files#diff-a36e9a0a75d46d9f20fb71c6f9de5d7d71f3e36004416838c00bfe2aa13d8c51R37

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,454 @@
...
+    /**
+     * Provide InputStream through which the data can be read starting from the
+     * begining.
+     *
+     * <p>Stream is not limited meaning that it would return any new data
+     * written to the buffer after stream creation.
+     *
+     * @return InputStream instance.
+     */
+    public InputStream getInputStream() {
```
nizhikov, line 118: https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines#CodingGuidelines-Gettersandsetters

> Getters and setters should be implemented without is/get/set prefixes in internal APIs.

Please, check other part of your PR, also.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBlob.java
```diff
@@ -74,55 +101,56 @@ public JdbcBlob(byte[] arr) {
...
+        int blobLen = buf.length();
+
+        if (start < 1)
+            throw new SQLException("Invalid argument. Start position can't be less than 1 [start=" + start + "]");
+
+        if (start > blobLen || ptrn.length() == 0 || ptrn.length() > blobLen)
             return -1;
 
-        return position(ptrn.getBytes(1, (int)ptrn.length()), start);
+        long idx = position(ptrn.getBinaryStream(), (int)ptrn.length(), (int)start - 1);
```
nizhikov, line 151: let's move all logic inside `position` implementation to reduce code duplication.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBlob.java
```diff
@@ -74,55 +101,56 @@ public JdbcBlob(byte[] arr) {
...
-        return -1;
+        return idx == -1 ? -1 : idx + 1;
     }
 
     /** {@inheritDoc} */
     @Override public long position(Blob ptrn, long start) throws SQLException {
         ensureNotClosed();
 
-        if (start < 1 || start > arr.length || ptrn.length() == 0 || ptrn.length() > arr.length)
+        int blobLen = buf.length();
```
nizhikov, line 143: Can you, please, don't introduce `blobLen` variable here and in another parts of class as it will reduce changes and simplify review. `buf.length()` are simple enough to keep it straight.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,454 @@
...
+     */
+    int read(int pos, byte[] resBuf, int resOff, int resLen) {
+        Objects.checkFromIndexSize(resOff, resLen, resBuf.length);
+
+        if (pos >= len)
+            return -1;
+
+        int bufOff = pos + off;
+
+        int size = Math.min(resLen, len - pos);
```
nizhikov, line 196: `size` may be inlined.
`U.arrayCopy` returns number of written bytes. We can use it to return restul.

```
return U.arrayCopy(arr, pos + off, resBuf, resOff, Math.min(resLen, len - pos);
```

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,454 @@
...
+
+        len = newLen;
+    }
+
+    /**
+     * Ensure capacity.
+     *
+     * @param newLen The new data length the buffer should be able to hold.
+     */
+    private void ensureCapacity(int newLen) {
```
nizhikov, line 267: ensureCapacity -> reallocateIfRequired

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,454 @@
...
+        off = 0;
+        isReadOnly = false;
+    }
+
+    /**
+     * Input stream to read data from buffer.
+     */
+    private class BufferInputStream extends InputStream {
+        /** Stream starting position. */
+        private final int start;
```
nizhikov, line 315: We can eliminate this.

`pos` means current position in buffer.
`limit` - max position in buffer.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
skorotkov: `U.arrayCopy` doesn't return number of written bytes.

Instead it returns new offset in the result array:

```
public static int arrayCopy(byte[] src, int off, byte[] resBuf, int resOff, int len) {
    .....
    return resOff + len;
}
```

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,449 @@
...
+
+    /**
+     * Truncate data in this buffer to specified length.
+     *
+     * @param len New length.
+     */
+    void truncate(int len) {
+        this.len = len;
+
+        reallocate(Math.max(MIN_CAP, len));
```
nizhikov, line 169: Why we want to keep MIN_CAP size even if user ask to trunkcate data to `len`?

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,449 @@
...
+     * @param newCapacity New capacity.
+     */
+    private void reallocate(int newCapacity) {
+        byte[] newBuf = new byte[newCapacity];
+
+        U.arrayCopy(arr, off, newBuf, 0, len);
+
+        arr = newBuf;
+        off = 0;
+        isReadOnly = false;
```
nizhikov, line 307: We must change `len` here.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,449 @@
...
+        @Override public int read(byte[] res, int off, int cnt) {
+            Objects.checkFromIndexSize(off, cnt, res.length);
+
+            int toRead = cnt;
+
+            if (limit != -1) {
+                if (pos >= limit)
+                    return -1;
+
+                int availableBytes = limit - pos;
```
nizhikov, line 368: Looks like logic to calculate amount of bytes that can be read included in `JdbcBinaryBuffer#read`, already.

Do we really need to calculate `toRead` here? 
Can we just call `JdbcBinaryBuffer#read` and rely on the logic inside it?

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,449 @@
...
+            pos = start;
+
+            markedPos = start;
+
+            this.limit = limit;
+        }
+
+        /** {@inheritDoc} */
+        @Override public int read() {
+            if (limit != -1 && pos >= limit)
```
nizhikov, line 347: Looks like the same check exists in `JdbcBinaryBuffer#read`.
Can we remove this and rely to the check inside `JdbcBinaryBuffer#read`?

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
skorotkov: `JdbcBinaryBuffer#read` knows nothing about the limit imposed on this stream instance.  There may be several instances of input stream created for the same `JdbcBinaryBuffer` with different limits.

So `toRead` should be calculated in the input stream class.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
skorotkov: Same logic as above.  Limits are imposed to stream instances, not to `JdbcBinaryBuffer`.

nizhikov: We implement same logic twice - at least we can introduce some methods to perform calculation.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
nizhikov: `newLen`

skorotkov: Sorry, but there is no such variable.

nizhikov: I mean `newCapacity`.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
nizhikov: > We never have buffer smaller then MIN_CAP.

How you come to that requirement?
I don't think we have it.

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
skorotkov: newCapacity is a length of the new byte[] internal buffer.

len is a length of actual data in this buffer.

Why do you think one should be changed to value of another?

nizhikov: Because, in the all cases except the one with truncate it's the same value

skorotkov: No.  Usually buffer is larger then data.  We allocate more space in advance to support streaming way of operations - we expect that more data would be written very very soon.

nizhikov: Two places where we call reallocationIfRequired:

```
        reallocateIfRequired(newLen);

        U.arrayCopy(inpBuf, inpOff, arr, pos, inpLen);

        len = newLen;
...

        reallocateIfRequired(newLen);

        arr[pos] = (byte)b;

        len = newLen;

```

Both of them change `len` to the parameter of the `reallocateIfRequired`.

Third case when we call reallocation directly:

```
        this.len = len;

        reallocate(Math.max(MIN_CAP, len));
```

We also change `len` value to the parameter of `reallocate` method.
So, we can move change of len to the one of methods.

From the code correction point of view - we must change buffere parameters in the fewest possible places.
Looks like we must add one more parameter to the `reallocate` method - `len`

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,449 @@
...
+        len = newLen;
+    }
+
+    /**
+     * Ensure capacity.
+     *
+     * @param newLen The new data length the buffer should be able to hold.
+     */
+    private void reallocateIfRequired(int newLen) {
+        if (newLen - arr.length > 0 || isReadOnly)
```
nizhikov, line 268: `newLen - arr.length > 0` -> `newLen > arr.length`

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
skorotkov: Okay, if you want to. But IMHO it's wrong.  

`reallocate` and  `reallocateIfRequired` do now exactly what names suggest - reallocate internal buffer. But do not change the buffer state.  It's callers who know how to change the buffer state (deal with `len` in particular).

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
skorotkov: In one of the previous comments you say about method which writes 1 byte to stream:

> We don't want to create byte[1] to write data.
> Let's create new array of reasonable size here.

So, MIN_CAP == 256 is looks reasonable.

***
But we can respect user here (but not in write method) and truncate to exact value.

And afterwards if user would start write increase at least for MIN_CAP

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBlob.java
```diff
@@ -18,54 +18,79 @@
...
-                "greater than size of underlying byte array. Requested length also can't be negative " + "" +
-                "[pos=" + pos + ", len=" + len + ']');
+                "greater than Blob length. Requested length also can't be negative " +
+                "[pos=" + pos + ", len=" + len + ", blobLen=" + buf.length() + "]");
+        }
 
         int idx = (int)(pos - 1);
 
-        int size = len > arr.length - idx ? arr.length - idx : len;
+        int size = len > buf.length() - idx ? buf.length() - idx : len;
```
nizhikov, line 86: int size = Math.min(len, buf.length() - idx);

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBinaryBuffer.java
```diff
@@ -0,0 +1,443 @@
...
+
+            int toRead = cnt;
+
+            if (limit != -1) {
+                if (pos >= limit)
+                    return -1;
+
+                int availableBytes = limit - pos;
+
+                if (cnt > availableBytes)
```
nizhikov, line 364: toRead = Math.min(limit - pos, cnt);

modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcBlob.java
```diff
@@ -134,57 +150,110 @@ public JdbcBlob(byte[] arr) {
...
+     * Actial implementation of the pattern search.
+     *
+     * @param ptrn InputStream containing the pattern.
+     * @param ptrnLen Pattern length.
+     * @param idx 1-based index in Blob to start search from.
+     * @return 1-based position at which the pattern appears, else -1.
+     */
+    private long position(InputStream ptrn, int ptrnLen, int idx) throws SQLException {
+        try {
+            InputStream blob = buf.inputStream(idx - 1, buf.length() - idx + 1);
```
nizhikov, line 208: Let's use try with resource here.

nizhikov (approved):

PR #11525, opened 2024-09-10
IGNITE-23179 : Calcite. Wrong numeric type coercion with 'IS DISTINCT FROM'

Vladsz83:

### Notes


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteTypeCoercion.java
```diff
@@ -55,6 +58,50 @@ public IgniteTypeCoercion(RelDataTypeFactory typeFactory, SqlValidator validator
...
+            if (targetType == null || targetType.getFamily() == SqlTypeFamily.ANY)
+                return super.binaryComparisonCoercion(binding);
+            else {
+                boolean coerced = false;
+
+                if (!leftType.equals(targetType))
+                    coerced = coerceOperandType(scope, call, 0, targetType);
+
+                if (!rightType.equals(targetType))
+                    coerced &= coerceOperandType(scope, call, 1, targetType);
```
alex-plekhanov, line 98: Why "&"? Looks like we should return true if any of operands is coerced ("or" condition)
Also looks like `coerceOperandsType` can be used instead of manual operands check.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ExpressionFactoryImpl.java
```diff
@@ -237,7 +239,8 @@ else if (c2 != HIGHEST_VALUE)
                 }
 
                 // If compared rows contain NULLs, they shouldn't be treated as equals, since NULL <> NULL in SQL.
-                return hasNulls ? 1 : 0;
+                // Expect for cases with IS DISTINCT / IS NOT DISTINCT.
+                return hasNulls ? nullComparison : 0;
```
alex-plekhanov, line 243: Let's use just `return hasNulls && !nullsEqual ? 1 : 0;`
`left.get(0).nullDirection.nullComparison` - null comparison for first field, I think it's not better than constant.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
```diff
@@ -138,11 +138,10 @@ public static boolean needCast(RelDataTypeFactory factory, RelDataType fromType,
         // No need to cast if the source type precedence list
         // contains target type. i.e. do not cast from
         // tinyint to int or int to bigint.
-        if (fromType.getPrecedenceList().containsType(toType)
-            && SqlTypeUtil.isIntType(fromType)
-            && SqlTypeUtil.isIntType(toType)) {
+        // Currently, RelDataTypeFactoryImpl#CLASS_FAMILIES doesn't consider the byte type.
+        if (SqlTypeUtil.isIntType(fromType) && SqlTypeUtil.isIntType(toType) && (fromType.getPrecedenceList().containsType(toType))
+            || fromType.getSqlTypeName() == SqlTypeName.TINYINT || toType.getSqlTypeName() == SqlTypeName.TINYINT)
```
alex-plekhanov, line 143: `&&` has higher priority than `||`, parentheses are required. 
In your case any call (for example with `VARCHAR` type) will return `false` if `fromType` or `toType` will be `TINYINT`.
Also, I don't understand why we refer to `RelDataTypeFactoryImpl#CLASS_FAMILIES`. 
`getPrecedenceList()` returns `SqlTypeExplicitPrecedenceList` for integer type, which contains `TINYINT` as well. I've commented this change and test still working. Do we really need this change?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DataTypesTest.java
```diff
@@ -450,6 +453,105 @@ public void testDecimalScale() {
             .check();
     }
 
+    /** */
+    @Test
+    public void testIsNotDistinctFromTypeConversion() {
+        SqlTypeName[] numerics = new SqlTypeName[] {SqlTypeName.TINYINT, SqlTypeName.SMALLINT, SqlTypeName.BIGINT,
+            SqlTypeName.DECIMAL, SqlTypeName.FLOAT, SqlTypeName.DOUBLE, SqlTypeName.BIGINT};
```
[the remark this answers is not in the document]
Vladsz83: INTEGER is not required. The cast is from INTEGER. Bu ok, for the full set.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
Vladsz83: Without this change `StatisticsPlannerTest#testBorders()` will fail on the `Byte` table field. Rewrote.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteTypeCoercion.java
Vladsz83: Good question. Actually, Calcite doesn't take in account the result of this call:

```
if (b) {
      // Coerce type first.
      if (callBinding.isTypeCoercionEnabled()) {
        TypeCoercion typeCoercion = callBinding.getValidator().getTypeCoercion();
        // For comparison operators, i.e. >, <, =, >=, <=.
        typeCoercion.binaryComparisonCoercion(callBinding);
      }
b = super.checkOperandTypes(callBinding, false);
```

'OR' works too. I checked. Fixed.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ExpressionFactoryImpl.java
Vladsz83: Nulls comparison must be the same to the `right`. It is checked above:

```
        // Check that collations is correct.
        for (int i = 0; i < left.size(); i++) {
            if (left.get(i).nullDirection.nullComparison != right.get(i).nullDirection.nullComparison)
                throw new IllegalArgumentException("Can't be compared: left=" + left.get(i) + ", right=" + right.get(i));

            if (left.get(i).direction != right.get(i).direction)
                throw new IllegalArgumentException("Can't be compared: left=" + left.get(i) + ", right=" + right.get(i));
        }
```

While returning `1` seems to work, taking a real comparison value covers theoretical common case. Let's keep it. WDYT?

alex-plekhanov: But you always return nullComparison for field 0, while nulls can be on another field. So it's not better than just return 1, which means that it's not equals.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteTypeCoercion.java
```diff
@@ -55,6 +58,41 @@ public IgniteTypeCoercion(RelDataTypeFactory typeFactory, SqlValidator validator
...
+            //
+            // A least restrictive type between types of different type families does not exist -
+            // the method returns null (See SqlTypeFactoryImpl::leastRestrictive).
+            //
+            RelDataType targetType = factory.leastRestrictive(Arrays.asList(leftType, rightType));
+
+            if (targetType == null || targetType.getFamily() == SqlTypeFamily.ANY)
+                return super.binaryComparisonCoercion(binding);
+            else
+                return coerceOperandType(scope, call, 0, targetType) || coerceOperandType(scope, call, 1, targetType);
```
alex-plekhanov, line 92: Why not `coerceOperandsType`?

PR #11566, opened 2024-10-02
IGNITE-23278 : Calcite. Incorrect error with TIMESTAMP WITH TIME ZONE type

Vladsz83:


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/type/IgniteTypeFactory.java
```diff
@@ -42,12 +43,23 @@
...
 import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.ignite.IgniteException;
 import org.apache.ignite.internal.util.typedef.F;
 
 /**
  * Ignite type factory.
  */
 public class IgniteTypeFactory extends JavaTypeFactoryImpl {
+    /** */
+    private static final Map<SqlTypeName, String> UNSUPPORTED_TYPES = new HashMap<>();
```
alex-plekhanov, line 54: HashMap -> EnumMap

PR #11581, opened 2024-10-09
IGNITE-23361: Calcite. Lost precision of dynamic parameters

Vladsz83:


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/IgniteSqlFunctions.java
```diff
@@ -64,46 +73,42 @@ public static String toString(BigDecimal x) {
...
     public static BigDecimal toBigDecimal(float val, int precision, int scale) {
-        BigDecimal decimal = new BigDecimal(String.valueOf(val));
-        return setScale(precision, scale, decimal);
+        return removeDefaultScale(precision, scale, toBigDecimal(BigDecimal.valueOf(val), precision, scale));
+    }
+
+    /** Removes redundant scale in case of default DECIMAL (without passed precision and scale). */
+    private static BigDecimal removeDefaultScale(int precision, int scale, BigDecimal val) {
+        if (precision == DFLT_NUM_PRECISION && scale == 0 && val.compareTo(val.setScale(0, NUMERIC_ROUNDING_MODE)) == 0)
+            return val.setScale(0, NUMERIC_ROUNDING_MODE);
```
alex-plekhanov, line 89: `val.setScale(0, NUMERIC_ROUNDING_MODE)` executed twice, we can cache first value. Can we somehow define that BigDecimal has a scale but fraction is 0?
Do we need this method at all? Can you provide examples when double/float is converted to BigDecimal without fraction but with scale? I've commented this method, but tests are passed.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/IgniteSqlFunctions.java
```diff
@@ -42,6 +42,15 @@
...
  */
 public class IgniteSqlFunctions {
+    /** */
+    public static final String NUMERIC_OVERFLOW_ERROR = "Numeric field overflow.";
+
+    /** */
+    private static final int DFLT_NUM_PRECISION = IgniteTypeSystem.INSTANCE.getDefaultPrecision(SqlTypeName.DECIMAL);
+
+    /** */
+    private static final RoundingMode NUMERIC_ROUNDING_MODE = RoundingMode.HALF_UP;
```
alex-plekhanov, line 52: We use CEILING rounding mode, but for some reason tests are passed with this rounding mode when I add cases like convertation from 100.16 to 100.1.

Vladsz83: Didn't get. We use HALF_UP

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/IgniteSqlFunctions.java
Vladsz83: Without this, the tests fail: `testDecimalScale()`, `testNumericConversion()`. Actually, there are questions about current precisions of numeric. But in this ticket I didn't fix anithing about it.

alex-plekhanov (approved):

PR #11600, opened 2024-10-16
IGNITE-23468 : Calcite. Fix error of unsupported 'FORMAT' in 'CAST'.

Vladsz83:


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -281,6 +285,10 @@ else if (n instanceof SqlDynamicParam) {
             if (isSystemFieldName(alias))
                 throw newValidationError(call, IgniteResource.INSTANCE.illegalAlias(alias));
         }
+        else if (call.getKind() == SqlKind.CAST) {
+            if (call.getOperandList().size() > 2)
+                throw new IgniteSQLException(ERR_TXT_CAST_OPERATORS);
```
alex-plekhanov, line 290: It's better to use something like `throw newValidationError(n, IgniteResource.INSTANCE.invalidCastParameters(nodeName));`

alex-plekhanov (approved):

PR #11770, opened 2024-12-23
IGNITE-23799 : Calcite. Hash join.

Vladsz83:


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/HashJoinConverterRule.java
```diff
@@ -0,0 +1,89 @@
...
+
+        try {
+            node.accept(v);
+
+            return true;
+        }
+        catch (Util.FoundOne e) {
+            return false;
+        }
+    }
```
alex-plekhanov, line 65: Conditions already updated in AI3, and they look more correct. Let's synchronize it.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/ExecutionTest.java
```diff
@@ -538,6 +534,119 @@ public void testCorrelatedNestedLoopJoin() {
         }
     }
 
+    /** Tests 'Select e.id, e.name, d.name as dep_name from DEP d <JOIN TYPE> join EMP e on e.DEPNO = d.DEPNO'. */
+    @Test
+    public void testHashJoin() {
```
alex-plekhanov: ExecutionTest already contains a lot of unsorted tests, maybe it's time to move join tests to another class. But we also have MergeJoinExecutionTest/NestedLoopJoinExecutionTest they look almost identical and looks like they cover more cases than testHashJoin test. Maybe it's better to extract common part and add new HashJoinExecutionTest class.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/CalciteBasicSecondaryIndexIntegrationTest.java
```diff
@@ -278,37 +279,40 @@ public void testIndexLoopJoin() {
...
-            .returns("Yiruma", "Yiruma")
-            .returns("Cacciapaglia", "Cacciapaglia")
-            .returns("Einaudi", "Einaudi")
-            .returns("Hasaishi", "Hasaishi")
-            .returns("Marradi", "Marradi")
-            .returns("Musorgskii", "Musorgskii")
-            .returns("Rihter", "Rihter")
-            .returns("Zimmer", "Zimmer")
-            .check();
+    public void testMergeAndHashJoin() {
```
alex-plekhanov: Two different methods? Also, perhaps it's worth to extract test to some other class.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/HashJoinPlannerTest.java
```diff
@@ -0,0 +1,209 @@
...
+        String sql = "select t1.ID, t2.ID1 "
+                + "from TEST_TBL_CMPLX t2 "
+                + "join TEST_TBL t1 on t1.id = t2.id1 "
+                + "order by t2.ID1 NULLS LAST, t2.ID2 NULLS LAST";
+
+        RelNode plan = physicalPlan(sql, schema, DISABLED_RULES);
+
+        assertEquals(0, findNodes(plan, byClass(IgniteSort.class)).size());
+        assertEquals(1, findNodes(plan, byClass(IgniteHashJoin.class)).size());
+        assertNotNull(findFirstNode(plan, byClass(IgniteHashJoin.class)));
```
alex-plekhanov:
```
        assertPlan(sql, schema, nodeOrAnyChild(isInstanceOf(IgniteHashJoin.class))
            .and(nodeOrAnyChild(isInstanceOf(IgniteSort.class)).negate()), DISABLED_RULES);
```

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/HashJoinPlannerTest.java
```diff
@@ -0,0 +1,209 @@
...
+
+        String sql = "select t1.ID, t2.ID1 "
+                + "from TEST_TBL t1 "
+                + "join TEST_TBL_CMPLX t2 on t1.id = t2.id1 "
+                + "order by t2.ID1 NULLS LAST, t2.ID2 NULLS LAST";
+
+        IgniteRel plan = physicalPlan(sql, schema, DISABLED_RULES);
+
+        assertNotNull(findFirstNode(plan, byClass(IgniteHashJoin.class)));
+        assertNotNull(sortOnTopOfJoin(plan));
```
alex-plekhanov:
```
        assertPlan(sql, schema, nodeOrAnyChild(isInstanceOf(IgniteSort.class)
            .and(input(isInstanceOf(IgniteHashJoin.class)))), DISABLED_RULES);
```

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/HashJoinPlannerTest.java
```diff
@@ -0,0 +1,209 @@
...
+                }
+            }
+        }
+    }
+
+    /** */
+    private static List<List<Object>> testJoinIsAppliedParameters() {
+        return F.asList(
+            F.asList("select t1.c1 from t1 %s join t1 t2 using(c1)", true),
+            F.asList("select t1.c1 from t1 %s join t1 t2 on t1.c1 = t2.c1", true),
```
alex-plekhanov: Please add two columns test. 
Also, in cases like `t1.c1 = t2.c2 and t1.id = 1` second condition can be pushed down to table scan (at least for some types of join), perhaps these cases should be tested too.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/JoinColocationPlannerTest.java
```diff
@@ -62,16 +66,18 @@ public void joinSameTableSimpleAff() throws Exception {
...
+            AbstractIgniteJoin join = findFirstNode(phys, byClass(AbstractIgniteJoin.class));
 
-        String invalidPlanMsg = "Invalid plan:\n" + RelOptUtil.toString(phys);
+            String invalidPlanMsg = "Invalid plan:\n" + RelOptUtil.toString(phys);
 
-        assertThat(invalidPlanMsg, join, notNullValue());
-        assertThat(invalidPlanMsg, join.distribution().function().affinity(), is(true));
-        assertThat(invalidPlanMsg, join.getLeft(), instanceOf(IgniteIndexScan.class));
-        assertThat(invalidPlanMsg, join.getRight(), instanceOf(IgniteIndexScan.class));
+            assertThat(invalidPlanMsg, join, notNullValue());
```
alex-plekhanov: Maybe rewrite with assertPlan?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinNode.java
```diff
@@ -0,0 +1,761 @@
...
+                    TouchedCollection<RowT> coll = it.next();
+
+                    if (!coll.touched && !coll.items().isEmpty()) {
+                        innerIt = coll.items().iterator();
+
+                        break;
+                    }
+                }
+            }
+        };
```
alex-plekhanov:
```
        return F.flat(F.iterator(entries.values(), TouchedCollection::items, true, v -> !v.touched));
```

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinNode.java
```diff
@@ -0,0 +1,761 @@
...
+    private final int[] leftKeys;
+
+    /** */
+    private final int[] rightKeys;
+
+    /** Output row handler. */
+    protected final RowHandler<Row> outRowHnd;
+
+    /** */
+    protected final Map<Key, TouchedCollection<Row>> hashStore = new HashMap<>(INITIAL_CAPACITY);
```
alex-plekhanov: Consider reusing RuntimeHashIndex

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/HashJoinPlannerTest.java
Vladsz83: There are two-column tests below. Yes, some scans are pushed down and join uses only one equi join pair.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/ExecutionTest.java
Vladsz83: I think you are right. But this might be other tests refactoring ticket.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteHashJoin.java
```diff
@@ -0,0 +1,113 @@
...
+        if (Double.isInfinite(leftRowCnt) || Double.isInfinite(rightRowCnt))
+            return planner.getCostFactory().makeInfiniteCost();
+
+        double rowCnt = leftRowCnt + rightRowCnt;
+
+        int rightKeysSize = joinInfo.rightKeys.size();
+
+        double rightSize = rightRowCnt * IgniteCost.AVERAGE_FIELD_SIZE * getRight().getRowType().getFieldCount();
+
+        double distRightRows = Util.first(mq.getDistinctRowCount(right, ImmutableBitSet.of(joinInfo.rightKeys), null), 0.9 * rightRowCnt);
```
alex-plekhanov: Exeeds line length limit
Let's introduce some constant for 0.9

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/MappingRowHandler.java
```diff
@@ -28,22 +29,28 @@ public class MappingRowHandler<Row> implements RowHandler<Row> {
     private final RowHandler<Row> delegate;
 
     /** */
-    private final int[] mapping;
+    private final ImmutableIntList mapping;
```
alex-plekhanov: `ImmutableIntList` - it's an redundant wrapper here. We can leave int array and use `toIntArray()` in second constructor or use int[] as second constructor argument

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/MappingRowHandler.java
```diff
@@ -28,22 +29,28 @@ public class MappingRowHandler<Row> implements RowHandler<Row> {
...
 
-    /** */
+    /** Creates a mapping handler based on a bit set mapping. Keeps columns unique, but may change required colums order. */
     public MappingRowHandler(RowHandler<Row> delegate, ImmutableBitSet requiredColumns) {
         this.delegate = delegate;
-        mapping = requiredColumns.toArray();
+        mapping = ImmutableIntList.of(requiredColumns.toArray());
+    }
+
+    /** Creates a mapping handler based on a array mapping as is. */
```
alex-plekhanov: `on a array` -> `on an array`

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/MappingRowHandler.java
```diff
@@ -28,22 +29,28 @@ public class MappingRowHandler<Row> implements RowHandler<Row> {
     private final RowHandler<Row> delegate;
 
     /** */
-    private final int[] mapping;
+    private final ImmutableIntList mapping;
 
-    /** */
+    /** Creates a mapping handler based on a bit set mapping. Keeps columns unique, but may change required colums order. */
```
alex-plekhanov: Columns order can't be changed. Perhaps this part of comment rely to second constructor.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/RuntimeHashIndex.java
```diff
@@ -89,40 +137,71 @@ public Iterable<Row> scan(Supplier<Row> searchRow) {
...
+        private RemappedSearcher(RuntimeHashIndex<Row> o, int[] remappedKeys) {
+            super(o.ectx, o.allowNulls, o.nullsMatch, new MappingRowHandler<>(o.ectx.rowHandler(), ImmutableIntList.of(remappedKeys)),
+                o.rows, o.collectionFactory);
+
+            this.origin = o;
+        }
+
+        /** {@inheritDoc} */
+        @Override public void push(Row r) {
+            origin.push(r);
```
alex-plekhanov: Can't understand how it works. It pushes row to the `origin`, but iterates over `this`.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinNode.java
```diff
@@ -0,0 +1,724 @@
...
+        leftKeys = info.leftKeys.toIntArray();
+        rightKeys = info.rightKeys.toIntArray();
+
+        assert leftKeys.length == rightKeys.length;
+
+        this.outRowHnd = outRowHnd;
+
+        this.nonEqCond = nonEqCond;
+
+        rightHashStore = new RuntimeHashIndex<>(ctx, info.rightKeys, keepRowsWithNull, info.hasMatchingNulls(),
```
alex-plekhanov: Since we use filter only for non-eq columns, we should correctly process each "IS NOT DISTINCT FROM" condition, it's not enough to just set `allowNulls = true`, each column should be checked separately for `null` value

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinNode.java
```diff
@@ -0,0 +1,724 @@
...
+            downstream().end();
+
+            return;
+        }
+
+        if (waitingRight == 0 && requested > 0)
+            rightSource().request(waitingRight = IN_BUFFER_SIZE);
+
+        if (waitingLeft == 0 && leftInBuf.size() <= HALF_BUF_SIZE)
+            leftSource().request(waitingLeft = IN_BUFFER_SIZE - leftInBuf.size());
```
alex-plekhanov: Let's move NestedLoopJoinNode#tryToRequestInputs to abstract class and reuse it

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinNode.java
```diff
@@ -0,0 +1,724 @@
...
+        waitingRight--;
+
+        rightHashStore.push(row);
+
+        if (waitingRight == 0)
+            rightSource().request(waitingRight = IN_BUFFER_SIZE);
+    }
+
+    /** */
+    protected void requestMoreOrEnd() throws Exception {
```
alex-plekhanov: Perhaps reuse checkJoinFinished()/tryToRequestInputs() pattern from other join types?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinNode.java
```diff
@@ -0,0 +1,724 @@
...
+
+        /** {@inheritDoc} */
+        @Override protected void join() throws Exception {
+            if (waitingRight == NOT_WAITING) {
+                inLoop = true;
+
+                try {
+                    while (requested > 0 && (left != null || !leftInBuf.isEmpty())) {
+                        // Proceed with next left row, if previous was fully processed.
+                        if (!rightIt.hasNext()) {
```
alex-plekhanov: Looks like `left == null` is more correct condition here

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinNode.java
```diff
@@ -0,0 +1,724 @@
...
+
+                        if (rightIt.hasNext()) {
+                            // Emits matched rows.
+                            while (requested > 0 && rightIt.hasNext()) {
+                                checkState();
+
+                                RowT right = rightIt.next();
+
+                                if (nonEqCond != null && !nonEqCond.test(left, right))
+                                    continue;
```
alex-plekhanov, line 255: We can block the thread for a long time here, not allowing other queries to proceed. We should leave after some count of rows and schedule the next task (the same for other join types).

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/HashJoinConverterRule.java
```diff
@@ -0,0 +1,78 @@
...
+    @Override public boolean matchesJoin(RelOptRuleCall call) {
+        LogicalJoin join = call.rel(0);
+
+        IgniteJoinInfo joinInfo = IgniteJoinInfo.of(join);
+
+        if (joinInfo.pairs().isEmpty())
+            return false;
+
+        // IS NOT DISTINCT is currently not supported simultaneously with equi conditions.
+        if (joinInfo.hasMatchingNulls() && joinInfo.matchingNullsCnt() != joinInfo.pairs().size())
```
alex-plekhanov: Can't understand how it worked for example for `JoinIntegrationTest#testIsNotDistinctWithEquiConditionFrom` (and other) it's two key pairs in the test (IS NOT DISTINCT FROM and EQUALS) and other join converter rules are disabled.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/RuntimeHashIndex.java
```diff
@@ -17,48 +17,91 @@
...
 import org.jetbrains.annotations.Nullable;
 
 /**
  * Runtime hash index based on on-heap hash map.
  */
 public class RuntimeHashIndex<Row> implements RuntimeIndex<Row> {
+    /** Allowed key for null values but matching no any other key. */
+    private static final GroupKey NON_MATCHING_NULLS_KEY = new GroupKey<>(null, null) {
+        @Override public boolean equals(Object o) {
+            throw new UnsupportedOperationException("Rows with null values must not be compared at all.");
```
alex-plekhanov: It still used as key and can be compared, see reproducer:
```
        sql("create table tbl1 (id int primary key, val int)");
        sql("create table tbl2 (id int primary key, val int)");

        sql("insert into tbl1 values (0, 0), (1, null)");
        sql("insert into tbl2 values (0, 0), (1, null)");

        sql("SELECT /*+ HASH_JOIN */ * FROM tbl1 RIGHT JOIN tbl2 ON tbl1.val = tbl2.val");
```

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/HashJoinConverterRule.java
Vladsz83: It didn't work. There were no these tests yet.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinNode.java
Vladsz83: But may block thread similary in many places. We could take known https://issues.apache.org/jira/browse/IGNITE-23960 later and adjust it. WDYT?

modules/calcite/src/test/java/org/apache/ignite/testsuites/ExecutionTestSuite.java
```diff
@@ -47,7 +46,6 @@
     ContinuousExecutionTest.class,
     MergeJoinExecutionTest.class,
     NestedLoopJoinExecutionTest.class,
-    HashJoinExecutionTest.class,
```
alex-plekhanov: Why it was excluded?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/NestedLoopJoinNode.java
```diff
@@ -17,212 +17,71 @@
...
-    protected final RowHandler<Row> handler;
-
-    /** */
-    protected int requested;
-
-    /** */
-    protected Row left;
-
-    /** */
-    protected int rightIdx;
```
alex-plekhanov, line 54: Why this field was moved to children classes?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/AbstractRightMaterializedJoinNode.java
```diff
@@ -0,0 +1,202 @@
...
+    private void doJoin() throws Exception {
+        checkState();
+
+        join();
+    }
+
+    /** */
+    protected boolean rescheduleJoin() {
+        if (processed++ > IN_BUFFER_SIZE) {
+            processed = 0;
```
alex-plekhanov: Maybe reset to zero at least at `doJoin`? Or before any `join` call? It will prevent premature exit on exceed processed rows count on previous step.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/NestedLoopJoinNode.java
```diff
@@ -517,10 +385,13 @@ private static class FullOuterJoin<Row> extends NestedLoopJoinNode<Row> {
         /** */
         private int lastPushedInd;
 
-        /**
-         * @param ctx Execution context.
-         * @param cond Join expression.
-         */
+        /** */
+        private Row left;
```
alex-plekhanov: Why new field was introduced intead of `left` from `AbstractRightMaterializedJoinNode`?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/AbstractRightMaterializedJoinNode.java
Vladsz83: We reset where we reschedule and free the thread.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/NestedLoopJoinNode.java
```diff
@@ -179,9 +179,6 @@ private static class LeftJoin<Row> extends NestedLoopJoinNode<Row> {
         /** Shows whether current left row was matched. */
         private boolean matched;
 
-        /** */
-        private int rightIdx;
```
alex-plekhanov: `rewindInternal` should be fixed too

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/AbstractRightMaterializedJoinNode.java
```diff
@@ -134,7 +134,7 @@ private void pushLeft(Row row) throws Exception {
 
         leftInBuf.add(row);
 
-        join();
+        join0();
```
alex-plekhanov, line 133: Now we do `checkState` twice, on each row

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinNode.java
```diff
@@ -0,0 +1,772 @@
...
+        @Override protected void join() throws Exception {
+            if (waitingRight == NOT_WAITING) {
+                inLoop = true;
+
+                try {
+                    while (requested > 0 && (left != null || !leftInBuf.isEmpty())) {
+                        checkState();
+
+                        // Proceed with next left row, if previous was fully processed.
+                        if (!rightIt.hasNext()) {
```
alex-plekhanov: The same as for inner join `left == null` (and for other join types)

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinNode.java
```diff
@@ -0,0 +1,772 @@
...
+                            left = null;
+                    }
+                }
+                finally {
+                    inLoop = false;
+                }
+            }
+
+            // Emit unmatched right rows.
+            if (left == null && leftInBuf.isEmpty() && waitingLeft == NOT_WAITING && waitingRight == NOT_WAITING && requested > 0) {
```
alex-plekhanov: leftFinished() && ...

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/RuntimeHashIndex.java
```diff
@@ -89,40 +121,92 @@ public Iterable<Row> scan(Supplier<Row> searchRow) {
      * IS NOT DISTINCT FROM condition).
      */
     private @Nullable GroupKey<Row> key(Row r) {
-        if (!allowNulls) {
-            for (int i = 0; i < keysRowHnd.columnCount(r); i++) {
-                if (keysRowHnd.get(i, r) == null)
-                    return null;
+        return new NullsCheckingGroupKey<>(r, keysRowHnd);
```
alex-plekhanov: Now key is never nullable and always stored in hash map. There was an optimization before, if we never can't find row we don't store it.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/HashJoinPlannerTest.java
```diff
@@ -0,0 +1,213 @@
...
+        assertPlan(
+            "select t1.ID, t1.INT_VAL, t2.ID, t2.INT_VAL from SMALL_TBL t1 join LARGE_TBL t2 on t1.ID = t2.ID",
+            schema,
+            nodeOrAnyChild(isInstanceOf(IgniteHashJoin.class).negate()),
+            "JoinCommuteRule"
+        );
+    }
+
+    /** */
+    private static @Nullable IgniteSort sortOnTopOfJoin(IgniteRel root) {
```
alex-plekhanov: Never used

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/JoinColocationPlannerTest.java
```diff
@@ -143,22 +146,24 @@ public void joinComplexToSimpleAff() throws Exception {
...
 
-        assertThat(invalidPlanMsg, exchanges, hasSize(1));
-        assertThat(invalidPlanMsg, exchanges.get(0).getInput(0), instanceOf(IgniteIndexScan.class));
-        assertThat(invalidPlanMsg, exchanges.get(0).getInput(0)
-            .getTable().unwrap(TestTable.class), equalTo(complexTbl));
+            assertThat(invalidPlanMsg, exchanges, hasSize(1));
+            assertThat(invalidPlanMsg, exchanges.get(0).getInput(0), instanceOf(IgniteIndexScan.class));
+            assertThat(invalidPlanMsg, exchanges.get(0).getInput(0)
+                .getTable().unwrap(TestTable.class), equalTo(complexTbl));
+        }
```
alex-plekhanov: Maybe use the same plan checker as for joinSameTableComplexAff (rewritten with `assertPlan`)?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinNode.java
```diff
@@ -0,0 +1,772 @@
...
+                                rightIt = rightRows.iterator();
+                        }
+
+                        if (rightIt.hasNext()) {
+                            // Emits matched rows.
+                            while (requested > 0 && rightIt.hasNext()) {
+                                if (rescheduleJoin())
+                                    return;
+
+                                checkState();
```
alex-plekhanov: Here we have `checkState()` on `join0`, `checkState()` on each left row, `checkState()` on each right row. So on one produced row there can be up to 3 state checks. Let's reduce count of state checks (and in other joins too). I think 1 check state on `join0` is enough (since we have processed rows limit and reschedule join on exceed of this limit).

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinNode.java
```diff
@@ -0,0 +1,750 @@
...
+                            while (requested > 0 && rightIt.hasNext()) {
+                                if (rescheduleJoin())
+                                    return;
+
+                                RowT right = rightIt.next();
+
+                                if (nonEqCond != null && !nonEqCond.test(left, right))
+                                    continue;
+
+                                 --requested;
```
alex-plekhanov: Wrong indent

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinNode.java
```diff
@@ -0,0 +1,750 @@
...
+                        if (rightIt.hasNext()) {
+                            // Emit unmatched left row.
+                            while (requested > 0 && rightIt.hasNext()) {
+                                if (rescheduleJoin())
+                                    return;
+
+                                RowT right = rightIt.next();
+
+                                if (nonEqCond != null && !nonEqCond.test(left, right))
+                                    continue;
```
alex-plekhanov: For left join we should emit row with empty right part, but only once per one left row

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinExecutionTest.java
```diff
@@ -0,0 +1,422 @@
...
+    @Test
+    public void testInnerJoinWithPostFiltration() {
+        doTestJoinWithPostFiltration(INNER, new Object[][]{{3, "Alexey", 1, 1, "Core"}});
+    }
+
+    /** */
+    @Test
+    public void testLeftJoinWithPostFiltration() {
+        Object[][] expected = new Object[][]{
+            {2, "Ival", 5, null, null},
```
alex-plekhanov: `Ival`? There are no such value in input data, something wrong with the check.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/HashJoinExecutionTest.java
```diff
@@ -0,0 +1,422 @@
...
+    /** */
+    @Test
+    public void testInnerJoinWithPostFiltration() {
+        doTestJoinWithPostFiltration(INNER, new Object[][]{{3, "Alexey", 1, 1, "Core"}});
+    }
+
+    /** */
+    @Test
+    public void testLeftJoinWithPostFiltration() {
+        Object[][] expected = new Object[][]{
```
alex-plekhanov: Every row from the left side should be in the set

Vladsz83: There is a condition `((CharSequence)l[1]).length() > 4;`. However, Left is reverted for now.

PR #11777, opened 2024-12-26
IGNITE-23598 Fix LengthPrefixCodec issue with 3 byte buffer

isapego:
- Added test;
- Fixed the big.

### DIFF
--- a/modules/platforms/cpp/network/src/network/length_prefix_codec.cpp
+++ b/modules/platforms/cpp/network/src/network/length_prefix_codec.cpp
@@ -45,7 +45,7 @@ namespace ignite
 
         DataBuffer LengthPrefixCodec::Decode(DataBuffer& data)
         {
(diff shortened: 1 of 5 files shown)


tmgodinho (approved):
LGTM

modules/platforms/cpp/network/src/network/length_prefix_codec.cpp
```diff
@@ -45,7 +45,7 @@ namespace ignite
 
         DataBuffer LengthPrefixCodec::Decode(DataBuffer& data)
         {
-            if (packet.IsValid() && packet.Get()->Length() == (PACKET_HEADER_SIZE + packetSize))
+            if (packet.IsValid() && packetSize != -1 && packet.Get()->Length() == (PACKET_HEADER_SIZE + packetSize))
```
tmgodinho, line 48: Probably just this nitpick. Joining these two branches with an if-else and putting the 2nd case first would be easier to follow.

PR #12121, opened 2025-06-03
Calcite. TpchTest may not clean queries #15

Vladsz83:


modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/tpch/TpchTest.java
```diff
@@ -58,6 +60,7 @@ public static Collection<Object> params() {
      */
     @Test
     public void test() {
-        sql(TpchHelper.getQuery(qryId));
+        for (FieldsQueryCursor<List<?>> cur : queryProcessor(client).query(queryContext(), "PUBLIC", TpchHelper.getQuery(qryId)))
```
alex-plekhanov: Maybe it's better to move view creation to `ddl.sql` file?

PR #12133, opened 2025-06-10
IGNITE-25389 Explicit index cache path

nizhikov:

### Notes


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/filename/FileTreeUtils.java
```diff
@@ -70,6 +75,11 @@ public static void removeTmpSnapshotFiles(SnapshotFileTree sft, boolean err, Ign
      * @return Storage path from config for partition.
      */
     public static @Nullable String partitionStorage(CacheConfiguration<?, ?> ccfg, int part) {
+        if (part == INDEX_PARTITION) {
+            if (F.isEmpty(ccfg.getIndexPath()))
```
sergey-chugunov-1985: Maybe we should use `!F.isEmpty(...)` condition here? Because right now logic looks counter-intuitive: if index path is not set, return index path.

If this comment makes sense I think we should add a test for it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/filename/FileTreeUtils.java
```diff
@@ -78,6 +88,23 @@ public static void removeTmpSnapshotFiles(SnapshotFileTree sft, boolean err, Ign
         return resolveStorage(csp, part);
     }
 
+    /**
+     * @param dsCfg Data storage configuration.
+     * @return All known storages.
+     * @throws IgniteCheckedException
+     */
+    public static Set<String> nodeStorages(@Nullable DataStorageConfiguration dsCfg) throws IgniteCheckedException {
```
sergey-chugunov-1985: I think using `@Nullable` is misleading here as the method itself doesn't handle null values "normally" but throws exception instead.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/filename/CacheConfigStoragePathTest.java
```diff
@@ -75,6 +79,39 @@ public class CacheConfigStoragePathTest extends AbstractDataRegionRelativeStorag
         };
     }
 
+    /** Sanity checks - all pathes for all partitions are different and contains cacheDir. */
```
sergey-chugunov-1985: Minor notes: `paths` instead of `pathes`, `contain` instead of `contains` (as subject is in plural form).

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/filename/NodeFileTree.java
```diff
@@ -881,9 +925,10 @@ else if (name.equals(METASTORAGE_DIR_NAME))
      * @param storagePath Value from config.
      * @return File storage.
      * @see CacheConfiguration#getStoragePaths()
+     * @see CacheConfiguration#getIndexPath()
      */
-    private File cacheStorageRoot(@Nullable String storagePath) {
-        return storagePath == null ? nodeStorage : extraStorages.getOrDefault(storagePath, nodeStorage);
+    private File cacheStorage(@Nullable String storagePath, String cacheDirName) {
```
sergey-chugunov-1985, line 932: I suggest to add javadoc for new parameter `cacheDirName`. As far as I can see it is a subdirectory name under `storagePath` parent directory, but it is not possible to deduce this fact from param name itself.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/filename/NodeFileTree.java
```diff
@@ -640,19 +656,35 @@ public File tmpCacheConfigurationFile(CacheConfiguration<?, ?> ccfg) {
...
     public File partitionFile(CacheConfiguration<?, ?> ccfg, int part) {
-        return new File(resolveStorage(cacheStorages(ccfg), part), partitionFileName(part));
+        if (part == INDEX_PARTITION) {
+            String idxPath = ccfg.getIndexPath();
+
+            if (!F.isEmpty(idxPath))
+                return new File(cacheStorage(idxPath, cacheDirName(ccfg)), partitionFileName(INDEX_PARTITION));
+        }
+
+        return new File(resolveStorage(cacheStorages(ccfg, false), part), partitionFileName(part));
```
sergey-chugunov-1985, line 666: Side note (don't feel obligated to fix it right now or at all). It seems to me that we have somewhat disorienting naming of two methods here, `cacheStorage` and `resolveStorage` - one private and one static from utils class. They look very similar to each other, but their semantics is very different.
If we can come up with better naming, it would be great. I'm ready to participate in a brainstorm when I get more familiar with this class and its neighbours.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/filename/NodeFileTree.java
sergey-chugunov-1985, line 1: Side note (don't feel obligated to fix it right now or at all). One particular feature of this class makes its review a challenging exercise: there are pairs of methods named like `cacheStorage` and `cacheStorages` with overloaded versions on top of that. Difference in name in just one letter makes it quite hard to navigate the context and I'm afraid error-prone too.

If we can rename more specific methods (by this I mean `cacheStorage` here) to something `storageForCache` without compromizing its semantics, I believe it would simplify reading and modifying this class in the future.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/filename/CacheConfigStoragePathTest.java
```diff
@@ -156,18 +193,27 @@ else if (storage == null)
...
         // Root -> cache -> partition set.
         Map<File, Map<String, Set<Integer>>> snpFiles = new HashMap<>();
 
         // Collecting all partition files under each snapshot root.
-        roots.forEach(snpRoot -> {
+        for (File snpRoot : roots) {
             assertTrue(snpRoot.exists());
             assertTrue(snpRoot.isDirectory());
 
+            Predicate<Path> pathPredicate = p -> NodeFileTree.partitionFile(p.toFile())
```
sergey-chugunov-1985: Side note. What do you think about renaming `NodeFileTree.partitionFile` into `NodeFileTree.isPartitionFile`? Not necessarily in this PR but in general.

My reason for that: when reviewing this PR I had to consult with source code to figure out that semantics of `partitionFile` and `partitionFileName` is very different despite their names are rather similar. Name like `isPartitionFile` immediately reveals that this method is for pattern matching and not for constructing a new File from an old one.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/filename/FileTreeUtils.java
nizhikovprivate: Removed

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/filename/CacheConfigStoragePathTest.java
nizhikovprivate: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/filename/NodeFileTree.java
nizhikovprivate: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/filename/NodeFileTree.java
nizhikovprivate: First name of `resolveStorage` was `oneOf` as it don't relates to storage itself, but just selects one element from array.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/filename/CacheConfigStoragePathTest.java
nizhikovprivate: Personally, I don't have any preferences about naming.
But, our style guide states:

> Getters and setters should be implemented without is/get/set prefixes in internal APIs.

https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=57901455#CodingGuidelines-Gettersandsetters

Don't know, if this method match "getter" :)
Feel free to come up with the better naming.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/filename/FileTreeUtils.java
nizhikovprivate: > Maybe we should use !F.isEmpty(...) condition here?

Good catch! Thanks.

> If this comment makes sense I think we should add a test for it.

Actually, can't do it.
This method used to generate storagePath for SnapshotSender.
But, index partition loaded from local snapshot, only.

And result of this method used for loading partitions from remote node.

So, I fixed it, but can't test it.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/filename/AbstractDataRegionRelativeStoragePathTest.java
```diff
@@ -86,14 +117,23 @@ public static List<Object[]> params() {
...
-        else {
+        else if (pathMode == PathMode.RELATIVE) {
             U.delete(new File(U.defaultWorkDirectory(), STORAGE_PATH));
             U.delete(new File(U.defaultWorkDirectory(), STORAGE_PATH_2));
+            U.delete(new File(U.defaultWorkDirectory(), IDX_PATH));
         }
+        else if (pathMode == PathMode.SEPARATE_ROOT) {
+            for (File f : new File(U.defaultWorkDirectory()).listFiles(f -> !f.getName().equals("log")))
+                U.delete(f);
+
```
sergey-chugunov-1985: Unnecessary empty line

PR #12147, opened 2025-06-17
IGNITE-25652 Fix BPlusTree corruption during concurrent removes

Phillippko:
https://issues.apache.org/jira/browse/IGNITE-25652

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/BPlusTree.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/BPlusTree.java
@@ -653,14 +653,13 @@ public abstract class BPlusTree<L, T extends L> extends DataStructure implements
                 }
 
                 // Retry must reset these fields when we release the whole branch without remove.
-                assert r.needReplaceInner == FALSE : "needReplaceInner";
-                assert r.needMergeEmptyBranch == FALSE : "needMergeEmptyBranch";
+                assert !r.needReplaceInner && r.needMergeEmptyBranch == FALSE
+                        : "needReplaceInner=" + r.needReplaceInner + ", needMergeEmptyBranch=" + r.needMergeEmptyBranch;
 
                 if (cnt == rmvCnt) // It was the last element on the leaf.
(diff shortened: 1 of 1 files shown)


sashapolo (approved):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/BPlusTree.java
```diff
@@ -4571,16 +4570,17 @@ protected boolean releaseForRetry(Tail<L> t) {
...
+                    + ", needMergeEmptyBranch=" + needMergeEmptyBranch;
 
             return false;
         }
 
         /** {@inheritDoc} */
         @Override protected Result finishTail() throws IgniteCheckedException {
-            assert !isFinished();
-            assert tail.type == Tail.EXACT && tail.lvl >= 0 : tail;
+            assert !isFinished() && tail.type == Tail.EXACT && tail.lvl >= 0 && needMergeEmptyBranch != READY
```
sergey-chugunov-1985, line 4582: @Phillippko, which tree invariant is protected by this new addition to assertion? There was no assertion about `needMergeEmptyBranch` in original code - was something new discovered?

Phillippko: @sergey-chugunov-1985 Just a precaution. ```needMergeEmptyBranch``` uses 4 state boolean and this assertion should catch similar bugs

```finishTail()``` doesn't expect ```needMergeEmptyBranch``` to be READY, and I think it would be READY only if something went wrong

sergey-chugunov-1985: Ok, I see. It looks like when `finishTail()` is called `needMergeEmptyBranch` indeed cannot be `READY`. It can get this value briefly inside `mergeEmptyBranch` and subsequent calls but shortly after that it transitions to `DONE` or `FALSE` if retry is needed.

Than no more questions from me, I'm going to approve this PR.

PR #12155, opened 2025-06-23
IGNITE-23120 Improve inline index feature

Phillippko:
https://issues.apache.org/jira/browse/IGNITE-23120


modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -349,6 +349,14 @@ public abstract class IgniteUtils extends CommonUtils {
     /** Ignite package. */
     public static final String IGNITE_PKG = "org.apache.ignite.";
 
+    /**
+     * To avoid tree corruption, at least two items should fit into one page of H2 index tree.
+     * So maximum payload size equals: P = (PS - H - 3L) / 2 - X , where P - Payload size, PS - page size, H - page
+     * header size, L - size of the child link, X - overhead per item. Calculated for pageSize = 1KB with
+     * KeystoreEncryptionSpi and MVCC enabled.
```
timoninmaxim: This constant looks weird

1. We dropped MVCC support - delete most of MVCC code, no need to include it to calculations.
2. Why 1KB? Can we use configured page size instead?
3. Whether 2 entries per page works MUCH better than 1 entry? 

Why this is a problem of Ignite? Users can configure inline size for their needs. If performance is bad - then this is a problem of configuration, isn't it?

Phillippko: Answered questions 2 and 3 in ticket (tl;dr: 2. Can't calculate on client nodes 3. Node could fail)

Will change to constant without MVCC support

Phillippko: Discussed in the ticket with @timoninmaxim , decided to just print WARN on server nodes, use real page size and provide a constant to easily configure min recommended number of entries in the tree. Updated the PR

PR #12166, opened 2025-07-01
IGNITE-25769 ComputeTaskContinuousMapper in Ignite.NET

kukushal:
### PR Highlights

* No documentation changes are needed, as the only [description of this functionality on the official documentation site](https://ignite.apache.org/docs/latest/resources-injection.html#pre-defined-resources) also applies to the new .NET API.
* New example `examples/Thick/Compute/Streaming/Program.cs` is a .NET version of the existing [ComputeContinuousMapperExample.java](https://github.com/apache/ignite/blob/2.17.0/examples/src/main/java/org/apache/ignite/examples/computegrid/ComputeContinuousMapperExample.java)
* I chose to enhance the existing .NET `ComputeTaskSession` and its Java counterpart `PlatformComputeTaskSession` to support the continuous mapper API, rather than creating separate implementations for Java and .NET. The reason is that a separate mapper implementation would require adding a second object argument to the `inLongLongLongObjectOutLong` callback, which I wanted to avoid


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Compute/ComputeTaskHolder.cs
```diff
@@ -521,4 +491,101 @@ private void Clean(long taskHandle)
             handleRegistry.Release(taskHandle, true);
         }
     }
+
+    /// <summary>
+    /// Shared compute task utilities.
+    /// </summary>
+    internal static class ComputeTaskUtils
```
ptupitsyn: Let's move this class to a separate file.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Compute/ComputeTaskHolder.cs
```diff
@@ -419,6 +376,19 @@ public void CompleteWithError(long taskHandle, PlatformMemoryStream stream)
             CompleteWithError(taskHandle, err);
         }
 
+        /// <inheritdoc />
+        public void AddJobs(ICollection<long> jobHandles)
+        {
+            if (_jobHandles != null)
```
ptupitsyn: Is this thread-safe? `_jobHandles` is `volatile`, which is suspicious.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Compute/ComputeTaskHolder.cs
```diff
@@ -419,6 +376,19 @@ public void CompleteWithError(long taskHandle, PlatformMemoryStream stream)
...
+            {
+                _jobHandles.AddRange(jobHandles);
```
ptupitsyn: Also probably not thread-safe.

modules/platforms/dotnet/examples/Thick/Compute/Streaming/Program.cs
```diff
@@ -0,0 +1,148 @@
...
+        // Send the first word
+        SendWord();
+        
+        // Return null from the Map method since we have sent at least one job
+        return null;
+    }
+
+    public override ComputeJobResultPolicy OnResult(IComputeJobResult<int> res, IList<IComputeJobResult<int>> rcvd)
+    {
+        // Fail-over to another node ff there is an error
```
ptupitsyn:
```suggestion
        // Fail-over to another node if there is an error
```

chesnokoff:
Hello @kukushal @ptupitsyn, what is the status of this PR?
It looks almost ready. Do you want to finish it?
If you are busy, I can help. I can open a new PR from this branch and keep your commits. Then I will fix the review comments and the TC tests. Aleksey Plekhanov (@alex-plekhanov) can review it.
What do you think?

PR #12178, opened 2025-07-08
IGNITE-25641 Fix discrepancy with CacheStore in case of an error during tx commit phase

sergey-chugunov-1985:

### Notes

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxAdapter.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxAdapter.java
@@ -512,11 +512,12 @@ public abstract class IgniteTxAdapter extends GridMetadataAwareAdapter implement
             try {
                 GridCacheEntryEx entry = e.cached();
 
-                if (e.op() != NOOP)
+                if (e.op() != NOOP) {
                     entry.invalidate(xidVer);
 
-                if (e.context().readThrough())
-                    entry.clear(xidVer, true);
+                    if (e.context().readThrough())
+                        entry.clear(xidVer, true);
+                }
             }
             catch (Throwable t) {
                 U.error(log, "Failed to invalidate transaction entries while reverting a commit.", t);


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxAdapter.java
```diff
@@ -504,6 +505,46 @@ public void storeEnabled(boolean storeEnabled) {
...
+     */
+    protected void clearCacheStoreBackedEntries() {
+        if (txState().cacheIds() == null)
+            return;
+
+        GridIntIterator iter = txState().cacheIds().iterator();
+        while (iter.hasNext()) {
+            int cacheId = iter.next();
+
+            if (cctx.cacheContext(cacheId) == null)
```
alex-plekhanov: Looks weird. Is it really possible? Maybe get config from cache descriptor? (`cctx.cache().cacheDescriptor(cacheId).cacheConfiguration()`)
If there are caches in tx, that is not isReadThrough, entries will not be cleared. Is it ok?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxAdapter.java
```diff
@@ -504,6 +505,46 @@ public void storeEnabled(boolean storeEnabled) {
...
+            if (!cctx.cacheContext(cacheId).config().isReadThrough())
+                return;
+        }
+
+        for (IgniteTxEntry e : writeMap().values()) {
+            try {
+                GridCacheEntryEx entry = e.cached();
+
+                if (e.op() != NOOP) {
+                    entry.clear(xidVer, true);
```
alex-plekhanov: Redundant braces

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+            }
+        }
+    }
+
+    /** */
+    private void updateKeysInTx(Ignite ig, List<Integer> keys) {
+        IgniteCache<Object, Object> cache = ig.cache(DEFAULT_CACHE_NAME);
+
+        try (Transaction tx = ig.transactions().txStart()) {
+            for (Integer key : keys) {
```
alex-plekhanov: Redundant braces

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+        for (int i = 0; i < numOfKeys; i++) {
+            cache.put(i, i);
+        }
+    }
+
+    /** */
+    private void checkKeysOnFaultyNode(List<Integer> keysToCheck) {
+        IgniteCache<Object, Object> cache = grid(FAULTY_NODE_IDX).cache(DEFAULT_CACHE_NAME);
+
+        for (Integer key : keysToCheck) {
```
alex-plekhanov: Redundant braces

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+        else {
+            checkKeysOnFaultyNode(keysOnFaultyNode);
+        }
+
+        checkKeysOnHealthyNodes(keysOnFaultyNode);
+    }
+
+    /** */
+    private void fillCache(IgniteCache<Integer, Integer> cache, int numOfKeys) {
+        for (int i = 0; i < numOfKeys; i++) {
```
alex-plekhanov: Redundant braces

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override protected NearCacheConfiguration nearConfiguration() {
+        return withNearCacheConfiguration ? super.nearConfiguration() : null;
+    }
+
+    /** {@inheritDoc} */
+    @Override protected CacheConfiguration cacheConfiguration(String igniteInstanceName) throws Exception {
+        CacheConfiguration ccfg = super.cacheConfiguration(igniteInstanceName);
```
alex-plekhanov: Please use generics to avoid IDE warnings

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+
+        IgniteEx txCoordinator = faultyNodeRole == FaultyNodeRole.TX_COORDINATOR ? grid(FAULTY_NODE_IDX) : startClientGrid();
+        if (faultyNodeType == FaultyNodeType.PRIMARY)
+            updateKeysInTxWithExceptionCatching(txCoordinator, keysOnFaultyNode);
+        else
+            updateKeysInTx(txCoordinator, keysOnFaultyNode);
+
+        if (withFaulireHandler) { // FH doesn't fail tx coordinator
+            if (faultyNodeRole == FaultyNodeRole.TX_COORDINATOR) {
+                if (faultyNodeType == FaultyNodeType.BACKUP) {
```
alex-plekhanov: Redundant braces

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+            this.instanceName = instanceName;
+            this.faultyNodeRole = faultyNodeRole;
+        }
+
+        /** {@inheritDoc} */
+        @Override public @Nullable Integer onBeforePut(Cache.Entry<Integer, Integer> entry, Integer newVal) {
+            if (newVal < 2 * KEYS_NUMBER)
+                return newVal;
+
+            if (instanceName.contains(FAULTY_NODE_SUFFIX)) {
```
alex-plekhanov: Fragile. Test will fail if test name fill be refactored or if someone change ignite instance name format for tests. Maybe it's better to declare `boolean faultyInstance`? And pass it like `getTestIgniteInstanceIndex(igniteInstanceName) == FAULTY_NODE_IDX`?

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+
+                assertTrue("Client node should survive test scenario",
+                    G.allGrids()
+                        .stream()
+                        .filter(ignite -> ((IgniteEx)ignite).context().clientNode())
+                        .count() == 1);
+            }
+
+        }
+        else {
```
alex-plekhanov: Redundant braces

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+                // two servers and a client node which is a tx coordinator
+                waitForTopology(3);
+
+                assertTrue("Client node should survive test scenario",
+                    G.allGrids()
+                        .stream()
+                        .filter(ignite -> ((IgniteEx)ignite).context().clientNode())
+                        .count() == 1);
+            }
+
```
alex-plekhanov: Redundant NL

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+    @Parameterized.Parameter
+    public FaultyNodeType faultyNodeType;
+
+    /** */
+    @Parameterized.Parameter(1)
+    public FaultyNodeRole faultyNodeRole;
+
+    /** */
+    @Parameterized.Parameter(2)
+    public boolean withFaulireHandler;
```
alex-plekhanov: Abbreviation for `handler` should be used.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+        }
+
+        /** {@inheritDoc} */
+        @Override public @Nullable Integer onBeforePut(Cache.Entry<Integer, Integer> entry, Integer newVal) {
+            if (newVal < 2 * KEYS_NUMBER)
+                return newVal;
+
+            if (instanceName.contains(FAULTY_NODE_SUFFIX)) {
+                if (faultyNodeRole == FaultyNodeRole.TX_COORDINATOR) {
+                    if (!map.containsKey(newVal))
```
alex-plekhanov: For tx coordinator we have two calls of this method for each key? Is it expected? Can you please add a comment?
Also, maybe use `Set` and simplify to `if (!set.add(newVal)) throw new exception`?

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+
+        List<Integer> keysOnFaultyNode = findKeys(grid(FAULTY_NODE_IDX).localNode(), cache, 5, 0, keysType);
+
+        IgniteEx txCoordinator = faultyNodeRole == FaultyNodeRole.TX_COORDINATOR ? grid(FAULTY_NODE_IDX) : startClientGrid();
+        if (faultyNodeType == FaultyNodeType.PRIMARY)
+            updateKeysInTxWithExceptionCatching(txCoordinator, keysOnFaultyNode);
+        else
+            updateKeysInTx(txCoordinator, keysOnFaultyNode);
+
+        if (withFaulireHandler) { // FH doesn't fail tx coordinator
```
alex-plekhanov: Please fix comments (all) according to Ignite code style (start with upper-case, end with point).

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+        }
+    }
+
+    /** */
+    private void updateKeysInTxWithExceptionCatching(Ignite ig, List<Integer> keys) {
+        try {
+            updateKeysInTx(ig, keys);
+        }
+        catch (Exception ignored) {
+            // No-op
```
alex-plekhanov: Point at the end

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+
+        fillCache(cache, KEYS_NUMBER);
+
+        int keysType = faultyNodeType == FaultyNodeType.PRIMARY ? 0 : 1;
+
+        List<Integer> keysOnFaultyNode = findKeys(grid(FAULTY_NODE_IDX).localNode(), cache, 5, 0, keysType);
+
+        IgniteEx txCoordinator = faultyNodeRole == FaultyNodeRole.TX_COORDINATOR ? grid(FAULTY_NODE_IDX) : startClientGrid();
+        if (faultyNodeType == FaultyNodeType.PRIMARY)
+            updateKeysInTxWithExceptionCatching(txCoordinator, keysOnFaultyNode);
```
alex-plekhanov, line 181: Is it expected behaviour that commit throws an exception only when failed to store on primary nodes?

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+        else
+            updateKeysInTx(txCoordinator, keysOnFaultyNode);
+
+        if (withFaulireHandler) { // FH doesn't fail tx coordinator
+            if (faultyNodeRole == FaultyNodeRole.TX_COORDINATOR) {
+                if (faultyNodeType == FaultyNodeType.BACKUP) {
+                    waitForTopology(2); // two servers - tx coordinator hosting backup partition fails
+                }
+                else {
+                    waitForTopology(3); // three servers - tx coordinator hosting primary partition doesn't fail
```
alex-plekhanov: Is it expected behaviour, that some nodes fail with FH, some node not, some nodes fail without FH? Or we check this case just because it works that way? If it works that way and we don't know the reason, we shouldn't check this cases in tests. Someone could make an incorrect assumption based on the scenario described in this test and think that's how it's supposed to work. Perhaps here we can simplify the code and check keys only on alive nodes.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -0,0 +1,322 @@
...
+
+        params.add(new Object[] {FaultyNodeType.PRIMARY, FaultyNodeRole.REGULAR, true, true});
+        params.add(new Object[] {FaultyNodeType.PRIMARY, FaultyNodeRole.REGULAR, false, true});
+        params.add(new Object[] {FaultyNodeType.BACKUP, FaultyNodeRole.REGULAR, true, true});
+        params.add(new Object[] {FaultyNodeType.BACKUP, FaultyNodeRole.REGULAR, false, true});
+
+        params.add(new Object[] {FaultyNodeType.PRIMARY, FaultyNodeRole.TX_COORDINATOR, false, true});
+        params.add(new Object[] {FaultyNodeType.BACKUP, FaultyNodeRole.TX_COORDINATOR, false, true});
+        // TODO https://issues.apache.org/jira/browse/IGNITE-25924
+        // params.add(new Object[] {FaultyNodeType.PRIMARY, FaultyNodeRole.TX_COORDINATOR, true, true});
```
alex-plekhanov: The only difference: tx coordinator hosting primary partition fails when near cache is used. Perhaps it's not exceptional case (we have almost random node failures for other scenarios)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxAdapter.java
```diff
@@ -504,6 +505,46 @@ public void storeEnabled(boolean storeEnabled) {
...
+
+            if (!cctx.cacheContext(cacheId).config().isReadThrough())
+                return;
+        }
+
+        for (IgniteTxEntry e : writeMap().values()) {
+            try {
+                GridCacheEntryEx entry = e.cached();
+
+                if (e.op() != NOOP) {
```
alex-plekhanov: Looks pretty similar to `uncommit()` method. Maybe move this code to `uncommit`? 
Like additional condition:
```
if (entry.context().readThrough())
    entry.clear(xidVer, true);
````

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxAdapter.java
sergey-chugunov-1985: I removed this null check as it indeed looks redundant. 

About not read-through caches. It is possible of cource to clear all involved caches unconditionally but I tried to avoid touching caches that are not backed by CacheStore thus we cannot recover their values easily. That was my motivation.

But it is possible to clear only read-through caches and leave all others - do you think it is a better idea?

Clear only the data you can easily recover.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxAdapter.java
sergey-chugunov-1985: Fixed.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
sergey-chugunov-1985: Fixed.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
sergey-chugunov-1985: Fixed.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
sergey-chugunov-1985: Fixed.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
sergey-chugunov-1985: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxAdapter.java
sergey-chugunov-1985: I agree that this implementation looks similar to `uncommit` but the reason I decided to implement it in a separate method was that I wanted to keep `uncommit` logic untouched so no possible exceptions thrown inside `clearCacheStoreBackedEntries` affect `uncommit`. Does it make sense to you?

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
sergey-chugunov-1985: This piece of code is rather entangled and hard to decipher, but the core distinction here is simple. When FH is not configured, no nodes are failing. When FH is configured nodes are always failing *except* a scenario when tx_coordinator is the same node as primary or backup. Only in that case things get tricky and personally I didn't quite get to the bottom of it.

I can create a ticket for this scenario and title it like "Errors during tx commit are not propagated to FH on Tx Coordinator". But I left this case intentionally out of scope of this PR as fixing it in a right way could require a lot of efforts.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
sergey-chugunov-1985: At least this is how tx protocol works right now: it propagates exception from commit phase only when it happens on primary node.

Exceptions from backup are not propagated - I suspect implementing it that way was a deliberate choice but I don't have a proof of that.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
sergey-chugunov-1985: Fixed.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
sergey-chugunov-1985: Fixed.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
sergey-chugunov-1985: Fixed.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
sergey-chugunov-1985: Fixed.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
sergey-chugunov-1985: I added a comment and simplified the logic with a Set instead of Map.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
sergey-chugunov-1985: Done.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
sergey-chugunov-1985: Fixed: failures in this scenario indeed were because of data races in test code and not a sign of another problem. Thus I mitigated this race and removed tests for near configuration setting completely.

PR #12301, opened 2025-08-27
IGNITE-25534 Update Ignite versions validation for cluster

chesnokoff:

### Notes

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
@@ -17,6 +17,7 @@
 
 package org.apache.ignite.internal.processors.nodevalidation;
 
+import java.util.concurrent.atomic.AtomicReference;
 import org.apache.ignite.cluster.ClusterNode;
 import org.apache.ignite.internal.GridKernalContext;
 import org.apache.ignite.internal.processors.GridProcessorAdapter;
@@ -33,7 +34,10 @@ import static org.apache.ignite.internal.IgniteNodeAttributes.ATTR_BUILD_VER;
  */
 public class OsDiscoveryNodeValidationProcessor extends GridProcessorAdapter implements DiscoveryNodeValidationProcessor {
     /** */
-    private static final int MAX_VER_DIFF_FOR_RU = 2;
+    private static final int MAX_MINOR_DIFF = 1;
+
+    /** */
+    private final AtomicReference<IgniteProductVersion> rmtVerAllowedRef = new AtomicReference<>();
 
     /**
      * @param ctx Kernal context.
@@ -56,12 +60,24 @@ public class OsDiscoveryNodeValidationProcessor extends GridProcessorAdapter imp
         IgniteProductVersion rmtVer = IgniteProductVersion.fromString(rmtBuildVer);
 
         if (!isRollingUpgradeEligible(locVer, rmtVer)) {
-            String errMsg = "Remote node [rmtBuildVer=" + rmtBuildVer + ", rmtNodeAddrs=" + U.addressesAsString(node) +
-                ", rmtNodeId=" + node.id() + "] rejected: Incompatible version for cluster join. Allowed major version " +
-                "range: " + locVer.major() + '.' + (locVer.minor() - 2) + '.' + locVer.maintenance() +
-                " to " + locVer.major() + '.' + (locVer.minor() + 2) + '.' + locVer.maintenance() +
-                " [locBuildVer=" + locBuildVer + ", locNodeAddrs=" + U.addressesAsString(locNode) +
-                ", locNodeId=" + locNode.id() + ']';
+            String errMsg = "Remote node rejected due to incompatible version for cluster join.\n"
+                + "Remote node info:\n"
+                + "  - Version     : " + rmtBuildVer + "\n"
+                + "  - Addresses   : " + U.addressesAsString(node) + "\n"
+                + "  - Node ID     : " + node.id() + "\n"
+                + "Local node info:\n"
+                + "  - Version     : " + locBuildVer + "\n"
+                + "  - Addresses   : " + U.addressesAsString(locNode) + "\n"
+                + "  - Node ID     : " + locNode.id() + "\n"
+                + "Allowed versions for joining:\n"
+                + "  - " + locVer.major() + '.' + locVer.minor() + ".X";
+
+            IgniteProductVersion curRmtVerAllowed = rmtVerAllowedRef.get();
+
+            if (curRmtVerAllowed == null) {
+                errMsg += "\n  - " + locVer.major() + '.' + (locVer.minor() + 1) + ".X";
+                errMsg += "\n  - " + locVer.major() + '.' + (locVer.minor() - 1) + ".X";
+            }
 
             LT.warn(log, errMsg);
 
@@ -75,13 +91,51 @@ public class OsDiscoveryNodeValidationProcessor extends GridProcessorAdapter imp
     }
 
     /**
-     * Checks whether remote node is eligible for rolling upgrade.
+     * Determines whether the remote node's version is eligible for rolling upgrade with the local node.
+     * <p>
+     * A remote version is considered eligible if:
+     * <ul>
+     *   <li>It exactly matches the local version; or</li>
+     *   <li>It has the same major version as the local node and the difference in minor versions is within {@link #MAX_MINOR_DIFF}.</li>
+     * </ul>
+     * If the version is deemed eligible, this method attempts to store the remote version for future validation.
      *
-     * @param locVer - Ignite build version for local node.
-     * @param rmtVer - Ignite build version for remote node.
-     * @return True - if remote node is eligible for rolling upgrade, thus can enter the cluster. False - otherwise.
+     * @param locVer Local node's Ignite product version.
+     * @param rmtVer Remote node's Ignite product version.
+     * @return {@code true} if the remote version is compatible and eligible for rolling upgrade, {@code false} otherwise.
      */
     private boolean isRollingUpgradeEligible(IgniteProductVersion locVer, IgniteProductVersion rmtVer) {
-        return Math.abs(locVer.minor() - rmtVer.minor()) <= MAX_VER_DIFF_FOR_RU;
+        if (locVer.major() == rmtVer.major() && locVer.minor() == rmtVer.minor())
+            return true;
+
+        boolean isEligible = locVer.major() == rmtVer.major() && Math.abs(locVer.minor() - rmtVer.minor()) <= MAX_MINOR_DIFF;
+
+        if (isEligible)
+            return updateEligibleRemoteVersion(rmtVer);
+
+        return false;
+    }
+
+    /**
+     * Atomically updates the allowed remote version for rolling upgrade, if not already set.
+     * <p>
+     * This method ensures that only one remote version is accepted as eligible across multiple threads.
+     * If the reference is already set, it checks whether the incoming version matches the stored one.
+     *
+     * @param rmtVer Remote node's Ignite product version to check or set.
+     * @return {@code true} if the version was successfully set or matches the already accepted version;
+     *         {@code false} if the version conflicts with an already accepted different version.
+     */
+    private boolean updateEligibleRemoteVersion(IgniteProductVersion rmtVer) {
+        while (true) {
+            IgniteProductVersion curRmtVerAllowed = rmtVerAllowedRef.get();
+
+            if (curRmtVerAllowed == null) {
+                if (rmtVerAllowedRef.compareAndSet(null, rmtVer))
+                    return true;
+            }
+            else
+                return curRmtVerAllowed.major() == rmtVer.major() && curRmtVerAllowed.minor() == rmtVer.minor();
+        }
     }
 }
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
@@ -17,12 +17,12 @@
 
 package org.apache.ignite.internal.processors.nodevalidation;
 
-import java.util.Objects;
 import org.apache.ignite.cluster.ClusterNode;
 import org.apache.ignite.internal.GridKernalContext;
 import org.apache.ignite.internal.processors.GridProcessorAdapter;
 import org.apache.ignite.internal.util.typedef.internal.LT;
 import org.apache.ignite.internal.util.typedef.internal.U;
+import org.apache.ignite.lang.IgniteProductVersion;
 import org.apache.ignite.spi.IgniteNodeValidationResult;
 import org.jetbrains.annotations.Nullable;
 
@@ -32,6 +32,9 @@ import static org.apache.ignite.internal.IgniteNodeAttributes.ATTR_BUILD_VER;
  * Node validation.
  */
 public class OsDiscoveryNodeValidationProcessor extends GridProcessorAdapter implements DiscoveryNodeValidationProcessor {
+    /** */
+    private static final int MAX_VER_DIFF_FOR_RU = 2;
+
     /**
      * @param ctx Kernal context.
      */
@@ -43,31 +46,42 @@ public class OsDiscoveryNodeValidationProcessor extends GridProcessorAdapter imp
     @Nullable @Override public IgniteNodeValidationResult validateNode(ClusterNode node) {
         ClusterNode locNode = ctx.discovery().localNode();
 
-        // Check version.
         String locBuildVer = locNode.attribute(ATTR_BUILD_VER);
         String rmtBuildVer = node.attribute(ATTR_BUILD_VER);
 
-        if (!Objects.equals(rmtBuildVer, locBuildVer)) {
-            // OS nodes don't support rolling updates.
-            if (!locBuildVer.equals(rmtBuildVer)) {
-                String errMsg = "Local node and remote node have different version numbers " +
-                    "(node will not join, Ignite does not support rolling updates, " +
-                    "so versions must be exactly the same) " +
-                    "[locBuildVer=" + locBuildVer + ", rmtBuildVer=" + rmtBuildVer +
-                    ", locNodeAddrs=" + U.addressesAsString(locNode) +
-                    ", rmtNodeAddrs=" + U.addressesAsString(node) +
-                    ", locNodeId=" + locNode.id() + ", rmtNodeId=" + node.id() + ']';
+        assert locBuildVer != null : ATTR_BUILD_VER + " is not set on local node!";
+        assert rmtBuildVer != null : ATTR_BUILD_VER + " is not set on remote node!";
 
-                LT.warn(log, errMsg);
+        IgniteProductVersion locVer = IgniteProductVersion.fromString(locBuildVer);
+        IgniteProductVersion rmtVer = IgniteProductVersion.fromString(rmtBuildVer);
 
-                // Always output in debug.
-                if (log.isDebugEnabled())
-                    log.debug(errMsg);
+        if (!isRollingUpgradeEligible(locVer, rmtVer)) {
+            String errMsg = "Remote node [rmtBuildVer=" + rmtBuildVer + ", rmtNodeAddrs=" + U.addressesAsString(node) +
+                ", rmtNodeId=" + node.id() + "] rejected: Incompatible version for cluster join. Allowed major version " +
+                "range: " + locVer.major() + '.' + (locVer.minor() - 2) + '.' + locVer.maintenance() +
+                " to " + locVer.major() + '.' + (locVer.minor() + 2) + '.' + locVer.maintenance() +
+                " [locBuildVer=" + locBuildVer + ", locNodeAddrs=" + U.addressesAsString(locNode) +
+                ", locNodeId=" + locNode.id() + ']';
 
-                return new IgniteNodeValidationResult(node.id(), errMsg);
-            }
+            LT.warn(log, errMsg);
+
+            if (log.isDebugEnabled())
+                log.debug(errMsg);
+
+            return new IgniteNodeValidationResult(node.id(), errMsg);
         }
 
         return null;
+    }
+
+    /**
+     * Checks whether remote node is eligible for rolling upgrade.
+     *
+     * @param locVer - Ignite build version for local node.
+     * @param rmtVer - Ignite build version for remote node.
+     * @return True - if remote node is eligible for rolling upgrade, thus can enter the cluster. False - otherwise.
+     */
+    private boolean isRollingUpgradeEligible(IgniteProductVersion locVer, IgniteProductVersion rmtVer) {
+        return Math.abs(locVer.minor() - rmtVer.minor()) <= MAX_VER_DIFF_FOR_RU;
     }
 }
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
@@ -17,7 +17,8 @@
 
 package org.apache.ignite.internal.processors.nodevalidation;
 
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.Set;
+import java.util.stream.Collectors;
 import org.apache.ignite.IgniteSystemProperties;
 import org.apache.ignite.SystemProperty;
 import org.apache.ignite.cluster.ClusterNode;
@@ -41,9 +42,6 @@ public class OsDiscoveryNodeValidationProcessor extends GridProcessorAdapter imp
 
     /** */
     private static final int MAX_MINOR_DIFF = 1;
-
-    /** */
-    private final AtomicReference<IgniteProductVersion> rmtVerAllowedRef = new AtomicReference<>();
 
     /**
      * @param ctx Kernal context.
@@ -76,14 +74,9 @@ public class OsDiscoveryNodeValidationProcessor extends GridProcessorAdapter imp
                 + "  - Addresses   : " + U.addressesAsString(locNode) + "\n"
                 + "  - Node ID     : " + locNode.id() + "\n"
                 + "Allowed versions for joining:\n"
-                + "  - " + locVer.major() + '.' + locVer.minor() + ".X";
-
-            IgniteProductVersion curRmtVerAllowed = rmtVerAllowedRef.get();
-
-            if (curRmtVerAllowed == null) {
-                errMsg += "\n  - " + locVer.major() + '.' + (locVer.minor() + 1) + ".X";
-                errMsg += "\n  - " + locVer.major() + '.' + (locVer.minor() - 1) + ".X";
-            }
+                + "  - " + locVer.major() + '.' + locVer.minor() + ".X\n"
+                + "  - " + locVer.major() + '.' + (locVer.minor() + 1) + ".X\n"
+                + "  - " + locVer.major() + '.' + (locVer.minor() - 1) + ".X";
 
             LT.warn(log, errMsg);
 
@@ -104,7 +97,6 @@ public class OsDiscoveryNodeValidationProcessor extends GridProcessorAdapter imp
      *   <li>It exactly matches the local version; or</li>
      *   <li>It has the same major version as the local node and the difference in minor versions is within {@link #MAX_MINOR_DIFF}.</li>
      * </ul>
-     * If the version is deemed eligible, this method attempts to store the remote version for future validation.
      *
      * @param locVer Local node's Ignite product version.
      * @param rmtVer Remote node's Ignite product version.
@@ -115,38 +107,17 @@ public class OsDiscoveryNodeValidationProcessor extends GridProcessorAdapter imp
             return true;
 
         boolean enableCheck = IgniteSystemProperties.getBoolean(IGNITE_ROLLING_UPGRADE_VERSION_CHECK);
-
         if (!enableCheck)
             return false;
 
-        boolean isEligible = locVer.major() == rmtVer.major() && Math.abs(locVer.minor() - rmtVer.minor()) <= MAX_MINOR_DIFF;
+        Set<Object> versions = ctx.discovery().allNodes().stream().map(node -> IgniteProductVersion.fromString(node.attribute(ATTR_BUILD_VER)).minor()).collect(Collectors.toSet());
 
-        if (isEligible)
-            return updateEligibleRemoteVersion(rmtVer);
+        if (versions.contains(rmtVer.minor()))
+            return true;
 
-        return false;
-    }
+        if (versions.size() > 1)
+            return false;
 
-    /**
-     * Atomically updates the allowed remote version for rolling upgrade, if not already set.
-     * <p>
-     * This method ensures that only one remote version is accepted as eligible across multiple threads.
-     * If the reference is already set, it checks whether the incoming version matches the stored one.
-     *
-     * @param rmtVer Remote node's Ignite product version to check or set.
-     * @return {@code true} if the version was successfully set or matches the already accepted version;
-     *         {@code false} if the version conflicts with an already accepted different version.
-     */
-    private boolean updateEligibleRemoteVersion(IgniteProductVersion rmtVer) {
-        while (true) {
-            IgniteProductVersion curRmtVerAllowed = rmtVerAllowedRef.get();
-
-            if (curRmtVerAllowed == null) {
-                if (rmtVerAllowedRef.compareAndSet(null, rmtVer))
-                    return true;
-            }
-            else
-                return curRmtVerAllowed.major() == rmtVer.major() && curRmtVerAllowed.minor() == rmtVer.minor();
-        }
+        return locVer.major() == rmtVer.major() && Math.abs(locVer.minor() - rmtVer.minor()) <= MAX_MINOR_DIFF;
     }
 }
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
@@ -18,6 +18,8 @@
 package org.apache.ignite.internal.processors.nodevalidation;
 
 import java.util.concurrent.atomic.AtomicReference;
+import org.apache.ignite.IgniteSystemProperties;
+import org.apache.ignite.SystemProperty;
 import org.apache.ignite.cluster.ClusterNode;
 import org.apache.ignite.internal.GridKernalContext;
 import org.apache.ignite.internal.processors.GridProcessorAdapter;
@@ -33,6 +35,10 @@ import static org.apache.ignite.internal.IgniteNodeAttributes.ATTR_BUILD_VER;
  * Node validation.
  */
 public class OsDiscoveryNodeValidationProcessor extends GridProcessorAdapter implements DiscoveryNodeValidationProcessor {
+    /** Enables version check for rolling upgrade. */
+    @SystemProperty(value = "Enables version check for rolling upgrade.")
+    public static final String IGNITE_ROLLING_UPGRADE_VERSION_CHECK = "IGNITE.ROLLING.UPGRADE.VERSION.CHECK";
+
     /** */
     private static final int MAX_MINOR_DIFF = 1;
 
@@ -107,6 +113,11 @@ public class OsDiscoveryNodeValidationProcessor extends GridProcessorAdapter imp
     private boolean isRollingUpgradeEligible(IgniteProductVersion locVer, IgniteProductVersion rmtVer) {
         if (locVer.major() == rmtVer.major() && locVer.minor() == rmtVer.minor())
             return true;
+
+        boolean enableCheck = IgniteSystemProperties.getBoolean(IGNITE_ROLLING_UPGRADE_VERSION_CHECK);
+
+        if (!enableCheck)
+            return false;
 
         boolean isEligible = locVer.major() == rmtVer.major() && Math.abs(locVer.minor() - rmtVer.minor()) <= MAX_MINOR_DIFF;
 
--- a/modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
+++ b/modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
@@ -174,6 +174,7 @@ import static org.apache.ignite.events.EventType.EVT_NODE_METRICS_UPDATED;
 import static org.apache.ignite.events.EventType.EVT_NODE_SEGMENTED;
 import static org.apache.ignite.failure.FailureType.CRITICAL_ERROR;
 import static org.apache.ignite.failure.FailureType.SYSTEM_WORKER_TERMINATION;
+import static org.apache.ignite.internal.IgniteNodeAttributes.ATTR_BUILD_VER;
 import static org.apache.ignite.internal.IgniteNodeAttributes.ATTR_LATE_AFFINITY_ASSIGNMENT;
 import static org.apache.ignite.internal.IgniteNodeAttributes.ATTR_MARSHALLER;
 import static org.apache.ignite.internal.IgniteNodeAttributes.ATTR_MARSHALLER_COMPACT_FOOTER;
@@ -1703,9 +1704,12 @@ class ServerImpl extends TcpDiscoveryImpl {
 
             NavigableMap<Long, Collection<ClusterNode>> hist = updateTopologyHistory(topVer, top);
 
-            lsnr.onDiscovery(
+            IgniteFuture<?> fut = lsnr.onDiscovery(
                 new DiscoveryNotification(type, topVer, node, top, hist, null, spanContainer)
             );
+
+            if (!Objects.equals(node.attribute(ATTR_BUILD_VER), locNode.attribute(ATTR_BUILD_VER)))
+              fut.get();
 
             return true;
         }
--- a/modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
@@ -28,7 +28,7 @@ import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
 import org.junit.Test;
 
 /**
- * Test grids starting with non compatible release types.
+ * Test Rolling Upgrade release types.
  */
 public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
     /** */
@@ -63,13 +63,13 @@ public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
      * @throws Exception If failed.
      */
     @Test
-    public void testOsEditionDoesNotSupportRollingUpdates() throws Exception {
-        nodeVer = "1.0.0";
+    public void testRollingUpgradeConflictVersions() throws Exception {
+        nodeVer = "2.20.0";
 
         startGrid(0);
 
         try {
-            nodeVer = "1.0.1";
+            nodeVer = "2.17.1";
 
             startGrid(1);
 
@@ -82,7 +82,7 @@ public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
 
             String stackTrace = errors.toString();
 
-            if (!stackTrace.contains("Local node and remote node have different version numbers"))
+            if (!stackTrace.contains("Incompatible version for cluster join"))
                 throw e;
         }
     }
@@ -91,13 +91,13 @@ public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
      * @throws Exception If failed.
      */
     @Test
-    public void testOsEditionDoesNotSupportRollingUpdatesClientMode() throws Exception {
-        nodeVer = "1.0.0";
+    public void testRollingUpgradeConflictVersionsWithClient() throws Exception {
+        nodeVer = "2.20.0";
 
         startGrid(0);
 
         try {
-            nodeVer = "1.0.1";
+            nodeVer = "2.17.1";
 
             startClientGrid(1);
 
@@ -110,8 +110,36 @@ public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
 
             String stackTrace = errors.toString();
 
-            if (!stackTrace.contains("Local node and remote node have different version numbers"))
+            if (!stackTrace.contains("Incompatible version for cluster join"))
                 throw e;
         }
+    }
+
+    /**
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testRollingUpgradeCompatibleVersions() throws Exception {
+        nodeVer = "2.20.0";
+
+        startGrid(0);
+
+        nodeVer = "2.18.1";
+
+        startGrid(1);
+    }
+
+    /**
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testRollingUpgradeCompatibleVersionsWithClient() throws Exception {
+        nodeVer = "2.20.0";
+
+        startGrid(0);
+
+        nodeVer = "2.18.1";
+
+        startClientGrid(1);
     }
 }
--- a/modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
@@ -21,11 +21,16 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.util.Map;
 import org.apache.ignite.IgniteCheckedException;
+import org.apache.ignite.Ignition;
 import org.apache.ignite.configuration.IgniteConfiguration;
+import org.apache.ignite.internal.util.function.ThrowableSupplier;
 import org.apache.ignite.lang.IgniteProductVersion;
 import org.apache.ignite.spi.discovery.tcp.TcpDiscoverySpi;
 import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
 import org.junit.Test;
+
+import static org.apache.ignite.testframework.GridTestUtils.assertThrows;
+import static org.apache.ignite.testframework.GridTestUtils.waitForCondition;
 
 /**
  * Test Rolling Upgrade release types.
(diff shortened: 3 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
```diff
@@ -43,31 +54,72 @@ public OsDiscoveryNodeValidationProcessor(GridKernalContext ctx) {
...
+        if (locVer.minor() == rmtVer.minor())
+            return true;
+
+        boolean enableCheck = IgniteSystemProperties.getBoolean(IGNITE_ROLLING_UPGRADE_VERSION_CHECK);
+        if (!enableCheck)
+            return false;
+
+        Set<Byte> versions = ctx.discovery().allNodes().stream()
+            .map(node -> IgniteProductVersion.fromString(node.attribute(ATTR_BUILD_VER)).minor())
+            .collect(Collectors.toSet());
```
chesnokoff: Previous patch used field to store available version only first time. But there are two problems:
1) If we change coordinator to node with higher version, it may accept +2 version so there is a probability to have three versions of nodes in cluster
2) If +1 version node was accepted, after that it disconnected and we want to add -1 version it will be rejected

So we have to scan available versions in `discovery().allNodes()` every time

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -1703,10 +1704,24 @@ private boolean notifyDiscovery(int type, long topVer, TcpDiscoveryNode node, Sp
 
             NavigableMap<Long, Collection<ClusterNode>> hist = updateTopologyHistory(topVer, top);
 
-            lsnr.onDiscovery(
+            IgniteFuture<?> fut = lsnr.onDiscovery(
```
chesnokoff: The future is responsible for updating `discovery().allNodes()` that is used in `allowedMinorVersions` so we need to wait it. Otherwise, there is race codition, when we can accept +1 and -1 version

modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
```diff
@@ -59,59 +68,260 @@ public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
...
+    private void assertRemoteRejected(ThrowableSupplier<IgniteEx, Exception> gridStart) {
+        Throwable e = assertThrows(log, gridStart::get, IgniteCheckedException.class, null);
+
+        StringWriter errors = new StringWriter();
+
+        e.printStackTrace(new PrintWriter(errors));
+
+        String stackTrace = errors.toString();
+
+        assert stackTrace.contains("Remote node rejected due to incompatible version for cluster join");
```
timoninmaxim: Can be replaced with `X.hasCause`

modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
```diff
@@ -59,59 +68,260 @@ public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
...
+        IgniteEx ign1 = startGrid(1, "2.18.0", false);
+        IgniteEx ign2 = startGrid(2, "2.18.0", false);
+
+        assertTrue(waitForCondition(() -> Ignition.allGrids().size() == 3, getTestTimeout()));
+
+        ign0.close();
+
+        assertTrue(waitForCondition(() -> Ignition.allGrids().size() == 2, getTestTimeout()));
+
+        startGrid(3, "2.19.0", false);
```
timoninmaxim: For RU you should reuse node ids. grid(0) is restarted with new version

modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
```diff
@@ -59,59 +68,260 @@ public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
...
+
+        testCompatibleVersions("2.18.0", "2.19.2", "2.19.6", true);
+        testCompatibleVersions("2.18.0", "2.17.3", "2.17.1", true);
+
+        testCompatibleVersions("2.18.1", "2.18.2", "2.18.3", true);
+    }
+
+    /** */
+    @Test
+    public void testRollingUpgrade0() throws Exception {
```
timoninmaxim: Let's add more tests:
1. For downgrade version of cluster
2. Upgrade and then downgrade back
3. Start upgrade not from coordinator node

modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
```diff
@@ -43,31 +54,72 @@ public OsDiscoveryNodeValidationProcessor(GridKernalContext ctx) {
...
+                + "  - Addresses   : " + U.addressesAsString(node) + "\n"
+                + "  - Node ID     : " + node.id() + "\n"
+                + "Local node info:\n"
+                + "  - Version     : " + locBuildVer + "\n"
+                + "  - Addresses   : " + U.addressesAsString(locNode) + "\n"
+                + "  - Node ID     : " + locNode.id() + "\n"
+                + "Allowed versions for joining:\n"
+                + "  - " + locVer.major() + '.' + locVer.minor() + ".X\n"
+                + "  - " + locVer.major() + '.' + (locVer.minor() + 1) + ".X\n"
+                + "  - " + locVer.major() + '.' + (locVer.minor() - 1) + ".X";
```
timoninmaxim: In case cluster has 2.18 (locVer) and 2.19 versions, then we can't accept 2.17 while this log promises it.

modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
```diff
@@ -59,59 +68,260 @@ public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
...
+
+    /** */
+    private void testCompatibleVersions(
+        String acceptedVer1,
+        String acceptedVer2,
+        String acceptedVer3,
+        boolean withClients
+    ) throws Exception {
+        startGrid(0, acceptedVer1, false);
+        startGrid(1, acceptedVer2, withClients);
```
timoninmaxim: Let's have a cluster with multiple server nodes. In this case different client nodes can be connected to different server nodes. We should check that all client nodes versions are participating in the check.

modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
```diff
@@ -17,19 +17,29 @@
...
 import org.junit.Test;
 
+import static org.apache.ignite.testframework.GridTestUtils.assertThrows;
+import static org.apache.ignite.testframework.GridTestUtils.waitForCondition;
+
 /**
- * Test grids starting with non compatible release types.
+ * Test Rolling Upgrade release types.
  */
+@WithSystemProperty(key = "IGNITE.ROLLING.UPGRADE.VERSION.CHECK", value = "true")
```
wernerdv: Let's use `OsDiscoveryNodeValidationProcessor#IGNITE_ROLLING_UPGRADE_VERSION_CHECK` as a key.

modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
```diff
@@ -59,59 +69,338 @@ public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
...
+            return ign;
+        };
+
+        assertRemoteRejected(sup);
+
+        stopAllGrids();
+    }
+
+    /** Checks that the third grid is not compatible. */
+    private void testConflictVersions(String acceptedVer1, String acceptedVer2, String rejVer, boolean withClients) {
```
wernerdv: `withClients` -> `isClient`

modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
```diff
@@ -59,59 +69,338 @@ public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
...
+        assertRemoteRejected(() -> startClientGridWithConnectionTo(4, "2.20.0", server1));
+
+        assertRemoteRejected(() -> startGrid(4, "2.20.0", false));
+        assertRemoteRejected(() -> startGrid(4, "2.17.0", false));
+
+        assertTrue(Ignition.allGrids().size() == 4);
+    }
+
+    /** Tests that starting a node with rejected version fails with remote rejection. */
+    private void testConflictVersions(String acceptedVer, String rejVer, boolean withClient) {
```
wernerdv: `withClient` -> `isClient`

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
chesnokoff: I have not implemented test for this because I think usually there will no be race condition with `discovery().allNodes()`. It can be reproduced only for high load systems. For instance when ten nodes with +1 and -1 versions are trying to connect to slow master at the same time. 

I think long TCP communication will usually save us from the race. So a reproducer would be flaky

modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
```diff
@@ -32,6 +36,10 @@
  * Node validation.
  */
 public class OsDiscoveryNodeValidationProcessor extends GridProcessorAdapter implements DiscoveryNodeValidationProcessor {
+    /** Enables version check for rolling upgrade. */
+    @SystemProperty(value = "Enables version check for rolling upgrade.")
+    public static final String IGNITE_ROLLING_UPGRADE_VERSION_CHECK = "IGNITE.ROLLING.UPGRADE.VERSION.CHECK";
```
sergey-chugunov-1985: Underscore is used as a separator in system properties accross ignite codebase. Please fix this in string literal: `IGNITE_ROLLING_UPGRADE_VERSION_CHECK` instead of `IGNITE.ROLLING.UPGRADE.VERSION.CHECK`

chesnokoff: Initially, I found `LocalDeploymentSpi#IGNITE_DEPLOYMENT_ADDITIONAL_CHECK`:
```java
/** Enables additional check for resource name on resources removal. */
@SystemProperty(value = "Enables an additional check of a resource name on resources removal")
public static final String IGNITE_DEPLOYMENT_ADDITIONAL_CHECK = "IGNITE.DEPLOYMENT.ADDITIONAL.CHECK";
```
That's why I used dots. But now I see that other properties use underscores. Resolved

modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
```diff
@@ -32,42 +35,76 @@
...
+        return new IgniteNodeValidationResult(node.id(), errMsg);
+    }
+
+    /** Checks if versions has same major, minor and maintenance versions. */
+    private boolean versionsMatch(IgniteProductVersion locVer, IgniteProductVersion rmtVer) {
+        if (locVer == null || rmtVer == null)
+            return false;
 
-        return null;
+        return locVer.major() == rmtVer.major() && locVer.minor() == rmtVer.minor() && locVer.maintenance() == rmtVer.maintenance();
```
chesnokoff: Probably it's better to use IgniteProductVersion#equals but it includes revision timestamp to compare. I'm not sure that we want to compare it here. Also we are expecting that user will provide the rollUpVerCheck and it may become a challenge for him to find out a necessary revision and specify it

modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
```diff
@@ -32,42 +35,76 @@
...
+        IgniteProductVersion locVer = IgniteProductVersion.fromString(locBuildVer);
+        IgniteProductVersion rmtVer = IgniteProductVersion.fromString(rmtBuildVer);
 
-                LT.warn(log, errMsg);
+        IgniteProductVersion rollUpVerCheck = this.rollUpVerCheck.get();
 
-                // Always output in debug.
-                if (log.isDebugEnabled())
-                    log.debug(errMsg);
+        if (versionsMatch(rmtVer, locVer) || versionsMatch(rmtVer, rollUpVerCheck))
```
timoninmaxim: We need support only 1 version for joining nodes.

modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
```diff
@@ -57,61 +81,258 @@ public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
...
-                throw e;
+        allowRollingUpgradeVersionCheck(grid(0), "2.18.1");
+
+        for (int i = 0; i < 3; i++)
+            grid(i).close();
+
+        assertClusterSize(0);
+
+        for (int i = 0; i < 3; i++)
+            startGrid(i, "2.18.0", false, cfgOp);
```
chesnokoff: startGrid(1) fails with
```
Remote node rejected due to incompatible version for cluster join.
Remote node info:
  - Version     : 2.18.0
  - Addresses   : [127.0.0.1]
  - Node ID     : 28e3551d-8372-450e-8c21-4b6da0b00001
Local node info:
  - Version     : 2.18.0
  - Addresses   : [127.0.0.1]
  - Node ID     : 9b1de5c0-2699-4680-aa2a-885aa7500000
Allowed version for joining: 2.18.1
```

modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
```diff
@@ -32,42 +38,85 @@
...
+    private final AtomicReference<IgnitePair<IgniteProductVersion>> verPairHolder = new AtomicReference<>();
+
     /**
      * @param ctx Kernal context.
      */
     public OsDiscoveryNodeValidationProcessor(GridKernalContext ctx) {
         super(ctx);
+
+        ctx.internalSubscriptionProcessor().registerDistributedMetastorageListener(new DistributedMetastorageLifecycleListener() {
+            @Override public void onReadyForRead(ReadableDistributedMetaStorage metastorage) {
```
chesnokoff: I decided not to implement onReadyForWrite because it looks like there can be race conditions then

modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
```diff
@@ -32,42 +38,85 @@
...
+        IgniteProductVersion rmtVer = IgniteProductVersion.fromString(rmtBuildVer);
 
-                LT.warn(log, errMsg);
+        IgnitePair<IgniteProductVersion> pair = verPairHolder.get();
 
-                // Always output in debug.
-                if (log.isDebugEnabled())
-                    log.debug(errMsg);
+        if (pair == null)
+            pair = F.pair(IgniteProductVersion.fromString(locBuildVer), null);
```
chesnokoff: I remember that we don't want to use locBuildVer but there can be a situation when we started validating node before metastorage initialization

chesnokoff: My suggestion is to implement logic with locNodeVer + loc store before the Processor

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Returns a pair containing the current and target versions of the cluster.
+     *
+     * @return A pair where:
+     *     <ul>
+     *         <li><b>First element</b> — current version of the cluster.</li>
+     *         <li><b>Second element</b> — target version to which the cluster is being upgraded.</li>
+     *     </ul>
+     */
+    public IgnitePair<IgniteProductVersion> versions() {
```
chesnokoff: Probably it is confusing to understand what is get0 and what is get1 but we need pair for synchronous access to current and target versions

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * @param current Current cluster version.
+     * @param target Target cluster version.
+     */
+    private boolean checkVersions(IgniteProductVersion current, IgniteProductVersion target) {
+        assert current != null && target != null;
+
+        return target.compareTo(current) >= 0;
+    }
```
chesnokoff: compareTo includes revision check. Do we want to compare it?

chesnokoff: Should we throw assertion here or just log warning and skip suggested version?

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,132 @@
...
+                RollingUpgradeProcessor.this.metastorage = metastorage;
+            }
+        });
+    }
+
+    /**
+     * @param current Current version.
+     * @param target Target version.
+     */
+    public void versions(@NotNull IgniteProductVersion current, @NotNull IgniteProductVersion target) throws IgniteCheckedException {
```
timoninmaxim: Current version can be retrieved from `ctx`

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,132 @@
...
+
+        return false;
+    }
+
+    /**
+     * @param current Current cluster version.
+     * @param target Target cluster version.
+     */
+    private boolean checkVersions(IgniteProductVersion current, IgniteProductVersion target) {
+        assert current != null && target != null;
```
timoninmaxim: no need in asserts here

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
timoninmaxim: we should check that target version is greater than current on 1 (minor, or hotfix).

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * @param current Current version.
+     * @param target Target version.
+     */
+    public void versions(@NotNull IgniteProductVersion current, @NotNull IgniteProductVersion target) throws IgniteCheckedException {
+        A.notNull(metastorage, "Metastorage not ready. Node not started?");
+
+        if (checkVersions(current, target))
```
timoninmaxim: We should throw an IgniteCheckedException here - to make API useful for user. Let's add WARN log here

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.ignite.lang.IgniteProductVersion;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import static org.apache.ignite.internal.IgniteNodeAttributes.ATTR_BUILD_VER;
+
+/** Rolling upgrade processor. Manages current and target versions of cluster. */
+public class RollingUpgradeProcessor extends GridProcessorAdapter {
+    /** Key for the distributed property that holds current and target versions. */
+    private static final String ROLL_UP_VERSIONS = "rolling.upgrade.versions";
```
timoninmaxim: AFAIU, we use the prefix for keys for internal usages.

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * @param current Current version.
+     * @param target Target version.
+     */
+    public void versions(@NotNull IgniteProductVersion current, @NotNull IgniteProductVersion target) throws IgniteCheckedException {
+        A.notNull(metastorage, "Metastorage not ready. Node not started?");
+
+        if (checkVersions(current, target))
+            metastorage.write(ROLL_UP_VERSIONS, F.pair(current, target));
```
timoninmaxim: Should check whether we already is in RU process with other version?

2.18.0 -> 2.18.1
2.18.0 -> 2.19.0

timoninmaxim: If we the same version is already set - then do nothing.

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,132 @@
...
+        ctx.internalSubscriptionProcessor().registerDistributedMetastorageListener(new DistributedMetastorageLifecycleListener() {
+            @Override public void onReadyForRead(ReadableDistributedMetaStorage metastorage) {
+                try {
+                    IgnitePair<IgniteProductVersion> pair = metastorage.read(ROLL_UP_VERSIONS);
+
+                    if (verPairHolder.compareAndSet(null, pair) && log.isDebugEnabled())
+                        log.debug("Read current and target versions from metastore: " + pair);
+                }
+                catch (IgniteCheckedException e) {
+                    if (log.isDebugEnabled())
```
timoninmaxim: In which case we throw an exception here?

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,132 @@
...
+/** Rolling upgrade processor. Manages current and target versions of cluster. */
+public class RollingUpgradeProcessor extends GridProcessorAdapter {
+    /** Key for the distributed property that holds current and target versions. */
+    private static final String ROLL_UP_VERSIONS = "rolling.upgrade.versions";
+
+    /** Metastorage with the write access. */
+    @Nullable private volatile DistributedMetaStorage metastorage;
+
+    /** Distributed property that holds current and target version. */
+    private final AtomicReference<IgnitePair<IgniteProductVersion>> verPairHolder = new AtomicReference<>();
```
timoninmaxim: How to clean up it?

chesnokoff: Do we need clean up here? It is null only until the first assignment

If rolling upgrade is disabled, we will just hold verPair with two identical versions

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
chesnokoff: Thanks for noticing. We really need IGNITE_INTERNAL_KEY_PREFIX

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
chesnokoff: From ReadableDistributedMetaStorage#read docs:

> Throws: IgniteCheckedException – If reading or unmarshalling failed.

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,180 @@
...
+    private boolean checkVersions(IgniteProductVersion cur, IgniteProductVersion target) throws IgniteCheckedException {
+        IgnitePair<IgniteProductVersion> pair = verPairHolder.get();
+
+        if (pair != null && pair.get2().equals(target))
+            return false;
+
+        if (cur.major() != target.major()) {
+            String errMsg = "Major versions are different.";
+
+            LT.warn(log, errMsg);
```
timoninmaxim: no need in log throttling here

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,180 @@
...
+
+    /** {@inheritDoc} */
+    @Override public void start() throws IgniteCheckedException {
+        ctx.internalSubscriptionProcessor().registerDistributedMetastorageListener(new DistributedMetastorageLifecycleListener() {
+            @Override public void onReadyForRead(ReadableDistributedMetaStorage metastorage) {
+                try {
+                    IgnitePair<IgniteProductVersion> pair = metastorage.read(ROLLING_UPGRADE_VERSIONS_KEY);
+
+                    if (verPairHolder.compareAndSet(null, pair) && log.isDebugEnabled())
+                        log.debug("Read current and target versions from metastore: " + pair);
```
timoninmaxim: This log is written when node is started in moment of RU. Looks like this log can be useful, let's make it INFO

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,180 @@
...
+                }
+                catch (IgniteCheckedException e) {
+                    throw new IgniteException(e);
+                }
+
+                metastorage.listen(ROLLING_UPGRADE_VERSIONS_KEY::equals, (key, oldVal, newVal) ->
+                    verPairHolder.compareAndSet((IgnitePair<IgniteProductVersion>)oldVal, (IgnitePair<IgniteProductVersion>)newVal)
```
timoninmaxim: Add log here too

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,180 @@
...
+                log.info("Rolling upgrade enabled [current=" + curtVer + ", target=" + target + ']');
+        }
+    }
+
+    /**
+     * Disables rolling upgrade.
+     *
+     * @throws IgniteCheckedException If metastorage is not available.
+     */
+    public void disable() throws IgniteCheckedException {
```
timoninmaxim, line 213: It must be prohibited to disable RU if there are different versions in cluster

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,180 @@
...
+     *         <li><b>Second element</b> — target version to which the cluster is being upgraded.</li>
+     *     </ul>
+     *     or {@code null} if rolling upgrade is not active.
+     */
+    public IgnitePair<IgniteProductVersion> versions() {
+        return verPairHolder.get();
+    }
+
+    /** Сhecks whether the cluster is in the rolling upgrade mode. */
+    public boolean isRollingUpgradeEnabled() {
```
timoninmaxim: We don't use 'is' prefix for internal packages

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,180 @@
...
+     * @throws IgniteCheckedException If versions are incorrect or metastorage is not available.
+     */
+    public void enable(IgniteProductVersion target) throws IgniteCheckedException {
+        A.notNull(metastorage, "Metastorage not ready. Node not started?");
+
+        String curBuildVer = ctx.discovery().localNode().attribute(ATTR_BUILD_VER);
+        IgniteProductVersion curtVer = IgniteProductVersion.fromString(curBuildVer);
+
+        if (checkVersions(curtVer, target)) {
+            metastorage.write(ROLLING_UPGRADE_VERSIONS_KEY, F.pair(curtVer, target));
```
timoninmaxim: Must validate written value before writing new one

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,163 @@
...
+        String curBuildVer = ctx.discovery().localNode().attribute(ATTR_BUILD_VER);
+        IgniteProductVersion curVer = IgniteProductVersion.fromString(curBuildVer);
+
+        if (!checkVersions(curVer, target))
+            return;
+
+        IgnitePair<IgniteProductVersion> newPair = F.pair(curVer, target);
+
+        if (!metastorage.compareAndSet(ROLLING_UPGRADE_VERSIONS_KEY, null, newPair))
+            throw new IgniteCheckedException("Rolling upgrade is already enabled with a different target version.");
```
timoninmaxim: Let's specify in error message the version that enabled.

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+            vers.add(minMaxVerPair.get2());
+
+            if (lastJoiningNode != null)
+                vers.add(lastJoiningNode.version());
+
+            if (vers.size() > 1) {
+                String msg = "Can't disable rolling upgrade with different versions in cluster: " + vers;
+                log.warning(msg);
+                throw new IgniteCheckedException(msg);
+            }
```
chesnokoff: We cannot test this if branch here, because tcp ring uses real IgniteProductVersions, not from attributes

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+import static org.apache.ignite.internal.IgniteNodeAttributes.ATTR_BUILD_VER;
+import static org.apache.ignite.internal.processors.metastorage.DistributedMetaStorage.IGNITE_INTERNAL_KEY_PREFIX;
+
+/** Rolling upgrade processor. Manages current and target versions of cluster. */
+public class RollingUpgradeProcessor extends GridProcessorAdapter {
+    /** Key for the distributed property that holds current and target versions. */
+    private static final String ROLLING_UPGRADE_VERSIONS_KEY = IGNITE_INTERNAL_KEY_PREFIX + "rolling.upgrade.versions";
+
+    /** Joining timeout. */
+    private static final long JOINING_TIMEOUT = 2_000;
```
chesnokoff: WDYT about value?

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+        }
+        catch (IgniteCheckedException e) {
+            throw new IgniteException(e);
+        }
+    }
+
+    /** Сhecks whether the cluster is in the rolling upgrade mode. */
+    public boolean enabled() {
+        return versions() != null;
+    }
```
chesnokoff, line 274: There was a suggestion to introduce `disabled` boolean flag to monitor state of the processor and use async methods of `metastorage`. But in that case we would have some so-called _"source of true"_ problems about state of the processor. To be more concise, we would have problems with synchronization between `metastorage` and `disabled` especially in cases of coordinator change

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -2997,6 +2997,9 @@ private RingMessageWorker(IgniteLogger log) {
 
                 runTasks();
             });
+
+            if (spi.ignite() instanceof IgniteEx)
+                ((IgniteEx)spi.ignite()).context().rollingUpgrade().minMaxVersionCallback(ring::minMaxNodeVersions);
```
chesnokoff: Probably we want to move the line down to `body` method to update callback on every message

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+            }
+        });
+    }
+
+    /** {@inheritDoc} */
+    @Override public @Nullable IgniteNodeValidationResult validateNode(ClusterNode node) {
+        synchronized (lock) {
+            lastJoiningNode = node;
+
+            lastJoiningNodeTimestamp = U.currentTimeMillis();
```
chesnokoff, line 116: Small optimization: we store only last joining node because if we meet new joining node, it means that old joining node is already in ring

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
timoninmaxim: I think we should get it from IgniteConfiguration

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+    /** Min max version of nodes in cluster supplier. */
+    private Supplier<IgnitePair<IgniteProductVersion>> minMaxVersionSupplier;
+
+    /** Last joining node. */
+    private ClusterNode lastJoiningNode = null;
+
+    /** Last joining node timestamp. */
+    private long lastJoiningNodeTimestamp;
+
+    /** */
```
timoninmaxim: Let's add a doc which threads we synchronize and for which reason

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+     *
+     * @throws IgniteCheckedException If metastorage is not available.
+     */
+    public void disable() throws IgniteCheckedException {
+        A.notNull(metastorage, "Metastorage not ready. Node not started?");
+
+        if (metastorage.read(ROLLING_UPGRADE_VERSIONS_KEY) == null) {
+            if (log.isInfoEnabled())
+                log.info("Rolling upgrade is already disabled");
+            return;
```
timoninmaxim, line 221: add NL before

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+     * Disables rolling upgrade.
+     *
+     * @throws IgniteCheckedException If metastorage is not available.
+     */
+    public void disable() throws IgniteCheckedException {
+        A.notNull(metastorage, "Metastorage not ready. Node not started?");
+
+        if (metastorage.read(ROLLING_UPGRADE_VERSIONS_KEY) == null) {
+            if (log.isInfoEnabled())
+                log.info("Rolling upgrade is already disabled");
```
timoninmaxim: I suppose we should throw here an exception instead

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+    private static final long JOINING_TIMEOUT = 2_000;
+
+    /** Metastorage with the write access. */
+    @Nullable private volatile DistributedMetaStorage metastorage;
+
+    /** Min max version of nodes in cluster supplier. */
+    private Supplier<IgnitePair<IgniteProductVersion>> minMaxVersionSupplier;
+
+    /** Last joining node. */
+    private ClusterNode lastJoiningNode = null;
```
timoninmaxim: no need initialize it with null

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+            if (lastJoiningNode != null)
+                vers.add(lastJoiningNode.version());
+
+            if (vers.size() > 1) {
+                String msg = "Can't disable rolling upgrade with different versions in cluster: " + vers;
+                log.warning(msg);
+                throw new IgniteCheckedException(msg);
+            }
+
+            metastorage.remove(ROLLING_UPGRADE_VERSIONS_KEY);
```
timoninmaxim: Blocking operation that might block discovery thread

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+
+            vers.add(minMaxVerPair.get1());
+
+            vers.add(minMaxVerPair.get2());
+
+            if (lastJoiningNode != null)
+                vers.add(lastJoiningNode.version());
+
+            if (vers.size() > 1) {
+                String msg = "Can't disable rolling upgrade with different versions in cluster: " + vers;
```
timoninmaxim: This message is for admin only, let's not WARN here.

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
chesnokoff: Just found `IgniteConfiguratoin#getFailureDetectionTimeout` with docs 

> Returns failure detection timeout used by TcpDiscoverySpi and  TcpCommunicationSpi.

Looks like that's what we need but default value is 10 seconds

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+        return null;
+    }
+
+    /**
+     * Enables rolling upgrade with specified target version.
+     *
+     * @param target Target version.
+     * @throws IgniteCheckedException If versions are incorrect or metastorage is not available.
+     */
+    public void enable(IgniteProductVersion target) throws IgniteCheckedException {
```
timoninmaxim, line 166: Let's forbid for non-TcpDiscoverySpi

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+            log.info("Rolling upgrade enabled [current=" + curVer + ", target=" + target + ']');
+    }
+
+    /**
+     * Disables rolling upgrade.
+     *
+     * @throws IgniteCheckedException If metastorage is not available.
+     */
+    public void disable() throws IgniteCheckedException {
+        A.notNull(metastorage, "Metastorage not ready. Node not started?");
```
timoninmaxim: Run on coordinator only

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
timoninmaxim: see TcpDiscoverySpi#getJoinTimeout

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+    private static final String ROLLING_UPGRADE_VERSIONS_KEY = IGNITE_INTERNAL_KEY_PREFIX + "rolling.upgrade.versions";
+
+    /** Joining timeout. */
+    private static final long JOINING_TIMEOUT = 2_000;
+
+    /** Metastorage with the write access. */
+    @Nullable private volatile DistributedMetaStorage metastorage;
+
+    /** Min max version of nodes in cluster supplier. */
+    private Supplier<IgnitePair<IgniteProductVersion>> minMaxVersionSupplier;
```
timoninmaxim: Version -> Ver

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+
+    /** Last joining node. */
+    private ClusterNode lastJoiningNode = null;
+
+    /** Last joining node timestamp. */
+    private long lastJoiningNodeTimestamp;
```
timoninmaxim, line 60: Must validate it with timeout

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+
+        IgnitePair<IgniteProductVersion> newPair = F.pair(curVer, target);
+
+        if (!metastorage.compareAndSet(ROLLING_UPGRADE_VERSIONS_KEY, null, newPair)) {
+            IgnitePair<IgniteProductVersion> existingVerPair = metastorage.read(ROLLING_UPGRADE_VERSIONS_KEY);
+
+            String errMsg = "Rolling upgrade is already enabled with a different current and target version: " +
+                existingVerPair.get1() + " , " + existingVerPair.get2();
+
+            log.warning(errMsg);
```
timoninmaxim: Let's log only cluster-specific events - enable / disable. Other messages for admin only

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,264 @@
...
+     *         <li><b>First element</b> — current version of the cluster.</li>
+     *         <li><b>Second element</b> — target version to which the cluster is being upgraded.</li>
+     *     </ul>
+     *     or {@code null} if rolling upgrade is not active.
+     */
+    public IgnitePair<IgniteProductVersion> versions() {
+        A.notNull(metastorage, "Metastorage not ready. Node not started?");
+
+        try {
+            return metastorage.read(ROLLING_UPGRADE_VERSIONS_KEY);
```
timoninmaxim: Let's use volatile versions value on class-level

modules/core/src/main/java/org/apache/ignite/internal/processors/nodevalidation/OsDiscoveryNodeValidationProcessor.java
```diff
@@ -43,31 +45,50 @@ public OsDiscoveryNodeValidationProcessor(GridKernalContext ctx) {
     @Nullable @Override public IgniteNodeValidationResult validateNode(ClusterNode node) {
         ClusterNode locNode = ctx.discovery().localNode();
 
-        // Check version.
```
timoninmaxim, line 46: You don't need this processor anymore

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,310 @@
...
+    public void disable() throws IgniteCheckedException {
+        if (!U.isLocalNodeCoordinator(ctx.discovery()))
+            throw new IgniteCheckedException("Rolling upgrade can be disabled only on coordinator node");
+
+        A.notNull(metastorage, "Metastorage not ready. Node not started?");
+
+        if (verPair == null)
+            return;
+
+        IgnitePair<IgniteProductVersion> minMaxVerPair = ring.minMaxNodeVersions();
```
chesnokoff, line 223: Need ring for that operation

modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
```diff
@@ -51,67 +84,406 @@ public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
...
+        assertClusterSize(3);
+
+        if (isClient)
+            grid(0).context().rollingUpgrade().disable();
+        else
+            grid(2).context().rollingUpgrade().disable();
+
+        for (int i = 0; i < 3; i++) {
+            if (!grid(i).localNode().isClient())
+                assertFalse(grid(i).context().rollingUpgrade().enabled());
```
chesnokoff, line 195: Fails if grid is client and waitForCondition does not help. Looks like metastorage gives us less guarantee about client grids

modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
```diff
@@ -51,67 +84,406 @@ public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
...
+        assertRemoteRejected(() -> startGrid(3, "2.18.0", isClient));
     }
 
-    /**
-     * @throws Exception If failed.
-     */
+    /** */
+    @Test
+    public void testJoiningNodeFailed() throws Exception {
+        int joinTimeout = 15_000;
```
chesnokoff: TcpDiscoverySpi#DFLT_JOIN_TIMEOUT is zero.

```java
    /**
     * Join timeout, in milliseconds. Time to wait for joining. If node cannot connect to any address from the IP
     * finder, the node continues to try to join during this timeout. If all addresses still do not respond, an
     * exception will occur and the node will fail to start. If 0 is specified, it means wait forever.
     */
    protected long joinTimeout = DFLT_JOIN_TIMEOUT;

File: modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java [hunk 54]
```diff
@@ -0,0 +1,310 @@
...
+
+            @Override public void onReadyForRead(ReadableDistributedMetaStorage metastorage) {
+                try {
+                    verPair = metastorage.read(ROLLING_UPGRADE_VERSIONS_KEY);
+                }
+                catch (IgniteCheckedException e) {
+                    throw new IgniteException(e);
+                }
+
+                metastorage.listen(ROLLING_UPGRADE_VERSIONS_KEY::equals, (key, oldVal, newVal) -> {
```
timoninmaxim, line 104: Add a comment for a reason for listening (in case coordinator changes)

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,310 @@
...
+                    verPair = (IgnitePair<IgniteProductVersion>)newVal;
+                });
+            }
+        });
+    }
+
+    /** {@inheritDoc} */
+    @Override public @Nullable IgniteNodeValidationResult validateNode(ClusterNode node) {
+        synchronized (lock) {
```
timoninmaxim, line 113: Add a comment for a reason to store this node in local var

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        if (cur.maintenance() + 1 != target.maintenance())
+            throw new IgniteCheckedException("Patch version can only be incremented by 1.");
+
+        return true;
+    }
+
+    /** Checks if versions have same major, minor and maintenance versions. */
+    private boolean validateVersionsForJoining(IgniteProductVersion ver1, IgniteProductVersion ver2) {
```
timoninmaxim: validate -> check, or vice versa

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,310 @@
...
+        if (ver1 == null || ver2 == null)
+            return false;
+
+        return ver1.major() == ver2.major() && ver1.minor() == ver2.minor() && ver1.maintenance() == ver2.maintenance();
```
timoninmaxim: equals (let's keep old logic here)

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,310 @@
...
+
+        if (cur.maintenance() + 1 != target.maintenance())
+            throw new IgniteCheckedException("Patch version can only be incremented by 1.");
+
+        return true;
+    }
+
+    /** Checks if versions have same major, minor and maintenance versions. */
+    private boolean validateVersionsForJoining(IgniteProductVersion ver1, IgniteProductVersion ver2) {
+        if (ver1 == null || ver2 == null)
```
timoninmaxim: When ver1 is null?

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,310 @@
...
+            + "  - Addresses   : " + U.addressesAsString(node) + "\n"
+            + "  - Node ID     : " + node.id() + "\n"
+            + "Local node info:\n"
+            + "  - Version     : " + locBuildVer + "\n"
+            + "  - Addresses   : " + U.addressesAsString(locNode) + "\n"
+            + "  - Node ID     : " + locNode.id() + "\n"
+            + "Allowed versions for joining: \n"
+            + " - " + curVer.major() + "." + curVer.minor() + "." + curVer.maintenance()
+            + (targetVer == null || targetVer.equals(curVer) ? "" :
+            "\n - " + targetVer.major() + "." + targetVer.minor() + "." + targetVer.maintenance());
```
timoninmaxim: use comma as a separator for second version

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    /**
+     * Enables rolling upgrade with specified target version.
+     * This method can only be called on coordinator node with {@link TcpDiscoverySpi}.
+     *
+     * @param target Target version.
+     * @throws IgniteCheckedException If current and target versions are not compatible
+     * or if node is not coordinator or if discovery SPI is not {@link TcpDiscoverySpi}.
+     * @throws NullPointerException If metastorage is not available.
```
timoninmaxim: Throwing NPE within Ignite node is bad practice.

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,310 @@
...
+            if (vers.size() > 1)
+                throw new IgniteCheckedException("Can't disable rolling upgrade with different versions in cluster: " + vers);
+
+            verPair = null;
+        }
+
+        metastorage.remove(ROLLING_UPGRADE_VERSIONS_KEY);
+
+        if (log.isInfoEnabled())
+            log.info("Rolling upgrade disabled");
```
timoninmaxim: Lets's mention version of cluster

timoninmaxim (changes requested):

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -2997,6 +2998,12 @@ private RingMessageWorker(IgniteLogger log) {
 
                 runTasks();
             });
+
+            if (spi.ignite() instanceof IgniteEx) {
+                RollingUpgradeProcessor rollingUpgradeProcessor = ((IgniteEx)spi.ignite()).context().rollingUpgrade();
+                if (rollingUpgradeProcessor != null)
```
chesnokoff: If we remove this null check, `TcpDiscoverySpiFailureTimeoutSelfTest#testConnectionCheckMessage` will fail

Looks like in some cases RingMessageWorker can be initialized before IgniteKernal

Because of that I've added null check for `ring` in `enable` method

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,310 @@
...
+        IgnitePair<IgniteProductVersion> pair = verPair;
+
+        if (pair == null)
+            pair = F.pair(IgniteProductVersion.fromString(locBuildVer), null);
+
+        IgniteProductVersion curVer = pair.get1();
+        IgniteProductVersion targetVer = pair.get2();
+
+        if (Objects.equals(rmtVer, curVer) || Objects.equals(rmtVer, targetVer))
+            return null;
```
chesnokoff, line 132: It was suggested using equals as is in OsDiscoveryNodeValidation so I think it's better to use equals for both `curVer` and `targetVer`

modules/core/src/main/java/org/apache/ignite/internal/GridKernalContextImpl.java
```diff
@@ -362,6 +363,10 @@ public class GridKernalContextImpl implements GridKernalContext, Externalizable
     @GridToStringExclude
     private PerformanceStatisticsProcessor perfStatProc;
 
+    /** Rolling upgrade processor. */
+    @GridToStringExclude
+    private RollingUpgradeProcessor rolUpProc;
```
timoninmaxim: rollUpProc

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,310 @@
...
+    private ClusterNode lastJoiningNode;
+
+    /** Last joining node timestamp. */
+    private long lastJoiningNodeTimestamp;
+
+    /** Lock for synchronization between tcp-disco-msg-worker thread and management operations. */
+    private final Object lock = new Object();
+
+    /** Pair with current and target versions. */
+    private volatile IgnitePair<IgniteProductVersion> verPair = null;
```
timoninmaxim: Not need to initialize with null

timoninmaxim: rollUpVers?

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,310 @@
...
+    /** Last joining node. */
+    private ClusterNode lastJoiningNode;
+
+    /** Last joining node timestamp. */
+    private long lastJoiningNodeTimestamp;
+
+    /** Lock for synchronization between tcp-disco-msg-worker thread and management operations. */
+    private final Object lock = new Object();
+
+    /** Pair with current and target versions. */
```
timoninmaxim: @Nullable. {@code null} when rolling upgrade is disabled.

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,310 @@
...
+
+        String locBuildVer = locNode.attribute(ATTR_BUILD_VER);
+        String rmtBuildVer = node.attribute(ATTR_BUILD_VER);
+
+        IgniteProductVersion rmtVer = IgniteProductVersion.fromString(rmtBuildVer);
+
+        IgnitePair<IgniteProductVersion> pair = verPair;
+
+        if (pair == null)
+            pair = F.pair(IgniteProductVersion.fromString(locBuildVer), null);
```
timoninmaxim: No need to build this variable:

```
curVer = pair == null ? IgniteProductVersion.fromString(locBuildVer) : pair.get1();
targetVer = pair == null ? null : pair.get2()
```

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,310 @@
...
+     *         <li>The current and target versions are incompatible;</li>
+     *         <li>The local node is not a coordinator;</li>
+     *         <li>The discovery SPI is not {@link TcpDiscoverySpi};</li>
+     *         <li>The distributed metastorage is not ready;</li>
+     *         <li>{@link TcpDiscoveryNodesRing} was not provided.</li>
+     *     </ul>
+     */
+    public void enable(IgniteProductVersion target) throws IgniteCheckedException {
+        if (ring == null)
+            throw new IgniteCheckedException("TCP discovery nodes ring was not provided.");
```
timoninmaxim: What admin should do with this message? And for "Meta storage is not ready"? 

Let's write a suggestion for an admin.

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,310 @@
...
+        vers.add(minMaxVerPair.get1());
+
+        vers.add(minMaxVerPair.get2());
+
+        if (vers.size() > 1)
+            throw new IgniteCheckedException("Can't disable rolling upgrade with different versions in cluster: " + vers);
+
+        synchronized (lock) {
+            if (lastJoiningNode != null) {
+                long timeout = ((TcpDiscoverySpi)ctx.config().getDiscoverySpi()).getJoinTimeout();
```
timoninmaxim: let's use more than single timeout, at least 3.

Also should handle special case with timeout = 0.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
timoninmaxim: Let's move this code to ServerImpl#spiStart before `joinTopology()` call. We should call callback to spi here. TestTcpDiscoverySpi in the test should do nothing in this callback

modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
```diff
@@ -17,23 +17,56 @@
...
 /**
- * Test grids starting with non compatible release types.
+ * Test Rolling Upgrade release types.
  */
+@RunWith(Parameterized.class)
 public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
     /** */
     private String nodeVer;
 
+    /** Is client. */
```
timoninmaxim: Who?

modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
```diff
@@ -51,67 +84,406 @@ public class GridReleaseTypeSelfTest extends GridCommonAbstractTest {
...
         stopAllGrids();
+        cleanPersistenceDir();
     }
 
-    /**
-     * @throws Exception If failed.
-     */
+    /** */
+    @Test
+    public void testTwoConflictVersions() {
```
timoninmaxim, line 107: Why this test runs 25sec? And other tests for 5 sec? Is there any timeout we are waiting for?

chesnokoff: JFY,  I have these stats on my machine
image

In `ServerImpl#spiStop0` we have
``` java
while (spiState != LEFT && timeout > 0) { // timeout == 5_000
                    try {
                        mux.wait(timeout);

                        timeout = U.nanosToMillis(thresholdNanos - System.nanoTime());
                    }
                    catch (InterruptedException ignored) {
                        Thread.currentThread().interrupt();

                        break;
                    }
                }
```

Java profilers show that we waste time on mutex waitng. Stop is called from `IgniteKernal#start` at line 1267. In this place we catch e == `IgniteCheckedException: Failed to start manager: GridManagerAdapter [enabled=true, name=org.apache.ignite.internal.managers.discovery.GridDiscoveryManager]` and its cause is  our `Remote node rejected due to incompatible version for joining`

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
timoninmaxim: You missed this comment

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,308 @@
...
+                "Check that TCP discovery SPI is configured properly and try again");
+
+        if (!U.isLocalNodeCoordinator(ctx.discovery()))
+            throw new IgniteCheckedException("Rolling upgrade can be enabled only on coordinator node");
+
+        if (metastorage == null)
+            throw new IgniteCheckedException("Metastorage is not ready yet. Try again later");
+
+        if (!(ctx.config().getDiscoverySpi() instanceof TcpDiscoverySpi))
+            throw new IgniteCheckedException("Rolling upgrade is supported only with TCP discovery SPI");
```
timoninmaxim, line 177: TcpDiscoverySpi is a public class, we can mention it directly

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,308 @@
...
+     *         <li>The local node is not a coordinator;</li>
+     *         <li>The discovery SPI is not {@link TcpDiscoverySpi};</li>
+     *         <li>The distributed metastorage is not ready;</li>
+     *         <li>{@link TcpDiscoveryNodesRing} was not provided.</li>
+     *     </ul>
+     */
+    public void enable(IgniteProductVersion target) throws IgniteCheckedException {
+        if (ring == null)
+            throw new IgniteCheckedException("No TCP discovery node ring provided. " +
+                "Check that TCP discovery SPI is configured properly and try again");
```
timoninmaxim: What is a proper way to configure the spi?

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,308 @@
...
+            if (vers.size() > 1)
+                throw new IgniteCheckedException("Can't disable rolling upgrade with different versions in cluster: " + vers);
+
+            rollUpVers = null;
+        }
+
+        metastorage.remove(ROLLING_UPGRADE_VERSIONS_KEY);
+
+        if (log.isInfoEnabled())
+            log.info("Rolling upgrade disabled. Current version of nodes in cluster: " + vers.stream().findAny().get());
```
timoninmaxim: `minMaxVerPair.get1()`?

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -0,0 +1,308 @@
...
+     * or if version information has not been read from the distributed metastorage.
+     *
+     * @return A pair where:
+     *     <ul>
+     *         <li><b>First element</b> — current version of the cluster.</li>
+     *         <li><b>Second element</b> — target version to which the cluster is being upgraded.</li>
+     *     </ul>
+     *     or {@code null} if rolling upgrade is not active.
+     */
+    public IgnitePair<IgniteProductVersion> versions() {
```
timoninmaxim: `@Nullable`

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -475,6 +476,12 @@ class ServerImpl extends TcpDiscoveryImpl {
             statsPrinter.start();
         }
 
+        if (spi.ignite() instanceof IgniteEx) {
+            RollingUpgradeProcessor rollUpProc = ((IgniteEx)spi.ignite()).context().rollingUpgrade();
```
timoninmaxim: Let's add a callback to spi to setup the ring. Failed test should override this callback for its spi implementation

modules/core/src/test/java/org/apache/ignite/internal/GridReleaseTypeSelfTest.java
timoninmaxim: Let's set this timeout to 1 sec for this test, and add a comment for reason

timoninmaxim (changes requested):

timoninmaxim (approved):

PR #12329, opened 2025-09-09
IGNITE-26266: Mark IndexQuery#getCriteria with @Nullable

korot-7:
Mark IndexQuery#getCriteria with @Nullable to avoid warning and highlight warning to a user.


modules/core/src/main/java/org/apache/ignite/cache/query/IndexQuery.java
```diff
@@ -138,6 +138,7 @@ public IndexQuery<K, V> setCriteria(List<IndexQueryCriterion> criteria) {
      *
      * @return List of criteria for this index query.
      */
+    @Nullable
```
timoninmaxim: Move the annotation to the method line. See, how it is done for other methods, for example `IndexQuery#getPartition`.

timoninmaxim (approved):

PR #12429, opened 2025-10-16
IGNITE-26608 Added a unified mechanism for propagating Context Attributes

petrov-mg:

### Notes

### DIFF
--- a/modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Scope.java
+++ b/modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Scope.java
@@ -18,8 +18,22 @@
 package org.apache.ignite.internal.thread.context;
 
 /**
- * Represents an arbitrary Scope. A Scope is active from the moment it is created until the {@link #close()} method is
- * called on the {@link Scope} instance. It is strongly encouraged to use a try-with-resources block to close a Scope.
+ * Represents the Scope of {@link OperationContext} attributes update. Explicitly calling {@link #close()} method undoes
+ * the applied changes and restores previous attribute values, if any. Note that every Scope relating to a specific
+ * {@link OperationContext} update must be closed to free up thread-bound resources and avoid memory leaks, so it is
+ * highly encouraged to use a try-with-resource block with Scope instances.
+ * <p>
+ * Scope is result of the following {@link OperationContext} update operations:
+ * <ul>
+ *     <li>{@link OperationContext#set(OperationContextAttribute, Object)} - creates a new or update an existing mapping
+ *     between specified {@link OperationContextAttribute} and its value</li>
+ *     <li>{@link OperationContext#restoreSnapshot(OperationContextSnapshot)} - updates {@link OperationContextAttribute}
+ *     values to match the values stored in {@link OperationContextSnapshot}</li>
+ * </ul>
+ * </p>
+ *
+ * @see OperationContext#set(OperationContextAttribute, Object)
+ * @see OperationContext#restoreSnapshot(OperationContextSnapshot)
  */
 public interface Scope extends AutoCloseable {
     /** Scope instance that does nothing when closed. */


modules/core/src/main/java/org/apache/ignite/internal/processors/security/IgniteSecurity.java
```diff
@@ -42,30 +44,43 @@
...
-     * Creates {@link OperationSecurityContext}. All calls of methods {@link #authorize(String, SecurityPermission)} or {@link
-     * #authorize(SecurityPermission)} will be processed into the context of passed {@link SecurityContext} until
-     * holder {@link OperationSecurityContext} will be closed.
+     * Creates new {@link Scope} with provided {@link SecurityContext}.
      *
      * @param secCtx Security Context.
-     * @return Security context holder.
+     * @return Thread Context Scope.
+     *
+     * @see #withContext(Scope, SecurityContext
```
NSAmelchev:
```suggestion
     * @see #withContext(Scope, SecurityContext)
```

petrov-mg: Done.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/ThreadContextData.java
```diff
@@ -0,0 +1,142 @@
...
+
+    /** */
+    private int activeAttrsCnt;
+
+    /** */
+    private ScopedAttributeValueStack<?>[] attrs = EMPTY;
+
+    /** */
+    <T> T get(ThreadContextAttribute<T> attr) {
+        if (attr.id() >= attrs.length)
```
timoninmaxim: Shoulud we throw assert instead? Is it correct to access to the attribute we actually didn't set?

Could it lead to an error, when user forget to set the var and received default value instead?

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Scope.java
```diff
@@ -15,32 +15,18 @@
...
-     * @param proc Ignite Security.
-     * @param secCtx Security context.
-     */
-    OperationSecurityContext(IgniteSecurity proc, SecurityContext secCtx) {
-        this.proc = proc;
-        this.secCtx = secCtx;
-    }
+public interface Scope extends AutoCloseable {
+    /** Binds attribute with specified value to the current scope. */
+    public <T> Scope withAttribute(ThreadContextAttribute<T> attr, T val);
```
timoninmaxim: Broken incapsulation. ThreadContextScope inherits Scope, then ThreadContextAttribute must be part of ThreadContextScrope.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/ThreadContextData.java
petrov-mg: No, we shouldn't. The initial value of an attribute literally means - the value returned by the ThreadContext#get method if the attribute's value is not explicitly set.

It can be used to skip explicitly setting attributes for some internal operations/workers and still get some value for attribute.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Scope.java
petrov-mg: It’s more of a naming problem than an “encapsulation.” Anyway, I’ve updated the PR and removed the `Scope` interface entirely — on second thought, it’s redundant for the current patch. So now only `ThreadContextScope` remains.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Scope.java
```diff
@@ -15,32 +15,13 @@
...
+    Scope NOOP_SCOPE = () -> {};
 
     /** {@inheritDoc} */
-    @Override public void close() {
-        if (secCtx == null)
-            ((IgniteSecurityProcessor)proc).restoreDefaultContext();
-        else
-            proc.withContext(secCtx);
-    }
+    @Override void close();
```
timoninmaxim, line 43: Let's add a comment why we override AutoCloseable#close. AFAIU, it's required to avoid declaring Exception?

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/ScopedContext.java
```diff
@@ -0,0 +1,85 @@
...
+        this.val = val;
+    }
+
+    /** */
+    public <T> ScopedContext with(ContextAttribute<T> attr, T val) {
+        return attr.get() == val ? this : new ScopedContext(attr, val, this);
+    }
+
+    /** */
+    public static ScopedContext create() {
```
timoninmaxim: Looks as misnaming. This method doesn't create anything

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/ContextAttribute.java
```diff
@@ -0,0 +1,84 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.thread.context;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+/** */
+public class ContextAttribute<T> {
```
timoninmaxim: I suppose you publicate the "Scope" term. Let's use at least public terms if possible. `ScopeAttribute` looks better.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/ContextAttribute.java
```diff
@@ -0,0 +1,84 @@
...
+    /** */
+    private final int bitmask;
+
+    /** */
+    private final T initVal;
+
+    /** */
+    private ContextAttribute(byte id, T initVal) {
+        this.id = id;
+        this.bitmask = 1 << id;
```
timoninmaxim: Looks that storing ID is enough. The bitmask is a parameter of ContextDataChain only.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/ContextAttribute.java
```diff
@@ -0,0 +1,84 @@
...
+    public static <T> ContextAttribute<T> newInstance() {
+        return newInstance(null);
+    }
+
+    /** */
+    public static <T> ContextAttribute<T> newInstance(T initVal) {
+        int id = ID_GEN.getAndIncrement();
+
+        if (MAX_ATTRIBUTE_CNT <= id) {
+            throw new RuntimeException("Exceeded maximum supported number of created Context Attributes instances" +
```
timoninmaxim: assert is better, as we should not handle this exception. Developers must redesign this module, if amount of attributes too high. It even looks like we need a repository of known attributes with their IDs.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/ScopedContext.java
```diff
@@ -0,0 +1,85 @@
...
+    /** */
+    private <T> ScopedContext(ContextAttribute<T> attr, T val, ScopedContext prev) {
+        super(prev.storedAttributeBits() | attr.bitmask(), prev);
+        
+        this.attr = attr;
+        this.val = val;
+    }
+
+    /** */
+    public <T> ScopedContext with(ContextAttribute<T> attr, T val) {
```
timoninmaxim: Looks weird, an object that stores an attribute value accepts another attribute (name + value). AFAIU, it's a responsibility of Scope to store all attributes within single scope.

timoninmaxim (changes requested):

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Scope.java
petrov-mg: Done.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/ScopedContext.java
petrov-mg: The whole method removed.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/ContextAttribute.java
petrov-mg: Nope. The Scope has nothing to do with the Attribute. Context does.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/ContextAttribute.java
petrov-mg: Nope. The bitmask is used in many places to check for the presence of an attribute value. Therefore, it's better to calculate it once and store it in the attribute instance.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/ContextAttribute.java
petrov-mg: Done.

`It even looks like we need a repository of known attributes with their IDs`

Definitely not in the current patch.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/ScopedContext.java
petrov-mg: `it's a responsibility of Scope to store all attributes within single scope`

Nope. See comments above.

I changed class naming a bit so it may be less confusing now.

`Looks weird, an object that stores an attribute value accepts another attribute (name + value)`
Also keep in mind that the current patch is based on intrusive lists and it is unlikely to be possible to avoid the "weirdness" you mentioned.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Context.java
```diff
@@ -0,0 +1,113 @@
...
+     * more mappings form {@link ContextAttribute} to its value.
+     *
+     * @see AttributeValueHolder#with(ContextAttribute, Object)
+     */
+    public static <T> AttributeValueHolder with(ContextAttribute<T> attr, T val) {
+        return AttributeValueHolder.ROOT.with(attr, val);
+    }
+
+    /** */
+    public static final class AttributeValueHolder extends ContextDataChainNode<AttributeValueHolder> {
```
timoninmaxim: Public nested classes must be regular classes

timoninmaxim: package-private

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Context.java
```diff
@@ -0,0 +1,113 @@
...
+ */
+public final class Context {
+    /**
+     * Creates a new Context containing a single mapping from the specified {@link AttributeValueHolder} to its value.
+     * The returned {@link AttributeValueHolder} represents the added mapping and can be used to accumulate to Context
+     * more mappings form {@link ContextAttribute} to its value.
+     *
+     * @see AttributeValueHolder#with(ContextAttribute, Object)
+     */
+    public static <T> AttributeValueHolder with(ContextAttribute<T> attr, T val) {
```
timoninmaxim: this class must not escape from this package

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Context.java
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * Attaches {@link ContextAttribute} values stored in current Context to the thread from which this method is
+     * called. If {@link ContextAttribute} value was already attached for the current thread, its value will be
+     * stashed and replaced by the new one.
+     *
+     * @return {@link Scope} instance that, when closed, resets the values for all {@link ContextAttribute}s added
+     * to the current Context and restores them to the previously attached values, if any.
+     */
+    public Scope attach() {
```
timoninmaxim: can be inlined in Builder.build()

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Context.java
```diff
@@ -0,0 +1,321 @@
...
+    /**
+     * Updates the value of the specified attribute for the Context bound to the thread this method is called from.
+     *
+     * @param attr Context Attribute.
+     * @return Scope instance that, when closed, undoes the applied update. It is crucial to undo all applied Context
+     * updates to free up thread-bound resources and avoid memory leaks, so it is highly encouraged to use a
+     * try-with-resource block to close the returned Scope. Note, updates must be undone in the same order they were applied.
+     */
+    public static <T> Scope set(ContextAttribute<T> attr, T val) {
+        if (get(attr) == val)
```
timoninmaxim: WDYT, can we move this optimization into `applyUpdates`? Currently, there are 2 calls of `INSTANCE.get()`, that the most expensive part.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Context.java
```diff
@@ -0,0 +1,321 @@
...
+    private static class ContextUpdater {
+        /** */
+        private static final int INIT_UPDATES_CAPACITY = 3;
+
+        /** */
+        private List<AttributeValueHolder<?>> updates;
+
+        /** */  
+        <T> ContextUpdater set(ContextAttribute<T> attr, T val) {
+            if (get(attr) == val)
```
timoninmaxim: Same here, can we check it later, within Context?

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/OperationContext.java
```diff
@@ -0,0 +1,321 @@
...
+
+            AttributeValueHolder<?>[] sealedUpdates = new AttributeValueHolder[updates.size()];
+
+            updates.toArray(sealedUpdates);
+
+            return INSTANCE.get().applyUpdates(sealedUpdates);
+        }
+
+        /** */
+        static ContextUpdater create() {
```
timoninmaxim, line 366: can be inlined

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Context.java
```diff
@@ -0,0 +1,321 @@
...
+ *         +-----------+   +-----------+
+ *         |           |   | A1 -> V2  |
+ * null <--| A1 -> V1  |<--|           |
+ *         |           |   | A2 -> V3  |
+ *         +-----------+   +-----------+
+ *</pre>
+ * Context Updates can be undone in the same order they were applied by closing the {@link Scope} associated with each
+ * update (see {@link #set(ContextAttribute, Object)} and related methods).
+ */
+public class Context {
```
timoninmaxim: WDYT, if Context will implement ContextSnapshot instead of Update? Looks like this makes snapshotting more clear.

```
    private ContextSnapshot createSnapshotInternal() {
        return lastUpd == null ? null : new Context(lastUpd);
    }

    private Scope restoreSnapshotInternal(ContextSnapshot newSnp) {
        ContextSnapshot prevSnp = createSnapshotInternal();

        if (newSnp == prevSnp)
            return Scope.NOOP_SCOPE;

        INSTANCE.set((Context)newSnp);

        return () -> INSTANCE.set((Context)prevSnp);
    }
```

modules/core/src/main/java/org/apache/ignite/internal/processors/pool/PoolProcessor.java
```diff
@@ -280,7 +280,7 @@ public PoolProcessor(GridKernalContext ctx) {
                         throw new IgniteException("Failed to register IO executor pool because its ID as " +
                             "already used: " + id);
 
-                    extPools[id] = ctx.security().enabled() ? new SecurityAwareIoPool(ctx.security(), ex) : ex;
+                    extPools[id] = ctx.security().enabled() ? ContextAwareIoPool.wrap(ex) : ex;
```
timoninmaxim: Here and below: looks like we should not check security anymore and use ContextAwareIoPool everywhere?

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/function/ContextAwareCallable.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    /**
+     * Creates a wrapper that stores a specified {@link Callable} along with the Snapshot of the Context bound
+     * to the thread when this method is called. Captured Context will be restored before {@link Callable} execution,
+     * potentially in another thread.
+     */
+    public static <T> Callable<T> wrap(Callable<T> delegate) {
+        return wrap(delegate, ContextAwareCallable::new);
```
timoninmaxim: Let's check should we cache lambda `ContextAwareCallable::new`. This lambda must not create every `wrap` call

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Context.java
```diff
@@ -0,0 +1,331 @@
...
+        AttributeValueHolder<T> valHolder = findAttributeValue(attr);
+
+        assert valHolder != null;
+        assert valHolder.attribute().equals(attr);
+
+        return valHolder.value();
+    }
+
+    /** Updates the current context with the specified attributes and their corresponding values. */
+    private Scope applyAttributeUpdates(AttributeValueHolder<?>... atrVals) {
```
timoninmaxim: Misprint, atrVals -> attrVals

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/OperationContext.java
```diff
@@ -0,0 +1,331 @@
...
+            // We iterate in reverse order to correctly handle the case when the value for the same attribute is
+            // specified multiple times.
+            for (int i = attrVals.length - 1; i >= 0; i--) {
+                AttributeValueHolder<?> valHolder = attrVals[i];
+
+                if (valHolder.attribute().equals(attr))
+                    return ((AttributeValueHolder<T>)valHolder);
+            }
+
+            return null;
```
timoninmaxim, line 290: This line doesn't covered within ContextAttributeTest

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/OperationContext.java
```diff
@@ -0,0 +1,331 @@
...
+    /** Iterates over the currently applied context updates and finds the latest value associated with the specified attribute. */
+    private <T> AttributeValueHolder<T> findAttributeValue(ContextAttribute<T> attr) {
+        for (Update upd = lastUpd; upd != null; upd = upd.prev) {
+            if (!upd.holdsValueFor(attr))
+                continue;
+
+            return upd.value(attr);
+        }
+
+        return null;
```
timoninmaxim, line 204: This line doesn't covered within ContextAttributeTest

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Context.java
```diff
@@ -0,0 +1,331 @@
...
+    /** */
+    private Scope restoreSnapshotInternal(ContextSnapshot newSnp) {
+        ContextSnapshot prevSnp = createSnapshotInternal();
+
+        if (newSnp == prevSnp)
+            return NOOP_SCOPE;
+
+        changeState(prevSnp, newSnp);
+
+        return () -> INSTANCE.get().changeState(newSnp, prevSnp);
```
timoninmaxim: It looks it should work with `() -> changeState(newSnp, prevSnp)`. Because we close Scope within the same thread that restores snapshot.

petrov-mg: Done.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Context.java
petrov-mg: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/pool/PoolProcessor.java
petrov-mg: We will definitely do this, but in a separate ticket.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Context.java
petrov-mg: Done.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Context.java
petrov-mg: Done.

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/Context.java
```diff
@@ -0,0 +1,346 @@
...
+ * @see ContextSnapshot
+ * @see ContextAwareWrapper
+ * @see ContextAwareExecutor
+ */
+public class Context {
+    /** */
+    private static final ThreadLocal<Context> INSTANCE = ThreadLocal.withInitial(Context::new);
+
+    /**
+     * Sequence of updated applied to the Context. Each update holds a link to the previous Update, so we store only
```
sergey-chugunov-1985:
```suggestion
     * Sequence of updates applied to the Context. Each update has a link to the previous Update, so we store only
```

modules/commons/src/main/java/org/apache/ignite/internal/thread/context/ContextSnapshot.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package org.apache.ignite.internal.thread.context;
+
+import org.apache.ignite.internal.thread.context.function.ContextAwareCallable;
+import org.apache.ignite.internal.thread.context.function.ContextAwareRunnable;
+
+/**
+ * Represents snapshot of all Attributes and their corresponding values for a particular {@link Context} instance.
+ * Its main purpose to save Context state and restore it later, possible for Context bound to another thread.
```
sergey-chugunov-1985:
```suggestion
 * The main purpose of the interface is to save Context state and restore it later, possibly in a Context bound to another thread.
```

PR #12517, opened 2025-11-18
IGNITE-26584 Discovery optimizations for MultiDC

anton-vinogradov:

### Notes

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/MdcAwareNodesComparator.java
+++ b/modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/MdcAwareNodesComparator.java
@@ -23,10 +23,15 @@ import java.util.Comparator;
 public class MdcAwareNodesComparator implements Comparator<TcpDiscoveryNode> {
     /** */
     @Override public int compare(TcpDiscoveryNode n1, TcpDiscoveryNode n2) {
-        String n1DcId = n1.dataCenterId() == null ? "" : n1.dataCenterId();
-        String n2DcId = n2.dataCenterId() == null ? "" : n2.dataCenterId();
+        int res = 0;
 
-        int res = n1DcId.compareTo(n2DcId);
+        String n1DcId = n1.dataCenterId();
+        String n2DcId = n2.dataCenterId();
+
+        assert (n1DcId != null && n2DcId != null) || (n1DcId == null && n2DcId == null) : "[node1=" + n1 + ", node2=" + n2 + "]";
+
+        if (n1DcId != null)
+            res = n1DcId.compareTo(n2DcId);
 
         if (res == 0)
             res = n1.compareTo(n2);


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/MdcAwareNodesComparator.java
```diff
@@ -0,0 +1,37 @@
...
+/** Compares nodes using the Data Center id as a primary factor. */
+public class MdcAwareNodesComparator implements Comparator<TcpDiscoveryNode> {
+    /** */
+    @Override public int compare(TcpDiscoveryNode n1, TcpDiscoveryNode n2) {
+        String n1DcId = n1.dataCenterId() == null ? "" : n1.dataCenterId();
+        String n2DcId = n2.dataCenterId() == null ? "" : n2.dataCenterId();
+
+        int res = n1DcId.compareTo(n2DcId);
+
+        if (res == 0) {
```
alex-plekhanov: Redundant braces

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNodesRing.java
```diff
@@ -506,7 +506,10 @@ public void clear() {
             if (filtered.size() < 2)
                 return null;
 
-            Iterator<TcpDiscoveryNode> iter = filtered.iterator();
+            NavigableSet<TcpDiscoveryNode> sorted = new TreeSet<>(new MdcAwareNodesComparator());
+            sorted.addAll(filtered);
```
alex-plekhanov: I'm not sure if `TreeSet` always preserves order of added elements if all elements are equal according to comparator, so it potentianly can break something in non-multi-DC cluster. Let's use sorting only for multi-DC environment. Also it's a potentially dangerous feature even for multi-DC, I think it's worth to have a flag to disable it explicitely.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNodesRing.java
```diff
@@ -506,7 +506,10 @@ public void clear() {
             if (filtered.size() < 2)
                 return null;
 
-            Iterator<TcpDiscoveryNode> iter = filtered.iterator();
```
alex-plekhanov, line 509: Methodd `TcpDiscoverySpi#forceNextNodeFailure` -> `ServerImpl#forceNextNodeFailure` depend on `nextNode`,  these methods are never used and can be cleaned up.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/MdcAwareNodesComparator.java
```diff
@@ -0,0 +1,37 @@
...
+package org.apache.ignite.spi.discovery.tcp.internal;
+
+import java.util.Comparator;
+
+/** Compares nodes using the Data Center id as a primary factor. */
+public class MdcAwareNodesComparator implements Comparator<TcpDiscoveryNode> {
+    /** */
+    @Override public int compare(TcpDiscoveryNode n1, TcpDiscoveryNode n2) {
+        String n1DcId = n1.dataCenterId() == null ? "" : n1.dataCenterId();
+        String n2DcId = n2.dataCenterId() == null ? "" : n2.dataCenterId();
```
alex-plekhanov: For multi-DC environment we have a check, that all `dataCenterId` is provided. We can ommit nullability check if comparator will be used only for multi-DC.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNodesRing.java
anton-vinogradov: MdcAwareNodesComparator guarantee elements to be non-equal since it compares dcId and if they are equal it compares nodes as usual.

```
if (res == 0)
   res = n1.compareTo(n2);
```

sergey-chugunov-1985: I agree that having a switch to turn this feature off and fall back to a manual management of ring structure is necessary.

There is always a chance that we miss an edge case during testing and a critical bug in discovery makes it to production. Having a flag as a way to get back to a suboptimal but working implementation is critical in such situation.

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/MultiDataCenterRignTest.java
```diff
@@ -0,0 +1,149 @@
...
+import org.apache.ignite.internal.util.typedef.internal.U;
+import org.apache.ignite.lang.IgniteInClosure;
+import org.apache.ignite.spi.discovery.DiscoverySpi;
+import org.apache.ignite.spi.discovery.tcp.messages.TcpDiscoveryAbstractMessage;
+import org.apache.ignite.spi.discovery.tcp.messages.TcpDiscoveryNodeAddedMessage;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.junit.Test;
+
+/** */
+public class MultiDataCenterRignTest extends GridCommonAbstractTest {
```
sergey-chugunov-1985:
```suggestion
public class MultiDataCenterRingTest extends GridCommonAbstractTest {
```

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -2733,42 +2699,15 @@ private static class PendingMessages implements Iterable<TcpDiscoveryAbstractMes
          */
         void add(TcpDiscoveryAbstractMessage msg) {
             msgs.add(new PendingMessage(msg));
-
-            while (msgs.size() > MAX) {
```
sergey-chugunov-1985, line 2737: As far as I can see, the old implementation have this logic to maintain internal queue size and not let it get close to MAX * 2 size right on add operation.

Now there is no such logic, we clean up internal collection on discard call itself.

Do we understand the logic behing this old behavior? What are situations when this logic starts working? Does new implementation perform better or worse in these situations?

This whole discard logic looks very important to me, and I want to make sure that this change doesn't break some corner case with lagging network which is not covered with our test base.

anton-vinogradov: Old sollution (queue limiting) seems to be a buggy. In case of a queue overfill we just lose the consistency.
New solution guarantee the consistency and cleans the elements only when it's safe to clean.
Not sure we have deployments for 1K+ nodes where this will NOT happen now, but code becomes more stable, I think.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/TcpDiscoveryNodesRing.java
anton-vinogradov: Fixed with

```
private static final boolean mdcAwareRing = IgniteSystemProperties.getBoolean("MDC_AWARE_RING", true);
```

```
Collection<TcpDiscoveryNode> sorted;

if (mdcAwareRing) {
    sorted = new TreeSet<>(new MdcAwareNodesComparator());
    sorted.addAll(nodes);
}
else
    sorted = nodes;
```

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/internal/MdcAwareNodesComparator.java
anton-vinogradov: Removed the "" defaults.

Issue #11960, opened 2025-03-23
Automatic Node Discovery fails in .NET Client with HTTPS enabled


martinsuchan:
We've discovered a bug in Ignite 2.15-2.17 if `EnableClusterDiscovery` property is set in .NET Client when connecting to a cluster with HTTPS enabled on each host.
If we try to connect form .NET Thin client in this configuration, it works without any certificate issues, but then if the **Automatic Server Node Discovery** is enabled (IgniteClientConfiguration.EnableClusterDiscovery), it fails on `RemoteCertificateMismatch`. Interestingly this error is only logged in the background, we can stilll connect to the cluster.

We've tried to debug it and we discovered, that requests in the server discovery background process are done by targeting IP addresses of our nodes, rather than using host names or DNS aliases of nodes.
Because of that these requests fail via HTTPS, because the IP address is not inlcluded in our host certificates in the SAN list. Including IP addresses in node certificates is not an option in our case.

Expected behavior: Automatic Server Node Discovery over HTTPS should not fail if each host has a trusted certificate with proper host name in the SAN list. **Node Discovery requests should use hostnames instead of IP addresses for connections.**

See this line in `ClientFailoverSocket.cs` where this behavior starts:
https://github.com/apache/ignite/blob/be1f4bc6378c0ceb75a16c286a1a6ee00875d624/modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/ClientFailoverSocket.cs#L133

ptupitsyn:
Thank you for the detailed report, ticket created: https://issues.apache.org/jira/browse/IGNITE-24906

martinsuchan:
Update, the same error is also logged when using this property in .NET Thin client and using HTTPS on server:
`IgniteClientConfiguration.EnableHeartbeats = true`
even when automatic node discovery and partition awareness is set to false.

ptupitsyn:
@martinsuchan do you have `IgniteConfiguration.localhost` configured on the server side, set to the correct host name?

martinsuchan:
@ptupitsyn we don't use this property anywhere. Also I don't see how this property is related to Node Discovery making HTTPS requests to IP addresses instead of to hostnames?

ptupitsyn:
@martinsuchan I'm trying to understand the expectations - where should those host names come from?
 
* Node discovery asks the cluster for the list of active nodes
[...]

PR #12549, opened 2025-12-02
IGNITE-24906 Improve client cluster discovery docs

ptupitsyn:
Clarify scenarios like #11960

### DIFF
--- a/docs/_docs/thin-clients/java-thin-client.adoc
+++ b/docs/_docs/thin-clients/java-thin-client.adoc
@@ -86,7 +86,6 @@ Additionally, thin client receives topology updates only when it performs some o
 
 [WARNING]
 ====
-[discrete]
 Server discovery may not work when servers are behind a NAT server or a proxy.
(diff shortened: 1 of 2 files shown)


docs/_docs/thin-clients/java-thin-client.adoc
```diff
@@ -75,6 +75,34 @@ include::{sourceCodeFile}[tag=connect-to-many-nodes,indent=0]
...
+
+Thin client can discover server nodes in the cluster automatically.
+This behavior is enabled when `ClientConfiguration.clusterDiscoveryEnabled` is enabled (`true` by default).
+
+Server discovery is an asynchronous process - it happens in the background.
+Additionally, thin client receives topology updates only when it performs some operations (to minimize server load and network traffic from idle connections).
+
+[WARNING]
+====
+[discrete]
```
IgGusev: I think we can drop this. There is no header in the warning.

ptupitsyn: Fixed here and on .NET page.

PR #12568, opened 2025-12-09
IGNITE-27294 : Use MessageSerializer for TcpDiscoveryHandshakeRequest

Vladsz83:


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryAbstractMessage.java
```diff
@@ -293,8 +293,11 @@ public void addFailedNode(UUID nodeId) {
     /**
      * @param failedNodes Failed nodes.
      */
-    public void failedNodes(@Nullable Set<UUID> failedNodes) {
-        this.failedNodes = failedNodes;
+    public void failedNodes(@Nullable Collection<UUID> failedNodes) {
```
shishkovilja: Let's wait for merging of #12581.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryHandshakeRequest.java
```diff
@@ -67,7 +79,7 @@ public UUID checkPreviousNodeId() {
      *
      * @param prevNodeId If not {@code null}, will set topology check flag and set node ID to check.
      */
-    public void changeTopology(UUID prevNodeId) {
+    public void previousNodeId(@Nullable UUID prevNodeId) {
         setFlag(CHANGE_TOPOLOGY_FLAG_POS, prevNodeId != null);
```
shishkovilja: This logic should be moved outside of setter. In old serde via Java serialization flag was not set during deserialization of message, so it should not being set here in setter, which is used in new serialization framework.

I think, it is quite easy to move setting of flag outside of the method.

PR #12574, opened 2025-12-10
IGNITE-27317 Refactor serialization for TcpDiscoveryAbstractTraceableMessage

Vladsz83:


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryIoSession.java
```diff
@@ -192,37 +194,26 @@ <T> T readMessage() throws IgniteCheckedException, IOException {
...
 
-                msgBuf.clear();
+                if (finished)
+                    break;
 
-                if (unprocessedTail != null)
-                    msgBuf.put(unprocessedTail);
+                msgBuf.clear();
             }
             while (true);
```
shishkovilja: `while (!finished)` ?

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryIoSession.java
```diff
@@ -192,37 +194,26 @@ <T> T readMessage() throws IgniteCheckedException, IOException {
...
+                    in.acceptPrefixBuffer(unprocessedReadTail);
                 }
 
-                msgBuf.clear();
+                if (finished)
+                    break;
 
-                if (unprocessedTail != null)
-                    msgBuf.put(unprocessedTail);
+                msgBuf.clear();
```
shishkovilja: Can be moved to start of cycle. Also, we can remove clearing of buffer just before cycle (line 189 now)

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryIoSession.java
```diff
@@ -192,37 +194,26 @@ <T> T readMessage() throws IgniteCheckedException, IOException {
...
+                    msgBuf.get(unprocessedReadTail, 0, msgBuf.remaining());
 
-                if (msgBuf.remaining() > 0) {
-                    unprocessedTail = new byte[msgBuf.remaining()];
-                    msgBuf.get(unprocessedTail, 0, msgBuf.remaining());
+                    in.acceptPrefixBuffer(unprocessedReadTail);
                 }
 
-                msgBuf.clear();
+                if (finished)
```
shishkovilja: Can be removed, if we replace `while(true)` to `while(!finished)`

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryIoSession.java
```diff
@@ -192,37 +194,26 @@ <T> T readMessage() throws IgniteCheckedException, IOException {
...
+                // Server Discovery only sends next message to next Server upon receiving a receipt for the previous one.
                 // This behaviour guarantees that we never read a next message from the buffer right after the end of
-                // the previous message.
-                assert msgBuf.remaining() == 0 || !finished : "Some data was read from the socket but left unprocessed.";
-
-                if (finished)
-                    break;
+                // the previous message. But it is not guaranteed with Client Discovery where messages aren't acknowledged.
+                // Thus, we have to keep the uprocessed bytes read from the socket. It won't return them again.
+                if (msgBuf.remaining() > 0) {
```
shishkovilja: `msgBuf.hasRemaining()` ?

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryNodeLeftMessage.java
```diff
@@ -39,6 +46,11 @@ public TcpDiscoveryNodeLeftMessage(UUID creatorNodeId) {
         super(creatorNodeId);
     }
 
+    /** {@inheritDoc} */
+    @Override public short directType() {
+        return 14;
```
wernerdv: It should be `16`.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryAbstractTraceableMessage.java
```diff
@@ -29,8 +31,12 @@ public abstract class TcpDiscoveryAbstractTraceableMessage extends TcpDiscoveryA
     /** Container. */
     private SpanContainer spanContainer = new SpanContainer();
 
+    /** Serialization holder of {@link #spanContainer}'s bytes. */
+    @Order(value = 5, method = "spanBytes")
```
wernerdv, line 36: Please add `@SuppressWarnings("unused")`

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryAbstractTraceableMessage.java
```diff
@@ -67,6 +74,21 @@ public Object readResolve() {
...
+    public @Nullable byte[] spanBytes() {
+        return spanContainer == null ? null : spanContainer.serializedSpanBytes();
+    }
+
+    /** @param spanBytes {@link #spanContainer}'s bytes. */
+    public void spanBytes(@Nullable byte[] spanBytes) {
+        if (spanBytes == null)
+            return;
+
+        readResolve();
```
shishkovilja: `readResolve` ensures, that `SpanContainer` will be **always** initialized during deserialization via Java deserialization.

I suggest to call `readResolve` in the beginning of the method.

Vladsz83: The java serializatiob, JDKMarshaller calls readResolve() itself. In `ObjectStreamClass`:
`readResolveMethod = getInheritableMethod(cl, "readResolve", null, Object.class)`

shishkovilja: @Vladsz83 , that's what I am talking about. But now behaviour changed with new framework and java serde is not used.

Vladsz83: With the new serialization we create a message object normally, by a constructor. The field initialization `private SpanContainer spanContainer = new SpanContainer();` should work.

shishkovilja: Yes, you are right. We don't need call `readResolve` in this setter at all.

PR #12578, opened 2025-12-11
IGNITE-27318 Use MessageSerializer for TcpDiscoveryRingLatencyCheckMessage

Vladsz83:


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryAbstractMessage.java
```diff
@@ -293,8 +293,10 @@ public void addFailedNode(UUID nodeId) {
     /**
      * @param failedNodes Failed nodes.
      */
-    public void failedNodes(@Nullable Set<UUID> failedNodes) {
-        this.failedNodes = failedNodes;
+    public void failedNodes(@Nullable Collection<UUID> failedNodes) {
```
shishkovilja: Let's wait for #12581, which brings support of `Set` serde.

PR #12598, opened 2025-12-22
IGNITE-27417 : Use MessageSerializer for TcpDiscoveryClientMetricsUpdateMessage

Vladsz83:


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryIoSession.java
```diff
@@ -185,19 +184,39 @@ <T> T readMessage() throws IgniteCheckedException, IOException {
...
                 if (read == -1)
                     throw new EOFException("Connection closed before message was fully read.");
 
-                msgBuf.limit(read);
+                msgBuf.limit(msgBuf.position() > 0 ? msgBuf.position() + read + 1 : read);
 
                 finished = msgSer.readFrom(msg, msgReader);
-            } while (!finished);
+
+                // We assume that there is only one message in the socket because Discovery is a serial protocol with
```
sergey-chugunov-1985:
```suggestion
                // We rely on the fact that Discovery only sends next message upon receiving a receipt for the previous one. This behaviour guarantees that we never read from msgBuf a beginning of a next message right after the end of the previous one.
```

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryClusterMetricsHolderMessage.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.ignite.cluster.ClusterMetrics;
+import org.apache.ignite.internal.managers.discovery.DiscoveryMessageFactory;
+import org.apache.ignite.internal.processors.cluster.NodeMetricsMessage;
+import org.apache.ignite.internal.util.typedef.internal.S;
+
+/**
+ * Utility container message of the node metrics. Is not a pure {@link TcpDiscoveryAbstractMessage}.
+ * Reuses Communication's {@link NodeMetricsMessage}.
+ */
+public class TcpDiscoveryClusterMetricsHolderMessage extends NodeMetricsMessage {
```
sergey-chugunov-1985:
```suggestion
public class TcpDiscoveryNodeMetricsMessage extends NodeMetricsMessage {
```

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryClusterMetricsHolderMessage.java
```diff
@@ -0,0 +1,49 @@
...
+
+package org.apache.ignite.spi.discovery.tcp.messages;
+
+import org.apache.ignite.cluster.ClusterMetrics;
+import org.apache.ignite.internal.managers.discovery.DiscoveryMessageFactory;
+import org.apache.ignite.internal.processors.cluster.NodeMetricsMessage;
+import org.apache.ignite.internal.util.typedef.internal.S;
+
+/**
+ * Utility container message of the node metrics. Is not a pure {@link TcpDiscoveryAbstractMessage}.
```
sergey-chugunov-1985:
```suggestion
 * We cannot directly reuse `NodeMetricsMessage` in Discovery as it is registered in a message factory of Communication component 
 *  and thus is unavailable in Discovery.
 * We have to extend `NodeMetricsMessage` and register this subclass in message factory of Discovery component.
```

PR #12609, opened 2025-12-26
IGNITE-27341 .ΝΕΤ: Propagate compute task names to Java

ravxz:
Ensure that the correct .NET task or closure name is visible in system views and internal APIs.

### DIFF
--- a/modules/platforms/dotnet/Apache.Ignite.Core.Tests/Compute/ComputeTaskName.cs
+++ b/modules/platforms/dotnet/Apache.Ignite.Core.Tests/Compute/ComputeTaskName.cs
@@ -17,6 +17,7 @@
 
 namespace Apache.Ignite.Core.Tests.Compute
 {
+    using System;
     using System.Collections.Generic;
     using System.Linq;
     using System.Threading;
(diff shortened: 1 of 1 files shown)


modules/platforms/dotnet/Apache.Ignite.Core/Compute/ICompute.cs
```diff
@@ -140,6 +140,19 @@ public interface ICompute
...
+        /// refer to <see cref="IComputeTask{A,T,R}"/> documentation.
+        /// </summary>
+        /// <param name="taskName">Task name</param>
+        /// <param name="task">Task to execute.</param>
+        /// <param name="taskArg">Optional task argument.</param>
+        /// <returns>Task result.</returns>
+        /// <typeparam name="TArg">Argument type.</typeparam>
+        /// <typeparam name="TJobRes">Type of job result.</typeparam>
+        /// <typeparam name="TRes">Type of final task result.</typeparam>
+        TRes Execute<TArg, TJobRes, TRes>(string taskName, IComputeTask<TArg, TJobRes, TRes> task, TArg taskArg);
```
ptupitsyn: This is different from the Java API, which has only one `execute` method with a task name:
```
public <T, R> R execute(String taskName, @Nullable T arg)
```

Can you please explain the idea here?

ravxz: I referred to the methods ExecuteJavaTask() and ExecuteJavaTaskAsync() upon enhancing similar methods for .Net tasks invocation, since the new argument 'taskName' was present in all versions of java task call methods.

The initial idea was to promote task name from .Net task name to Java task. Then I introduced this argument to all .Net task call methods. May be the right place for this change is a different PR.

ptupitsyn: I don't think we need to change anything in .NET, especially in the public API. The ticket is about passing correct task name to system views, metrics, and logs. `PlatformFullTask#taskName` already contains what we need, and the fix in `GridTaskProcessor` should be enough.

ravxz: I rolled back changes in .Net Compute Task API.

ptupitsyn (commented):
Please add tests.

ravxz:
> Please add tests.

Introduced auto test on .Net compute task name propagation to Platform Task.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Compute/ComputeTaskName.cs
```diff
@@ -0,0 +1,173 @@
...
+            _grid1.GetCompute().ExecuteAsync(task, 123);
+
+            // Check task in system views via SQL
+            var results = _grid1
+                .GetOrCreateCache<string, string>("test")
+                .Query(new SqlFieldsQuery("SELECT * FROM SYS.TASKS", null)).GetAll();
+            
+            Assert.AreEqual(1, results.Count());
+            Assert.AreEqual("Apache.Ignite.Core.Tests.Compute.ComputeTaskNameTest+LongTask", results[0][3]);
+            Assert.AreEqual("org.apache.ignite.internal.processors.platform.compute.PlatformFullTask", results[0][4]);
```
ptupitsyn: Let's use explicit column names:

```suggestion
            var res = _grid1
                .GetOrCreateCache<string, string>("test")
                .Query(new SqlFieldsQuery("SELECT TASK_NAME, TASK_CLASS_NAME FROM SYS.TASKS", null))
                .GetAll()
                .Single();
            
            Assert.AreEqual("Apache.Ignite.Core.Tests.Compute.ComputeTaskNameTest+LongTask", res[0]);
            Assert.AreEqual("org.apache.ignite.internal.processors.platform.compute.PlatformFullTask", res[1]);
```

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Compute/ComputeTaskName.cs
```diff
@@ -0,0 +1,173 @@
...
+    using Apache.Ignite.Core.Cache.Query;
+    using Apache.Ignite.Core.Client;
+    using Apache.Ignite.Core.Compute;
+    using NUnit.Framework;
+    using static TestUtils;
+
+    /// <summary>
+    /// Tests for <see cref="IComputeTaskSession"/>
+    /// </summary>
+    public class ComputeTaskNameTest
```
ptupitsyn: Let's inherit from `TestBase` to reduce boilerplate.

ptupitsyn:
Looks good to me. Waiting for CI build.

ravxz:
Enhanced task name propagation for other compute methods. It seems to fix similar cases for closures.

PR #12620, opened 2025-12-30
IGNITE-27542 : Use MessageSerializer for TcpDiscoveryMetricsUpdateMessage

Vladsz83:


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryNodesMetricsMapMessage.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.Map;
+import java.util.UUID;
+import org.apache.ignite.internal.Order;
+import org.apache.ignite.internal.managers.discovery.DiscoveryMessageFactory;
+import org.apache.ignite.internal.util.typedef.internal.S;
+import org.apache.ignite.plugin.extensions.communication.Message;
+
+/** Holds map of nodes metrics messages per node id. */
+public class TcpDiscoveryNodesMetricsMapMessage implements Message {
```
sergey-chugunov-1985: From usages of this class I can see that it is used solely for metrics from client nodes. I think we should capture this information right in the name of the class to make it easier for a developer to understand its purpose.

What do you think about a name like `TcpDiscoveryClientNodesMetricsMessage`?

In that case it may make sense to rewrite javadoc as well.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryMetricsUpdateMessage.java
```diff
@@ -79,130 +83,136 @@ public TcpDiscoveryMetricsUpdateMessage(UUID creatorNodeId) {
...
-     * @param nodeId Server node ID.
-     * @param clientNodeId Client node ID.
-     * @param metrics Node metrics.
+     * @param srvrId Server node ID.
+     * @param clientNodeId Connected client node ID.
+     * @param clientMetrics Client metrics.
      */
-    public void setClientMetrics(UUID nodeId, UUID clientNodeId, ClusterMetrics metrics) {
-        assert nodeId != null;
+    public void setClientMetrics(UUID srvrId, UUID clientNodeId, ClusterMetrics clientMetrics) {
```
sergey-chugunov-1985:
```suggestion
    public void addClientMetrics(UUID srvrId, UUID clientNodeId, ClusterMetrics clientMetrics) {
```

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryMetricsUpdateMessage.java
```diff
@@ -79,130 +83,136 @@ public TcpDiscoveryMetricsUpdateMessage(UUID creatorNodeId) {
...
-     * @param metrics Node metrics.
+     * @param srvrId Server ID.
+     * @param newMetrics New server metrics to add.
      */
-    public void setMetrics(UUID nodeId, ClusterMetrics metrics) {
-        assert nodeId != null;
-        assert metrics != null;
-        assert !this.metrics.containsKey(nodeId);
+    public void addServerMetrics(UUID srvrId, ClusterMetrics newMetrics) {
+        assert srvrId != null;
```
sergey-chugunov-1985, line 90: Do we really need to protect something in metric message with asserts? I believe that metrics are not that important, and even if some parameters are null, we could skip any actions with metrics without big harm to functionality of the cluster.

At the same time triggered assertion could cause a node or ever a cluster to shut down.

I don't think we should tie stability of the cluster to possible issues with metrics gathering. Importance of metrics is lower than an overall stability of the cluster.

Vladsz83: We still can drop a node by the metrics. At least a client. Look for '_Failing client node due to not receiving metrics updates_ '. It traverses a map. If key not found, node is considered as failed. Unfortunatelly. I would not touch this logic and what related to it. Instead, we might move these metrics into Communication. And/or remove them at all.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryNodesMetricsMapMessage.java
Vladsz83: I thought of it. Not sure. `Where we use` and `what it can do` are different. In the scope of this class, metrics of nodes are held. It can be either server metrics or client metrics. Would work with both.

Where it is actually used, the notation and/or javadoc 'client' or 'clients' is present.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -3399,10 +3390,11 @@ private void sendMessageToClients(TcpDiscoveryAbstractMessage msg) {
                 for (ClientMessageWorker clientMsgWorker : clientMsgWorkers.values()) {
                     if (msgBytes == null) {
                         try {
-                            msgBytes = U.marshal(spi.marshaller(), msg);
+                            msgBytes = clientMsgWorker.ses.serializeMessage(msg);
                         }
-                        catch (IgniteCheckedException e) {
-                            U.error(log, "Failed to marshal message: " + msg, e);
+                        catch (IgniteCheckedException | IOException e) {
+                            U.error(log, "Failed to serialize message to a client: " + msg + ", client id: "
```
sergey-chugunov-1985:
```suggestion
                            U.error(log, "Failed to serialize message: " + msg + ", recepient client id: "
```

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -6073,9 +6066,11 @@ private void processMetricsUpdateMessage(TcpDiscoveryMetricsUpdateMessage msg) {
                 }
                 else {
                     // Message is on its second ring.
-                    removeMetrics(msg, locNodeId);
+                    msg.removeServerMetrics(locNodeId);
 
-                    Collection<UUID> clientNodeIds = msg.clientNodeIds();
+                    Collection<UUID> clientNodeIds = msg.clientNodeIds() == null
```
sergey-chugunov-1985: Is `F.emptyIfNull` applicable here?

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryNodesMetricsMapMessage.java
sergey-chugunov-1985: But at the same time we use it only for collecting and distributing client metrics, and there are no plans to develop discovery-based metrics further.

So I still think giving this class a more specific wouldn't harm existing codebase but would make it more readable instead.

There are five other classes matching "TcpDiscovery*MetricsMessage", and it is already hard to build a mental model of what's going on. Giving a sixth one a vague name makes this situation even worse.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryIoSession.java
```diff
@@ -320,4 +306,147 @@ private void detectSslAlert(byte firstByte, InputStream in) throws IOException {
         if (hex.matches("15....00"))
             throw new StreamCorruptedException("invalid stream header: " + hex);
     }
+
+    /**
+     * Input stream allowing to read some data first as a prefix of an original input stream.
+     * Supports only basic read methods.
+     */
+    private static class PrefixedBufferedInputStream extends BufferedInputStream {
```
sergey-chugunov-1985:
```suggestion
    private static class CompositeInputStream extends BufferedInputStream {
```

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryIoSession.java
```diff
@@ -320,4 +306,147 @@ private void detectSslAlert(byte firstByte, InputStream in) throws IOException {
         if (hex.matches("15....00"))
             throw new StreamCorruptedException("invalid stream header: " + hex);
     }
+
+    /**
+     * Input stream allowing to read some data first as a prefix of an original input stream.
```
sergey-chugunov-1985:
```suggestion
     * Input stream implementation that combines a byte array and a regular InputStream allowing to read
     * bytes from the array first and then proceed with reading from InputStream.
```

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryIoSession.java
```diff
@@ -320,4 +306,147 @@ private void detectSslAlert(byte firstByte, InputStream in) throws IOException {
...
+        /** Prefix data input stream to read before the original input stream. */
+        @Nullable private ByteArrayInputStream prefixIs;
+
+        /** @param srcIs Original input stream to read when {@link #prefixIs} is empty. */
+        private PrefixedBufferedInputStream(InputStream srcIs) {
+            super(srcIs);
+        }
+
+        /** @param prefixData Prefix data to read before the original input stream. */
+        private void acceptPrefixBuffer(byte[] prefixData) {
```
sergey-chugunov-1985:
```suggestion
        private void attachByteArray(byte[] attachment) {
```

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryIoSession.java
```diff
@@ -320,4 +306,147 @@ private void detectSslAlert(byte firstByte, InputStream in) throws IOException {
...
             throw new StreamCorruptedException("invalid stream header: " + hex);
     }
+
+    /**
+     * Input stream allowing to read some data first as a prefix of an original input stream.
+     * Supports only basic read methods.
+     */
+    private static class PrefixedBufferedInputStream extends BufferedInputStream {
+        /** Prefix data input stream to read before the original input stream. */
+        @Nullable private ByteArrayInputStream prefixIs;
```
sergey-chugunov-1985:
```suggestion
        @Nullable private ByteArrayInputStream attachedBytesIs;
```

sergey-chugunov-1985 (approved):

PR #12651, opened 2026-01-22
IGNITE-27626 Use MessageSerializer for custom messages

shishkovilja:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheStatisticsModeChangeMessage.java
```diff
@@ -73,7 +87,8 @@ private CacheStatisticsModeChangeMessage(CacheStatisticsModeChangeMessage req) {
      *
      * @param caches Collection of cache names.
      */
-    public CacheStatisticsModeChangeMessage(UUID reqId, Collection<String> caches, boolean enabled) {
+    public CacheStatisticsModeChangeMessage(IgniteUuid id, UUID reqId, Collection<String> caches, boolean enabled) {
```
wernerdv: We always pass `IgniteUuid.randomUuid()` for the `id` argument in the constructor.

Maybe we should remove id from the constructor arguments and just initialize it inside the constructor?

This also applies to the private constructor.

By the way, we could also remove the `UUID reqId` argument and initialize it as `UUID.randomUUID()` instead.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryCustomEventMessage.java
```diff
@@ -73,47 +89,91 @@ public TcpDiscoveryCustomEventMessage(TcpDiscoveryCustomEventMessage msg) {
...
      */
-    @Nullable public DiscoverySpiCustomMessage message(@NotNull Marshaller marsh, ClassLoader ldr) throws Throwable {
-        if (msg == null) {
+    //TODO: Should be removed in https://issues.apache.org/jira/browse/IGNITE-27627
+    public void prepareMarshal(Marshaller marsh) throws IgniteCheckedException {
+        if (msg instanceof Message)
+            serMsg = (Message)msg;
+        else {
+            assert msgBytes == null || msg.isMutable() : "Message bytes are not null for immutable message: bytes=" +
+                Arrays.toString(msgBytes) + "]";
```
wernerdv: Let's remove `+ “]”`

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryCustomEventMessage.java
```diff
@@ -73,47 +89,91 @@ public TcpDiscoveryCustomEventMessage(TcpDiscoveryCustomEventMessage msg) {
...
+    //TODO: Should be removed in https://issues.apache.org/jira/browse/IGNITE-27627
+    public void finishUnmarhal(Marshaller marsh, ClassLoader ldr) throws IgniteCheckedException {
+        if (msg != null)
+            return;
+
+        if (serMsg != null)
+            msg = (DiscoveryCustomMessage)serMsg;
+        else {
             try {
                 msg = U.unmarshal(marsh, msgBytes, ldr);
```
wernerdv, line 160: Should we add a check that msgBytes != null?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheStatisticsModeChangeMessage.java
shishkovilja: Sounds reasonable.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryCustomEventMessage.java
shishkovilja: In fact, if `msgBytes == null`, it is an incorrect state of message. Unmarshal will fail with an asserion error.

PR #12674, opened 2026-01-29
IGNITE-27699 Make "metastorage" and "distributedMetastorage" system v…

alex-plekhanov:
…iews filtrable

### Notes


modules/core/src/main/java/org/apache/ignite/internal/managers/systemview/walker/SqlQueryHistoryViewWalker.java
```diff
@@ -24,7 +24,7 @@
 /**
  * Generated by {@code org.apache.ignite.codegen.SystemViewRowAttributeWalkerGenerator}.
  * {@link SqlQueryHistoryView} attributes walker.
- *
+ *
```
chesnokoff: Do we really need this change?

modules/core/src/main/java/org/apache/ignite/internal/managers/systemview/walker/MetastorageViewWalker.java
```diff
@@ -27,6 +30,19 @@
  * @see MetastorageView
  */
 public class MetastorageViewWalker implements SystemViewRowAttributeWalker<MetastorageView> {
+    /** Filter key for attribute "name" */
+    public static final String NAME_FILTER = "name";
+
+    /** List of filtrable attributes. */
+    private static final List<String> FILTRABLE_ATTRS = Collections.unmodifiableList(F.asList(
+        "name"
+    ));
```
chesnokoff:
```
public static <T> List<T> asList(@Nullable T t) {
        return t == null ? Collections.<T>emptyList() : Collections.singletonList(t);
    }
```
F.asList already returns singletonList which is immutable

modules/core/src/main/java/org/apache/ignite/internal/processors/metastorage/persistence/DistributedMetaStorageImpl.java
```diff
@@ -324,12 +324,17 @@ public void inMemoryReadyForRead() {
...
+                String name = (String)filter.get(MetastorageViewWalker.NAME_FILTER);
+
+                if (name == null)
+                    name = "";
+
                 try {
                     List<MetastorageView> data = new ArrayList<>();
 
-                    iterate("", (key, val) -> data.add(new MetastorageView(key, val == null || !val.getClass().isArray()
+                    iterate(name, (key, val) -> data.add(new MetastorageView(key, val == null || !val.getClass().isArray()
```
chesnokoff: `iterate` finds all keys with the given prefix. So if we pass something like `rolling.upgrade`, it will also return `rolling.upgrade.version`.

This could be confusing, especially since in `org.apache.ignite.internal.processors.cluster.GridClusterStateProcessor#nodeAttributeViewSupplier` we look up an attribute only by the exact `attrName`

1) Do we have keys where one key is prefix of another key?
2) Will such behavior be clear for user?

modules/core/src/main/java/org/apache/ignite/internal/managers/systemview/walker/SqlQueryHistoryViewWalker.java
alex-plekhanov: It's a generated code, this view was somehow changed after generation and now every time SystemViewRowAttributeWalkerGenerator is called this change is reappears again.

modules/core/src/main/java/org/apache/ignite/internal/processors/metastorage/persistence/DistributedMetaStorageImpl.java
alex-plekhanov: We never state that `filter` is for exact match. So, formally, we are not violating the contract. It can be exect value, for some views, it can be prefix for another. In SQL, where contract required that filtering return exact value - there is post filtration based on condition after lookup by filter. So filtering it's a good way to narrow the scope.
1. I don't know, but I think it's possible.
2. I'm not sure anyone is using iterators direcly, views filtration are made usually using SQL, in SQL this is not an issue. But ok, I think we can fix it and make behaviour the same for different views.

Fixed.

modules/core/src/main/java/org/apache/ignite/internal/managers/systemview/walker/SqlQueryHistoryViewWalker.java
chesnokoff: Nice catch then!

chesnokoff (approved):
LGTM!

PR #12717, opened 2026-02-10
IGNITE-27791 Added support for char arrays for IDTO code generator

petrov-mg:


modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -3625,6 +3625,41 @@ public static void writeLongArray(DataOutput out, @Nullable long[] arr) throws I
...
+            out.writeInt(arr.length);
+
+            for (int c : arr)
+                out.writeChar(c);
+        }
+    }
+
+    /**
+     * @param in Stream to read from.
+     * @return Read byte array, possibly <tt>null</tt>.
```
NSAmelchev:
```suggestion
     * @return Read char array, possibly <tt>null</tt>.
```

petrov-mg: Done.

PR #12726, opened 2026-02-11
IGNITE-27817 Add REGEXP_SUBSTR/REGEXP_EXTRACT function for Calcite en…

vldpyatkov:
…gine

https://issues.apache.org/jira/browse/IGNITE-27817

### DIFF
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteConvertletTable.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteConvertletTable.java
@@ -17,8 +17,8 @@
 
 package org.apache.ignite.internal.processors.query.calcite.prepare;
 
-import com.google.common.collect.ImmutableList;
 import java.math.BigDecimal;
+import com.google.common.collect.ImmutableList;
 import org.apache.calcite.avatica.util.DateTimeUtils;
 import org.apache.calcite.avatica.util.TimeUnit;
 import org.apache.calcite.rel.type.RelDataType;
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/RexImpTable.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/RexImpTable.java
@@ -16,7 +16,6 @@
  */
 package org.apache.ignite.internal.processors.query.calcite.exec.exp;
 
-import com.google.common.collect.ImmutableList;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
@@ -30,6 +29,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
+import com.google.common.collect.ImmutableList;
 import org.apache.calcite.adapter.enumerable.EnumUtils;
 import org.apache.calcite.adapter.enumerable.NullPolicy;
 import org.apache.calcite.adapter.java.JavaTypeFactory;


zstan:
1. update all corresponding documentation, search, for example : "REGEXP_REPLACE" in *.adoc
2. move all newly passed tests from regex_replace.test_ignore into regex_replace.test (probably you can remove *_ignore file after it ? it all passed there ?)
3. Check https://issues.apache.org/jira/browse/IGNITE-15001probably we can close it ? or you can extend it
and overall good practise for such kind of issues is to extend tests coverage (i mean - append adiitional tests) into appropriate **.test** file

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/StdSqlOperatorsTest.java
```diff
@@ -275,6 +275,17 @@ public void testPosixRegex() {
         assertExpression("REGEXP_REPLACE('aAaA', '[Aa]+', 'X', 2)").returns("aX").check();
         assertExpression("REGEXP_REPLACE('aAbaAcaA', '[Aa]+', 'X', 2, 2)").returns("aAbXcaA").check();
         assertExpression("REGEXP_REPLACE('aAbaAcaA', 'a', 'X', 2, 0, 'i')").returns("aXbXXcXX").check();
+        assertExpression("REGEXP_SUBSTR('hello world', 'l.o')").returns("llo").check();
```
alex-plekhanov, line 278: This class is not for testing all cases, see javadoc. Only one case per implementor in this class please. Other cases can be moved to some other classes.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/sql/fun/IgniteStdSqlOperatorTable.java
```diff
@@ -230,6 +230,8 @@ public IgniteStdSqlOperatorTable() {
         register(SqlLibraryOperators.REGEXP_REPLACE_4);
         register(SqlLibraryOperators.REGEXP_REPLACE_5_ORACLE);
         register(SqlLibraryOperators.REGEXP_REPLACE_6);
+        register(SqlLibraryOperators.REGEXP_SUBSTR);
+        register(SqlLibraryOperators.REGEXP_EXTRACT);
```
alex-plekhanov: Let's add only REGEXP_SUBSTR

vldpyatkov: Why not add both?
Because REGEXP_SUBSTR is an alias for REGEXP_EXTRACT in Calcite.

alex-plekhanov: REGEXP_EXTRACT not in standard
REGEXP_EXTRACT not used by major DB vendors except BigQuery
Built-in function is a public API, and add duplicated functionality to public API it's not a good idea, we need to support more functions in the future and can't remove it (we already have made such a mistake with LENGTH, CHAR_LENGTH, CHARACTER_LENGTH or some other functions)

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/StdSqlOperatorsTest.java
vldpyatkov: Done.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteConvertletTable.java
```diff
@@ -42,13 +46,29 @@
...
+    };
+
     /** Instance. */
     public static final IgniteConvertletTable INSTANCE = new IgniteConvertletTable();
 
     /** */
     protected IgniteConvertletTable() {
         // Replace Calcite's convertlet with our own.
         registerOp(SqlStdOperatorTable.TIMESTAMP_DIFF, new TimestampDiffConvertlet());
+        registerOp(SqlLibraryOperators.REGEXP_SUBSTR, StandardConvertletTable.INSTANCE.get(REGEXP_EXTRACT_CALL));
```
alex-plekhanov: Maybe something like:
```
        registerOp(SqlLibraryOperators.REGEXP_SUBSTR, (cx, call) -> cx.getRexBuilder().makeCall(
            SqlLibraryOperators.REGEXP_SUBSTR, Commons.transform(call.getOperandList(), cx::convertExpression)));
```
?

vldpyatkov: Ok

vldpyatkov:
@zstan I added a dedicated file with tests on REGEXP_SUBSTR (test/sql/function/string/test_regexp_substr.test).
I also check all the ignored tests for REGEXP_SUBSTR. Unfortunately, all issues are still there.

PR #12740, opened 2026-02-13
IGNITE-27833 [ducktests] Add Multi-DC Integration Test for Cross-DC Network Partition Resilience

maksaska:

### Notes


modules/ducktests/tests/ignitetest/services/utils/ignite_configuration/communication.py
```diff
@@ -53,7 +53,8 @@ def __init__(self,
                  connections_per_node: int = None,
                  use_paired_connections: bool = None,
                  message_queue_limit: int = None,
-                 unacknowledged_messages_buffer_size: int = None):
+                 unacknowledged_messages_buffer_size: int = None,
+                 connect_timeout: int = 5_000):
```
anton-vinogradov: Defaulting `connect_timeout` to `5_000` (instead of `None`) changes behavior of **all** existing ducktests, not only MDC ones: `communication_macro.j2` renders every non-empty field, so every test now gets an explicit `<property name="connectTimeout" value="5000"/>`, and `TcpCommunicationSpi#setConnectTimeout()` disables the failure-detection-timeout logic for the communication SPI (`failureDetectionTimeoutEnabled(false)` in `TcpCommunicationConfigInitializer`). Today ducktests run comm timeouts derived from `failure_detection_timeout=10000` (the `IgniteConfiguration` default here); with this change they silently switch to the `connTimeout`/`maxConnectTimeout`/`reconCnt` scheme.

Quick fix: keep `connect_timeout: int = None` here — `MdcCluster` already passes it explicitly (`TcpCommunicationSpi(connect_timeout=tcp_connect_timeout)`), so MDC tests keep their value while the rest of the suite stays untouched.

modules/ducktests/tests/ignitetest/services/utils/control_utility.py
```diff
@@ -404,7 +483,19 @@ def __parse_cluster_state(output):
...
+
+        match = coordinator_pattern.search(output)
+
+        if match:
+            order = int(match.group("order"))
+
+            coordinator = next((node for node in baseline if node.order == order), None)
+
+            if coordinator is None:
+                raise AssertionError(f"Coordinator with order={order} is not found in baseline [baseline={baseline}]")
```
anton-vinogradov: This turns a working `cluster_state()` into a crash for legitimate topologies. `--baseline` prints the coordinator as the min-order node among **all** server nodes, including non-baseline ones (`AbstractBaselineCommand`), while "Other nodes:" lines carry no `State=`, so they don't match `baseline_pattern` and never make it into `baseline`. If the coordinator happens to be a non-baseline server (joined, not yet in baseline), every `cluster_state()`/`baseline()` call in any existing test now raises here.

Quick fix: leave `coordinator=None` in that case instead of raising — `verify_split_brain()` already asserts coordinator presence on its own.

modules/ducktests/tests/ignitetest/services/utils/control_utility.py
```diff
@@ -181,6 +181,83 @@ def idle_verify_dump(self, node=None):
...
+                                                   partitions={})
+                groups[cur_group.name] = cur_group
+                continue
+
+            match = row_pattern.match(line)
+            if match and cur_group is not None:
+                attrs = {}
+                if user_attributes and match.group("attr_values") is not None:
+                    values = [v.strip() for v in match.group("attr_values").split(",")]
+                    attrs = dict(zip(user_attributes, values))
```
anton-vinogradov: The task stores user attributes in a `TreeMap` (`CacheDistributionTask#run`), so values are printed in *alphabetical* key order, not in `--user-attributes` argument order. `dict(zip(user_attributes, values))` mis-assigns values whenever 2+ attributes are requested in non-alphabetical order (fine for the single-attribute MDC usage, but the helper is generic).

Quick fix: `dict(zip(sorted(user_attributes), values))`, or parse the names back from the header row, which prints them in output order.

modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/utils/Utils.java
```diff
@@ -0,0 +1,92 @@
...
+        JsonNode field = jNode.path(fieldName);
+
+        if (field.isMissingNode() || field.isNull())
+            return dfltVal;
+
+        try {
+            return Enum.valueOf(dfltVal.getDeclaringClass(), field.asText().toUpperCase(Locale.ROOT));
+        }
+        catch (IllegalArgumentException e) {
+            return dfltVal; // Fallback if string doesn't match any enum constant.
```
anton-vinogradov: Silently falling back on an unrecognized value contradicts the rationale of the required-field overload below ("a typo must fail the run, not pick a default"): e.g. `"atomicity": "TRANSACTONAL"` would silently create an ATOMIC cache and invalidate the scenario the test believes it is running, with no signal anywhere. Suggest defaulting only on missing/null and letting `Enum.valueOf` throw for a present-but-invalid value.

anton-vinogradov (commented):
Reviewed the whole PR. The netem/iptables layering (tcset baseline + per-pair DROP chains, so healing is a pure chain flush) is nicely designed, and the tests read well. A few findings below, most with one-line fixes; the two `control_utility.py` ones and the `connect_timeout` default affect shared infrastructure beyond MDC.

Nits not tied to a diff line:
- `modules/ducktests/tests/tox.ini` `envlist` still contains `py38`, while the workflow dropped it and `tcconfig==0.30.1` requires Python >= 3.9 — a local `tox` run will fail on the py38 env. Worth removing it there too.
- `thin_client_test.py` reserves `@cluster(num_nodes=8)` but allocates 7 (2+2 servers, 1 runner, 2 thin clients).

modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/tests/mdc/MdcContinuousLoadApplication.java
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    /**
+     * Executes a single operation of the given mode against the given key. Throws a
+     * {@link CacheException} or {@link IgniteException} on operation failure (e.g. a write
+     * rejected by the topology validator).
+     *
+     * @return {@code true} if the operation succeeded and (for reads) returned the expected value.
+     */
+    private boolean doOperation(LoadMode mode, int key) {
```
anton-vinogradov: A missed/corrupted read (GET / SQL_SELECT) returns `ok=false` and only bumps `errCnt`: unlike an exception, it neither fails the application when `tolerateErrors=false` nor cuts the load when `stopOnError=true`. So a data-integrity failure — the worst outcome a read load can produce — is reduced to a counter the python side may or may not assert on.

Suggest routing verification failures through the same policy as exceptions (fail when `!tolerateErrors`, cut when `stopOnError`), or at least spelling the distinction out in the class javadoc.

modules/ducktests/tests/ignitetest/tests/mdc/thin_client_test.py
```diff
@@ -0,0 +1,149 @@
...
+
+            svc = mdc.run_service(cli_dc1, {"mode": "GET", "cacheName": CACHE_NAME, "keyFrom": 0,
+                                            "keyTo": KEYS, "iterations": PINNED_GET_ITERS,
+                                            "resultPrefix": "pinnedGet"})
+
+            avg_cli_dc_1 = mdc.result_float(svc, "pinnedGetAvgOpMs")
+
+            self.logger.info(f"Thin client routing latency [delayMs={cross_dc_latency_ms}, getAvgOpMs={avg_cli_dc_1}]")
+
+            assert avg_cli_dc_1 < cross_dc_latency_ms, \
```
anton-vinogradov: Worth also asserting `pinnedGetErrCnt == 0`: missed/corrupted reads only increment ErrCnt in `MdcThinClientLoadApplication`, and a broken read path (all gets returning null) would still satisfy the latency assertion — likely even faster than healthy local reads.

modules/ducktests/tests/ignitetest/services/mdc/mdc_cluster.py
```diff
@@ -0,0 +1,531 @@
...
+        cfg_kwargs = {
+            "version": IgniteVersion(ignite_version),
+            "network_timeout": network_timeout,
+            "communication_spi": TcpCommunicationSpi(connect_timeout=tcp_connect_timeout)
+        }
+
+        if client_connector:
+            cfg_kwargs["client_connector_configuration"] = ClientConnectorConfiguration()
+
+        self.ignite_config = IgniteConfiguration(**cfg_kwargs)
```
anton-vinogradov, line 124: The mechanism that makes two DCs form ONE cluster is completely invisible here and worth an explicit comment (or an explicit shared `discovery_spi=TcpDiscoverySpi()` in `cfg_kwargs`): both `IgniteService`s share this single config object, hence one `TcpDiscoveryVmIpFinder`, and `prepare_on_start()` memoizes the *first started* DC's addresses into it — DC2 then discovers through DC1's nodes, and `restart(DC_2)` re-joins the same way. That is also exactly why `test_main_dc_loss_and_return` must call `sync_service_discovery()` before restarting the seed DC. Making this load-bearing detail explicit would save the next reader a debugging session.

.github/workflows/commit-check.yml
```diff
@@ -147,7 +147,6 @@ jobs:
       fail-fast: false
       matrix:
         cfg:
-          - { python: "3.8", toxenv: "py38" }
           - { python: "3.9", toxenv: "py39" }
```
anton-vinogradov, line 162: `modules/ducktests/tests/tox.ini` still lists `py38` in `envlist` — with `tcconfig==0.30.1` requiring Python >= 3.9, the py38 env can no longer install the deps. Suggest dropping it from tox.ini in the same commit that drops it here.

modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/tests/mdc/MdcDataGeneratorApplication.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.function.IntFunction;
+import com.fasterxml.jackson.databind.JsonNode;
+import org.apache.ignite.IgniteCache;
+import org.apache.ignite.internal.IgniteInterruptedCheckedException;
+import org.apache.ignite.internal.ducktest.tests.dto.IndexedDataRecord;
+
+/**
+ * Populates the MDC cache with a deterministic data set: keys in {@code [from, to)},
+ * values {@code IndexedDataRecord(key)} (or the key itself in {@code sqlMode}).
+ * Run it before the network partition.
```
sergey-chugunov-1985:
```suggestion
 * Runs it until the network partition event.
```

modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/utils/Utils.java
```diff
@@ -0,0 +1,84 @@
...
+            return null;
+        });
+    }
+
+    /** Formats a nanosecond latency as milliseconds with 3 decimal places. */
+    public static String fmtMs(double ns) {
+        return String.format(Locale.US, "%.3f", ns < 0 ? -1.0 : ns / 1e6);
+    }
+
+    /** Parses an optional enum-valued field.  Defaults only on missing/null. */
```
sergey-chugunov-1985:
```suggestion
    /** Parses an optional enum-valued field. Defaults only on missing/null. */
```

modules/ducktests/tests/docker/Dockerfile
```diff
@@ -125,31 +126,31 @@ RUN echo 'PermitUserEnvironment yes' >> /etc/ssh/sshd_config
...
-
-# Install Kafka
-ARG KAFKA_VERSION="3.9.1"
-ARG KAFKA_NAME="kafka"
-ARG KAFKA_RELEASE_NAME="${KAFKA_NAME}_2.13-$KAFKA_VERSION"
-ARG KAFKA_RELEASE_ARTIFACT="$KAFKA_RELEASE_NAME.tgz"
-RUN cd /opt && curl -O $APACHE_ARCHIVE/kafka/$KAFKA_VERSION/$KAFKA_RELEASE_ARTIFACT \
-    && tar xvf $KAFKA_RELEASE_ARTIFACT && rm $KAFKA_RELEASE_ARTIFACT \
-    && mv /opt/$KAFKA_RELEASE_NAME /opt/$KAFKA_NAME-$KAFKA_VERSION
+## Install Ignite
```
anton-vinogradov: **[blocker]** This block (together with the Zookeeper and Kafka ones below) got commented out by the "codestyle" commit (69b6538) - looks like a local image-rebuild speed-up that slipped into the branch. With this image every existing ducktest that runs released versions (`@ignite_versions(str(LATEST))` - baseline, discovery, compatibility suites), the ZooKeeper discovery suites and the CDC-Kafka tests lose their binaries. Needs to be restored before merge.

modules/ducktests/tests/ignitetest/services/utils/control_utility.py
```diff
@@ -481,6 +569,7 @@ class ClusterState(NamedTuple):
     state: str
     topology_version: int
     baseline: list
+    coordinator: BaselineNode = None
```
anton-vinogradov, line 594: **[blocker]** Adding a 4th field to `ClusterState` breaks positional unpacking in existing tests: `control_utility/baseline_test.py:66/128/134` and `index_rebuild_test.py:83` do `_, version, _ = control_utility.cluster_state()`. NamedTuple unpacking always yields **all** fields regardless of defaults, so these raise `ValueError: too many values to unpack (expected 3)` on the first call. Please update those four call sites within this PR (named access is future-proof: `control_utility.cluster_state().topology_version`).

modules/ducktests/tests/ignitetest/services/utils/control_utility.py
```diff
@@ -181,6 +181,83 @@ def idle_verify_dump(self, node=None):
...
+                                                   partitions={})
+                groups[cur_group.name] = cur_group
+                continue
+
+            match = row_pattern.match(line)
+            if match and cur_group is not None:
+                attrs = {}
+                if user_attributes and match.group("attr_values") is not None:
+                    values = [v.strip() for v in match.group("attr_values").split(",")]
+                    attrs = dict(zip(sorted(user_attributes), values))
```
anton-vinogradov: **[major]** My round-1 suggestion here was itself incomplete - sorry for the detour. The `TreeMap` only exists on the server side: `CacheDistributionNode.userAttrs` is an IDTO field serialized via `U.writeStringMap`/`U.readStringMap`, and `readStringMap` materializes a plain `HashMap`. `CacheDistributionTaskResult.print()` runs in the control.sh JVM **after** deserialization, so both the header and the per-row values iterate that `HashMap` - the real order is HashMap iteration order: deterministic, but neither the `--user-attributes` argument order nor alphabetical. With a single attribute (all current usages) any zip works, but for 2+ attributes `sorted()` can still mis-assign values.

The output already carries the truth: the header line (`[groupId,partition,...,nodeAddresses,<attr names>]`) is printed by iterating the same map as the row values, so header order == value order always. Robust fix: capture the trailing attribute names from the header line and zip with those instead of `sorted(user_attributes)`. The comment above `row_pattern` ("in the same order they were passed") needs the same correction, otherwise the next editor will "fix" the code back to the comment.

modules/ducktests/tests/ignitetest/tests/mdc/transactional_partition_test.py
```diff
@@ -0,0 +1,146 @@
...
+            for dc in [DC_1, DC_2]:
+                svc = mdc.stop_loader(dc)
+
+                inserts = mdc.result_int(svc, f"txLoad{dc}OpsCnt")
+                errors = mdc.result_int(svc, f"txLoad{dc}ErrCnt")
+
+                self.logger.info(f"Transactional load cut by the partition "
+                                 f"[txLoad{dc}OpsCnt={inserts}, txLoad{dc}ErrCnt={errors}]")
+
+                assert inserts > 0, "The transactional load performed no successful inserts"
```
anton-vinogradov: **[major]** The scenario's central claim - the load was cut off by the partition - is never asserted: `errors` is only logged, and `inserts > 0` also holds when nothing was cut at all (with `iterations=0` the loader runs until `stop_loader()`, so e.g. a broken/absent topology validator keeps this test green). The application already records the flag:

```suggestion
                assert inserts > 0, "The transactional load performed no successful inserts"

                assert svc.extract_result(f"txLoad{dc}StoppedOnError") == "true", \
                    f"The load was expected to be cut off by the partition [dc={dc}]"
```

modules/ducktests/tests/ignitetest/tests/mdc/partition_resilience_test.py
```diff
@@ -0,0 +1,215 @@
...
+from ignitetest.utils import cluster, ignite_versions
+from ignitetest.utils.ignite_test import IgniteTest
+from ignitetest.utils.version import DEV_BRANCH
+
+CACHE_NAME = "mdc-resilience"
+
+BACKUPS = 1
+
+# Time for discovery to detect the partition and for both half-rings to complete PME.
+SPLIT_SETTLE_SECS = 10
```
anton-vinogradov: **[minor, flakiness]** `SPLIT_SETTLE_SECS = 10` equals the default `failure_detection_timeout=10_000` exactly. With iptables DROP the surviving ring needs the *full* FDT just to detect the loss, then process the node failures and finish PME - so `verify_split_brain()` fires right at the theoretical minimum, and on a slow CI runner this is a coin flip (false-red). Consider waiting deterministically (e.g. `await_event` for the node-failed message on the surviving DC before verifying) or at least a comfortable margin above FDT.

modules/ducktests/tests/ignitetest/tests/mdc/thin_client_test.py
```diff
@@ -0,0 +1,154 @@
...
+                                            "keyTo": KEYS, "iterations": PINNED_GET_ITERS,
+                                            "resultPrefix": "pinnedGet"})
+
+            avg_cli_dc_1 = mdc.result_float(svc, "pinnedGetAvgOpMs")
+            err_cnt_cli_dc_1 = mdc.result_int(svc, "pinnedGetErrCnt")
+
+            self.logger.info(f"Thin client routing latency [delayMs={cross_dc_latency_ms}, getAvgOpMs={avg_cli_dc_1}, "
+                             f"getErrCnt={err_cnt_cli_dc_1}]")
+
+            assert avg_cli_dc_1 < cross_dc_latency_ms, \
```
anton-vinogradov: **[minor]** This proves DC-locality only for the DC1-pinned client, and DC1 happens to be the first/main DC - a routing regression of the form "always prefer the first configured DC" would still pass. `cli_dc2` is already registered: a symmetric `pinnedGet` burst from it before the partition, with the same `< cross_dc_latency_ms` assert, would pin down actual DC-awareness at the cost of a few seconds.

modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/tests/mdc/MdcContinuousLoadApplication.java
```diff
@@ -0,0 +1,295 @@
...
+        while (!terminated() && (iterations == 0 || opsCnt + errCnt < iterations)) {
+            boolean ok = true;
+
+            try {
+                doOperation(mode, key);
+            }
+            catch (CacheException | IgniteException e) {
+                ok = false;
+
+                if (inadmissible)
```
anton-vinogradov: **[minor]** The rename dropped the `mode.isWrite()` guard the old `mode.isWrite() && !expectAdmissible` had (and the key-advance condition below still keeps): a read-mode run with a stray `inadmissible=true` now swallows every failure - including the missed/corrupted-read `IgniteException` - logging it as "Write rejected as expected", and finishes green, because the final admissibility check is write-only. Restoring the guard keeps the javadoc's "write modes only" honest:

```suggestion
                if (mode.isWrite() && inadmissible)
```

modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/tests/mdc/MdcThinClientLoadApplication.java
```diff
@@ -0,0 +1,143 @@
...
+
+                    if (val == null || !val.equals(new IndexedDataRecord(key)))
+                        throw new IgniteException("Read entry is missed or corrupted [key=" + key +
+                            ", val=" + val + "]");
+                }
+            }
+            catch (ClientException | CacheException | IgniteException e) {
+                ok = false;
+
+                if (inadmissible)
```
anton-vinogradov: **[minor]** Same as in `MdcContinuousLoadApplication`: without the write guard a GET run with a stray `inadmissible=true` swallows corrupted reads as "Put rejected as expected".

```suggestion
                if (put && inadmissible)
```

modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/tests/mdc/MdcCacheAwareApplication.java
```diff
@@ -0,0 +1,170 @@
...
+
+        if (jNode.hasNonNull("datacenters")) {
+            Set<String> dcs = new HashSet<>();
+
+            jNode.get("datacenters").forEach(dc -> dcs.add(dc.asText()));
+
+            topValidator.setDatacenters(dcs);
+        }
+        else
+            topValidator.setMainDatacenter(jNode.path("mainDc").asText());
```
anton-vinogradov: **[minor]** `path().asText()` turns a missing `mainDc` into `""`, which passes `MdcTopologyValidator.checkConfiguration()` (it only guards `null`) and then can never match any node's `dataCenterId` in `validate()`: a test that forgets the parameter gets a cache that silently rejects every write instead of a configuration error. Fail fast:

```suggestion
        else if (jNode.hasNonNull("mainDc"))
            topValidator.setMainDatacenter(jNode.get("mainDc").asText());
        else
            throw new IllegalArgumentException("Either 'datacenters' or 'mainDc' must be specified.");
```

modules/ducktests/tests/ignitetest/services/network_group/manager.py
```diff
@@ -0,0 +1,420 @@
...
+            remote_ips = self._resolve_group_ips(dst_group)
+
+            cmd = to_partition_enable_cmd(chain, remote_ips)
+
+            for node in self._iter_group_nodes(src_group):
+                tasks.append((node, cmd))
+
+        self._ssh_parallel(tasks, tag="PARTITION_ON")
+
+        self._log_network(f"PARTITION {group_a} <-> {group_b}")
```
anton-vinogradov, line 147: **[minor]** `_log_network` runs two sequential SSH round-trips (`tc qdisc show` / `tc filter show`) per node *inside* the partition window - after the DROP rules are installed but before the caller's `sleep(BLIP_SECS)` starts. With 8+ nodes that inflates every "1-second" blip by several seconds, uncomfortably close to the 10s failure detection timeout the blips scenario relies on staying under. The tc state doesn't change when toggling iptables, so the dump adds no information here - consider dropping it from `enable_network_partition`/`disable_network_partition` (or making it opt-in) so the blip lasts what the test asked for.

modules/ducktests/tests/ignitetest/services/mdc/mdc_cluster.py
```diff
@@ -0,0 +1,542 @@
...
+
+        assert len(cluster_state.baseline) == exp_total, \
+            f"Cluster baseline should cover both DCs [exp={exp_total}, actual={cluster_state.baseline}]"
+
+    def verify_servers_log_clean(self):
+        """
+        Verifies the negative invariants on all server nodes: no long running transactions
+        were detected, no PME hang and no lost partitions were reported.
+        """
+        for pattern in (LRT_PATTERN, PME_FREEZE_PATTERN, LOST_PARTITIONS_PATTERN):
```
anton-vinogradov: **[minor]** After a `restart()` the previous incarnation's log is rotated to `ignite.log.N` (`IgniteAwareService.__update_node_log_file`), and the default `node.log_file` glob `ignite*.log` does not match it. So the final `verify_servers_log_clean()` in the transactional test never sees what DC2 logged between the pre-restart check and the restart - exactly the partition-heal PME window. An explicit glob covers both: `svc.check_event_absent(pattern, log_file="ignite.log*")`.

modules/ducktests/src/main/java/org/apache/ignite/internal/ducktest/tests/mdc/MdcDataGeneratorApplication.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.function.IntFunction;
+import com.fasterxml.jackson.databind.JsonNode;
+import org.apache.ignite.IgniteCache;
+import org.apache.ignite.internal.IgniteInterruptedCheckedException;
+import org.apache.ignite.internal.ducktest.tests.dto.IndexedDataRecord;
+
+/**
+ * Populates the MDC cache with a deterministic data set: keys in {@code [from, to)},
+ * values {@code IndexedDataRecord(key)} (or the key itself in {@code sqlMode}).
+ * Runs it until the network partition event.
```
anton-vinogradov: **[nit]** The new wording promises more than the code does: the generator writes `[from, to)` and finishes, and `run_app()` waits for that completion - in every scenario it is done before the partition is even enabled; nothing here runs "until the network partition event".

```suggestion
 * Run it to completion before enabling the network partition.
```

anton-vinogradov (commented):
Round 2: verified all previous threads against the new head - every one of them is properly fixed (including the `tcconfig==0.29.1` route: it requires only Python >=3.7, so the py38 env is consistent again; also compiled with `-Pducktests` and the strict `-Pcheckstyle` profile locally - clean).

The final pass still found a few things, two of them merge-blocking:

* **blocker**: the "codestyle" commit accidentally commented out the Ignite/Zookeeper/Kafka installs in the ducktests Docker image;
* **blocker**: the new `ClusterState.coordinator` field breaks positional unpacking of `cluster_state()` in existing tests;
* one correction to my own round-1 comment about `--user-attributes` ordering (it is client-side `HashMap` order, not `TreeMap` order - details inline, sorry for the detour);
* plus a handful of test-strength/robustness notes and two doc nits.

Details inline.

modules/ducktests/tests/ignitetest/tests/mdc/transactional_partition_test.py
```diff
@@ -0,0 +1,146 @@
...
+class MdcTransactionalPartitionTest(IgniteTest):
+    """
+    Transactional load resilience to a cross-DC network partition.
+    """
+    @cluster(num_nodes=8)
+    @ignite_versions(str(DEV_BRANCH))
+    @parametrize(cross_dc_latency_ms=100)
+    def test_transactional_load_cut_on_partition(self, ignite_version, cross_dc_latency_ms):
+        """
+        Continuous implicit-transaction insert load from the main DC is cut off by the first
```
anton-vinogradov: **[nit]** "implicit-transaction" (here, in the module docstring and in the comments below): `TX_PUT` actually runs *explicit* `txStart(PESSIMISTIC, REPEATABLE_READ)` transactions (`MdcContinuousLoadApplication`). Either the wording should say "explicit pessimistic transactions", or - if implicit txs (a plain `cache.put` on a transactional cache) were the intent - the load mode should change; the two abort differently mid-prepare, so for a recovery scenario the distinction isn't cosmetic.

modules/ducktests/tests/ignitetest/tests/mdc/transactional_partition_test.py
```diff
@@ -0,0 +1,147 @@
...
+        cache exception the cross-DC partition triggers; afterwards no transaction is left
+        hanging on either half-ring and the server logs are clean.
+        """
+        mdc = MdcCluster(self, ignite_version, srv_per_dc=2, runners_per_dc=1, loaders_per_dc={DC_2: 1},
+                         network_timeout=20_000, tcp_connect_timeout=10_000)
+
+        with cross_dc_network(self.logger, mdc, delay_ms=cross_dc_latency_ms) as net:
+            mdc.start_servers()
+
+            # Continuous single-threaded explicit-transaction insert load from the backup DC.
```
sergey-chugunov-1985, line 81: This comment contradicts a high-level comment on the test. The high-level comment says that the load runs in the main DC while from what I see in the code the actual runner is started in the backup DC.

modules/ducktests/tests/ignitetest/tests/mdc/transactional_partition_test.py
```diff
@@ -0,0 +1,147 @@
...
+copies and fails with a cache exception. The load cuts itself off on that first
+exception and records how many inserts had succeeded.
+
+After the split settles the test asserts that the cluster really split-brained,
+that the load stopped because of the partition (not because it ran out of work),
+and - the point of the scenario - that the aborted explicit transactions left
+nothing hanging on either half-ring and no suspicious entries in the server logs.
+
+Data accessibility during the split is deliberately NOT checked: a transactional
+cache needs all partition copies available, which a split-brained half-ring cannot
```
sergey-chugunov-1985, line 33: I suspect this statement isn't true: a transactional cache can handle data modification requests if all nodes which backup copies are assigned to are declared FAILED by the discovery component.

So I think we could add a data accessibility check and even a check that data is possible to modify in this scenario as well.

sergey-chugunov-1985 (approved):

sergey-chugunov-1985 (approved):

PR #12755, opened 2026-02-17
IGNITE-27889 Test TcpDiscoveryCoordinatorFailureTest#testClusterFailedNewCoordinatorInitialized hangs and locks suite SPI(Discovery) until timeout

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-27889


zstan:
i`m ok with this change if green visa will be obtained and several sequential test run will pass.

PR #12760, opened 2026-02-17
IGNITE-27871 Improve deployment lookup to reduce deploy() contention …

oleg-vlsk:
…for locally available tasks with peerClassLoadingEnabled=true

### Notes

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
@@ -22,7 +22,6 @@ import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Set;
 import java.util.UUID;
 import org.apache.ignite.IgniteCheckedException;
@@ -249,9 +248,9 @@ class GridDeploymentLocalStore extends GridDeploymentStoreAdapter {
             if (meta.classLoader() != null)
                 dep = deps.get(meta.classLoader());
 
-            if (dep == null && meta.classLoaderId() != null) {
+            if ((dep == null || dep.undeployed()) && meta.classLoaderId() != null) {
                 for (GridDeployment d : deps.values()) {
-                    if (d.classLoaderId().equals(meta.classLoaderId())) {
+                    if (d.classLoaderId().equals(meta.classLoaderId()) && !d.undeployed()) {
                         dep = d;
 
                         break;
@@ -535,37 +534,29 @@ class GridDeploymentLocalStore extends GridDeploymentStoreAdapter {
      * @param ldr Class loader to undeploy.
      */
     private void undeploy(ClassLoader ldr) {
-        Collection<GridDeployment> doomed = U.newIdentityHashSet();
+        GridDeployment dep;
 
         synchronized (mux) {
-            for (Iterator<Map<ClassLoader, GridDeployment>> i1 = depsByAlias.values().iterator(); i1.hasNext();) {
-                Map<ClassLoader, GridDeployment> deps = i1.next();
+            dep = depByLdr.remove(ldr);
 
-                for (Iterator<Entry<ClassLoader, GridDeployment>> i2 = deps.entrySet().iterator(); i2.hasNext();) {
-                    Entry<ClassLoader, GridDeployment> entry = i2.next();
+            if (dep != null) {
+                dep.undeploy();
 
-                    GridDeployment dep = entry.getValue();
+                if (log.isInfoEnabled())
+                    log.info("Removed undeployed class: " + dep);
 
-                    if (dep.classLoader() == ldr) {
-                        dep.undeploy();
+                for (Iterator<Map<ClassLoader, GridDeployment>> it = depsByAlias.values().iterator(); it.hasNext();) {
+                    Map<ClassLoader, GridDeployment> deps = it.next();
 
-                        i2.remove();
+                    deps.remove(ldr);
 
-                        doomed.add(dep);
-
-                        if (log.isInfoEnabled())
-                            log.info("Removed undeployed class: " + dep);
-                    }
+                    if (deps.isEmpty())
+                        it.remove();
                 }
-
-                if (deps.isEmpty())
-                    i1.remove();
             }
-
-            depByLdr.remove(ldr);
         }
 
-        for (GridDeployment dep : doomed) {
+        if (dep != null) {
             if (dep.obsolete()) {
                 // Resource cleanup.
                 ctx.resource().onUndeployed(dep);


modules/core/src/test/java/org/apache/ignite/testsuites/IgniteP2PSelfTestSuite.java
```diff
@@ -82,7 +83,8 @@
     GridDifferentLocalDeploymentSelfTest.class,
     P2PUnsupportedClassVersionTest.class,
     P2PClassLoadingFailureHandlingTest.class,
-    P2PClassLoadingIssuesTest.class
+    P2PClassLoadingIssuesTest.class,
+    GridDeploymentLocalStoreReuseTest.class
```
alex-plekhanov: Add comma to the end of line please (to reduce conflicts on merge)

modules/core/src/test/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStoreReuseTest.java
```diff
@@ -0,0 +1,177 @@
...
+    /** */
+    private static void executeTasksOnClient(IgniteClient client, int cnt, long timeout) {
+        for (int i = 0; i < cnt; i++) {
+            CompletableFuture<T2<UUID, Set<UUID>>> fut = client.compute()
+                .withTimeout(timeout).
+                    <T2<UUID, Set<UUID>>, T2<UUID, Set<UUID>>>executeAsync2(TestTask.class.getName(), null)
+                .toCompletableFuture();
+
+            try {
+                fut.get();
```
alex-plekhanov:
```
                client.compute().execute(TestTask.class.getName(), null);
```

modules/core/src/test/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStoreReuseTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+            for (int i = 0; i < CLIENT_CNT; i++)
+                clients.add(startClient(allServerNodes));
+
+            List<IgniteInternalFuture<Void>> futs = new ArrayList<>(CLIENT_CNT);
+
+            for (IgniteClient client : clients)
+                futs.add(runAsync(() -> executeTasksOnClient(client, EXEC_CNT, 5_000L)));
+
+            waitForAllFutures(futs.toArray(new IgniteInternalFuture[0]));
```
alex-plekhanov:
```
            runMultiThreaded(i -> executeTasksOnClient(clients.get(i), EXEC_CNT), CLIENT_CNT, "worker");
```

modules/core/src/test/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStoreReuseTest.java
```diff
@@ -0,0 +1,177 @@
...
+     * Verifies that multiple task executions do not cause excessive local deployment cache misses. The "deployment not
+     * found ... clsLdrId=null" message is allowed only once per thin client (initial task execution).
+     */
+    @Test
+    public void testNoExcessiveLocalDeployment() {
+        try {
+            ClusterNode[] allServerNodes = grid(0).cluster().forServers().nodes().toArray(new ClusterNode[0]);
+
+            for (int i = 0; i < CLIENT_CNT; i++)
+                clients.add(startClient(allServerNodes));
```
alex-plekhanov: You can connect to any server node,it's not necessary to provide all nodes, one is enough, i.e. `clients.add(startClient(0));`

modules/core/src/test/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStoreReuseTest.java
```diff
@@ -0,0 +1,177 @@
...
+        @Override public ListeningTestLogger getLogger(Object ctgr) {
+            return this;
+        }
+
+        /** */
+        public List<String> depNotFound() {
+            return depNotFound.stream().collect(Collectors.toUnmodifiableList());
+        }
+    }
+}
```
alex-plekhanov: It's incorrect usage of listening logger, all you need is register listener like:
```
            LogListener lsnr = LogListener.matches(notFoundMsg).times(CLIENT_CNT).build();

            listeningTestLog.registerListener(lsnr);
```
listeningTestLog should be created on top of standard logger, for example:
```
        setLoggerDebugLevel();

        listeningTestLog = new ListeningTestLogger(log);
```
And passed to ignite configuration. No need for logger for each node.

modules/core/src/test/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStoreReuseTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+            for (DeploymentListeningLogger log : logs)
+                allNotFound.addAll(log.depNotFound());
+
+            String taskClsName = TestTask.class.getName();
+
+            String notFoundMsg = String.format(
+                "Deployment was not found for class with specific class loader [alias=%s, clsLdrId=null]", taskClsName);
+
+            assertEquals(CLIENT_CNT, Collections.frequency(allNotFound, notFoundMsg));
```
alex-plekhanov: It's not a correct check. You get CLIENT_CNT occurances here just becouse clients start tasks concurrently and you enter the same section by different threads before class was deployed to cluster. In case of any delay test become flaky. For example, if you add something like 
```
            clients.get(0).compute().execute(TestTask.class.getName(), null);
            doSleep(1000);
```
Before massive client tasks start, you will get just one occurance of message independent of clients count.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentManager.java
```diff
@@ -391,13 +391,16 @@ private GridDeployment checkDeployment(GridDeployment deployment, String store)
...
+        ClassLoader ldr = Thread.currentThread().getContextClassLoader();
+
         GridDeploymentMetadata meta = new GridDeploymentMetadata();
 
         meta.record(true);
         meta.deploymentMode(ctx.config().getDeploymentMode());
         meta.alias(rsrcName);
         meta.className(clsName);
         meta.senderNodeId(ctx.localNodeId());
+        meta.classLoader(ldr);
```
alex-plekhanov: Setting classloader disables deployment SPI as far as I understand. See https://github.com/apache/ignite/blob/master/modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java#L174

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
```diff
@@ -63,6 +63,9 @@ class GridDeploymentLocalStore extends GridDeploymentStoreAdapter {
     /** Deployment cache by class name. */
     private final ConcurrentMap<String, Deque<GridDeployment>> cache = new ConcurrentHashMap<>();
 
+    /** Deployment cache by classloader. */
+    private final ConcurrentMap<ClassLoader, Deque<GridDeployment>> cacheByLdr = new ConcurrentHashMap<>();
```
alex-plekhanov: cacheByLdr always used under the lock `mux`, no ConcurrentMap overhead required here.
Also maybe it worth to use IdentityHashMap in case someone redefine classloader's `equals()` in a wrong way.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
```diff
@@ -296,23 +299,49 @@ private GridDeployment deploy(
...
-                            fireEvt = d.addDeployedClass(cls, alias);
+                Deque<GridDeployment> depsByLdr = cacheByLdr.get(ldr);
 
-                            cachedDeps = deps;
+                if (depsByLdr != null) {
+                    GridDeployment candidate = null;
 
-                            dep = d;
+                    for (GridDeployment d : depsByLdr) {
+                        if (!d.undeployed() && d.classLoader() == ldr) {
```
alex-plekhanov: If it's undeployed, it's cleaned from cache, how we can find it?
Why do we need to check classloader if we put in cache only items with exactly this classloader?

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
```diff
@@ -296,23 +299,49 @@ private GridDeployment deploy(
...
-                        break;
+                    if (candidate != null) {
+                        fireEvt = candidate.addDeployedClass(cls, alias);
+
+                        cachedDeps = depsByLdr;
+
+                        dep = candidate;
+                    }
+                }
+                else {
```
alex-plekhanov: Do we still need this check? If deployment not found by classloader in classloader cache it can't be found in aliases cache. We preserve both caches synchronized and modify it only under the lock.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
```diff
@@ -296,23 +299,49 @@ private GridDeployment deploy(
             try {
                 Deque<GridDeployment> cachedDeps = null;
 
-                // Find existing class loader info.
-                for (Deque<GridDeployment> deps : cache.values()) {
-                    for (GridDeployment d : deps) {
-                        if (d.classLoader() == ldr) {
-                            // Cache class and alias.
-                            fireEvt = d.addDeployedClass(cls, alias);
+                Deque<GridDeployment> depsByLdr = cacheByLdr.get(ldr);
```
alex-plekhanov: Looks like it's one-to-one relation for deployment and classloader. Did I miss something?

modules/core/src/test/java/org/apache/ignite/testsuites/IgniteP2PSelfTestSuite.java
oleg-vlsk: Done.

modules/core/src/test/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStoreReuseTest.java
oleg-vlsk: Used this snippet, thank you.

modules/core/src/test/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStoreReuseTest.java
oleg-vlsk: I decided not to go for multi-threaded execution as the perpose of the test is to verify certain behaviour during subsequent executions of the same task. So I ended up using a simple for-loop.

modules/core/src/test/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStoreReuseTest.java
oleg-vlsk: Done.

modules/core/src/test/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStoreReuseTest.java
oleg-vlsk: Done.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentManager.java
oleg-vlsk: Moved the local app classloader check to `GridDeploymentLocalStore#deployment` so that in the initial call the meta does not contains classloader.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
oleg-vlsk: Done, thank you for the hint.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
oleg-vlsk: Yes, those were 'extra safety' checks. Changed the lookup logic altogether (see below).

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
oleg-vlsk: Removed this else block.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
oleg-vlsk: Not necessarily. In `GridDeploymentLocalStore#cache` we can have several deployments with the same classloader associated with one alias/class name (see attached screenshots). Most recent deployment are added to the beginning of the queue (the `addFirst()` call in `GridDeploymentLocalStore#deploy`).

local deployments cache - 1

local deployments cache - 2

modules/core/src/test/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStoreReuseTest.java
oleg-vlsk: Changed the test logic.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
```diff
@@ -263,6 +267,20 @@ class GridDeploymentLocalStore extends GridDeploymentStoreAdapter {
 
                     return dep;
                 }
+                else {
+                    ClassLoader ldr = Thread.currentThread().getContextClassLoader();
+
+                    if (ldr == null)
+                        ldr = U.resolveClassLoader(ctx.config());
```
alex-plekhanov: 1. Let's move ldr initialization outside the loop.
2. Just add `|| dep.classLoader() == ldr` to the `if` condition

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
alex-plekhanov: I'm talking about `cacheByLdr`, not `cache`. For `cacheByLdr` it looks like only one deployment is possible for one classloader.

modules/core/src/test/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStoreReuseTest.java
```diff
@@ -0,0 +1,114 @@
...
+        testLog.registerListener(lsnr0);
+        testLog.registerListener(lsnr1);
+
+        try (IgniteClient client = startClient(0)) {
+            for (int i = 0; i < EXEC_CNT; i++)
+                client.compute().execute(TestTask.class.getName(), null);
+        }
+
+        assertTrue(lsnr0.check(5_000));
+        assertTrue(lsnr1.check(5_000));
```
alex-plekhanov: Why do we need to wait here? As far as I understand here strict happens-before between task completion and log message.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
oleg-vlsk: Moved classloader initialization outside the for loop, but kept `dep.classLoader() == ldr` as a separate if-block to be able to log correct trace messages. As I understand, if we add the `|| dep.classLoader() == ldr` condition to the initial if-block, the trace message will contain either null as classloader id or incorrect classloader id. As well as it won’t be obvious from such message that the local app classloader was used in this case.

modules/core/src/test/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStoreReuseTest.java
oleg-vlsk: Removed timeout.

oleg-vlsk:
@alex-plekhanov Regarding you comment:

"I'm talking about cacheByLdr, not cache. For cacheByLdr it looks like only one deployment is possible for one classloader."

I added:

```
 private final Map<ClassLoader, GridDeployment> depsByLdr = new IdentityHashMap<>();
```
to map classloader to a specific deployment. But I laso kept:

```
private final Map<ClassLoader, Deque<GridDeployment>> cacheByLdr = new IdentityHashMap<>();
```

to have quick access to all deployments with the given classloader so we can add them all to `cache` in order to preserve the current contract:

```
if (cachedDeps != null) {
    assert dep != null;

    cache.put(alias, cachedDeps);

    if (!cls.getName().equals(alias)) 
        cache.put(cls.getName(), cachedDeps);
    
    ...
}
```

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
```diff
@@ -248,22 +235,55 @@ class GridDeploymentLocalStore extends GridDeploymentStoreAdapter {
...
+                for (GridDeployment d : deps.values()) {
+                    if (d.classLoaderId().equals(meta.classLoaderId())) {
+                        dep = d;
+
+                        break;
+                    }
                 }
             }
+
+            if (dep != null && !dep.undeployed()) {
```
alex-plekhanov, line 261: `dep.undeployed()` should be checked inside the loop, or we stop searching after first undeployed deployment

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
```diff
@@ -248,22 +235,55 @@ class GridDeploymentLocalStore extends GridDeploymentStoreAdapter {
...
-                // local or remote deployment.
-                if (dep.classLoaderId() == meta.classLoaderId() || dep.classLoader() == meta.classLoader()) {
-                    if (log.isTraceEnabled())
-                        log.trace("Deployment was found for class with specific class loader [alias=" + meta.alias() +
-                            ", clsLdrId=" + meta.classLoaderId() + "]");
+            if (meta.classLoader() != null)
+                dep = deps.get(meta.classLoader());
 
-                    return dep;
+            if (dep == null && meta.classLoaderId() != null) {
```
alex-plekhanov: (dep == null || dep.undeployed) && ...

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
```diff
@@ -543,14 +535,16 @@ private String getAlias(GridDeployment dep, Class<?> cls) {
      * @param ldr Class loader to undeploy.
      */
     private void undeploy(ClassLoader ldr) {
-        Collection<GridDeployment> doomed = new HashSet<>();
+        Collection<GridDeployment> doomed = U.newIdentityHashSet();
```
alex-plekhanov: Looks like there can be only one doomed instance? Or I miss something?

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
```diff
@@ -543,14 +535,16 @@ private String getAlias(GridDeployment dep, Class<?> cls) {
...
 
         synchronized (mux) {
-            for (Iterator<Deque<GridDeployment>> i1 = cache.values().iterator(); i1.hasNext();) {
-                Deque<GridDeployment> deps = i1.next();
+            for (Iterator<Map<ClassLoader, GridDeployment>> i1 = depsByAlias.values().iterator(); i1.hasNext();) {
```
alex-plekhanov: Maybe reverse delete from caches order? This will also allow to call undeploy only once per instance. Like this:
```
            GridDeployment dep = depByLdr.remove(ldr);
            
            if (dep != null) {
                dep.undeploy();
                
                doomed.add(dep);

                if (log.isInfoEnabled())
                    log.info("Removed undeployed class: " + dep);

                for (Iterator<Map<ClassLoader, GridDeployment>> i1 = depsByAlias.values().iterator(); i1.hasNext(); ) {
                    Map<ClassLoader, GridDeployment> deps = i1.next();
                    
                    deps.remove(ldr);

                    if (deps.isEmpty())
                        i1.remove();
                }
            }
```

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
oleg-vlsk: Added the `dep.undeployed` check to the loop.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
oleg-vlsk: Done.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
oleg-vlsk: Refactored the `undeploy()` method based on the 1:1 classloder to deployment relation.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
oleg-vlsk: Done as part of refactoring.

PR #12825, opened 2026-02-26
IGNITE-26060 FailureHandler isn't called on TX coordinator node in certain scenarios

zstan:
https://issues.apache.org/jira/browse/IGNITE-26060

### DIFF
--- a/modules/calcite/src/test/java/org/apache/ignite/internal/processors/tx/TxWithExceptionalInterceptorTest.java
+++ b/modules/calcite/src/test/java/org/apache/ignite/internal/processors/tx/TxWithExceptionalInterceptorTest.java
@@ -82,10 +82,10 @@ public class TxWithExceptionalInterceptorTest extends GridCommonAbstractTest {
     private static final AtomicInteger exceptionRaised = new AtomicInteger();
 
     /** Cache with or without writeThrough configured. */
-    private static final String processedCacheName = "processedCache";
+    private static final String PROC_CACHE_NAME = "PROC_CACHE";
 
     /** Tx involved cache. */
-    private static final String commonCacheName = "commonCache";
+    private static final String COMMON_CACHE_NAME = "COMMON_CACHE";
 
     /** Default client name. */
     private static final String CLIENT_NAME = "client";
@@ -155,12 +155,12 @@ public class TxWithExceptionalInterceptorTest extends GridCommonAbstractTest {
                     ));
         }
 
-        CacheConfiguration<Integer, Integer> ccfg1 = new CacheConfiguration<>(processedCacheName);
+        CacheConfiguration<Integer, Integer> ccfg1 = new CacheConfiguration<>(PROC_CACHE_NAME);
         ccfg1.setAtomicityMode(TRANSACTIONAL);
         ccfg1.setCacheMode(REPLICATED);
 
         ccfg1.setInterceptor(new CustomInterceptor(writeThrough, txCoord, getTestIgniteInstanceName(0)));
-        ccfg1.setQueryEntities(List.of(queryEntity(processedCacheName)));
+        ccfg1.setQueryEntities(List.of(queryEntity(PROC_CACHE_NAME)));
 
         // All server nodes besides zero-indexed
         if (txCoord == TxCoordNodeRole.COORDINATOR_NO_DATA)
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxStateImpl.java
```diff
@@ -293,6 +297,16 @@ public class IgniteTxStateImpl extends IgniteTxLocalStateAdapter {
 
     /** {@inheritDoc} */
     @Override public boolean storeWriteThrough(GridCacheSharedContext sctx) {
+        if (storeWriteThrough != null)
```
sergey-chugunov-1985, line 300: Do we actually need to optimize this part? Method's call tree makes me thinking that the method is called once on each node during execution of a specific part of tx's handling procedure.

If this is true we don't need this additional logic.

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -194,6 +192,8 @@ public void testSystemExceptionAfterCacheStoreCommit() throws Exception {
                         .filter(ignite -> ((IgniteEx)ignite).context().clientNode())
                         .count() == 1);
             }
+            else
+                waitForTopology(2);
```
sergey-chugunov-1985, line 192: Can we add some additional checks here to examine dead and alive nodes? Like it is done in `then` branch with check for client node.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxStateImpl.java
zstan: i made simple thing:
just write naive log here : System.err.println("!! storeWriteThrough " + hashCode());
run TxWithExceptionalInterceptorTest#testTxWithExceptionInterceptor
and found numerous of occurence with equal hash code, thus i suppose it helpful

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
zstan: done, thanks !

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -201,6 +200,23 @@ public void testSystemExceptionAfterCacheStoreCommit() throws Exception {
...
+    private void checkClientIsAvailable() {
+        assertTrue("Client node should survive test scenario",
+            G.allGrids()
+                .stream()
+                .filter(ignite -> ((IgniteEx)ignite).context().clientNode())
+                .count() == 1);
+    }
+
+    /** Check that faulty node is absent in current topology. */
+    private void checkFultyNode(int faultyNodeIdx) {
```
sergey-chugunov-1985:
```suggestion
    private void checkFаultyNodeLeftTopology(int faultyNodeIdx) {
```

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
```diff
@@ -201,6 +200,23 @@ public void testSystemExceptionAfterCacheStoreCommit() throws Exception {
...
+        assertTrue("Faulty node should survive test scenario, idx=" + faultyNodeIdx,
```
sergey-chugunov-1985:
```suggestion
        assertTrue("Faulty node should not survive test scenario, idx=" + faultyNodeIdx,
```

zstan: done

modules/core/src/test/java/org/apache/ignite/cache/store/CacheStoreWithIgniteTxFailureTest.java
zstan: done

PR #13002, opened 2026-04-09
IGNITE-28475 Document CacheInterceptor interaction with keepBinary cache wrapper

zstan:


docs/_docs/code-snippets/java/src/main/java/org/apache/ignite/snippets/WorkingWithBinaryObjects.java
```diff
@@ -142,6 +142,45 @@ public static void configuringBinaryObjects() {
...
+
+    private static class CustomInterceptor implements CacheInterceptor<Object, Object> {
+        boolean keepBinary;
+
+        CustomInterceptor(boolean keepBinary) {
+            this.keepBinary = keepBinary;
+        }
+
+        @Nullable @Override public Object onBeforePut(Cache.Entry<Object, Object> entry, Object newVal) {
+            assertEquals(keepBinary, newVal instanceof BinaryObject);
```
alex-plekhanov: assertEquals is a part of JUnit framework, шт "examples" let's use `assert` instead

docs/_docs/key-value-api/binary-objects.adoc
```diff
@@ -177,6 +177,21 @@ tab:C#/.NET[unsupported]
...
+--
+tab:Java[]
+[source,java]
+----
+include::{javaCodeDir}/WorkingWithBinaryObjects.java[tag=keepBinaryWithCacheInterceptor,indent=0]
+----
+
+tab:C#/.NET[unsupported]
+
+tab:C++[unsupported]
```
alex-plekhanov: I think this section can be included to `Enabling Binary Mode for Caches`, since it refer to `keepBinary`. But there are couple of sections between these sections now.

modules/core/src/main/java/org/apache/ignite/cache/CacheInterceptor.java
```diff
@@ -33,6 +33,9 @@
  * Cache interceptor is configured via {@link CacheConfiguration#getInterceptor()}
  * configuration property.
  * <p>
+ * The type of incoming values depends on cache operation types, that is, the {@link IgniteCache#withKeepBinary()}
```
alex-plekhanov: For me `cache operation types` are `put`, `get`, etc. Maybe some other description of `keepBinary`? `cache options`? `cache modifiers`? `runtime cache configuration`? `cache operation context` (like in private API)?

zstan: yep, done

docs/_docs/key-value-api/binary-objects.adoc
zstan: moved

docs/_docs/code-snippets/java/src/main/java/org/apache/ignite/snippets/WorkingWithBinaryObjects.java
zstan: fixed

zstan:
@alex-plekhanov all comments are fixed.

alex-plekhanov (approved):

PR #13010, opened 2026-04-10
IGNITE-28485 Fix 'illegal reflective access' during control utility startup

zstan:
before:
```
java -version
openjdk version "11.0.19" 2023-04-18

./control.sh  | grep illegal-access=warn
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.apache.ignite.internal.util.GridUnsafe$2 (file:/home/apache-ignite-slim-2.18.0-bin/libs/ignite-core-2.18.0.jar) to field java.nio.Buffer.address
WARNING: Please consider reporting this to the maintainers of org.apache.ignite.internal.util.GridUnsafe$2
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations

```

after - No warn found in logs.

### DIFF
--- a/bin/include/jvmdefaults.bat
+++ b/bin/include/jvmdefaults.bat
@@ -32,6 +32,7 @@ if %java_version% GEQ 11 if %java_version% LSS 14 (
     --add-exports=jdk.internal.jvmstat/sun.jvmstat.monitor=ALL-UNNAMED ^
     --add-exports=java.base/sun.reflect.generics.reflectiveObjects=ALL-UNNAMED ^
     --add-opens=jdk.management/com.sun.management.internal=ALL-UNNAMED ^
+    --add-opens=java.base/java.nio=ALL-UNNAMED ^
     --illegal-access=permit ^
     %current_value%
 )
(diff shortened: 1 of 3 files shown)


chesnokoff:
Do we also need to update additional files?

It seems that the following should be updated as well:
- `deliveries/docker/apache-ignite/run.sh`
- `docs/_docs/includes/java9.adoc`

Also, just for reference, some files already contain this flag:
- `modules/platforms/cpp/core/src/ignition.cpp`
- `modules/platforms/dotnet/Apache.Ignite.Core/Impl/Unmanaged/Jni/Jvm.cs`
- `parent/pom.xml`

zstan:
@chesnokoff thank you for review, but i`d prefer to store it as is:
1. part i mention in PR is visible for user, while "docker/apache-ignite/run.sh" is just internal stuff, also - i can`t check it right for now, just after release completness - i will fill additional issue in such a case.
2. I agree that java9.adoc need to be fixed, thanks !

chesnokoff (approved):

PR #13037, opened 2026-04-16
IGNITE-28556 Calcite. Support wrap_key and wrap_value flags

zstan:


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/ddl/DdlSqlToCommandConverter.java
```diff
@@ -74,20 +75,40 @@
...
 
+    /** Processor that validates that value can be parsed as boolean. */
+    private static final BiFunction<IgniteSqlCreateTableOption, PlanningContext, Boolean> VALUE_IS_BOOL_IDENTIFIER_VALIDATOR =
+        (opt, ctx) -> {
+            SqlNode val = opt.value();
+            if (!(val instanceof SqlIdentifier) || !((SqlIdentifier)val).isSimple())
+                throwOptionParsingException(opt, SIMPLE_PREDICATE, ctx.query());
+
+            String simple = ((SqlIdentifier)val).getSimple().toLowerCase(Locale.ROOT);
+            if (!"true".equals(simple) && !"false".equals(simple))
```
alex-plekhanov, line 109: equalsIgnoreCase?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/ddl/DdlSqlToCommandConverter.java
```diff
@@ -450,7 +473,7 @@ private AlterTableDropCommand convertAlterTableDrop(IgniteSqlAlterTableDropColum
      */
     private String paramIsSqlIdentifierValidator(IgniteSqlCreateTableOption opt, PlanningContext ctx) {
         if (!(opt.value() instanceof SqlIdentifier) || !((SqlIdentifier)opt.value()).isSimple())
-            throwOptionParsingException(opt, "a simple identifier", ctx.query());
+            throwOptionParsingException(opt, SIMPLE_PREDICATE, ctx.query());
```
alex-plekhanov, line 479: This method is not used, but can be used to simplify VALUE_IS_IDENTIFIER_VALIDATOR and VALUE_IS_BOOL_IDENTIFIER_VALIDATOR

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/ddl/CreateTableCommand.java
```diff
@@ -45,6 +46,12 @@ public class CreateTableCommand implements DdlCommand {
     /** Name of cache key type. */
     private String keyTypeName;
 
+    /** Wrap key flag. */
+    private Boolean wrapKey;
+
+    /** Wrap value flag. */
+    private boolean wrapValue = true;
```
alex-plekhanov, line 53: Maybe we should use Boolean here and move default logic to command handler? This class is a representation of parsed command, if WRAP_VALUE is not provided it's more correct to see null here.
Let's also add the same comment as for H2 somewhere:
`By default value is always wrapped to allow for ALTER TABLE ADD COLUMN commands.`

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/TableDdlIntegrationTest.java
```diff
@@ -347,6 +348,208 @@ public void createTableIfNotExists() {
...
+            "WRAP_VALUE parameter cannot be \"false\" with multiple columns.");
+
+        assertThrowsSqlException(() ->
+                sql("create table a (id int, x varchar, primary key(id)) with \"wrap_value=false, value_type=custom\""),
+            "WRAP_VALUE parameter cannot be \"false\" when VALUE_TYPE is defined.");
+    }
+
+    /** */
+    @SuppressWarnings("ThrowableNotThrown")
+    private static void assertThrowsSqlException(Callable<?> call, @Nullable String msg) {
```
alex-plekhanov, line 450: We have `assertThrows(sql, cls, msg)` in parent class, why do we need yet another method?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/TableDdlIntegrationTest.java
```diff
@@ -347,6 +348,208 @@ public void createTableIfNotExists() {
...
+        GridTestUtils.assertThrows(log, call, IgniteSQLException.class, msg);
+    }
+
+    /**
+     * Test that {@code ADD COLUMN} fails.
+     */
+    @Test
+    public void testAlterTableFailOnSingleCacheValue() {
+        CacheConfiguration c =
```
alex-plekhanov, line 459: Add generics

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
```diff
@@ -386,6 +423,22 @@ else if (!F.isEmpty(cmd.primaryKeyColumns()) && cmd.primaryKeyColumns().size() =
             res = new QueryEntityEx(res).implicitPk(true);
         }
 
+        if (!cmd.wrapValue()) {
```
alex-plekhanov, line 426: `wrapKey` is not handled. Only validated.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/TableDdlIntegrationTest.java
```diff
@@ -347,6 +348,208 @@ public void createTableIfNotExists() {
...
+                "  id2 uuid, " +
+                "  col varchar(100), " +
+                "  PRIMARY KEY(id, id2) )  WITH \"wrap_key=true\"");
+            assertEquals(35, sql(
+                "select INLINE_SIZE from SYS.INDEXES where TABLE_NAME = 'T' and IS_PK = true").get(0).get(0));
+        }
+        finally {
+            sql("DROP TABLE IF EXISTS T");
+        }
+    }
```
alex-plekhanov, line 551: It's not a correct tests for wrapping. PK index is always unwrapped if table was created via DDL. You can change wrap_key to false in all these test and get the same result. To test wrap_key correctly you can iterate over cache values and check key, if it primitive type or binary object.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/jdbc/JdbcQueryTest.java
```diff
@@ -477,6 +479,186 @@ public void testParametersMetadata() throws Exception {
...
+        doTestKeyValueWrap(true, true, true);
+    }
+
+    /**
+     * Test behavior for given combination of wrap flags.
+     * @param wrapKey Whether key wrap should be enforced.
+     * @param wrapVal Whether value wrap should be enforced.
+     * @throws SQLException if failed.
+     */
+    private void doTestKeyValueWrap(boolean wrapKey, boolean wrapVal, boolean testUuid) throws SQLException {
```
alex-plekhanov, line 564: Looks like it's not related to JDBC somehow, just ordinal integration test.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/jdbc/JdbcQueryTest.java
```diff
@@ -477,6 +479,186 @@ public void testParametersMetadata() throws Exception {
...
+        try {
+            String sql = testUuid ? String.format("CREATE TABLE T (\"Id\" UUID primary key, \"xX\" UUID) WITH " +
+                "\"wrap_key=%b,wrap_value=%b", wrapKey, wrapVal) :
+                String.format("CREATE TABLE T (\"Id\" int primary key, \"xX\" varchar) WITH " +
+                    "\"wrap_key=%b,wrap_value=%b", wrapKey, wrapVal);
+
+            UUID guid = UUID.randomUUID();
+
+            if (wrapKey)
```
alex-plekhanov, line 573: Case with wrap_key=true and without key_type is not covered (and has errors, since wrap_key=true never handled)

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/ddl/DdlSqlToCommandConverter.java
zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/ddl/DdlSqlToCommandConverter.java
zstan: done, thanks, i wonder why idea not highlights it for me, i check - options are ok

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/ddl/CreateTableCommand.java
zstan: done

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/TableDdlIntegrationTest.java
zstan: the idea here is sequentially move to logic that sql related operations need to throw IgniteSqlException (yes i know it internal but this is actual reality..) from high level and (in the ideal case) this check method will become commonly uset through sql tests. Move it into parent class.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/TableDdlIntegrationTest.java
zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
zstan: may be it`s enough? or i miss smth in tests (

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/TableDdlIntegrationTest.java
zstan: yep i know it, it`s not about wrap_key, i just come around h2 wrap(key|val) activity and found relates tests, i think it would be helpful to have them for calcite too, wdyt ?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/jdbc/JdbcQueryTest.java
zstan: jdbc metadata is also checked "for company", do you decide we need to cut this tests just for jdbc and 'ordinal' parts ?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/jdbc/JdbcQueryTest.java
zstan: i extend tests and seems it`s all ok there, or i miss smth (

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
```diff
@@ -212,6 +212,8 @@ private void checkKVWrappedParam(CreateTableCommand cmd) {
                     IgniteQueryErrorCode.PARSING);
             }
         }
+        else
+            cmd.wrapValue(true); // By default, value is always wrapped to allow for ALTER TABLE ADD COLUMN commands.
```
alex-plekhanov: Looks like setting to true is redundant, since we check only `if (cmd.wrapValue() != null && !cmd.wrapValue())`
Let's just move comment near to this check..

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/TableDdlIntegrationTest.java
alex-plekhanov: Than you need to reuse `assertThrows` with predefined `IgniteSqlException`. We don't need closure in this method, you never use it other way than `() -> sql(...)`.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/jdbc/JdbcQueryTest.java
alex-plekhanov: You still provide "key_type=" in WITH clause, that's why it works. Try without key_type and compare results with H2.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/TableDdlIntegrationTest.java
alex-plekhanov: But `wrap_key=true` in WITH clause and this confusing. Someone may think that inline size have such a value because wrap_key is true. This test maybe helpful, but we need to check with  wrap_key=true , wrap_key=false and without wrap_key. And ensure that inline sizes are equal for all these cases. And add coment that `PK index is always unwrapped if table was created via DDL`

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
alex-plekhanov: No, key type should be generated when wrapKey is true, see H2 related code.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/jdbc/JdbcQueryTest.java
zstan: ok, got it, done

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/TableDdlIntegrationTest.java
zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
zstan: done

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/TableDdlIntegrationTest.java
zstan: done

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/TableDdlIntegrationTest.java
```diff
@@ -489,57 +489,81 @@ private void doTestAlterTableOnFlatValue(String tblName) {
...
-            sql("DROP TABLE IF EXISTS T");
+    public void testInlineSizeKeyWrapParam() {
+        String query = "CREATE TABLE IF NOT EXISTS T ( " +
+            "  id varchar(15), " +
+            "  col varchar(100), " +
+            "  PRIMARY KEY(id) ) ";
+
+        {
+            try {
+                sql(query);
```
alex-plekhanov: Block duplicated 3 times, instead let's do something like:
```
            for (String ddl : F.asList(qry, qry + "WITH \"wrap_key=true\"", qry + "WITH \"wrap_key=false\"")) {
                sql(ddl);
            }
```

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/TableDdlIntegrationTest.java
```diff
@@ -489,57 +489,81 @@ private void doTestAlterTableOnFlatValue(String tblName) {
...
-                "select INLINE_SIZE from SYS.INDEXES where TABLE_NAME = 'T' and IS_PK = true").get(0).get(0));
+    public void testInlineSizeMultiKeyWrapParam() {
+        String query = "CREATE TABLE IF NOT EXISTS T ( " +
+            "  id varchar(15), " +
+            "  id2 uuid, " +
+            "  col varchar(100), " +
+            "  PRIMARY KEY(id, id2) )";
+        {
+            try {
+                sql(query);
```
alex-plekhanov: The same, duplication here.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/TableDdlIntegrationTest.java
zstan: done

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/TableDdlIntegrationTest.java
zstan: done

alex-plekhanov (approved):

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/sql/SqlReservedWordsTest.java
```diff
@@ -114,6 +114,8 @@ public class SqlReservedWordsTest extends GridCommonAbstractTest {
         "WHEN",
         "WHERE",
         "WITH",
+        "WRAP_KEY",
+        "WRAP_VALUE",
```
alex-plekhanov: Let's exclude it from reserved words instead of fixing test

alex-plekhanov: As already made for other WITH parameters (see nonReservedKeywordsToAdd)

alex-plekhanov (approved):

PR #13045, opened 2026-04-17
IGNITE-28570 Calcite. Fix failure on test_large_in.test

zstan:


zstan:
If someone try to run this test with: -Xss512k it will also fail, so i just cut a bit test items set.

PR #13058, opened 2026-04-21
IGNITE-28515 : Use MessageSerializer for QueryIndex - copy

Vladsz83:


anton-vinogradov (approved):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheIoManager.java
```diff
@@ -917,7 +917,7 @@ private void processFailedMessage(UUID nodeId,
 
             break;
 
-            case 10910: {
+            case 10911: {
```
shishkovilja, line 920: @Vladsz83 @anton-vinogradov , now it looks very ugly that we need to do this. It is necessary to refactor this case block.

Vladsz83: We should use smth. like msgFactory.messageType(msg). Looks like other ticket. Or I could place the message below in the registration list. But I could place it where it should be and make a fix.

PR #13099, opened 2026-05-04
IGNITE-28629 Support savepoint in SQL

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-28629

### DIFF
--- a/docs/_docs/sql-reference/transactions.adoc
+++ b/docs/_docs/sql-reference/transactions.adoc
@@ -1,0 +1,76 @@
+// Licensed to the Apache Software Foundation (ASF) under one or more
+// contributor license agreements.  See the NOTICE file distributed with
+// this work for additional information regarding copyright ownership.
+// The ASF licenses this file to You under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance with
+// the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+= Transactions
+
+This page describes SQL transaction commands supported by the Calcite-based SQL engine.
+
+== SAVEPOINT
+
+Creates a named savepoint in the current transaction.
+
+[source,sql]
+----
+SAVEPOINT savepointName
+----
+
+=== Parameters
+
+- `savepointName` - the name of the savepoint to create.
+
+=== Description
+
+`SAVEPOINT` can be used only inside an explicit `PESSIMISTIC` transaction.
+
+The command records the current transaction state. Later, you can use `ROLLBACK TO SAVEPOINT` to roll back all transaction changes made after the savepoint was created.
+
+If a savepoint with the same name already exists, `SAVEPOINT` replaces it. The command does not roll back any transaction changes. It removes the previous savepoint with that name and makes the name refer to the current transaction state.
+
+== ROLLBACK TO SAVEPOINT
+
+Rolls back transaction changes to a previously created savepoint.
+
+[source,sql]
+----
+ROLLBACK TO SAVEPOINT savepointName
+----
+
+=== Parameters
+
+- `savepointName` - the name of the savepoint to roll back to.
+
+=== Description
+
+`ROLLBACK TO SAVEPOINT` can be used only inside an explicit `PESSIMISTIC` transaction.
+
+The command rolls back all transaction changes made after the specified savepoint was created. The transaction remains active and can be committed or rolled back later.
+
+When a transaction is rolled back to a savepoint, savepoints created after the target savepoint are released. The target savepoint remains available and can be used again.
+
+If the specified savepoint does not exist, the command fails.
+
+== Example
+
+[source,sql]
+----
+INSERT INTO Person(id, name) VALUES (1, 'John');
+
+SAVEPOINT before_update;
+
+UPDATE Person SET name = 'Jane' WHERE id = 1;
+
+ROLLBACK TO SAVEPOINT before_update;
+----
+
+After `ROLLBACK TO SAVEPOINT before_update`, the `UPDATE` statement is rolled back, but the preceding `INSERT` statement remains part of the active transaction.
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
@@ -64,6 +64,7 @@ import org.apache.ignite.internal.util.typedef.F;
 import org.apache.ignite.lang.IgniteUuid;
 import org.apache.ignite.plugin.security.SecurityPermission;
 
+import static org.apache.ignite.internal.processors.cache.distributed.near.GridNearTxLocal.SAVEPOINTS_EXPLICIT_TX_ONLY;
 import static org.apache.ignite.internal.processors.query.QueryUtils.convert;
 import static org.apache.ignite.internal.processors.query.QueryUtils.isDdlOnSchemaSupported;
 
@@ -131,10 +132,8 @@ public class DdlCommandHandler {
 
         GridNearTxLocal tx = Commons.queryTransaction(qryCtx, cacheProc.context());
 
-        if (tx == null) {
-            throw new IgniteSQLException("Savepoints can be used only inside explicit transactions.",
-                IgniteQueryErrorCode.UNSUPPORTED_OPERATION);
-        }
+        if (tx == null)
+            throw new IgniteSQLException(SAVEPOINTS_EXPLICIT_TX_ONLY, IgniteQueryErrorCode.UNSUPPORTED_OPERATION);
 
         switch (cmd.type()) {
             case SAVEPOINT:
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
@@ -36,6 +36,7 @@ import org.apache.ignite.configuration.CacheConfiguration;
 import org.apache.ignite.internal.processors.cache.GridCacheContext;
 import org.apache.ignite.internal.processors.cache.GridCacheContextInfo;
 import org.apache.ignite.internal.processors.cache.GridCacheProcessor;
+import org.apache.ignite.internal.processors.cache.distributed.near.GridNearTxLocal;
 import org.apache.ignite.internal.processors.cache.query.IgniteQueryErrorCode;
 import org.apache.ignite.internal.processors.query.GridQueryProcessor;
 import org.apache.ignite.internal.processors.query.GridQueryTypeDescriptor;
@@ -43,6 +44,7 @@ import org.apache.ignite.internal.processors.query.IgniteSQLException;
 import org.apache.ignite.internal.processors.query.QueryEntityEx;
 import org.apache.ignite.internal.processors.query.QueryField;
 import org.apache.ignite.internal.processors.query.QueryUtils;
+import org.apache.ignite.internal.processors.query.calcite.prepare.BaseQueryContext;
 import org.apache.ignite.internal.processors.query.calcite.prepare.ddl.AlterTableAddCommand;
 import org.apache.ignite.internal.processors.query.calcite.prepare.ddl.AlterTableDropCommand;
 import org.apache.ignite.internal.processors.query.calcite.prepare.ddl.ColumnDefinition;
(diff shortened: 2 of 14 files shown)


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/sql/IgniteSqlRollbackToSavepoint.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /** */
+    public SqlIdentifier name() {
+        return name;
+    }
+
+    /** {@inheritDoc} */
+    @Override public List<SqlNode> getOperandList() {
+        return ImmutableList.of(name);
```
zstan: +1 with sonar here

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/sql/IgniteSqlRollbackToSavepoint.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.SqlSpecialOperator;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+/** Parse tree for {@code ROLLBACK TO SAVEPOINT name} statement. */
+public class IgniteSqlRollbackToSavepoint extends SqlDdl {
+    /** */
+    protected static final SqlOperator OPERATOR =
+        new SqlSpecialOperator("ROLLBACK TO SAVEPOINT", SqlKind.OTHER_DDL);
```
zstan, line 34: let`s move "ROLLBACK TO SAVEPOINT" to final static, cause it used more than once ?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/sql/IgniteSqlSavepoint.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /** */
+    public SqlIdentifier name() {
+        return name;
+    }
+
+    /** {@inheritDoc} */
+    @Override public List<SqlNode> getOperandList() {
+        return ImmutableList.of(name);
```
zstan: +1 here too, let`s reduce guava usage

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/sql/IgniteSqlRollbackToSavepoint.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override public List<SqlNode> getOperandList() {
+        return ImmutableList.of(name);
+    }
+
+    /** {@inheritDoc} */
+    @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+        writer.keyword("ROLLBACK TO SAVEPOINT");
```
zstan: why not ?
```suggestion
       writer.keyword(getOperator().getName());
```

modules/calcite/src/test/java/org/apache/ignite/internal/processors/tx/SqlTransactionsSavepointTest.java
```diff
@@ -0,0 +1,240 @@
...
+        sql("CREATE TABLE " + TBL + "(ID INT PRIMARY KEY, VAL VARCHAR) WITH atomicity=transactional");
+    }
+
+    /** */
+    @Test
+    public void testSavepointCommandsInSqlScript() {
+        try (Transaction tx = client.transactions().txStart(PESSIMISTIC, READ_COMMITTED)) {
+            sqlScript(
+                "INSERT INTO " + TBL + " VALUES (1, 'before_sp1');" +
+                "SAVEPOINT sp1;" +
```
zstan, line 75: i see - it`s possible to create savepoints with equal names, is it expected ? if so - such ambiguity need to be documented too !

modules/calcite/src/test/java/org/apache/ignite/internal/processors/tx/SqlTransactionsSavepointTest.java
```diff
@@ -0,0 +1,240 @@
...
+                "INSERT INTO " + TBL + " VALUES (2, 'after_sp1');" +
+                "SAVEPOINT sp2;" +
+                "UPDATE " + TBL + " SET VAL = 'after_sp2' WHERE ID = 1;" +
+                "DELETE FROM " + TBL + " WHERE ID = 2;" +
+                "ROLLBACK TO SAVEPOINT sp2"
```
zstan, line 80: savepoint`s need to be liner as mentioned ? thus such is impossible too ? 
```
                "ROLLBACK TO SAVEPOINT sp2;" +
                "ROLLBACK TO SAVEPOINT sp1;" +
                "ROLLBACK TO SAVEPOINT sp2;"
```

modules/calcite/src/test/java/org/apache/ignite/internal/processors/tx/SqlTransactionsSavepointTest.java
```diff
@@ -0,0 +1,240 @@
...
+            );
+
+            assertQuery(tx, "SELECT ID, VAL FROM " + TBL + " ORDER BY ID")
+                .returns(1, "before_sp1")
+                .returns(2, "after_sp1")
+                .ordered()
```
zstan, line 86: redundant, already ordered

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
```diff
@@ -123,6 +124,34 @@ else if (cmd instanceof NativeCommandWrapper)
         }
     }
 
+    /** */
+    private void handle0(BaseQueryContext qryCtx, TransactionCommand cmd) throws IgniteCheckedException {
+        if (cmd.type() == TransactionCommand.Type.NOOP)
```
zstan, line 140: how can we get NOOP here ? need test or remove.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
```diff
@@ -123,6 +124,34 @@ else if (cmd instanceof NativeCommandWrapper)
...
+            return;
+
+        GridNearTxLocal tx = Commons.queryTransaction(qryCtx, cacheProc.context());
+
+        if (tx == null) {
+            throw new IgniteSQLException("Savepoints can be used only inside explicit transactions.",
```
zstan: move to final static plz ? 

Image

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
```diff
@@ -123,6 +124,34 @@ else if (cmd instanceof NativeCommandWrapper)
         }
     }
 
+    /** */
+    private void handle0(BaseQueryContext qryCtx, TransactionCommand cmd) throws IgniteCheckedException {
```
zstan: we don`t need a whole **qryCtx** here only tx, also - tx is autoClosable - who is resbonsible for closing it ? need to document it somehow

docs/_docs/sql-reference/transactions.adoc
```diff
@@ -0,0 +1,76 @@
...
+
+When a transaction is rolled back to a savepoint, savepoints created after the target savepoint are released. The target savepoint remains available and can be used again.
+
+If the specified savepoint does not exist, the command fails.
+
+== Example
+
+[source,sql]
+----
+INSERT INTO Person(id, name) VALUES (1, 'John');
```
zstan, line 67: this is pure sql syntax, do we need additionally append : transactions().txStart(PESSIMISTIC, READ_COMMITTED) and tx.commit(); here ? Just for clarification ?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/ddl/TransactionCommand.java
```diff
@@ -16,8 +16,47 @@
...
+
+        /** Roll transaction changes back to a savepoint. */
+        ROLLBACK_TO_SAVEPOINT
+    }
+
+    /** */
+    private final Type type;
+
+    /** Savepoint name. */
+    private final String savepointName;
```
zstan: @Nullable ?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/sql/IgniteSqlRollbackToSavepoint.java
vldpyatkov: I am not sure that is right, but it is a style in our code. For example, IgniteSqlCommit, IgniteSqlRollback.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/sql/IgniteSqlRollbackToSavepoint.java
vldpyatkov: Ok.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/tx/SqlTransactionsSavepointTest.java
vldpyatkov: It is documented in  transactions.adoc
`If a savepoint with the same name already exists, `SAVEPOINT` replaces it. The command does not roll back any transaction changes. It removes the previous savepoint with that name and makes the name refer to the current transaction state.`

modules/calcite/src/test/java/org/apache/ignite/internal/processors/tx/SqlTransactionsSavepointTest.java
vldpyatkov: Your example looks wrong because you cannot use sp2 after you already rolled back to sp1.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
vldpyatkov: It is present behavior, a reaction to COMMIT/ROLLBAK SQL command.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
vldpyatkov: Extracted it in a const SAVEPOINTS_EXPLICIT_TX_ONLY

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
vldpyatkov: I do not see a way to avoid `qryCtx` because it is used to get a transaction.
As for the question, "Who is responsible for closing?" It is a transaction from the client layer (suspended and resumed to use in SQL flow), so the client is responsible for this one.

docs/_docs/sql-reference/transactions.adoc
vldpyatkov: We do not have a syntax for a SQL script to start/stop transactions.
I do not think that a part of Java syntax is appropriate here. But you're right; we would like to have BEGIN, COMMIT, and ROLLBACK functions in SQL.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/tx/SqlTransactionsSavepointTest.java
vldpyatkov: The behavior is also documented:
`When a transaction is rolled back to a savepoint, savepoints created after the target savepoint are released.`

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/sql/IgniteSqlRollbackToSavepoint.java
zstan: now - it`s ok

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
zstan: you can push not **qryCtx** but lambda or simple tx = Commons.queryTransaction(qryCtx, cacheProc.context()); I believe we need to reduce function parameters scope if it possible, wdyt ?

zstan (commented):
overall looks good besides 1 comment, i still believe it can be a bit clearer

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearTxLocal.java
```diff
@@ -3035,9 +3035,10 @@ private void resume(long threadId) throws IgniteCheckedException {
     /**
      * Creates savepoint for a pessimistic transaction.
      *
-     * @param name Savepoint name.
+     * @param name      Savepoint name.
```
zstan, line 3038: tabulation ?

zstan (approved):

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/ddl/DdlCommandHandler.java
vldpyatkov: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearTxLocal.java
vldpyatkov: It is not a tab, but a series of spaces. I do not know, but I just pressed a hot key, and the text was aligned in comparison with our stylesheet.
Anyway, I have to get a visa before merging it. We will look. Is it an issue or our old good style...

PR #13112, opened 2026-05-07
IGNITE-28649 Support JDBC API for savepoint

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-28649


docs/_docs/SQL/JDBC/jdbc-driver.adoc
```diff
@@ -555,6 +555,46 @@ In addition to generic DataSource properties, `IgniteJdbcThinDataSource` support
...
+
+    try (Statement stmt = conn.createStatement()) {
+        stmt.executeUpdate("INSERT INTO Person(id, name) VALUES (1, 'John')");
+
+        Savepoint savepoint = conn.setSavepoint("before_update");
+
+        stmt.executeUpdate("UPDATE Person SET name = 'Jane' WHERE id = 1");
+
+        conn.rollback(savepoint);
+        conn.releaseSavepoint(savepoint);
```
zstan, line 585: from this example is not clear - what was the final state of row **name**  where **id** = 1, can you extend this test plz ? I suppose pure java: **assert** expectation would be enough here

docs/_docs/SQL/JDBC/jdbc-driver.adoc
```diff
@@ -555,6 +555,46 @@ In addition to generic DataSource properties, `IgniteJdbcThinDataSource` support
...
+    }
+    catch (Throwable t) {
+        conn.rollback();
+
+        throw t;
+    }
+}
+----
+
+You can also use SQL savepoint commands, such as `SAVEPOINT` and `ROLLBACK TO SAVEPOINT`, from JDBC statements.
```
zstan: probably in the third person would be more accurate, smth like : SQL savepoint command can be used ...

modules/clients/src/test/java/org/apache/ignite/jdbc/thin/JdbcThinConnectionSavepointTest.java
```diff
@@ -0,0 +1,256 @@
...
+import org.apache.ignite.configuration.TransactionConfiguration;
+import org.junit.Test;
+
+/** Savepoint tests for thin JDBC connection. */
+public class JdbcThinConnectionSavepointTest extends JdbcThinAbstractSelfTest {
+    /** */
+    private static final String TBL = "SAVEPOINT_TEST_TABLE";
+
+    /** URL. */
+    private String url = partitionAwareness ?
```
zstan:
```suggestion
    private final String url = partitionAwareness ?
```

docs/_docs/SQL/JDBC/jdbc-driver.adoc
```diff
@@ -555,6 +555,46 @@ In addition to generic DataSource properties, `IgniteJdbcThinDataSource` support
...
+
+JDBC Thin Driver supports the standard JDBC savepoint API for explicit transactions:
+
+* `Connection.setSavepoint()`
+* `Connection.setSavepoint(String name)`
+* `Connection.rollback(Savepoint savepoint)`
+* `Connection.releaseSavepoint(Savepoint savepoint)`
+
+Savepoints are available for JDBC connections that use the Calcite-based SQL engine and explicit `PESSIMISTIC` transactions.
+Disable auto-commit before creating a savepoint.
```
zstan, line 568: Otherwize what ?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/jdbc/JdbcThinConnectionSavepointTest.java
```diff
@@ -0,0 +1,256 @@
...
+                conn.rollback(sp2);
+
+                assertQuery(conn, 1, "after_sp1", 2, "after_sp1");
+
+                conn.releaseSavepoint(sp2);
+                conn.rollback(sp1);
+
+                assertQuery(conn, 1, "before_sp1");
+
+                conn.releaseSavepoint(sp1);
```
zstan, line 110: plz append additional test, from java doc : java.sql.Connection#releaseSavepoint
```
Any reference to the savepoint after it have been removed will cause an SQLException to be thrown.
```

thus test like :

```
Savepoint sp1 = conn.setSavepoint("sp1");

conn.releaseSavepoint(sp1);
conn.rollback(sp1);
```
need to fail correctly

Probably test org.apache.ignite.jdbc.thin.JdbcThinConnectionSelfTest#testReleaseSavepoint
need to be expanded, i see only null and fake = not exist safePoints been checked there

modules/clients/src/test/java/org/apache/ignite/jdbc/thin/JdbcThinConnectionSelfTest.java
```diff
@@ -1695,11 +1695,17 @@ public void testReleaseSavepoint() throws Exception {
 
             final Savepoint savepoint = getFakeSavepoint();
 
-            checkNotSupported(new RunnableX() {
-                @Override public void runx() throws Exception {
-                    conn.releaseSavepoint(savepoint);
-                }
-            });
+            assertThrows(log,
+                new Callable<Object>() {
```
zstan:
```suggestion
                new Callable<>() {
```

modules/clients/pom.xml
```diff
@@ -73,6 +73,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>${project.groupId}</groupId>
```
zstan: probably it`s better to have Calcite related test in Calcite ? 
check:
package org.apache.ignite.internal.processors.query.calcite.jdbc;

zstan:
what about JdbcThinFeature and correct before\after savepoint jdbc vs old\new savepoint cluster correct handling tests ?

docs/_docs/SQL/JDBC/jdbc-driver.adoc
vldpyatkov: Added a check that it is not possible to roll back on an already rolled back savepoint.

docs/_docs/SQL/JDBC/jdbc-driver.adoc
vldpyatkov: Added note:
`If auto-commit remains enabled, JDBC savepoint API calls that require an explicit transaction fail with `SQLException`.`

zstan: It need to be smth like : 
If auto-commit is left enabled, JDBC savepoint API calls requiring an explicit transaction will fail with an SQLException.

zstan (approved):
overall looks good, plz check last 2 comments ?

PR #13136, opened 2026-05-14
IGNITE-28675 Fix flaky TxDeadlockCauseTest #testCause, #testCauseSeveralNodes

zstan:


zstan:
From documentation: 
https://ignite.apache.org/docs/ignite2/latest/key-value-api/transactions
However, if a deadlock is detected, the cause of the returned TransactionTimeoutException will be TransactionDeadlockException (**at least** for one transaction involved in the deadlock).

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/transactions/TxDeadlockCauseTest.java
```diff
@@ -204,7 +206,9 @@ private void checkCauseObject(
                     tx.commit();
                 }
                 catch (Exception e) {
-                    ex.compareAndSet(null, e);
+                    // TransactionDeadlockException raised at least for one transaction involved in the deadlock
```
zstan, line 209: this is from our own documentation

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/TxDeadlockDetection.java
```diff
@@ -55,7 +56,7 @@ public class TxDeadlockDetection {
     public static final int DFLT_TX_DEADLOCK_DETECTION_TIMEOUT = 60000;
 
     /** Deadlock detection maximum iterations. */
-    private static int deadLockTimeout =
+    private static final int DEAD_LOCK_TIMEOUT =
```
chesnokoff, line 59: Let's roll back adding the `final` qualifier here, because it changes the semantics of `TxDeadlockDetectionNoHangsTest` and is not the main goal of this patch.

chesnokoff: and roll back renaming too

zstan: disagree, why do we need to support non-final definition only for test cases ?
test work properly well without this reflection changes, i run it multiple times, you can check it through TC

chesnokoff: My position is not really about final/non-final definition itself. My point is that it is usually safer to keep PRs as minimal as possible for a specific purpose. The main goal here is to fix `TxDeadlockCauseTest`, and for now it is not very clear why we also need to refactor an unrelated field and affect existing test behavior.

Green TC runs are definitely a good sign, but some flaky tests fail once in weeks or even months. That's why I'd prefer to avoid introducing additional assumptions here unless this change is really required for the fix itself.

Could you provide more reasoning on why it is safe to remove before/afterTests from `TxDeadlockDetectionNoHangsTest`?

zstan: I appent additional checks into test, plz check last commit
Also, tests restore after for hardcoded - 60000, that equal to current default but can be changed in future thus there can be a situation when test restores not a default.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/transactions/TxDeadlockDetectionNoHangsTest.java
```diff
@@ -68,6 +74,9 @@ public class TxDeadlockDetectionNoHangsTest extends GridCommonAbstractTest {
 
         cfg.setCacheConfiguration(ccfg);
 
+        listeningLog.registerListener(lsnr);
```
chesnokoff:
```
public void registerListener(@NotNull LogListener lsnr) {
        lsnr.reset();

```

Can we lose an already caught "Deadlock detection was timed out" message here? `doTest` starts `grid(NODES_CNT)` many times and each start calls `getConfiguration()`. But `registerListener(LogListener)` resets the listener, so a later node start can clear a timeout message that was caught before

zstan: I missunderstand - lsnr.check() calls after each test on: afterTest()
Can you explain how it can be a problem ? Thanks !

chesnokoff: I mean that `lsnr.reset()` can clear the listener state. Can `getConfiguration()` be called after `Deadlock detection was timed out` was already logged?

`doTest()` runs for about 120 seconds. During this time, the restart thread starts `NODES_CNT` grids many times, and each start calls `getConfiguration()`. The deadlock detection timeout is 60 seconds, so the timeout log can appear first. After that, another node restart can call `getConfiguration()` and reset the listener before `afterTest()` calls `lsnr.check()`

zstan: got it ! thanks, if now it`s ok i will repeatedly run a TC

chesnokoff: Theoretically, we still have a small race:
```
assertFalse(DEAD_LOCK_LSNR.check());
// catch the log
listeningLog.registerListener(DEAD_LOCK_LSNR);
```

Probably it would be better to use
```
public void registerListener(@NotNull Consumer<String> lsnr)
```
and make smth like
```
registerListener(log -> <update atomicBoolean if catch the log>)
```
so the atomicBoolean would be a common state for all grids


But on the other hand it makes the test more complicated. WDYT?

zstan: you say the proper things of cource, thanks, change a tests.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/transactions/TxDeadlockDetectionNoHangsTest.java
```diff
@@ -82,20 +99,9 @@ public class TxDeadlockDetectionNoHangsTest extends GridCommonAbstractTest {
...
-
-        GridTestUtils.setFieldValue(TxDeadlockDetection.class, "deadLockTimeout", (int)(getTestTimeout() * 2));
-    }
-
-    /** {@inheritDoc} */
-    @Override protected void afterTestsStopped() throws Exception {
-        GridTestUtils.setFieldValue(TxDeadlockDetection.class, "deadLockTimeout",
-            getInteger(IGNITE_TX_DEADLOCK_DETECTION_TIMEOUT, 60000));
+        stopAllGrids();
     }
```
chesnokoff, line 105:
```suggestion
    @Override protected void afterTest() throws Exception {
        super.afterTest();

        stopAllGrids();

        assertFalse(DEAD_LOCK_FLAG.get());
    }
```

chesnokoff: We need to change order of `stopAllGrids` and `assertFalse` because otherwise we won't stop grids if the assertions fails

zstan: done

chesnokoff (approved):

PR #13140, opened 2026-05-15
IGNITE-28687 Remove MarshallableMessage from PartitionUpdateCountersMessage

Vladsz83:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/PartitionUpdateCountersMessage.java
```diff
@@ -158,10 +171,9 @@ private void ensureSpace(int newSize) {
         int req = newSize * ITEM_SIZE;
 
         if (data.length < req)
-            data = Arrays.copyOf(data, data.length << 1);
+            data = Arrays.copyOf(data, (int)(data.length * 1.33f));
```
shishkovilja, line 174: Are you trying to eliminate unnecessary allocations?

Vladsz83: There is no quarantine of invoking _finishUpdating()_. We lack a message finalization life cycle (not _prepareMarshall()_). If wasn't finished/optimized, less memory would wasted. Also, growing with factor of 1.3 or 1.5 is a common practice too. x2 looks too sweeping.

shishkovilja: Ok, thanks for clarification.

PR #13162, opened 2026-05-20
IGNITE-28706 Fix TxWithExceptionalInterceptorTest.testTxWithExceptionInterceptor fails

zstan:


zstan:
After fix: "Run until failure" is ok
image

modules/calcite/src/test/java/org/apache/ignite/internal/processors/tx/TxWithExceptionalInterceptorTest.java
```diff
@@ -363,8 +366,19 @@ public void testTxWithExceptionInterceptor() throws Exception {
...
 
+            Object sqlVal;
+
             // obtain sql results first, kv api can eventually recover results, thus for more clear test - let`s check sql first
-            Object sqlVal = getSqlResultByKey(node, PROC_CACHE_NAME, primaryKey, false);
+            try {
+                sqlVal = getSqlResultByKey(node, PROC_CACHE_NAME, primaryKey, false);
+            }
+            catch (IgniteSQLException ex) {
+                assertTrue(X.hasCause(ex, ClusterTopologyException.class));
```
alex-plekhanov: 1. Block under condition `txCoord == TxCoordNodeRole.PRIMARY` uses `getSqlResultByKey` method too, but not covered by the same check. Why?

2. There is a line in this method:
```
Integer primaryKeyCommon = primaryKeyCoordAware(PROC_CACHE_NAME);
```
Here PROC_CACHE_NAME is used, but not COMMON_CACHE_NAME. Is it intentional?

3. Several code style violations in this method (comment is not started with uppercase, and not ended with the point. Let's fix it.

zstan: fixed

alex-plekhanov (approved):

PR #13195, opened 2026-05-31
IGNITE-28711 Introduced  IgniteFeature related data structures

petrov-mg:

### Notes


modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -17,334 +17,480 @@
...
+    private final Set<ClusterNode> joiningNodes = new HashSet<>();
 
     /** */
-    private final CountDownLatch startLatch = new CountDownLatch(1);
+    private volatile boolean isNodeFenceActive;
 
-    /** Pair with current and target versions. {@code null} when rolling upgrade is disabled. */
-    @Nullable private volatile IgnitePair<IgniteProductVersion> rollUpVers;
+    /** */
+    private volatile boolean isVersionUpgradeEnabled;
```
chesnokoff: Should we keep these two fields in metastorage to restore them in persistence mode after full restart of cluster?

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -17,334 +17,480 @@
...
+                        ", joiningNode=" + joiningNode + ']');
             }
-        });
-    }
 
-    /** {@inheritDoc} The joining node is stored to verify later whether it successfully connected to the ring or failed to join. */
-    @Override public @Nullable IgniteNodeValidationResult validateNode(ClusterNode node) {
-        synchronized (lock) {
-            lastJoiningNode = node;
+            joiningNodes.add(joiningNode);
```
chesnokoff, line 281: The problem is that a joining node may leave the cluster without producing any discovery event.

See `org.apache.ignite.spi.discovery.tcp.ServerImpl.RingMessageWorker#processJoinRequestMessage`, around lines 4522-4595:
```java
err = spi.getSpiContext().validateNode(node);
...
if (!Objects.equals(locMarsh, rmtMarsh)) {
    ...
    // Send message "Local node's marshaller differs from remote node's marshaller"
    trySendMessageDirectly(node, new TcpDiscoveryCheckFailedMessage(locNodeId, sndMsg));
    return;
}
```
In this path the node has already passed component validation, but it is rejected later by TCP discovery checks and never joins the topology. As a result, the coordinator does not receive `NODE_JOINED`/`NODE_LEFT`/`NODE_FAILED` for this node, so it cannot remove it from `joiningNodes` by listening to discovery events only.

This is why the previous implementation tracked `lastJoiningNode` and `lastJoiningNodeTimestamp`. They were needed to eventually forget such abandoned join attempts

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
chesnokoff:
```suggestion
    private volatile boolean isVerUpgradeEnabled;
```

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
petrov-mg: Not relevant after [`IGNITE-28751`](https://issues.apache.org/jira/browse/IGNITE-28751)

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
petrov-mg: Done.

petrov-mg: Storing RU state in metastorage has a drawback - any data stored in metastorage requires full backward compatibility because it is written to PDS. Furthermore, there is currently no strong reasons to support RU state recovery after a full cluster restart. This has also been confirmed by potential users of the RU mechanism. In any case, if such a need arises, it can be easily implemented.

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -17,331 +17,514 @@
...
-            if (lastJoiningNode != null) {
-                IgniteProductVersion lastJoiningNodeVer = IgniteProductVersion.fromString(lastJoiningNode.attribute(ATTR_BUILD_VER));
+                isNodeFenceActive = true;
 
-                if (!minMaxVerPair.get1().equals(lastJoiningNodeVer))
-                    throw new IgniteCheckedException("Can't disable rolling upgrade with different versions in cluster: "
-                        + minMaxVerPair.get1() + ", " + lastJoiningNodeVer);
+                return new GridFinishedFuture<>();
             }
+        }
```
chesnokoff: Do we need to execute the prepare phase on every node?

Consider a 3-node cluster A, B, C where A is the coordinator. A starts finalization and sends the prepare phase to all nodes. If prepare succeeds on B but fails on C for some reasons, B has already set `isNodeFenceActive=true`, while the coordinator aborts finalization because one node failed. Since the complete phase is not started, B never resets the fence. A later finalization attempt will fail because B now reports "Cluster version finalization procedure is already in progress", and joins routed through B will also be rejected.

Could you clarify why prepare must validate cluster and set the fence independently on every node? If this is intended to protect coordinator changes, we probably still need a rollback mechanism for `isNodeFenceActive` when prepare fails on any node

petrov-mg: > If prepare succeeds on B but fails on C for some reasons

For what reasons? This is crucial, as we rely on message delivery guarantee provided by Discovery.

> coordinator aborts finalization because one node failed

Currently the distributed process is not aborted if nodes leave the cluster.

chesnokoff: My point is not about discovery messaging.

The prepare phase can finish with an exception in `GridFinishedFuture` (the version check). So `isNodeFenceActive` becomes `false` on the node that failed, but `true` on the nodes that passed.

In my scenario, coordinator A may fail its own prepare phase, if its local `joiningNodes` still has a node with a different version. Only the coordinator handles `joiningNodes`, so the discovery delivery guarantee does not help here.

Because A returns an error, `finishPreparePhase` does not start `completePhase`. And `completePhase` is the only place that sets `isNodeFenceActive` back to `false`. So B and C keep the fence forever, while A is not even fenced

petrov-mg: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -17,331 +17,529 @@
...
+        ctx.security().authorize(ADMIN_ROLLING_UPGRADE);
+
+        if (!isVerUpgradeEnabled)
+            return;
+
+        finalizeProc.start().get();
+
+        if (log.isInfoEnabled())
+            log.info("Cluster version was successfully finalized [activeLogicalVer=" + clusterLogicalVersion() + ']');
+    }
```
chesnokoff, line 152: Should we allow `finalizeProc.start()` from non-coordinator nodes?

`finalizeClusterVersion` can currently be called from different nodes at the same time. In that case we start two different distributed processes, for example `finProc1` and `finProc2`.

`finProc1` may execute prepare first and set `isInProgress=true` and `isNodeFenceActive=true`. Then `finProc2` executes prepare and fails with `finalization procedure is already in progress`. However, when `finProc2` finishes, finishProcess resets `isInProgress=false` for any `reqId`. Also, the prepare error path resets `isNodeFenceActive=false`.

As a result, the failed concurrent finalization can clear the fence of the still running finalization process.

Maybe we should allow finalize only from the coordinator, like the old RU implementation did, or keep `isInProgress` and `isNodeFenceActive` to a specific reqId and reset them only for the process that set them

petrov-mg: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/rollingupgrade/ClusterVersionsRollingUpgradeTest.java
```diff
@@ -0,0 +1,853 @@
...
+    @Test
+    public void testConcurrentFinalization() throws Exception {
+        startCluster();
+
+        ru(1).enableVersionUpgrade();
+
+        spi(grid(1)).blockMessages((node, msg) -> msg instanceof SingleNodeMessage);
+
+        try {
+
```
sergey-chugunov-1985:
```suggestion
```

sergey-chugunov-1985 (changes requested):

modules/core/src/test/java/org/apache/ignite/internal/processors/rollingupgrade/ClusterVersionsRollingUpgradeTest.java
petrov-mg: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -17,331 +17,564 @@
...
+            if (!F.isEmpty(errors)) {
+                if (reqId.equals(activeProcId))
+                    isNodeFenceActive = false;
 
-        if (log.isInfoEnabled())
-            log.info("Rolling upgrade disabled. Current version of nodes in cluster: " + minMaxVerPair.get1());
+                finishProcess(reqId, firstError(errors));
+            }
+            else if (U.isLocalNodeCoordinator(ctx.discovery()))
+                completePhase.start(reqId, null);
```
chesnokoff: I have some doubts about these lines.
Let's assume we start finalization and `executePreparePhase` completed on all nodes. After that we start `finishPreparePhase` on all nodes. If coordinator disconnects at this moment, other nodes may not execute start of `completePhase`. As a result, none of grids will start the `completePhase`. Moreover, if we try to call finalization again, it will be canceled because of enabled fences. And we also won't be able to connect new nodes to cluster. And probably the only way to fix it will be full restart of cluster during rolling upgrade 

Of course, looks like the scenario is rare and we have the same pattern in other processes. But here it will break concept of rolling upgrade

WDYT?

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -17,331 +17,564 @@
...
+        featureMgr.onGridDataReceived(gridData.activeFeatures());
+    }
+
+    /** {@inheritDoc} */
+    @Override public @Nullable IgniteNodeValidationResult validateNode(ClusterNode joiningNode) {
+        synchronized (topGuard) {
+            if (isNodeFenceActive) {
+                return new IgniteNodeValidationResult(
+                    joiningNode.id(),
+                    "Node joins are not allowed during cluster version finalization [joiningNode=" + joiningNode + ']');
```
chesnokoff: We have received user feedback that some error messages do not explain what to do next. In this case, the finalization process is an internal operation, so it is unclear whether the user should retry or take another action. Could we add an actionable hint, such as `Wait for the current finalization process to complete and try again later`? 

We can also add similar tips to other RU errors, but it is up to you

petrov-mg: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
petrov-mg: The issue you describe can indeed occur. A known example is when the coordinator enqueues an InitMessage corresponding to the finalization completion phase to the RingMessageWorker and then fails. There may be other such cases.

Even so, I propose keeping the current logic as is. There's no point in further complicating the RU logic to fix this issue. As you noted, the chances of such a scenario are low. 

I  think it worth it to implement @sergey-chugunov-1985 suggestion and introduce a mechanism for manually aborting the cluster version finalization process. This will give us a backup plan for such corner cases.

The PR contains an implementation of the mentioned proposal.

sergey-chugunov-1985 (approved):

chesnokoff (approved):
Just to record the current limitations for future investigation:

- Coordinator change during finalization may prevent the complete phase from being started
- Failure of an activation listener may interrupt notification of the remaining listeners and leave the node in a half-finalized state

PR #13217, opened 2026-06-07
IGNITE-28751 Refactored TCP Discovery SPI joining node validation

petrov-mg:

### Notes


modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -85,26 +78,25 @@ public RollingUpgradeProcessor(GridKernalContext ctx) {
...
     /** {@inheritDoc} */
     @Override public void start() throws IgniteCheckedException {
-        ctx.event().addLocalEventListener(new GridLocalEventListener() {
-            @Override public void onEvent(Event evt) {
+        ctx.event().addLocalEventListener(
+            evt -> {
                 UUID nodeId = ((DiscoveryEvent)evt).eventNode().id();
 
                 synchronized (lock) {
                     if (lastJoiningNode != null && lastJoiningNode.id().equals(nodeId))
```
chesnokoff, line 91:
```suggestion
                ClusterNode eventNode = ((DiscoveryEvent)evt).eventNode();
                synchronized (lock) {
                    if (lastJoiningNode == eventNode)
                        lastJoiningNode = null;
                }
```

Looks like there is a race here. Suppose coordinator node **A** has version 2.18, and node **B** with version 2.19 tries to join.

On the first attempt, node **B** passes `RollingUpgradeProcessor#validateNode`, so `lastJoiningNode` points to this attempt. Then it fails in a later validation step, and `EVT_NODE_VALIDATION_FAILED` is queued asynchronously. After that, node **B** immediately retries with the same node ID and passes `RollingUpgradeProcessor#validateNode` again, so `lastJoiningNode` points to the second attempt.

After that, the old `EVT_NODE_VALIDATION_FAILED` from the first attempt can be delivered. Since we compare only by node ID, this old event clears `lastJoiningNode` for the second attempt. If `disable` is called before the second attempt is added to the `ring`, it can succeed. As a result, rolling upgrade can be disabled while a 2.19 node is still joining a 2.18 cluster.

We should match the exact join attempt, not only the node ID.

chesnokoff:
Small typo in title: Refactord -> Refactor

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
petrov-mg: >  After that, node B immediately retries with the same node ID and passes

How is this possible? Even client nodes generate a new identifier when reconnecting to the cluster. The same for server nodes.

chesnokoff: As I understand it, node B may wait for `networkTimeout` milliseconds. If it does not receive the validation failure response during this time, it can send the join request again with the same node ID (Logic is in `ServerImpl.java:1263`) WDYT?

petrov-mg: > Then it fails in a later validation step
> it can send the join request again with the same node ID
> As a result, rolling upgrade can be disabled while a 2.19 node is still joining a 2.18 cluster

It does not look like a problem - the joining node will fail validation again and will not join the cluster.

Anyway I don't think that it is good idea to get around some specific Discovery implementation quirks in RollingUpgradeProcessor.

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
```diff
@@ -247,22 +237,13 @@ public void disable() throws IgniteCheckedException {
         if (rollUpVers == null)
             return;
 
-        IgnitePair<IgniteProductVersion> minMaxVerPair = ring.minMaxNodeVersions();
+        IgnitePair<IgniteProductVersion> minMaxVerPair = resolveMinMaxNodeVersions();
```
chesnokoff: Probably we should move `resolveMinMaxNodeVersions` under the lock. Otherwise, in theory, a joining node can be added to the discovery ring after `resolveMinMaxNodeVersions` takes its snapshot but before we check `lastJoiningNode`. If the corresponding discovery event clears `lastJoiningNode` in that window, `disable` may use outdated version snapshot and miss the joining node.

The probability looks low, but keeping the ring snapshot and lastJoiningNode check under the same lock makes the invariant explicit

chesnokoff (approved):

modules/core/src/main/java/org/apache/ignite/internal/processors/rollingupgrade/RollingUpgradeProcessor.java
petrov-mg: Done.

chesnokoff (approved):

sergey-chugunov-1985 (approved):

PR #13220, opened 2026-06-08
[MINOR] Compile time check for Message default constructor added

nizhikov:
`AbstractMarshallableMessageFactoryProvider` requires default (no-arg) constructor for all classes implementing `Message` interface. 

This PR moves check for default constructor to compile time.


modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
```diff
@@ -146,6 +151,34 @@ public class MessageProcessor extends AbstractProcessor {
...
+
+            boolean isDfltConstructor = F.isEmpty(c.getParameters());
+
+            if (isDfltConstructor && !isMarshallableMsg)
+                return true;
+        }
+
+        processingEnv.getMessager().printMessage(
+            Diagnostic.Kind.ERROR,
+            "Class must have default constructor: " + clazz.getQualifiedName(),
```
zstan:
```suggestion
            "A class must have a default constructor: " + clazz.getQualifiedName(),
```

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
```diff
@@ -146,6 +151,34 @@ public class MessageProcessor extends AbstractProcessor {
...
+
+        for (Element el : clazz.getEnclosedElements()) {
+            if (el.getKind() != ElementKind.CONSTRUCTOR)
+                continue;
+
+            ExecutableElement c = (ExecutableElement)el;
+
+            boolean isDfltConstructor = F.isEmpty(c.getParameters());
+
+            if (isDfltConstructor && !isMarshallableMsg)
```
shishkovilja: We don't need check that message is marshallable, because `MarshallableMessage` inheritors should have default constructor too: https://github.com/apache/ignite/blob/938452eb63a7646c9029a16c3e6a52aec2d51012/modules/core/src/main/java/org/apache/ignite/internal/plugin/AbstractMarshallableMessageFactoryProvider.java#L68

Also we need to check that constructor is pulbic, eg.:
```
c.getModifiers().contains(Modifier.PUBLIC);
```

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
```diff
@@ -146,6 +151,34 @@ public class MessageProcessor extends AbstractProcessor {
...
+                return true;
+        }
+
+        processingEnv.getMessager().printMessage(
+            Diagnostic.Kind.ERROR,
+            "A class must have a default constructor: " + clazz.getQualifiedName(),
```
shishkovilja:
```suggestion
            "A class must have a public default constructor: " + clazz.getQualifiedName(),
```

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
nizhikov: Fixed

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
nizhikov: Fixed

shishkovilja (approved):

PR #13238, opened 2026-06-11
IGNITE-28692 Add cache metrics for MDC-related checks

sergey-chugunov-1985:

### Notes


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ValidationOnNodeJoinUtils.java
```diff
@@ -631,6 +637,33 @@ private static void checkMemoryConfiguration(ClusterNode rmt, GridKernalContext
...
+        if (affFunc instanceof RendezvousAffinityFunction) {
+            IgniteBiPredicate<ClusterNode, List<ClusterNode>> filter = ((RendezvousAffinityFunction)affFunc).getAffinityBackupFilter();
+
+            if (filter instanceof ClusterNodeAttributeAffinityBackupFilter attrFilter) {
+                if (!F.asList(attrFilter.getAttributeNames()).contains(ATTR_DATA_CENTER_ID))
+                    return false;
+            }
+
+            if (!(filter instanceof MdcAffinityBackupFilter) && !(filter instanceof ClusterNodeAttributeColocatedBackupFilter))
+                return false;
```
anton-vinogradov: `ClusterNodeAttributeAffinityBackupFilter` places backups on nodes with a **different** value of the attribute, so with `DATA_CENTER_ID` it forces the backup into another DC — i.e. it *is* MDC-safe. But it's neither `MdcAffinityBackupFilter` nor `ClusterNodeAttributeColocatedBackupFilter`, so this `if` returns `false` for it. By the time we reach here such a filter is already guaranteed to contain `DATA_CENTER_ID` (otherwise we'd have returned at the check above), so it's safe to accept:

```suggestion
            if (!(filter instanceof MdcAffinityBackupFilter)
                && !(filter instanceof ClusterNodeAttributeColocatedBackupFilter)
                && !(filter instanceof ClusterNodeAttributeAffinityBackupFilter))
                return false;
```

This case isn't covered today: `CACHE_WITH_MDC_SAFE_ATTRIBUTE_FILTER` uses exactly this filter, but `MdcCacheMetricsTest` only checks its *distribution* metric, never `IsCacheAffinityConfigurationMdcSafe` — worth adding that assertion.

Minor follow-up for the multi-attribute case: `getAttributeNames().contains(DATA_CENTER_ID)` is necessary but not sufficient — with e.g. `[DATA_CENTER_ID, RACK]` the AND-semantics still allow a backup in the same DC but a different rack.

anton-vinogradov (commented):
Reviewed against IGNITE-28692. The two metrics are a reasonable direction, but a few things are worth addressing before merge — including one logic bug that the current tests don't catch and a per-exchange overhead that also hits non-MDC clusters. Details inline.

Higher level:

1. The JIRA lists three deliverables — a warning on cache start **plus** the two metrics. The warning-on-cache-start isn't part of this PR. Is it intentionally split out? It's arguably the most visible signal for a user who simply forgot the cache-level configuration; a passive boolean metric is unlikely to reach exactly that user.

2. `IsCacheAffinityConfigurationMdcSafe` is a heuristic over filter types, so it can't be a real guarantee in either direction: `ClusterNodeAttributeColocatedBackupFilter` is only MDC-safe when the cells are actually stretched across DCs (a deployment property, invisible from the filter), and a non-Rendezvous custom affinity is currently always reported as safe. Might be worth reflecting that in the metric description / Javadoc so `Safe` isn't read as a guarantee.

3. For the distribution metric, a boolean loses diagnostic value — a count of partitions without full DC coverage would also serve as a severity signal. Just a thought, not blocking.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -2594,6 +2601,35 @@ private void updateDurationHistogram(long duration) {
...
+    /**
+     * Updates metric for a partition distribution across data centers for a given cache group.
+     *
+     * @param grp Cache group the metric should be recalculated for.
+     * @param assignment New assignment for the cache group.
+     */
+    private void updateMdcMetrics(CacheGroupContext grp, AffinityAssignment assignment) {
+        BaselineTopology top = cctx.discovery().discoCache().state().baselineTopology();
+        if (top != null) {
+            int numberOfDataCenters = top.numberOfDatacenters();
```
anton-vinogradov: `numberOfDatacenters()` returns `-1` when DCs aren't configured, but that isn't handled here: for a non-MDC cluster we still iterate every partition of every non-replicated group on the exchange thread (allocating a stream per partition) and then discard the result — on every PME. An early return honours the `-1` contract and skips the work for non-MDC clusters; it also avoids leaving the metric at its default `true` when the DC count is unknown:

```suggestion
            int numberOfDataCenters = top.numberOfDatacenters();

            if (numberOfDataCenters < 2)
                return;
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheMetricsImpl.java
```diff
@@ -246,6 +246,12 @@ public class CacheMetricsImpl implements CacheMetrics {
     /** Conflict resolver merged entries count. */
     private LongAdderMetric rslvrMergedCnt;
 
+    /** */
+    private Boolean affCfgMdcSafe;
+
+    /** */
+    private Boolean mdcSafePartDistrib;
```
anton-vinogradov: These are written from the exchange thread (`setMdcSafePartitionDistribution`) and read by metric exporters / JMX from other threads. Without `volatile` there's no happens-before, so readers may observe a stale value — more likely to actually surface on weakly-ordered (ARM) CPUs. The other metric fields here are thread-safe (`LongAdderMetric`); these should be too:

```suggestion
    private volatile Boolean affCfgMdcSafe;

    /** */
    private volatile Boolean mdcSafePartDistrib;
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/BaselineTopology.java
```diff
@@ -274,6 +276,20 @@ public Map<String, Object> attributes(Object consId) {
...
+     *
+     * @return Number of datacenters presented in the baseline or {@code -1} if unknown.
+     */
+    public int numberOfDatacenters() {
+        Collection<Map<String, Object>> allNodesAttrs = nodeMap.values();
+
+        if (!allNodesAttrs.isEmpty() && allNodesAttrs.iterator().next().get(ATTR_DATA_CENTER_ID) != null)
+            return (int)allNodesAttrs.stream().map(m -> m.get(ATTR_DATA_CENTER_ID)).distinct().count();
+
+        return -1;
```
anton-vinogradov, line 290: Inspecting only the first node's attributes makes this return `-1` whenever the *first* baseline node happens to lack the DC attribute — even if every other node has it. That's exactly the partial-misconfiguration case this feature is meant to catch, and it would silently mark the whole cluster as safe. Counting across all nodes (ignoring nulls) is more robust:

```suggestion
        long dcs = nodeMap.values().stream()
            .map(m -> m.get(ATTR_DATA_CENTER_ID))
            .filter(dc -> dc != null)
            .distinct()
            .count();

        return dcs == 0 ? -1 : (int)dcs;
```

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/MdcCacheMetricsTest.java
```diff
@@ -0,0 +1,506 @@
...
+            PARTITION_DISTRIBUTION_SAFE_METRIC_NAME);
+        BooleanMetric cacheWithMdcSafeAttrFilterDistributionSafeMetric = findMetricForCache(
+            grid(1),
+            CACHE_WITH_MDC_SAFE_ATTRIBUTE_FILTER,
+            PARTITION_DISTRIBUTION_SAFE_METRIC_NAME
+        );
+
+        assertNotNull(cacheWithMdcFilterDistributionSafeMetric);
+        assertNotNull(cacheWithColocatedFilterDistributionSafeMetric);
+        assertNotNull(cacheWithColocatedFilterDistributionSafeMetric);
```
anton-vinogradov: Copy-paste: this repeats the `colocated` assertion. It should null-check the third metric (currently `cacheWithMdcSafeAttrFilterDistributionSafeMetric` is never asserted non-null):

```suggestion
        assertNotNull(cacheWithMdcSafeAttrFilterDistributionSafeMetric);
```

anton-vinogradov (commented):
_Accidental double-submit — please disregard this review; see my other review on this PR (the one with the inline comments)._

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/MdcCacheMetricsTest.java
sergey-chugunov-1985: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/BaselineTopology.java
sergey-chugunov-1985: I don't consider this comment as valid because we have a validation at node join moment which guarantees that either all server nodes in the cluster have DC_ID attribute configured or none of them.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheMetricsImpl.java
sergey-chugunov-1985: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
sergey-chugunov-1985: This code was moved to CacheMetricsImpl itself, I took into account this idea there.

NSAmelchev (approved):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ValidationOnNodeJoinUtils.java
sergey-chugunov-1985: Accepted.

PR #13241, opened 2026-06-15
IGNITE-28737 Calcite. Cache with keep binary context need to be correctly handled through CacheInterceptor

zstan:


modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/KeepBinaryIntegrationTest.java
```diff
@@ -156,6 +156,53 @@ public void testDynamicParameters() {
...
+        };
+
+        if (sqlTxMode == SqlTransactionMode.NONE)
+            checker.get();
+        else
+            txAction(client, checker);
+    }
+
+    /** */
+    private void runQuery(int begin, int end, IgniteCache<?, ?> cache) {
```
alex-plekhanov: `runQuery` sounds like it executes only one select query. Maybe something like `checkDml`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/atomic/GridDhtAtomicCache.java
```diff
@@ -1352,7 +1355,7 @@ private IgniteInternalFuture removeAllAsync0(
 
         Collection<GridCacheVersion> drVers = null;
 
-        if (opCtx != null && keys != null && opCtx.hasDataCenterId()) {
+        if (opCtx != null && keys != null && opCtx.dataCenterId() != null) {
```
alex-plekhanov: Redundant change

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
```diff
@@ -64,6 +64,9 @@ public class CacheOperationContext implements Serializable {
     /** Application attributes. */
     private final Map<String, String> appAttrs;
 
+    /** Handle binary in interceptor operation flag. */
+    private final boolean handleBinaryInInterceptor;
```
alex-plekhanov: Maybe it's better to implement it with default value == keepBinary (when handleBinaryInInterceptor is not explicetly set)? For example, now I see only one place where this flag is set to true when keepBinary is set to true, but there are other places in code, where keepBinary flag changes. In other places maybe it's worth to set handleBinaryInInterceptor to true too.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/KeepBinaryIntegrationTest.java
zstan: done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/atomic/GridDhtAtomicCache.java
zstan: miss it, when change this PR into 2 parts .. fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
zstan: ok, done

alex-plekhanov (approved):

PR #13242, opened 2026-06-15
IGNITE-28773 CacheOperationContext refactoring

zstan:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
```diff
@@ -151,214 +131,291 @@ public CacheOperationContext keepBinary() {
...
+        /** Skip store. */
+        @GridToStringInclude
+        private boolean skipStore;
+
+        /** Skip read through. */
+        @GridToStringInclude
+        private boolean skipReadThrough;
+
+        /** No retries flag. */
+        @GridToStringInclude
```
alex-plekhanov: Why only these three fields marked as GridToStringInclude? AFAIK like this annotation is redundant for booleans and we can remove it (and from CacheOperationContext too). Let's check it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
```diff
@@ -108,38 +100,26 @@ public CacheOperationContext(
...
-            true,
-            expiryPlc,
-            noRetries,
-            dataCenterId,
-            recovery,
-            readRepairStrategy,
-            appAttrs);
+    /** Context with keepBinary flag. */
+    public CacheOperationContext withKeepBinary() {
+        return builder(this).keepBinary(true).build();
```
alex-plekhanov, line 127: Check for each method:
```
if (keepBinary)
  return this;
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -466,59 +466,53 @@ public void active(boolean active) {
...
-                null,
-                false,
-                null,
-                null);
+        if (opCtx == null)
+            opCtx = CacheOperationContext.builder().skipReadThrough(true).build();
+        else {
+            if (!opCtx.skipReadThrough())
+                opCtx = opCtx.withSkipReadThrough();
         }
```
alex-plekhanov: Maybe replace this boilerplate code in cache adapters/cache proxies with something like:
```
opCtx = CacheOperationContext.of(opCtx).withSkipReadThrough();
```
?

And implement `CacheOperationContext.of` like:
```
return opCtx == null ? INSTANCE : opCtx;
```
Count of allocated objects will be the same (one builder and one context object). Perhaps builder can be marked as private after such a change.

zstan: good catch ! done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
zstan: done except expiration for same behavior as in GridCacheAdapter#withExpiryPolicy, really don\`t know why equality is not checked there.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
zstan: you are right, remove it

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
```diff
@@ -32,68 +31,52 @@
...
-    @GridToStringInclude
     private final boolean noRetries;
 
-    /** */
+    /** Recovery flag. */
     private final boolean recovery;
 
-    /** Read-repair strategy. */
-    private final ReadRepairStrategy readRepairStrategy;
+    /** Read-repair strategy.*/
```
alex-plekhanov: Space before end of the comment

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
```diff
@@ -183,223 +136,311 @@ public CacheOperationContext withKeepBinaryInInterceptor() {
...
-            expiryPlc,
-            noRetries,
-            dataCenterId,
-            recovery,
-            readRepairStrategy,
-            appAttrs,
-            keepBinaryInInterceptor);
+    /** Context with read repair strategy. */
+    public CacheOperationContext withReadRepairStrategy(ReadRepairStrategy strategy) {
+        if (readRepairStrategy() != null && readRepairStrategy() == strategy)
```
alex-plekhanov: Condition `readRepairStrategy() != null` is redundant

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProxyImpl.java
```diff
@@ -252,26 +251,22 @@ public IgniteInternalCache<K, V> delegate() {
...
-            if (opCtx != null && opCtx.skipStore() == skipStore)
-                return this;
+            if (opCtx != null) {
+                if (opCtx.skipStore())
+                    return this;
+                else
+                    newOpCtx = opCtx.withSkipStore();
+            }
+            else
+                newOpCtx = CacheOperationContext.builder().skipStore(true).build();
```
alex-plekhanov:
```
newOpCtx = CacheOperationContext.of(opCtx).withSkipStore();

return newOpCtx == opCtx ? this : new GridCacheProxyImpl<>(ctx, delegate, newOpCtx);
```

zstan: done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
zstan: yep

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
zstan: thanks fixed

alex-plekhanov (approved):

PR #13243, opened 2026-06-15
IGNITE-28770 Added propagation of OperationContext attributes on remote nodes via Discovery

Vladsz83:


modules/commons/src/main/java/org/apache/ignite/internal/thread/context/OperationContextAttribute.java
```diff
@@ -32,7 +32,7 @@ public class OperationContextAttribute<T> {
     static final AtomicInteger ID_GEN = new AtomicInteger();
 
     /** */
-    static final int MAX_ATTR_CNT = Integer.SIZE;
+    public static final int MAX_ATTR_CNT = Integer.SIZE;
```
petrov-mg: We can revert this change.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import org.apache.ignite.IgniteException;
+import org.apache.ignite.internal.DistributedOperationContextAttributesMessage;
+import org.apache.ignite.internal.util.typedef.F;
+import org.apache.ignite.plugin.extensions.communication.Message;
+import org.jetbrains.annotations.Nullable;
+
+/** */
+public class DistributedOperationAttributeManager {
```
petrov-mg: DistributedOperationAttributeManager -> DistributedOperationContextAttributeManager

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
```diff
@@ -0,0 +1,117 @@
...
+    /** */
+    public static final byte MAX_DISTRIBUTED_ATTR_ID = 7;
```
petrov-mg: Let's use MAX_DISTRIBUTED_ATTR_**CNT** = 8;
And change it to package private

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
```diff
@@ -0,0 +1,117 @@
...
+
+    /** */
+    private static final DistributedOperationAttributeManager INSTANCE = new DistributedOperationAttributeManager();
+
+    /** Attributes by their id. */
+    private final Map<Byte, OperationContextAttribute<? extends Message>> attrs = new ConcurrentHashMap<>();
```
petrov-mg: We can replace it with an array and keep it sorted by ID.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
```diff
@@ -0,0 +1,117 @@
...
+    public static DistributedOperationAttributeManager instance() {
+        return INSTANCE;
+    }
+
+    /** */
+    public <T extends Message> OperationContextAttribute<T> createDistributedAttriubte(byte id, @Nullable T initVal) {
+        assert id >= 0;
+
+        if (attrs.size() == OperationContextAttribute.MAX_ATTR_CNT)
+            throw new IgniteException("Maximum number of ributed attributes is exceeded [" + OperationContextAttribute.MAX_ATTR_CNT + "].");
```
petrov-mg: Typo.

Let's use a single style - Error msg [param=val] and remove the dot at the end

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
```diff
@@ -0,0 +1,117 @@
...
+    /** */
+    public static DistributedOperationAttributeManager instance() {
+        return INSTANCE;
+    }
+
+    /** */
+    public <T extends Message> OperationContextAttribute<T> createDistributedAttriubte(byte id, @Nullable T initVal) {
+        assert id >= 0;
+
+        if (attrs.size() == OperationContextAttribute.MAX_ATTR_CNT)
```
petrov-mg: It seems that we should replace it with `id < MAX_DISTRIBUTED_ATTR_CNT`

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
```diff
@@ -0,0 +1,117 @@
...
+            throw new IgniteException("Maximum number of ributed attributes is exceeded [" + OperationContextAttribute.MAX_ATTR_CNT + "].");
+
+        attrs.compute(id, (id0, attr0) -> {
```
petrov-mg: We can make the method synchronized and simplify this logic. 

Check if distributed attribute ID is vacant.
Reserve attribute.
Register attribute.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationContextManager.java
```diff
@@ -0,0 +1,117 @@
...
+    /** */
+    public @Nullable DistributedOperationContextAttributesMessage collectDistributedAttributes() {
+        DistributedOperationContextAttributesMessage res = null;
+
+        for (Map.Entry<Byte, OperationContextAttribute<? extends Message>> e : attrs.entrySet()) {
+            OperationContextAttribute<? extends Message> attr = e.getValue();
+
+            Message curVal = OperationContext.get(attr);
+
+            assert attr.initialValue() == null || curVal == null || curVal.getClass().isAssignableFrom(attr.initialValue().getClass());
```
petrov-mg: Frankly, I don't understand the purpose of this assertion.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
```diff
@@ -0,0 +1,117 @@
...
+        DistributedOperationContextAttributesMessage res = null;
+
+        for (Map.Entry<Byte, OperationContextAttribute<? extends Message>> e : attrs.entrySet()) {
+            OperationContextAttribute<? extends Message> attr = e.getValue();
+
+            Message curVal = OperationContext.get(attr);
+
+            assert attr.initialValue() == null || curVal == null || curVal.getClass().isAssignableFrom(attr.initialValue().getClass());
+
+            if (!Objects.equals(curVal, attr.initialValue())) {
```
petrov-mg: We need to use  `==` not equals. The the same way the org.apache.ignite.internal.thread.context.OperationContext.ContextUpdater#set does.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
```diff
@@ -0,0 +1,117 @@
...
+        if (msg == null)
+            return Scope.NOOP_SCOPE;
+
+        assert msg.idBitmask != 0;
+        assert !F.isEmpty(msg.vals);
+        assert msg.vals.size() <= MAX_DISTRIBUTED_ATTR_ID;
+
+        OperationContext.ContextUpdater updater = OperationContext.ContextUpdater.create();
+
+        for (byte valIdx = 0, maskIdx = -1; valIdx < msg.vals.size(); ++valIdx) {
```
petrov-mg: Do we really need to start maskIdx at -1?

modules/core/src/main/java/org/apache/ignite/internal/DistributedOperationContextAttributesMessage.java
```diff
@@ -0,0 +1,38 @@
...
+
+/** Transport for {@link OperationContext} distibuted attributes. */
+public class DistributedOperationContextAttributesMessage implements Message {
+    /** Values of operation context attributes. */
+    @Order(0)
+    public List<Message> vals;
+
+    /** Bitmask of effective attributes ids. */
+    @Order(1)
+    public byte idBitmask;
```
petrov-mg: Bitmask -> bitmap

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryAbstractMessage.java
```diff
@@ -76,6 +77,11 @@ public abstract class TcpDiscoveryAbstractMessage implements Message {
     @Order(4)
     Set<UUID> failedNodes;
 
+    /** Operation context attributes message. */
+    @GridToStringInclude
+    @Order(5)
+    public @Nullable DistributedOperationContextAttributesMessage opCtxMsg;
```
petrov-mg: I would prefer to have getters and setters just like for all other variables.

modules/core/src/main/java/org/apache/ignite/internal/DistributedOperationContextMessage.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.List;
+import org.apache.ignite.internal.thread.context.OperationContext;
+import org.apache.ignite.plugin.extensions.communication.Message;
+
+/** Transport for {@link OperationContext} distibuted attributes. */
+public class DistributedOperationContextAttributesMessage implements Message {
+    /** Values of operation context attributes. */
+    @Order(0)
+    public List<Message> vals;
```
petrov-mg: I would prefer just plain array here.

modules/core/src/main/java/org/apache/ignite/internal/DistributedOperationContextAttributesMessage.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal;
+
+import java.util.List;
+import org.apache.ignite.internal.thread.context.OperationContext;
+import org.apache.ignite.plugin.extensions.communication.Message;
+
+/** Transport for {@link OperationContext} distibuted attributes. */
```
petrov-mg: Typo.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
Vladsz83: I suggest a sorted map. I don't like structures like List<T2<Byte, OperationContextAttribute<? extends Message>>>

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
Vladsz83: Why?

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
Vladsz83: We use instance() everywhere. I don't think that just one call should be without instance()

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationContextManager.java
Vladsz83: We check that a value of other type wouldn't set.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryAbstractMessage.java
Vladsz83: We do not use getters and setters in the messages

modules/core/src/main/java/org/apache/ignite/internal/DistributedOperationContextMessage.java
Vladsz83: I've reverted plain array. It cannot change its size while we do not know actual number of effective context attributes. Or we have to do another re-copy of it with actual size. It looks not better to me.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationContextManager.java
```diff
@@ -0,0 +1,117 @@
...
+    /** Attributes by their id. */
+    private final Map<Byte, OperationContextAttribute<? extends Message>> attrs = new ConcurrentHashMap<>();
+
+    /** */
+    public static DistributedOperationAttributeManager instance() {
+        return INSTANCE;
+    }
+
+    /** */
+    public <T extends Message> OperationContextAttribute<T> createDistributedAttriubte(byte id, @Nullable T initVal) {
```
petrov-mg: Typo.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
petrov-mg: ок.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationContextManager.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.ignite.IgniteException;
+import org.apache.ignite.internal.DistributedOperationContextAttributesMessage;
+import org.apache.ignite.internal.util.typedef.F;
+import org.apache.ignite.plugin.extensions.communication.Message;
+import org.jetbrains.annotations.Nullable;
+
+/** */
+public class DistributedOperationContextAttributeManager {
+    /** */
+    static final byte MAX_DISTRIBUTED_ATTR_CNT = 7;
```
petrov-mg: 7 -> 8 ?

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationContextManager.java
petrov-mg: We already have this guarantee - OperationContextAttribute is a template class. OperationContext#set is a template method.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
petrov-mg: So we declare that the distributed ID is 0 <= ID < MAX_DISTRIBUTED_ATTR_CNT

And the map ensures that attributes with duplicate identifiers are not created.

modules/core/src/main/java/org/apache/ignite/internal/DistributedOperationContextMessage.java
petrov-mg: > re-copy of it with actual size.

This doesn't seem to be a problem, since the List will be created on the stack for the duration of the collectDistributedAttributes method.

`DistributedOperationAttributesMessage` can be attached to ALL communication and discovery messages. 
But serializing a List currently requires one extra byte for the collection type. 
So i think it worth making `DistributedOperationAttributesMessage` as compact as possible.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
petrov-mg: The current implementation allows creating a distributed attribute with an index of '100', using it, and then getting a NullPointerException during mesage deserialization. This is unacceptable.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationContextManager.java
petrov-mg: I emphasize again that the asserted guarantee is already provided by OperationContext and is not necessary.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationContextManager.java
petrov-mg: It still exists.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationContextManager.java
petrov-mg: Event better - `Byte.MAX_VALUE`

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ClientImpl.java
```diff
@@ -2001,7 +2005,11 @@ else if (discoMsg instanceof TcpDiscoveryCheckFailedMessage)
                             }
                         }
 
-                        processDiscoveryMessage((TcpDiscoveryAbstractMessage)msg);
+                        TcpDiscoveryAbstractMessage msg0 = (TcpDiscoveryAbstractMessage)msg;
+
+                        try (Scope ignored = DistributedOperationContextManager.instance().restoreDistributedAttributes(msg0.opCtxMsg)) {
+                            processDiscoveryMessage(msg0);
```
petrov-mg: It looks like we need to wrap not just the `processDiscoveryMessage` call, but the entire processing of the message from the moment it is polled from the queue.

modules/core/src/main/java/org/apache/ignite/internal/DistributedOperationContextMessage.java
Vladsz83: One extra transport byte at kilobyte-like messages vs one new extra data copy, new operator, extra GC load. I'm not sure it is better.

petrov-mg: > extra GC load

This doesn't seem to be the case unless the created list goes outside the scope of the method. Even if this is not, short-lived objects are extremely optimized in modern JVMs.

But keeping tons of messages with unused arrays in discovery and communication queues seems to be a more realistic problem.

modules/core/src/main/java/org/apache/ignite/internal/DistributedOperationContextMessage.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.ignite.internal.thread.context.OperationContext;
+import org.apache.ignite.plugin.extensions.communication.Message;
+
+/** Transport for {@link OperationContext} distributed attributes. */
+public class DistributedOperationContextMessage implements Message {
+    /** Values of operation context attributes. */
+    @Order(0)
+    public List<Message> vals;
+
+    /** Bitmask of effective attributes ids. */
```
petrov-mg: Bitmask -> Bitmap.

petrov-mg:
It's also worth adding meaningful comments describing the purpose of DistributedOperationContextManager and its methods.

What is DistributedOperationContextMessage and why was it decided to use a bitmap.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationAttributeManager.java
petrov-mg: I'm still observing the mentioned issue. It looks like the patch requires more thorough testing.

Vladsz83: There are an assert now

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationContextManager.java
```diff
@@ -0,0 +1,114 @@
...
+    public static DistributedOperationContextManager instance() {
+        return INSTANCE;
+    }
+
+    /** */
+    public <T extends Message> OperationContextAttribute<T> createDistributedAttribute(byte id, @Nullable T initVal) {
+        assert id >= 0 && id < MAX_DISTRIBUTED_ATTR_CNT : "Invalid distributed attributed id [id=" + id + "].";
+
+        if (attrs.size() == MAX_DISTRIBUTED_ATTR_CNT)
+            throw new IgniteException("Maximum number of distributed attributes is exceeded [max=" + MAX_DISTRIBUTED_ATTR_CNT + "].");
```
petrov-mg: Let's remove dot from the end of the error message. 

It seems we can safely remove this error because in this case `Duplicated distributed attribute id` will be raised.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationContextManager.java
```diff
@@ -0,0 +1,114 @@
...
+
+        return (OperationContextAttribute<T>)attrs.compute(id, (id0, attr0) -> {
+            if (attr0 != null)
+                throw new IgniteException("Duplicated distributed attribute id [id=" + id + "].");
```
petrov-mg: Let's remove dot from the end of the error message.

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationContextManager.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentSkipListMap;
+import org.apache.ignite.IgniteException;
+import org.apache.ignite.internal.DistributedOperationContextMessage;
+import org.apache.ignite.internal.util.typedef.F;
+import org.apache.ignite.plugin.extensions.communication.Message;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A manager of {@link OperationContextAttribute} which are required to be propagated through a cluster.
```
petrov-mg:
```
/**
 * Provides the ability to manage {@link OperationContext} attributes in a distributed manner.
 *
 * <p>This mechanism is primarily used to propagate {@link OperationContext} state across the cluster by
 * capturing it before a message is sent, transferring it together with the message, and restoring it on
 * the receiving node before message processing begins.</p>
 *
 * <p>The implementation relies on a mapping between a distributed identifier and an
 * {@link OperationContextAttribute} instance that is consistent across all cluster nodes.</p>
 *
 * <p>To enable propagation of an {@link OperationContextAttribute} value across cluster nodes, the
 * attribute must be created using the {@link #createDistributedAttribute(byte, Message)} method.
 *
 * <p> Note, that the maximum number of distributed attribute instances that can be created is currently limited to
 * {@link #MAX_DISTRIBUTED_ATTR_CNT} for implementation reasons.</p>
 *
 * @see OperationContext
 * @see DistributedOperationContextMessage
 */
```

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationContextManager.java
```diff
@@ -0,0 +1,128 @@
...
+
+    /** Registered distributed attributes by their cluster-wide id. */
+    private final Map<Byte, OperationContextAttribute<? extends Message>> attrs = new ConcurrentSkipListMap<>();
+
+    /** */
+    public static DistributedOperationContextManager instance() {
+        return INSTANCE;
+    }
+
+    /**
```
petrov-mg, line 63:
```
/**
     * Creates a new {@link OperationContext} attribute with the specified distributed ID and initial value.
     *
     * <p>The distributed ID is used to consistently identify the attribute across all nodes in the cluster.
     * It must be unique, and its value must be in the range from {@code 0} (inclusive) to {@code Byte.SIZE} (exclusive).</p>
     *
     * <p>The value of the created attribute is automatically captured and propagated between cluster nodes
     * during message transmission.</p>
     *
     * @see OperationContextAttribute#newInstance(Object)
     */
```

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationContextManager.java
```diff
@@ -0,0 +1,128 @@
...
+        return (OperationContextAttribute<T>)attrs.compute(id, (id0, attr0) -> {
+            if (attr0 != null)
+                throw new IgniteException("Duplicated distributed attribute id [id=" + id + ']');
+
+            return OperationContextAttribute.newInstance(initVal);
+        });
+    }
+
+    /**
+     * Requests current {@link OperationContext} for its effective attributes and collects ones which are also registered
```
petrov-mg:
```
 /**
     * Collects the values of all distributed {@link OperationContextAttribute}s registered by this manager
     * in a format suitable for transmission between cluster nodes.
     *
     * @see OperationContext#get(OperationContextAttribute)
     */
```

modules/core/src/main/java/org/apache/ignite/internal/thread/context/DistributedOperationContextManager.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+
+        if (res != null)
+            res.vals = vals.toArray(vals.toArray(new Message[vals.size()]));
+
+        return res;
+    }
+
+    /** Sets the received distributed operation context attributes (if any) into current {@link OperationContext}. */
```
petrov-mg:
```
/** Restores distributed {@link OperationContextAttribute} values received from a remote node. */
```

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ClientImpl.java
```diff
@@ -1757,276 +1761,290 @@ private MessageWorker(IgniteLogger log) {
...
-                    if (msg instanceof JoinTimeout) {
-                        int joinCnt0 = ((JoinTimeout)msg).joinCnt;
-
-                        if (joinCnt == joinCnt0) {
-                            if (state == STARTING) {
-                                joinError(new IgniteSpiException("Join process timed out, did not receive response for " +
-                                    "join request (consider increasing 'joinTimeout' configuration property) " +
-                                    "[joinTimeout=" + spi.joinTimeout + ", sock=" + currSock + ']'));
-
+                    if (msg instanceof TcpDiscoveryAbstractMessage msg0 && msg0.opCtxMsg != null) {
```
petrov-mg: Let's just leave the method as is, but wrap all of its code in a try-with-resource block.

Vladsz83: I cannot wrap entire code because there would not be a message yet.

petrov-mg: Obviously, that's not what I meant. Let's wrap the code related to processing of the message from the moment it is polled from the queue, but not move anything to other new methods. 

In the current state of the code, moving the message handling logic into a separate method seems awkward. It's better to refactor this later in a separate ticket.

Vladsz83: We've already moved message processing in a dedicated methods in same manner at other places. Message is a Object here, not even a Message. I need to check its type first. An `if` is required. That is why I introduce a method or did the processing in the part of code before.

petrov-mg (approved):

PR #13263, opened 2026-06-22
IGNITE-28806 Fix dynamic cache group affinity init during first local join

oleg-vlsk:
… join

### Notes


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheAffinitySharedManager.java
```diff
@@ -2079,6 +2090,40 @@ public IgniteInternalFuture<?> initCoordinatorCaches(
...
+        if (!firstLocalJoinExchange(fut))
+            return false;
+
+        AffinityTopologyVersion grpStartTopVer = desc.startTopologyVersion();
+
+        if (grpStartTopVer == null)
+            return false;
+
+        return grpStartTopVer.after(fut.initialVersion());
+    }
```
chesnokoff: Can we make smth like
```java
/** */
private boolean skipNotStartedDynamicGroup(
    GridDhtPartitionsExchangeFuture fut,
    CacheGroupDescriptor desc,
    @Nullable CacheGroupContext grp,
    boolean newAff
) {
    if (grp != null || newAff)
        return false;

    AffinityTopologyVersion grpStartTopVer = desc.startTopologyVersion();

    return grpStartTopVer != null && grpStartTopVer.after(fut.initialVersion());
}
```

To make this check based on the cache group start version instead of the first join?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/DynamicCacheStartDuringRestartExchangeTest.java
```diff
@@ -0,0 +1,349 @@
...
+    private final Map<Integer, ClientCacheConfiguration> cacheConfs = new HashMap<>();
+
+    /** */
+    private Thread.UncaughtExceptionHandler oldUncaughtHnd;
+
+    /** */
+    private IgniteInternalFuture<?> startFut;
+
+    /** */
+    private IgniteInternalFuture<?> clientFut;
```
chesnokoff: I think this test is too heavy for this fix. It looks more like a stress/probabilistic test than a deterministic regression test. Even if the bug appears again, this test can still pass if the race does not happen during the run

I suggest reverting this test in a separate commit, in case another reviewer asks to restore it later. Instead, we can add a smaller happy-path test to check that we do not skip extra cache groups by mistake

oleg-vlsk: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheAffinitySharedManager.java
oleg-vlsk: Yes, makes sense. If we remove the first-join-check, the fix will be broader. Corrections added.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheAffinitySharedManager.java
```diff
@@ -1955,16 +1955,27 @@ public IgniteInternalFuture<?> initCoordinatorCaches(
...
             @Override public void applyx(CacheGroupDescriptor desc) throws IgniteCheckedException {
+                int grpId = desc.groupId();
+
+                CacheGroupContext grp = cctx.cache().cacheGroup(grpId);
+
+                if (skipNotStartedDynamicGroup(fut, desc, grp, newAff)) {
+                    if (log.isDebugEnabled()) {
+                        log.debug("Skip coordinator affinity initialization for cache group started after" +
+                            " current exchange [grp=" + desc.cacheOrGroupName() +
+                            ", grpId=" + desc.groupId() + ", curTopVer=" + topVer +
```
chesnokoff:
```suggestion
                            ", grpId=" + grpId + ", curTopVer=" + topVer +
```
desc.groupId() was already calculated at line
```java
int grpId = desc.groupId();
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheAffinitySharedManager.java
```diff
@@ -2079,6 +2090,21 @@ public IgniteInternalFuture<?> initCoordinatorCaches(
         return null;
     }
 
+    /** */
```
chesnokoff: Could you add a short doc to skipNotStartedDynamicGroup? Explain that the group has not started yet at this exchange, so it will be initialized later by its own exchange and that is why we can safely skip it here

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/DynamicCacheStartExchangeTest.java
```diff
@@ -0,0 +1,135 @@
...
+    /** {@inheritDoc} */
+    @Override protected void afterTest() throws Exception {
+        stopAllGrids(true);
+
+        super.afterTest();
+    }
+
+    /** */
+    @Test
+    public void testThinClientDynamicCacheStartCreatesCacheGroups() throws Exception {
```
chesnokoff: I'm not sure that this test covers the fix. `skipNotStartedDynamicGroup` is called only from `initCoordinatorCaches`, and this method runs only on a local join

Here all 3 servers are started before dynamic caches are created, so the new skip check code is not called

Could you start one more node after cache creation and check groups and caches on this new node too? Then the test will really check that we do not skip valid groups

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheAffinitySharedManager.java
oleg-vlsk: Javadoc added.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheAffinitySharedManager.java
oleg-vlsk: Done.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/DynamicCacheStartExchangeTest.java
oleg-vlsk: I thoroughly debugged the current version of the test and found out that it does not cover the skip check path at all:
- `CacheAffinitySharedManager#initCoordinatorCaches` is not called during the dynamic cache creation;
-  neither is it called during the proposed additional node startup.

It is only called during the initial cluster startup on the coordinator node upon static cache initialization. That’s why it was necessary to repeatedly shut down the cluster and start it up again in the reproducer and the initial stress test.

Based on that, a smaller happy-path test with several static cache groups seems more relevant: it verifies that cache groups which belong to the current coordinator initialization exchange are not skipped and are properly started, even though during their initialization `CacheAffinitySharedManager#skipNotStartedDynamicGroup` is called. Therefore I changed the happy-path test to `CacheAffinityCoordinatorInitTest#testStaticCacheGroupsOnCoordinatorStart`.

This still does not cover the exact “future dynamic group descriptor” scenario. Covering that would require either an internal hook, or the initial stress test.

chesnokoff (approved):
LGTM

One note for next reviewer: initial patch had a reproducer test
https://github.com/apache/ignite/pull/13263/changes/53bb6f938e42572e58b70ba65cb8937679c51c9e

But that test was more like a stress test. It was heavy and probabilistic, so it could make TC runs slower and still miss the bug in some runs

I think the current deterministic happy-path test is an acceptable compromise. It checks that we do not skip cache groups that should be started in the current exchange

But the final decision is up to committer

oleg-vlsk:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13263)** verdict · RunAll build [9201449](https://ci2.ignite.apache.org/build/9201449) · 147 suites ran, 0 reused

⚠️ **7 broken suite(s)** (no reliable run):
- Platform C++ CMake (Linux): Number of tests 595 is 43% less than 1046 in build #42424
- Platform C++ CMake (Linux Clang): Number of tests 595 is 43% less than 1046 in build #42405
- Platform .NET (Windows): execution timeout · non-zero exit code
- Platform C++ CMake (Win x64 | Release): Number of tests 595 is 43% less than 1046 in build #18140
- Cache (Failover) 4: execution timeout · non-zero exit code · Number of tests 17 is 55% less than 38 in build #159
- Platform .NET (Windows) 2: execution timeout · non-zero exit code · 3 failed tests detected
- Platform .NET (Windows) 3: execution timeout · non-zero exit code

✅ No test blockers otherwise; 14 pre-existing/flaky filtered out.

sergey-chugunov-1985 (approved):

PR #13265, opened 2026-06-22
IGNITE-28805 Fix flaky ContinuousQueryMarshallerTest#testRemoteFilterFactoryServer

chesnokoff:

### Notes

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/ContinuousQueryMarshallerTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/ContinuousQueryMarshallerTest.java
@@ -83,6 +83,8 @@ public class ContinuousQueryMarshallerTest extends GridCommonAbstractTest {
 
         final Ignite node2 = "client".equals(node2Name) ? startClientGrid(node2Name) : startGrid(node2Name);
 
+        awaitPartitionMapExchange();
+
         final ContinuousQuery<Integer, MarshallerCheckingEntry> qry = new ContinuousQuery<>();
 
         ScanQuery<Integer, MarshallerCheckingEntry> scanQry = new ScanQuery<>(new IgniteBiPredicate<Integer, MarshallerCheckingEntry>() {


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/ContinuousQueryMarshallerTest.java
```diff
@@ -83,6 +83,8 @@ private void check(String node1Name, String node2Name) throws Exception {
 
         final Ignite node2 = "client".equals(node2Name) ? startClientGrid(node2Name) : startGrid(node2Name);
 
+        awaitPartitionMapExchange();
```
chesnokoff, line 86: Test started `server2` and immediately executed CQ with an initial scan query. At this moment partition map exchange could still be in progress and the initial scan sometimes did not see all existing entries.

In failed run the initial query returned only 2 entries instead of 5. Then all update events arrived, but the latch was still not fully counted down and the test failed with `AssertionError: 4`

The fix is to wait for pme after starting `server2` and before executing the continuous query

anton-vinogradov:
LGTM — root cause and fix are both correct.

A couple of notes for the record:

**Why only the server variant flaked:** both `@Test`s share `check()`, but starting a *second server* triggers rebalancing/PME while starting a *client* moves no data. So the initial `ScanQuery` racing an in-progress PME cleanly explains why `testRemoteFilterFactoryServer` misses pre-loaded entries and the client variant doesn't. `awaitPartitionMapExchange()` blocks on a deterministic condition (no `MOVING` partitions), so this is a real fix rather than a sleep-style band-aid, and it matches the idiom already used by ~10 other tests in this package. Keeping the wait in the shared `check()` also hardens the client variant at no cost. 👍

**Latch math checks out:** 5 initial (even keys 0,2,4,6,8) + 10 updates (keys 10–19; `DummyEventFilterFactory` accepts all) = 15.

**Minor:** the numbers in the explanation don't quite reconcile — a 2-entry initial scan leaves the latch at 3 (15−2−10), but the message was `AssertionError: 4`, which implies either a 1-entry initial scan or that one *update* event was also missed in that run. `awaitPartitionMapExchange()` covers both, just worth confirming the missed-event path is the initial scan only and not also CQ registration.

**Residual:** the 5s `latch.await` for async CQ delivery is the only remaining timing dependency — standard and fine, flagging for completeness.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/ContinuousQueryMarshallerTest.java
anton-vinogradov, line 86: Optional nit: a one-line note on *why* the wait is here would protect it from being removed during a future "cleanup", since the call looks redundant out of context. Up to you — most of the other `awaitPartitionMapExchange()` calls in this package aren't commented either.

```suggestion
        // Wait for rebalance to finish so the initial scan query observes all pre-loaded entries.
        awaitPartitionMapExchange();
```

chesnokoff: I'd prefer not to add this comment. It mostly repeats the method name because`awaitPartitionMapExchange()` already says that we wait for PME

Also, "rebalance" is a bit too narrow here. The main point is that the initial scan should not start while the topology and  partition map is still changing after `server2` joins

anton-vinogradov, line 86: Compromise on the earlier comment thread: keep the WHY (so the call survives a future "cleanup"), while addressing both of your objections — it doesn't restate the method name, and it says "stable topology" instead of the narrower "rebalance".

```suggestion
        // Initial scan must observe all pre-loaded entries, so run it only on a stable topology.
        awaitPartitionMapExchange();
```

Non-blocking — fine to merge without it if you'd rather keep the line bare.

PR #13257, opened 2026-06-19
IGNITE-28800 Add documentation code snippets compilation check

chesnokoff:
Thank you for submitting the pull request to the Apache Ignite.

In order to streamline the review of the contribution 
we ask you to ensure the following steps have been taken:

### The Contribution Checklist
- [ ] There is a single JIRA ticket related to the pull request. 
- [ ] The web-link to the pull request is attached to the JIRA ticket.
- [ ] The JIRA ticket has the _Patch Available_ state.
- [ ] The pull request body describes changes that have been made. 
The description explains _WHAT_ and _WHY_ was made instead of _HOW_.
- [ ] The pull request title is treated as the final commit message. 
The following pattern must be used: `IGNITE-XXXX Change summary` where `XXXX` - number of JIRA issue.
- [ ] A reviewer has been mentioned through the JIRA comments 
(see [the Maintainers list](https://cwiki.apache.org/confluence/display/IGNITE/How+to+Contribute#HowtoContribute-ReviewProcessandMaintainers)) 
- [ ] The pull request has been checked by the Teamcity Bot and 
the `green visa` attached to the JIRA ticket (see tab `PR Check` at [TC.Bot - Instance 1](https://tcbot2.sbt-ignite-dev.ru/prs.html) or [TC.Bot - Instance 2](https://mtcga.gridgain.com/prs.html))

### Notes
- [How to Contribute](https://cwiki.apache.org/confluence/display/IGNITE/How+to+Contribute)
- [Coding abbreviation rules](https://cwiki.apache.org/confluence/display/IGNITE/Abbreviation+Rules)
- [Coding Guidelines](https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines)
- [Apache Ignite Teamcity Bot](https://cwiki.apache.org/confluence/display/IGNITE/Apache+Ignite+Teamcity+Bot)

If you need any help, please email dev@ignite.apache.org or ask anу advice on http://asf.slack.com _#ignite_ channel.

PR #13266, opened 2026-06-22
IGNITE-28799 Fix documentation code snippets compilation

chesnokoff:

### Notes

### DIFF
--- a/docs/_docs/code-snippets/java/pom.xml
+++ b/docs/_docs/code-snippets/java/pom.xml
@@ -80,6 +80,8 @@
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>
+            <!-- ignite-cloud-ext is not published to Maven Central, so use the last released legacy
+                 artifact that still provides TcpDiscoveryCloudIpFinder. -->
             <artifactId>ignite-cloud</artifactId>
             <version>2.13.0</version>
         </dependency>
--- a/docs/_docs/code-snippets/java/pom.xml
+++ b/docs/_docs/code-snippets/java/pom.xml
@@ -80,8 +80,7 @@
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>
-            <!-- ignite-cloud-ext is not published to Maven Central, so use the last released legacy
-                 artifact that still provides TcpDiscoveryCloudIpFinder. -->
+            <!-- ignite-cloud-ext is not published to Maven Central, so use the last released legacy artifact that still provides TcpDiscoveryCloudIpFinder. -->
             <artifactId>ignite-cloud</artifactId>
             <version>2.13.0</version>
         </dependency>
--- a/docs/_docs/code-snippets/java/pom.xml
+++ b/docs/_docs/code-snippets/java/pom.xml
@@ -25,6 +25,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <ignite.version>2.19.0-SNAPSHOT</ignite.version>
+        <ignite.extensions.version>1.0.0</ignite.extensions.version>
     </properties>
     <dependencies>
         <dependency>
@@ -67,11 +68,46 @@
             <artifactId>ignite-zookeeper</artifactId>
             <version>${ignite.version}</version>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>ignite-aws-ext</artifactId>
+            <version>${ignite.extensions.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>ignite-azure-ext</artifactId>
+            <version>${ignite.extensions.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>ignite-cloud</artifactId>
+            <version>2.13.0</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>ignite-gce-ext</artifactId>
+            <version>${ignite.extensions.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>ignite-kubernetes</artifactId>
+            <version>${ignite.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>ignite-zookeeper-ip-finder-ext</artifactId>
+            <version>${ignite.extensions.version}</version>
+        </dependency>
 
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>ignite-opencensus</artifactId>
             <version>${ignite.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.opencensus</groupId>
+            <artifactId>opencensus-exporter-trace-zipkin</artifactId>
+            <version>0.31.1</version>
         </dependency>
 
         <dependency>
--- a/docs/_docs/code-snippets/java/src/main/java/org/apache/ignite/snippets/Snapshots.java
+++ b/docs/_docs/code-snippets/java/src/main/java/org/apache/ignite/snippets/Snapshots.java
@@ -16,7 +16,10 @@
  */
 package org.apache.ignite.snippets;
 
+import java.io.File;
+import java.util.Collections;
 import org.apache.ignite.Ignite;
+import org.apache.ignite.IgniteCache;
 import org.apache.ignite.Ignition;
 import org.apache.ignite.configuration.CacheConfiguration;
 import org.apache.ignite.configuration.IgniteConfiguration;
@@ -28,7 +31,7 @@ public class Snapshots {
         //tag::config[]
         IgniteConfiguration cfg = new IgniteConfiguration();
 
-        File exSnpDir = U.resolveWorkDirectory(U.defaultWorkDirectory(), "ex_snapshots", true);
+        File exSnpDir = new File("work", "ex_snapshots");
 
         cfg.setSnapshotPath(exSnpDir.getAbsolutePath());
         //end::config[]
--- a/docs/_docs/code-snippets/java/src/main/java/org/apache/ignite/snippets/Snapshots.java
+++ b/docs/_docs/code-snippets/java/src/main/java/org/apache/ignite/snippets/Snapshots.java
@@ -31,6 +31,7 @@ public class Snapshots {
         //tag::config[]
         IgniteConfiguration cfg = new IgniteConfiguration();
 
+        // Any directory the Ignite process can write to (example value).
         File exSnpDir = new File("work", "ex_snapshots");
 
         cfg.setSnapshotPath(exSnpDir.getAbsolutePath());
--- a/docs/_docs/code-snippets/java/src/main/java/org/apache/ignite/snippets/WarmUpStrategy.java
+++ b/docs/_docs/code-snippets/java/src/main/java/org/apache/ignite/snippets/WarmUpStrategy.java
@@ -14,70 +14,91 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-//tag::warm-all-regions[]
-IgniteConfiguration cfg = new IgniteConfiguration();
+package org.apache.ignite.snippets;
 
-DataStorageConfiguration storageCfg = new DataStorageConfiguration();
+import org.apache.ignite.configuration.DataRegionConfiguration;
+import org.apache.ignite.configuration.DataStorageConfiguration;
+import org.apache.ignite.configuration.IgniteConfiguration;
+import org.apache.ignite.configuration.LoadAllWarmUpConfiguration;
+import org.apache.ignite.configuration.NoOpWarmUpConfiguration;
 
-//Changing the default warm-up strategy for all data regions
-storageCfg.setDefaultWarmUpConfiguration(new LoadAllWarmUpConfiguration());
+public class WarmUpStrategy {
+    public void warmAllRegions() {
+        //tag::warm-all-regions[]
+        IgniteConfiguration cfg = new IgniteConfiguration();
 
-cfg.setDataStorageConfiguration(storageCfg);
-//end::warm-all-regions[]
-//tag::warm-specific-region[]
-IgniteConfiguration cfg = new IgniteConfiguration();
+        DataStorageConfiguration storageCfg = new DataStorageConfiguration();
 
-DataStorageConfiguration storageCfg = new DataStorageConfiguration();
+        //Changing the default warm-up strategy for all data regions
+        storageCfg.setDefaultWarmUpConfiguration(new LoadAllWarmUpConfiguration());
 
-//Setting another warm-up strategy for a custom data region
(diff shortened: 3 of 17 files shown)


chesnokoff:
Run 
```script
./mvnw -f docs/_docs/code-snippets/java/pom.xml compile -B -V
```
to check compilation

Also try to run checks on https://github.com/apache/ignite/pull/13257 before merge

docs/_docs/code-snippets/java/pom.xml
```diff
@@ -67,12 +68,47 @@
...
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>ignite-azure-ext</artifactId>
+            <version>${ignite.extensions.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>ignite-cloud</artifactId>
+            <version>2.13.0</version>
```
zstan, line 85: is it from extensions ? if so - why ver is hardcoded ?

chesnokoff: Yes, `ignite-cloud` was moved from the main Ignite repository to `ignite-extensions` as `ignite-cloud-ext`, but it looks like `ignite-cloud-ext` was never published to Maven Central. As a result, the only resolvable artifact containing `TcpDiscoveryCloudIpFinder` is the legacy `ignite-cloud:2.13.0`.

chesnokoff: https://central.sonatype.com/search?q=ignite-cloud

chesnokoff: Started discussion on dev list https://lists.apache.org/thread/d214wbsp84gm6sc93woctf3q68dhrsch

anton-vinogradov (commented):
Solid and genuinely needed cleanup. The published snippets currently teach code that does **not** compile — wrong class names (`TcpDiscoveryAzureBlobStorageIpFinder`), an Azure var assigned `new TcpDiscoveryGoogleStorageIpFinder()`, missing semicolons in `JavaThinClient`, undefined `ccfg`/`part`/`PART_CUSTOM_AFFINITY_CACHE_NAME`, a wrong package on `EchoService*`, JUnit 4 `@Test` against a JUnit 5 module — and they even leak internal API (`U.resolveWorkDirectory`, `F.asMap`) into user-facing docs. Replacing those with public API and fixing the `CacheInterceptor`/generics signatures is the right call.

I checked the rendering side: all 95 snippet `include::`s use `indent=0`, so wrapping the former free-floating `WarmUpStrategy` statements into a class + methods does **not** change the rendered output. Good.

Three things to settle before merge:

**1. This fix is throwaway without the CI gate, and the two PRs aren't formally coupled.** IGNITE-28800 (#13257) adds the compile check and edits the *same* `pom.xml`. Without it the snippets rot again exactly as they did here, so the value of this PR decays in weeks. Please land them as a coordinated pair and pick the order explicitly: merging the gate first turns `master` red (this fix isn't in yet); merging this first leaves an unguarded window. Safest is to rebase #13257 on top of this, verify green, and merge back-to-back (resolving the inevitable `pom.xml` conflict).

**2. We're putting the docs build on external ignite-extensions artifacts, one of which doesn't exist.** `ignite-cloud-ext` is not on Maven Central, hence the legacy `ignite-cloud:2.13.0` fallback — good that you opened the [dev-list thread](https://lists.apache.org/thread/d214wbsp84gm6sc93woctf3q68dhrsch). Beyond that one: the gate's red/green now depends on Maven Central availability and on `*-ext:1.0.0` (released independently of core) staying API-compatible with `2.19.0-SNAPSHOT`. Worth confirming `ignite-cloud:2.13.0` actually compiles against 2.19 core, and deciding whether the cloud-discovery snippets should be in scope for the gate at all — otherwise a third-party (un)publish can turn core CI red.

**3. The acceptance criterion isn't demonstrated yet.** The ticket is literally "snippets compile", but there's no green TC visa / successful `mvn -f docs/_docs/code-snippets/java/pom.xml compile` attached, and the Copilot pass bailed on quota. Please attach the green run (ideally via #13257's workflow) before merge.

Nice-to-have while you're in these files: `memory-configuration/data-regions.adoc` prose says "pass `LoadAllWarmUpStrategy`/`NoOpWarmUpStrategy` to `setDefaultWarmUpConfiguration`", but the now-compiling snippet passes `LoadAllWarmUpConfiguration`/`NoOpWarmUpConfiguration` — the prose names the *strategy*, the API takes the *configuration*.

Inline notes below.

docs/_docs/code-snippets/java/pom.xml
```diff
@@ -25,6 +25,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <ignite.version>2.19.0-SNAPSHOT</ignite.version>
+        <ignite.extensions.version>1.0.0</ignite.extensions.version>
```
anton-vinogradov, line 28: Minor, but worth flagging here: ignite-extensions modules are released **independently** of each other (and of core), so a single `ignite.extensions.version` property is a latent trap — it only works because `ignite-aws-ext`, `ignite-azure-ext`, `ignite-gce-ext` and `ignite-zookeeper-ip-finder-ext` all happen to sit at `1.0.0` today. As soon as one of them publishes a new release, you're either stuck on it for all four or silently compiling against a stale one. Per-artifact version properties (or at least a comment stating the shared value is coincidental) will age better.

docs/_docs/code-snippets/java/pom.xml
```diff
@@ -67,12 +68,47 @@
...
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>ignite-azure-ext</artifactId>
+            <version>${ignite.extensions.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>ignite-cloud</artifactId>
+            <version>2.13.0</version>
+        </dependency>
```
anton-vinogradov, line 86: This is the one dependency that isn't a real extension artifact: `ignite-cloud-ext` was never published to Maven Central, so `ignite-cloud:2.13.0` is a deliberate legacy fallback (the only resolvable artifact that still ships `TcpDiscoveryCloudIpFinder`). Please record that in the POM so the next reader doesn't "fix" it by bumping the version or switching to a non-existent `ignite-cloud-ext`:

```suggestion
        <dependency>
            <groupId>${project.groupId}</groupId>
            <!-- ignite-cloud-ext (the ignite-extensions successor of this module) is not yet
                 published to Maven Central, so we deliberately fall back to the last released
                 legacy ignite-cloud artifact, which still ships TcpDiscoveryCloudIpFinder.
                 Revisit once ignite-cloud-ext is released (dev-list:
                 https://lists.apache.org/thread/d214wbsp84gm6sc93woctf3q68dhrsch). -->
            <artifactId>ignite-cloud</artifactId>
            <version>2.13.0</version>
        </dependency>
```

Also please double-check `2.13.0` actually compiles against `ignite-core:2.19.0-SNAPSHOT` — that's a wide gap on the discovery IP-finder SPI, and it's the dependency most likely to break the gate.

docs/_docs/code-snippets/java/src/main/java/org/apache/ignite/snippets/Snapshots.java
```diff
@@ -28,7 +31,7 @@ void configuration() {
         //tag::config[]
         IgniteConfiguration cfg = new IgniteConfiguration();
 
-        File exSnpDir = U.resolveWorkDirectory(U.defaultWorkDirectory(), "ex_snapshots", true);
+        File exSnpDir = new File("work", "ex_snapshots");
```
anton-vinogradov, line 35: Small doc-quality nit on the rendered output: this line lives inside `tag::config[]`, so it ships verbatim into `persistence/snapshot-directory.adoc`. `new File("work", "ex_snapshots")` is resolved against the process working directory, which *reads* like Ignite's work dir but isn't guaranteed to be it — the old `U.resolveWorkDirectory(...)` (rightly dropped, internal API) used to make that explicit. A one-line comment keeps the example unambiguous for readers:

```suggestion
        // Any directory the Ignite process can write to (example value).
        File exSnpDir = new File("work", "ex_snapshots");
```

docs/_docs/code-snippets/java/src/main/java/org/apache/ignite/snippets/WarmUpStrategy.java
```diff
@@ -14,70 +14,91 @@
...
 
-cfg.setDataStorageConfiguration(storageCfg);
-//end::warm-all-regions[]
-//tag::warm-specific-region[]
-IgniteConfiguration cfg = new IgniteConfiguration();
+        DataStorageConfiguration storageCfg = new DataStorageConfiguration();
 
-DataStorageConfiguration storageCfg = new DataStorageConfiguration();
+        //Changing the default warm-up strategy for all data regions
+        storageCfg.setDefaultWarmUpConfiguration(new LoadAllWarmUpConfiguration());
```
anton-vinogradov, line 33: Not introduced by this PR, but flagging since it's the same feature: this snippet is included by `memory-configuration/data-regions.adoc`, whose prose says to "pass the configuration parameter `LoadAllWarmUpStrategy` to `DataStorageConfiguration#setDefaultWarmUpConfiguration`" (and `NoOpWarmUpStrategy` further down). The actual public config — correctly used here — is `LoadAllWarmUpConfiguration` / `NoOpWarmUpConfiguration`. `*WarmUpStrategy` is the (internal) strategy, not what `setDefaultWarmUpConfiguration(WarmUpConfiguration)` accepts. Since the whole point of the ticket is that doc and code agree, worth fixing that prose in the same pass.

anton-vinogradov (changes requested):
Switching to *request changes* to track the two merge-readiness blockers from my detailed review above. This is **not** about the snippet fixes themselves — those are correct and a real improvement (they currently teach non-compiling code and internal API). Purely about readiness:

1. **Acceptance criterion isn't demonstrated.** The ticket is literally "snippets compile", but there's no green CI / TC visa and no attached `mvn -f docs/_docs/code-snippets/java/pom.xml compile -B -V` output (the Copilot pass bailed on quota). Please attach a green run before merge.

2. **Coordinate the merge with the CI gate (IGNITE-28800 / #13257).** Both PRs edit the same `docs/_docs/code-snippets/java/pom.xml`, and this fix is throwaway without the gate — the snippets rotted precisely because nothing compiled them. Please land them as a coordinated pair with an explicit order (rebase #13257 on top of this, verify green, merge back-to-back) so `master` never goes red and the snippets can't silently rot again.

Still open (not blocking on its own): the `ignite-cloud:2.13.0` legacy fallback, since `ignite-cloud-ext` isn't published — fine to proceed once the [dev-list thread](https://lists.apache.org/thread/d214wbsp84gm6sc93woctf3q68dhrsch) concludes, but please confirm `2.13.0` actually compiles against `ignite-core:2.19.0-SNAPSHOT`.

Happy to flip to approve once the green run is attached and the merge order with #13257 is settled.

docs/_docs/code-snippets/java/pom.xml
chesnokoff: I think keeping a shared property is fine here. This POM is only a compile harness for documentation snippets, not user-facing dependency management. The extension artifacts used here are currently all published as 1.0.0, so a single property avoids repeating the same literal. If these versions diverge later, we can split it into per-artifact properties when updating the affected dependency

chesnokoff:
Confirmed with `./mvnw -f docs/_docs/code-snippets/java/pom.xml compile -B -V`; the snippets compile successfully with `ignite-cloud:2.13.0` and current `ignite-core:2.19.0-SNAPSHOT`.

chesnokoff:
All actionable code comments should be addressed now

For the CI gate: the intended merge order is to land this snippets-fix PR first, then update #13257 on top of it and merge the CI gate immediately after it is green. Merging #13257 first would make master red because the snippets fixes are not there yet

zstan (approved):

docs/_docs/code-snippets/java/pom.xml
```diff
@@ -67,12 +68,48 @@
...
+            <groupId>${project.groupId}</groupId>
+            <artifactId>ignite-azure-ext</artifactId>
+            <version>${ignite.extensions.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <!-- ignite-cloud-ext is not published to Maven Central, so use the last released legacy artifact that still provides TcpDiscoveryCloudIpFinder. -->
+            <artifactId>ignite-cloud</artifactId>
+            <version>2.13.0</version>
+        </dependency>
```
anton-vinogradov, line 86: Optional hardening — not blocking, the build is green as-is. `ignite-cloud:2.13.0` also drags a transitive `ignite-core:2.13.0` (plus jclouds 2.0.0 / guava 16.0.1) onto the gate classpath. Maven mediation already prefers the direct 2.19 core, but for a *blocking* gate it's worth excluding it explicitly so a second, 6-year-old core can never shadow 2.19:

```suggestion
            <exclusions>
                <!-- Keep only the current ignite-core on the snippets classpath; this
                     legacy artifact otherwise also pulls in ignite-core:2.13.0. -->
                <exclusion>
                    <groupId>${project.groupId}</groupId>
                    <artifactId>ignite-core</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

anton-vinogradov (commented):
**Verified the acceptance criterion locally — it's green.**

Ran the canonical check from the ticket on the PR head against `ignite-core:2.19.0-SNAPSHOT`, with the extension artifacts resolved from Maven Central:

```
./mvnw -f docs/_docs/code-snippets/java/pom.xml -DskipTests compile -B -V
...
[INFO] Compiling 82 source files with javac [debug target 17] to target/classes
[INFO] BUILD SUCCESS
```

Only deprecation/unchecked warnings, no errors. This also settles my earlier open question: `ignite-cloud:2.13.0` **does** compile against `ignite-core:2.19.0-SNAPSHOT`. So blocker #1 is satisfiable — please attach an equivalent green run (ideally through #13257's workflow) to the ticket.

I re-checked the substance and it holds up: the Azure rename is correct (`azure-ext` really ships `TcpDiscoveryAzureBlobStoreIpFinder`, there is no `…BlobStorageIpFinder`); `opencensus-exporter-trace-zipkin:0.31.1` is genuinely required by `Tracing.java` and matches core's `opencensus.version`; and every added import is actually used.

**One consistency miss — flagging here because the file isn't in the diff.** `EchoService.java` and `EchoServiceImpl.java` were moved to `org.apache.ignite.snippets.services`, but their sibling in the same directory, `EchoServiceAccessInterceptor.java`, still declares `package org.apache.ignite.examples;`. It compiles (javac tolerates the package/dir mismatch) and the stale line is outside the rendered `service-call-interceptor` tag, so it doesn't reach the docs — but it's the same source/doc drift this ticket is about, and the class ends up under `org/apache/ignite/examples/`. One-liner:

```java
// EchoServiceAccessInterceptor.java
package org.apache.ignite.snippets.services;
```

**Remaining gating item:** the merge order with the CI gate (#13257) on the shared `pom.xml` — without the gate the snippets rot again. Happy to flip to approve once that's settled and a green run is attached. Optional pom hardening inline below.

zstan (approved):

PR #13281, opened 2026-06-25
IGNITE-28823 Reduce duration of the "Check java code" GitHub Actions job (commit-check.yml)

anton-vinogradov:
## [IGNITE-28823](https://issues.apache.org/jira/browse/IGNITE-28823)

The `check-java` job in `commit-check.yml` is the long pole of the "Code Style, Abandoned Tests, Javadocs" workflow (~8.9 min; every other job in the workflow finishes in under a minute). It runs three sequential Maven invocations and has two inefficiencies:

1. The build is single-threaded (no `-T`), although `ubuntu-latest` has 4 vCPUs.
2. The "abandoned tests" step recompiles the whole reactor a second time. `maven-compiler-plugin` 3.15 logs *"Recompiling the module because of changed dependency"* for 28 modules (including `ignite-core`, 3322 source files): the second invocation uses a different profile set, so the plugin's incremental engine treats reactor dependencies as changed and cascades a full recompile.

### Change

- `-T 1C` on the codestyle / `test-compile` step — parallel reactor build.
- `-Dmaven.compiler.useIncrementalCompilation=false` on the "abandoned tests" step — reuse the classes compiled by the previous step instead of recompiling.

The "abandoned tests" step is intentionally **not** parallelized: the orphaned-test collection writes to a shared file finalized by the last reactor module, so `-T` would race.

### Why not merge the two steps

Merging would pull `examples` / `lgpl` modules into the orphaned-test check (today it runs only over `-Pall-java,scala`) and could newly fail it. The chosen flag keeps the check scope identical — profile-conditional source roots (`java-lgpl`) exist only in the `examples` module, which is not part of the "abandoned tests" reactor.

### Verification (local, full reactor)

| | before | after |
|---|---|---|
| step 2 recompiled modules | 30 (28 "changed dependency") | 2 (0 "changed dependency") |
| step 1 under `-T 1C` | — | 47 modules SUCCESS, checkstyle incl., 0 failures |

The before/after wall-clock is visible on this PR's own CI run, since the change is to `commit-check.yml` itself.

### DIFF
--- a/.github/workflows/commit-check.yml
+++ b/.github/workflows/commit-check.yml
@@ -43,6 +43,7 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: ${{ matrix.java }}
+          cache: 'maven'
 
       - name: Install prerequisites
         run: |
@@ -96,11 +97,12 @@ jobs:
 
       - name: Run codestyle and licenses checks
         run: |
-          ./mvnw test-compile -Pall-java,licenses,lgpl,checkstyle,examples,scala,check-licenses -B -V
+          ./mvnw test-compile -Pall-java,licenses,lgpl,checkstyle,examples,scala,check-licenses -B -V -T 1C
 
       - name: Run abandoned tests checks.
+        # Reuse classes from the previous step; the differing profiles otherwise trigger a full reactor recompile.
         run : |
-          ./mvnw test -Pcheck-test-suites,all-java,scala -B -V
+          ./mvnw test -Pcheck-test-suites,all-java,scala -B -V -Dmaven.compiler.useIncrementalCompilation=false
 
       - name: Check javadocs.
         run : |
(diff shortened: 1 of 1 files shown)


.github/workflows/commit-check.yml
```diff
@@ -96,11 +97,12 @@ jobs:
 
       - name: Run codestyle and licenses checks
         run: |
-          ./mvnw test-compile -Pall-java,licenses,lgpl,checkstyle,examples,scala,check-licenses -B -V
+          ./mvnw test-compile -Pall-java,licenses,lgpl,checkstyle,examples,scala,check-licenses -B -V -T 1C
```
alex-plekhanov: Do we still need `scala` profile? Looks like this profile was deleted a long time ago (all scala modules were moved to extensions)

zstan: +1 no need in scala

anton-vinogradov: Good catch — removed. The `scala` profile no longer exists (deleted in IGNITE-18301 together with the visor console; scala modules live in extensions now), so `-Pscala` was a no-op. Dropped it from both invocations in 13a45303063.

zstan (approved):

PR #13285, opened 2026-06-26
IGNITE-28826 Ducktests development guide.

EgorBaranovEnjoysTyping:
Hi everyone. I propose adding a comprehensive README.md to the ducktest module. This would serve as a full documentation guide for developers writing new ducktests.


modules/ducktests/DEV_GUIDE.md
```diff
@@ -0,0 +1,541 @@
...
+        server_config = IgniteConfiguration(
+            version=IgniteVersion(ignite_version),
+            caches=[CacheConfiguration(name='test-cache', backups=1)]
+        )
+        servers = IgniteService(self.test_context, server_config, num_nodes=2)
+        servers.start()
+        # ── 2. Start a client application ──────────────────────
+        client_config = IgniteThinClientConfiguration(
+            version=IgniteVersion(ignite_version),
+            discovery_spi=from_ignite_cluster(servers)
```
anton-vinogradov: `IgniteThinClientConfiguration` is a standalone `NamedTuple` with only `addresses` and `version` — it has no `discovery_spi` field, so this raises `TypeError: __new__() got an unexpected keyword argument 'discovery_spi'`. Thin clients connect by address (cf. `thin_client_test.py`, `auth_test.py`). Suggested fix — set the connector on the server and pass `addresses`:

```python
server_config = IgniteConfiguration(
    version=IgniteVersion(ignite_version),
    client_connector_configuration=ClientConnectorConfiguration(),
    caches=[CacheConfiguration(name='test-cache', backups=1)]
)
...
addresses = [servers.nodes[0].account.hostname + ":" +
             str(server_config.client_connector_configuration.port)]
client_config = IgniteThinClientConfiguration(
    version=IgniteVersion(ignite_version),
    addresses=addresses
)
```

modules/ducktests/DEV_GUIDE.md
```diff
@@ -0,0 +1,541 @@
...
+        )
+        app = IgniteApplicationService(
+            self.test_context,
+            client_config,
+            java_class_name="org.apache.ignite.internal.ducktest.tests.mytest.MyTestApp",
+            params={"cacheName": "test-cache", "range": 1000}
+        )
+        app.start()  # blocks until IGNITE_APPLICATION_INITIALIZED
+        # ── 3. Verify results ──────────────────────────────────
+        result = app.extract_result("putCount")
```
anton-vinogradov, line 73: `app.start()` only waits for `IGNITE_APPLICATION_INITIALIZED`. Since the Java sample calls `markInitialized()` *before* the `put` loop and `recordResult(...)`, the `putCount->...` line may not be in the log yet — so `extract_result` greps zero lines and trips `assert len(results) == len(self.nodes)`. Read the result after the app has finished, i.e. after `app.stop()` (which waits for `IGNITE_APPLICATION_FINISHED`); cf. `pme_free_switch_test.py`:

```python
app.start()   # blocks until IGNITE_APPLICATION_INITIALIZED
app.stop()    # blocks until IGNITE_APPLICATION_FINISHED
result = app.extract_result("putCount")
assert result == "1000", f"Expected 1000, got {result}"
```

modules/ducktests/DEV_GUIDE.md
```diff
@@ -0,0 +1,541 @@
...
+5. One @cluster per method. The decorator must appear once per test method. It sets cluster metadata in ctx.cluster_use_metadata. If the context already has metadata, it is not overwritten.
+
+#### Errors and Limitations
+
+    ┌────────────────────────────────────────┬───────────────────────────────────────────────────────────────┐
+    │ Situation                              │ Result                                                        │
+    ├────────────────────────────────────────┼───────────────────────────────────────────────────────────────┤
+    │ num_nodes exceeds available containers │ Test will not start (ducktape reports insufficient resources) │
+    │ num_nodes <= 0                         │ Parsing error when reading global cluster_size                │
+    │ Multiple @cluster on one method        │ Only the top one applies                                      │
```
anton-vinogradov: This row contradicts key point #5 above ("if the context already has metadata, it is not overwritten"). Since `cluster()` wraps the function and decorators apply bottom-up, the inner (bottom) `@cluster` sets the metadata first and wins — and stacking also runs `before()/after()` twice. Suggest dropping this row or rewording to "the inner (bottom) `@cluster` applies".

modules/ducktests/DEV_GUIDE.md
```diff
@@ -0,0 +1,541 @@
...
+Purpose: @ignore_if marks a test method as IGNORED if a specific condition evaluates to True. It is used to skip tests for certain Ignite versions or global parameter configurations without removing the test code.
+
+#### Basic Usage
+```python
+from ignitetest.utils import cluster, ignite_versions, ignore_if
+from ignitetest.utils.version import V_2_11_0
+class MyTest(IgniteTest):
+    @cluster(num_nodes=2)
+    @ignite_versions(str(DEV_BRANCH), str(LATEST))
+    @ignore_if(lambda version, globals: version <= V_2_11_0)
```
anton-vinogradov: Nit: `globals` shadows the Python built-in. The other examples in this file use `g` for the second lambda argument — suggest the same here for consistency.

anton-vinogradov (commented):
Thanks for putting this together — the guide is accurate and genuinely useful. It nicely complements the existing `README.md` (which covers *running* tests) by documenting how to *write* them. I verified the API-level claims against the sources and they hold up: lifecycle methods, the `IGNITE_APPLICATION_*` tokens, the `recordResult` -> `name->val<-` format, `IgniteConfiguration` fields, discovery helpers, decorator defaults, and the 13-container default.

A few things before merge — two of them are functional bugs in the sample code that readers will copy verbatim (left inline):

**Blockers**
- Thin-client config passes a non-existent `discovery_spi` kwarg → `TypeError`.
- The result is read right after `start()`, which races with the Java app writing it.

**Suggestions**
- *README discoverability:* `modules/ducktests/README.md` doesn't link to this guide, so developers who land on the README won't find it. Consider a pointer under the `# Development` section, e.g. `See [DEV_GUIDE.md](DEV_GUIDE.md) for writing new ducktests.`
- *Tables:* they're drawn as ASCII art inside indented code blocks, so on GitHub they render as monospace and several have a broken right border (e.g. "Application Lifecycle Methods" and the `@cluster` "Errors and Limitations" table). Converting them to native Markdown tables would render cleanly and fix the borders.
- Two inline nits below.

Marking as Comment rather than Request changes — the polish is your call, but the two inline blockers should be fixed.

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: Done

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: Done

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: Done

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: Done

modules/ducktests/DEV_GUIDE.md
```diff
@@ -0,0 +1,518 @@
...
+| markSyncExecutionComplete() | Run-to-completion shortcut          | Use instead of init + finish pair         |
+
+---
+## Service Types (what connection the app gets)
+
+| Service Type     | Python config class           | Java field              |
+|------------------|-------------------------------|-------------------------|
+| Full server node | IgniteConfiguration           | this.ignite             |
+| Thin client      | IgniteThinClientConfiguration | this.client             |
+| Thin JDBC        | (same, mode=THIN_JDBC)        | this.thinJdbcDataSource |
```
maksaska: Writing "(same,
    mode=THIN_JDBC)" implies it's a flag on the same config as Thin Client, which is wrong. Should be:

| Thin JDBC        | IgniteThinJdbcConfiguration   | this.thinJdbcDataSource |

modules/ducktests/DEV_GUIDE.md
```diff
@@ -0,0 +1,518 @@
...
+@cluster(num_nodes=3)  # 2 servers + 1 client = 3
+def test_with_client(self, ignite_version):
+    servers = IgniteService(self.test_context, config, num_nodes=2)
+    client = IgniteService(self.test_context, client_config, num_nodes=1)
+    servers.start()
+    client.start()
+```
+3. Global cluster_size parameter. You can override num_nodes at runtime via a global parameter:
+   ./docker/run_tests.sh -g cluster_size=5 -t ./ignitetest/tests/my_test.py
+4. Instead of num_nodes, you can specify a ClusterSpec from ducktape for typed node specifications:
```
maksaska: It looks like the text cuts off after the fourth bullet point. Was there supposed to be a code snippet here?

modules/ducktests/DEV_GUIDE.md
```diff
@@ -0,0 +1,518 @@
...
+``` python
+@cluster(num_nodes=2)
+@ignite_versions(str(DEV_BRANCH), str(LATEST), version_prefix="server_version")
+@ignite_versions(str(DEV_BRANCH), str(LATEST), version_prefix="client_version")
+def test_cross_version(self, server_version, client_version):
+   # Generates a cartesian product of server/client versions
+```
+4. If a single test needs to spin up nodes with different versions simultaneously, pass a tuple/list of size ≥ 2 as
+   one argument: `@ignite_versions(("dev", "2.17.0"))`
+5. @ignite_versions must be placed below @cluster and above @matrix (or the method definition):
```
maksaska: I'm not sure the order matters. Nevertheless there is a trailing colon with nothing after

modules/ducktests/DEV_GUIDE.md
```diff
@@ -0,0 +1,518 @@
...
+        // 5. For long-running apps, loop until SIGTERM:
+        //    while (!terminated()) { U.sleep(100); }
+
+        // 6. Signal completion — Python side waits for this on stop()
+        markFinished();
+    }
+}
+```
+---
+## Application Lifecycle Methods
```
maksaska: It's a low-level implementation detail. Not sure it should be in dev guide

modules/ducktests/DEV_GUIDE.md
```diff
@@ -0,0 +1,518 @@
...
+def test_with_client(self, ignite_version):
+    servers = IgniteService(self.test_context, config, num_nodes=2)
+    client = IgniteService(self.test_context, client_config, num_nodes=1)
+    servers.start()
+    client.start()
+```
+3. Global cluster_size parameter. You can override num_nodes at runtime via a global parameter:
+   ./docker/run_tests.sh -g cluster_size=5 -t ./ignitetest/tests/my_test.py
+4. Instead of num_nodes, you can specify a ClusterSpec from ducktape for typed node specifications:
+5. One @cluster per method. The decorator must appear once per test method. It sets cluster metadata in ctx.cluster_use_metadata. If the context already has metadata, it is not overwritten.
```
maksaska: Maybe we should say something about metadata. Otherwise it's not clear what it is, and why it exists.

modules/ducktests/DEV_GUIDE.md
```diff
@@ -0,0 +1,518 @@
...
+
+#### Errors and Limitations
+
+| Situation                              | Result                                                        |
+|----------------------------------------|---------------------------------------------------------------|
```
maksaska: The result should explain the framework behavior with definite outcome for the developer.  
* Parsing error - sounds vague
* before()/after() run twice - what are those methods. 
* context - same question. Why we should worry about it? How does it affect the test.

A simple explanatory sentence should suffice.

modules/ducktests/DEV_GUIDE.md
```diff
@@ -0,0 +1,518 @@
...
+| Using @matrix on a non-method function                                        | Parameters are still injected, but self.test_context is None — runtime AttributeError |
+
+#### Behavior Limitations
+
+| Situation                                                | Behavior                                                                                   |
+|----------------------------------------------------------|--------------------------------------------------------------------------------------------|
+| One parameter list is empty                              | The entire @matrix produces zero test cases — the test is silently skipped                 |
+| Multiple @matrix decorators on one method                | Each @matrix is applied independently, creating nested cartesian products                  |
+| Large number of combinations                             | Exponential test growth — e.g., 3 params × 5 values each = 125 test executions per version |
+| Non-hashable parameter values (e.g., lists, dicts)       | Ducktape may fail to deduplicate test cases — unexpected duplicates or crashes             |
```
maksaska: "may fail" + "unexpected" + "or crashes" - sounds vague

modules/ducktests/DEV_GUIDE.md
maksaska: ClusterSpec - What is it, and why the user should think about it?

maksaska (changes requested):

modules/ducktests/DEV_GUIDE.md
```diff
@@ -0,0 +1,518 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+# How to Write a New DuckTest
+
+The ignite-ducktests framework is a bilingual integration testing framework:
+
+| Layer                | Language | Purpose                                                           |
+|----------------------|----------|-------------------------------------------------------------------|
+| Test orchestration   | Python   | Manages Docker containers, starts/stops nodes, asserts results    |
```
maksaska: This row implies ducktape is Docker-specific. It's a general multi-node orchestration framework (supports Vagrant, EC2, local processes, etc.). Docker is just the backend we use here. 

Consider:

    > | Test orchestration | Python | Manages cluster nodes, runs workloads, asserts results |

Or keep "Docker containers" if you want to set expectations for the local dev experience - just make sure it reads as "in this setup" rather than a framework capability.

modules/ducktests/DEV_GUIDE.md
```diff
@@ -0,0 +1,518 @@
...
+| In-cluster workloads | Java     | Runs inside Ignite nodes (cache ops, transactions, queries, etc.) |
+
+Each Ignite node runs in a separate Docker container. The Python layer manages container lifecycle and simulates network failures via iptables.
+
+---
+## Write a Java Application (if needed)
```
maksaska: I would start with python first

maksaska (changes requested):

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: Intentional: the Java section is marked "(if needed)" — readers who only need a Python wrapper can skip it. The Python test references the Java class by `java_class_name`, so it makes sense to see what Java class you're wiring in before writing the Python side.

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: This is the contract between the Java app and the Python runner. Without it, a developer writing a new Java app wouldn't know which token to print, or why `start()` doesn't return until `markInitialized()`. I'd argue it's the most important section for Java-side authors. Keeping it.

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: Agree, done.

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: Agree, done.

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: Agree, done.

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: Agree, done.

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: Agree, done.

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: Agree, done.

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: Agree, done.

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: Agree, done

modules/ducktests/DEV_GUIDE.md
EgorBaranovEnjoysTyping: Okey, now  python related section is followed by java related section

maksaska (approved):

PR #13295, opened 2026-06-29
IGNITE-28835 BinaryMarshaller: marshal directly into OutputStream to avoid a full-object copy

anton-vinogradov:
## Description

[IGNITE-28835](https://issues.apache.org/jira/browse/IGNITE-28835)

`BinaryMarshaller.marshal0(Object, OutputStream)` had two avoidable inefficiencies on the hot marshal-to-stream path:

1. **Full-size temporary copy.** It did `byte[] arr = marshal(obj); out.write(arr);` — `GridBinaryMarshaller.marshal` returns `writer.array()` (an `Arrays.copyOf` of the whole serialized object), which was then immediately streamed out and discarded.
2. **Redundant node-name scope.** It called the *public* `marshal(obj)`, which re-enters `AbstractNodeNameAwareMarshaller`'s `setCurrentIgniteName`/`restoreOldIgniteName` scope a second time, although `marshal0` already runs inside the scope set by the outer `marshal(obj, out)`.

## Change

- Added `GridBinaryMarshaller.marshal(obj, out, failIfUnregistered)` that writes the writer's internal buffer to the stream by length (`out.write(s.array(), 0, s.position())`) — no trim-copy, no extra allocation.
- `BinaryMarshaller.marshal0(obj, out)` now calls it directly, dropping the second name-scope.

Behavior-preserving: the bytes written to the stream are identical to the array path (the internal buffer `[0, position())` equals `writer.array()`); the `null` object still maps to a single `NULL` byte. Covered by a new test `BinaryMarshallerSelfTest#testMarshalToStreamMatchesArray`.

## Benchmark

Isolated JMH micro-benchmark of the avoided work (serialization excluded, identical in both variants; AverageTime, ns/op, JDK 17):

```
full-object copy in marshal0(obj, out) — copy+write vs write-directly:
  object size   current     proposed    gain
  64 B          4.96 ns     2.14 ns     -57%
  1 KB          43.7 ns     18.8 ns     -57%
  16 KB         667 ns      364 ns      -45%
  256 KB        10563 ns    3756 ns     -64%  (-6.8 us)

redundant name-scope (per marshal-to-stream call):
  current  : 13.00 ns
  proposed : 11.16 ns   (-14%)
```

The stream-tail cost roughly halves and scales with object size; additionally a full-size temporary allocation per call is eliminated (reduced GC pressure), which the ns/op figures do not fully capture.

## Checklist

- [x] No public API / wire-format change (local to the `binary` module).
- [x] New unit test added.

### DIFF
--- a/modules/binary/api/src/main/java/org/apache/ignite/internal/binary/BinaryMarshaller.java
+++ b/modules/binary/api/src/main/java/org/apache/ignite/internal/binary/BinaryMarshaller.java
@@ -83,14 +83,7 @@ public class BinaryMarshaller extends AbstractNodeNameAwareMarshaller {
 
     /** {@inheritDoc} */
     @Override protected void marshal0(@Nullable Object obj, OutputStream out) throws IgniteCheckedException {
-        byte[] arr = marshal(obj);
-
-        try {
-            out.write(arr);
-        }
-        catch (Exception e) {
-            throw new BinaryObjectException("Failed to marshal the object: " + obj, e);
-        }
+        impl.marshal(obj, out, false);
     }
 
     /** {@inheritDoc} */
--- a/modules/binary/api/src/main/java/org/apache/ignite/internal/binary/GridBinaryMarshaller.java
+++ b/modules/binary/api/src/main/java/org/apache/ignite/internal/binary/GridBinaryMarshaller.java
@@ -17,6 +17,8 @@
 
 package org.apache.ignite.internal.binary;
 
+import java.io.IOException;
+import java.io.OutputStream;
 import java.lang.reflect.Array;
 import java.util.Collection;
 import java.util.Map;
@@ -254,6 +256,35 @@ public class GridBinaryMarshaller {
             writer.marshal(obj);
 
             return writer.array();
+        }
+    }
+
+    /**
+     * Marshals the object directly into the given stream, without allocating a trimmed copy of the whole result.
+     *
+     * @param obj Object to marshal.
+     * @param out Output stream.
+     * @param failIfUnregistered Throw exception if class isn't registered.
+     * @throws BinaryObjectException In case of error.
+     */
+    public void marshal(@Nullable Object obj, OutputStream out, boolean failIfUnregistered) throws BinaryObjectException {
+        try {
+            if (obj == null) {
+                out.write(NULL);
+
+                return;
+            }
+
+            try (BinaryWriterEx writer = BinaryUtils.writer(ctx, failIfUnregistered, UNREGISTERED_TYPE_ID)) {
+                writer.marshal(obj);
+
+                BinaryOutputStream s = writer.out();
+
+                out.write(s.array(), 0, s.position());
+            }
+        }
+        catch (IOException e) {
+            throw new BinaryObjectException("Failed to marshal the object: " + obj, e);
         }
     }
 
--- a/modules/core/src/test/java/org/apache/ignite/internal/binary/BinaryMarshallerSelfTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/binary/BinaryMarshallerSelfTest.java
@@ -17,6 +17,7 @@
 
 package org.apache.ignite.internal.binary;
 
+import java.io.ByteArrayOutputStream;
 import java.io.Externalizable;
 import java.io.IOException;
 import java.io.ObjectInput;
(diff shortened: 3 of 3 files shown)


anton-vinogradov:
JMH benchmarks behind the numbers in the description. They are standalone (no running node): each replicates the exact hot-path step from the change in *current* vs *proposed* form, with serialization excluded (identical in both), so the measured delta is precisely the avoided work.

Run: drop into a JMH project (`jmh-core`/`jmh-generator-annprocess` 1.37), `mvn package`, `java -jar target/benchmarks.jar`. Measured on JDK 17, `Mode.AverageTime`, ns/op.

<details>
<summary><code>StreamCopyBench</code> — avoided full-object copy in <code>marshal0(obj, out)</code></summary>

```java
package bench;

import java.io.IOException;
import java.io.OutputStream;
import java.util.Arrays;
import java.util.concurrent.TimeUnit;
import org.openjdk.jmh.annotations.*;
import org.openjdk.jmh.infra.Blackhole;

/**
 * current  = Arrays.copyOf(buf, n) then out.write(arr)   // GridBinaryMarshaller.marshal -> writer.array() + out.write
 * proposed = out.write(buf, 0, n)                         // write writer's internal buffer directly, no trim-copy
 * Serialization is excluded (identical both ways); the OutputStream write cost is identical too.
 */
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@Warmup(iterations = 5, time = 1)
@Measurement(iterations = 8, time = 1)
@Fork(2)
@State(Scope.Thread)
public class StreamCopyBench {
    @Param({"64", "1024", "16384", "262144"})
    public int n;

    byte[] internalBuf;   // over-allocated writer buffer, n valid bytes
    SinkStream out;

    static final class SinkStream extends OutputStream {
        byte[] dst;
        int cnt;
        SinkStream(int cap) { dst = new byte[cap]; }
        @Override public void write(int b) { ensure(1); dst[cnt++] = (byte)b; }
        @Override public void write(byte[] b, int off, int len) {
            ensure(len);
            System.arraycopy(b, off, dst, cnt, len);
            cnt += len;
        }
        void ensure(int extra) {
            if (cnt + extra > dst.length)
                dst = Arrays.copyOf(dst, Math.max(dst.length << 1, cnt + extra));
        }
        void reset() { cnt = 0; }
    }

    @Setup(Level.Trial)
    public void setup() {
        internalBuf = new byte[(int)(n * 1.4) + 16];
        for (int i = 0; i < internalBuf.length; i++)
            internalBuf[i] = (byte)i;
        out = new SinkStream(n + 64);
    }

    @Benchmark
    public void current(Blackhole bh) throws IOException {
        out.reset();
        byte[] arr = Arrays.copyOf(internalBuf, n);   // writer.array() trim-copy
        out.write(arr);                                // marshal0(obj,out): out.write(arr)
        bh.consume(arr);
        bh.consume(out.dst);
    }

    @Benchmark
    public void proposed(Blackhole bh) throws IOException {
        out.reset();
        out.write(internalBuf, 0, n);                  // write internal buffer directly
        bh.consume(out.dst);
    }
}
```
</details>

<details>
<summary><code>NameScopeBench</code> — avoided redundant node-name scope</summary>

```java
package bench;

import java.util.concurrent.TimeUnit;
import org.openjdk.jmh.annotations.*;
import org.openjdk.jmh.infra.Blackhole;

/**
 * Replicates CommonUtils.setCurrentIgniteName/restoreOldIgniteName (identity-compare, set only if differs).
 * current  = marshal0(obj,out) -> public marshal(obj) -> marshal0(obj)   (scope entered twice)
 * proposed = marshal0(obj,out) -> marshal0(obj)                          (scope entered once)
 */
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@Warmup(iterations = 5, time = 1)
@Measurement(iterations = 8, time = 1)
@Fork(2)
@State(Scope.Thread)
public class NameScopeBench {
    static final String EMPTY = new String();
    static final ThreadLocal<String> LOC = ThreadLocal.withInitial(() -> EMPTY);

    @SuppressWarnings("StringEquality")
    static String setName(String n) {
        String old = LOC.get();
        if (old != n)
            LOC.set(n);
        return old;
    }

    @SuppressWarnings("StringEquality")
    static void restore(String old, String cur) {
        if (old != cur)
            LOC.set(old);
    }

    String nodeName = "node-instance-name";
    byte[] sink = new byte[256];
    private final Object obj = new Object();

    private byte[] doWork(Object obj) {           // representative small marshal payload, same for both
        byte[] a = new byte[48];
        int h = obj.hashCode();
        for (int i = 0; i < a.length; i++)
            a[i] = (byte)(h + i);
        return a;
    }

    // CURRENT: marshal0(obj,out) -> public marshal(obj) -> marshal0(obj)
    private byte[] marshalPublic(Object obj) {    // = AbstractNodeNameAwareMarshaller.marshal(obj)
        String old = setName(nodeName);           // redundant second scope
        try {
            return doWork(obj);
        }
        finally {
            restore(old, nodeName);
        }
    }

    @Benchmark
    public void current(Blackhole bh) {
        String old = setName(nodeName);           // outer scope (parent marshal(obj,out))
        try {
            byte[] arr = marshalPublic(obj);      // re-enters scope
            System.arraycopy(arr, 0, sink, 0, arr.length);
            bh.consume(sink);
        }
        finally {
            restore(old, nodeName);
        }
    }

    @Benchmark
    public void proposed(Blackhole bh) {
        String old = setName(nodeName);
        try {
            byte[] arr = doWork(obj);             // direct, no second scope
            System.arraycopy(arr, 0, sink, 0, arr.length);
            bh.consume(sink);
        }
        finally {
            restore(old, nodeName);
        }
    }
}
```
</details>

alex-plekhanov (approved):

PR #13298, opened 2026-06-30
IGNITE-28838 Fix flaky GridCacheVersionGenerationWithCacheStorageTest after IGNITE-27678

zstan:


zstan:
image

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
```diff
@@ -3134,13 +3133,16 @@ private TxRecoveryInitRunnable(ClusterNode node) {
                         sendTxSalvage(tx, evtNodeId);
                     }
 
+                    Supplier<Boolean> fullSyncedOp = () -> tx.writeEntries().stream().map(e ->
+                        cctx.cacheContext(e.cacheId())).allMatch(GridCacheContext::syncCommit);
```
alex-plekhanov: `Supplier` is redundant, we can use this call directly.
Logic is inconsistent with `syncMode` logic (in `syncMode` `FULL_SYNC` is returned if **any** of cache have `FULL_SYNC` mode, but in current statement `FULL_SYNC` is returned if **all** caches have `FULL_SYNC` mode)
Let's use loop instead of streams.
Let's describe in comment why we can't use `deriveSyncMode`

zstan: i fill not deep comment, if it\`s not enough - i need to dig deeper.
other -seems fixed, locally tests are ok

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
```diff
@@ -3139,8 +3137,18 @@ private TxRecoveryInitRunnable(ClusterNode node) {
...
                         && tx.nodeId().equals(evtNodeId) && tx.state() == PREPARED) {
                         // Delay a commit, on backup. It will be raised further after near or coord. node will confirm it.
-                        // In different from {@code FULL_SYNC} modes, appropriate recovery message can never be raized.
-                        if (deriveSyncMode(cctx, tx.txState().cacheIds()) != FULL_SYNC)
+                        // In modes different from FULL_SYNC, appropriate recovery message can never be raized.
+                        // Approach with {@code IgniteTxImplicitSingleStateImpl.syncMode} can`t be used here because
+                        // {@code IgniteTxRemoteStateAdapter#cacheIds} can be empty.
+                        boolean fullSyncedOp = false;
+                        for (IgniteTxEntry ent : tx.writeEntries()) {
+                            if (cctx.cacheContext(ent.cacheId()).syncCommit()) {
```
vldpyatkov, line 3145: Are you sure that you can commit a transaction beyond a near node or coordinator node?

alex-plekhanov (approved):

PR #13311, opened 2026-07-02
IGNITE-28844 Calcite. Improve type checking in LIMIT / OFFSET clauses

zstan:
https://issues.apache.org/jira/browse/IGNITE-28844


modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/StatementChecker.java
```diff
@@ -0,0 +1,194 @@
...
+ * <p><b>Schema initialization</b>
+ *
+ * <pre>
+ *     new StatementChecker()
+ *         .table("t1", "int_col", Integer.class)
+ *         .sql("SELECT int_col FROM t1")
+ *         .ok();
+ * </pre>
+ */
+public class StatementChecker {
```
zstan: this checker will be used not only for dyn params checks and will be extended in future

tkalkirill:
Hi, regarding https://issues.apache.org/jira/browse/CALCITE-7624, I implemented support for BigDecimal for the OFFSET/FETCH and LIMIT in https://github.com/apache/ignite/pull/13375.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
```diff
@@ -17,29 +17,31 @@
...
+    /** Offset param. */
+    private final long offset;
 
-    /** Fetch if its present, otherwise 0. */
-    private final int fetch;
+    /** Fetch param. */
+    private final long fetch;
+
+    /** Fetch can be unset. */
+    private final boolean fetchUndefined;
```
alex-plekhanov: Let's add `long limit = fetch == -1 ? Long.MAX_VALUE : IgniteMath.addExact(fetch, offset)` instead. To execute `IgniteMath.addExact(fetch, offset)` only once (and not for each request and each push), to simplify `hasMoreData` (`rowsProcessed < limit`).

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
```diff
@@ -17,29 +17,31 @@
...
 
     /** Already processed (pushed to upstream) rows count. */
     private int rowsProcessed;
```
alex-plekhanov: Comparing int with long, can overflow

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
```diff
@@ -85,38 +90,46 @@ public LimitNode(
...
-
         --waiting;
 
-        checkState();
-
-        if (rowsProcessed > offset) {
-            if (fetchNode == null || (fetchNode != null && rowsProcessed <= fetch + offset))
-                downstream().push(row);
+        if (rowsProcessed >= offset && hasMoreData()) {
+            // this two rows can`t be swapped, cause if all requested rows have been pushed it will trigger further request call.
```
alex-plekhanov: Comment should be started with an uppercase letter.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -241,54 +241,81 @@ private void validateTableModify(SqlNode table) {
...
     /**
-     * @param n Node to check limit.
+     * Invalidate fetch/offset params restrictions.
+     *
+     * @param n        Node to check limit.
      * @param nodeName Node name.
      */
-    private void checkIntegerLimit(SqlNode n, String nodeName) {
+    private void invalidateFetchOffset(@Nullable SqlNode n, String nodeName) {
+        if (n == null) {
```
alex-plekhanov: Redundant braces

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -241,54 +241,81 @@ private void validateTableModify(SqlNode table) {
...
 
     /**
-     * @param n Node to check limit.
+     * Invalidate fetch/offset params restrictions.
+     *
+     * @param n        Node to check limit.
      * @param nodeName Node name.
      */
-    private void checkIntegerLimit(SqlNode n, String nodeName) {
+    private void invalidateFetchOffset(@Nullable SqlNode n, String nodeName) {
```
alex-plekhanov: invalidate - means make something invalid. Here we validate parameters, let's use `validate`or `check`.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java
```diff
@@ -109,56 +108,21 @@ private void checkLimit(int offset, int fetch) {
...
         limitNode.register(srcNode);
 
-        if (fetch > 0) {
-            for (int i = offset; i < offset + fetch; i++) {
-                assertTrue(rootNode.hasNext());
-                assertEquals(i, rootNode.next()[0]);
-            }
-
-            assertFalse(rootNode.hasNext());
-            assertEquals(srcNode.requested.get(), offset + fetch);
```
alex-plekhanov, line 125: This check is crucial, it shows that we don't request redundant rows from source node. Why it was removed?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java
```diff
@@ -109,56 +108,21 @@ private void checkLimit(int offset, int fetch) {
...
-        }
-        else {
+        for (int i = offset; i < offset + fetch; i++) {
             assertTrue(rootNode.hasNext());
-            assertEquals(offset, rootNode.next()[0]);
```
alex-plekhanov, line 129: This check for case `fetch == 0` was removed to

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/AbstractPlannerTest.java
```diff
@@ -438,14 +470,31 @@ protected static TestTable createTable(IgniteSchema schema, String name, RelData
...
+    ) throws Exception {
+        assertPlan(ctxBuilder, rel -> true);
+    }
+
+    /** */
+    protected <T extends RelNode> void assertPlan(
+        TestPlanningContextBuilder ctxBuilder,
+        Predicate<T> predicate
+    ) throws Exception {
+        invalidatePlan(ctxBuilder, predicate);
```
alex-plekhanov: Why do we need this method (it just a wrapper for invalidatePlan)
Why `invalidatePlan`? We don't invalidate anything, we check plans (better to use assertPlan too)

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
```diff
@@ -101,40 +98,6 @@ public void testNestedLimitOffsetWithUnion() {
...
-
-        // Check with parameters
-        assertThrows("SELECT * FROM TEST_REPL OFFSET ? ROWS FETCH FIRST ? ROWS ONLY",
-            SqlValidatorException.class, "Illegal value of fetch / limit", -1, -1);
-
-        assertThrows("SELECT * FROM TEST_REPL OFFSET ? ROWS",
-            SqlValidatorException.class, "Illegal value of offset", -1);
-
-        assertThrows("SELECT * FROM TEST_REPL FETCH FIRST ? ROWS ONLY",
-            SqlValidatorException.class, "Illegal value of fetch / limit", -1);
```
alex-plekhanov, line 135: Looks like these checks are still valid, why they was removed?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/AbstractPlannerTest.java
```diff
@@ -220,20 +221,51 @@ protected PlanningContext plannerCtx(
         @Nullable RelOptListener planLsnr,
         String... disabledRules
     ) {
-        return plannerCtx(sql, Collections.singleton(publicSchema), planLsnr, disabledRules);
+        return plannerCtx(sql, Collections.singleton(publicSchema), planLsnr, null, disabledRules);
+    }
+
+    /** */
+    private PlanningContext plannerCtx(
```
alex-plekhanov, line 227: Fully duplicated another method except one parameter `disabledRules`. Let's unify these methods.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/AbstractPlannerTest.java
```diff
@@ -438,14 +470,31 @@ protected static TestTable createTable(IgniteSchema schema, String name, RelData
         return table;
     }
 
+    /** */
+    protected <T extends RelNode> void assertPlan(
```
alex-plekhanov: Redundant method. Used in assertThrows, but it's better to write own assertThrows method with parameters like PlannerContext or PlannerContextBuilder.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/AbstractPlannerTest.java
```diff
@@ -805,4 +857,79 @@ class TestFailureProcessor extends FailureProcessor {
...
+    }
+
+    /** */
+    public static TestPlanningContextBuilder contextBuilder() {
+        return new TestPlanningContextBuilder();
+    }
+
+    /** */
+    @SuppressWarnings("ThrowableNotThrown")
+    static void assertThrows(
```
alex-plekhanov: Let's rewrite to assertThrows(PlannerContext, cls, msg) or assertThrows(PlannerContextBuilder, cls, msg) and use physicalPlan(...) method inside (not assertPlan)

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/AbstractPlannerTest.java
```diff
@@ -455,20 +504,34 @@ protected <T extends RelNode> void assertPlan(
...
+
+        assertPlan(builder, predicate);
+    }
+
+    /** */
+    private <T extends RelNode> void invalidatePlan(
+        TestPlanningContextBuilder ctxBuilder,
+        Predicate<T> predicate
+    ) throws Exception {
+        IgniteRel plan = physicalPlan(plannerCtx(ctxBuilder.query, ctxBuilder.schemas, ctxBuilder.planListener,
```
alex-plekhanov: Move `plannerCtx(...)` to `TestPlanningContextBuilder.build()` method.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/HashJoinPlannerTest.java
```diff
@@ -28,10 +28,10 @@
 import org.apache.ignite.internal.processors.query.calcite.trait.IgniteDistributions;
 import org.apache.ignite.internal.util.typedef.F;
 import org.apache.ignite.internal.util.typedef.internal.CU;
+import org.apache.ignite.testframework.GridTestUtils;
```
alex-plekhanov, line 31: Redundant change

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java
```diff
@@ -58,21 +57,21 @@ public class SortNode<Row> extends MemoryTrackingNode<Row> implements SingleNode
...
-        assert fetch == null || fetch.get() >= 0;
-        assert offset == null || offset.get() >= 0;
+        assert fetch == -1 || fetch >= 0;
+        assert offset >= 0;
 
-        limit = fetch == null ? -1 : fetch.get() + (offset == null ? 0 : offset.get());
+        limit = fetch == -1 ? -1 : (fetch > Long.MAX_VALUE - offset ? -1 : fetch + offset);
 
-        if (limit < 0)
+        if (limit < 1 || limit > Integer.MAX_VALUE)
```
alex-plekhanov, line 76: Can we rich this line with limit == 0? If yes, it's more correct to use `limit <= 1' or add other logic to return empty result

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -84,9 +87,6 @@
 /** Validator. */
 @Value.Enclosing
 public class IgniteSqlValidator extends SqlValidatorImpl {
```
tkalkirill, line 88: Maybe we should use `SqlTypeName.DECIMAL`, similar to the CALCITE-7624?
org.apache.calcite.sql.validate.SqlValidatorImpl#handleOffsetFetch

A question regarding rounding might arise; we can use the Calcite `org.apache.calcite.adapter.enumerable.FetchOffsetRoundingPolicy` interface, and it will handle rounding when creating nodes.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -241,54 +241,81 @@ private void validateTableModify(SqlNode table) {
...
+        else if (n instanceof SqlDynamicParam dynamicParam) {
             if (F.isEmpty(parameters))
                 return;
 
-            int idx = ((SqlDynamicParam)n).getIndex();
+            if (dynamicParam.getIndex() < parameters.length) {
+                Object param = parameters[dynamicParam.getIndex()];
+
+                if (!(param instanceof Number)) {
+                    SqlTypeName expectType = SqlTypeName.BIGINT;
```
tkalkirill: If we support the `SqlTypeName.DECIMAL` as in Calcite, then it will need to be fixed.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -241,54 +241,81 @@ private void validateTableModify(SqlNode table) {
...
             }
         }
     }
 
+    /** */
+    private void checkLimitOffset(Number offsetFetchLimit, @Nullable SqlNode n, String nodeName) {
+        try {
+            long res = IgniteMath.convertToLongExact(offsetFetchLimit);
+
+            if (res < 0)
```
tkalkirill, line 305: If we support the `SqlTypeName.DECIMAL` as in Calcite, then it will need to be fixed.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
```diff
@@ -85,38 +90,46 @@ public LimitNode(
...
-        if (waiting == -1)
+        if (waiting == NOT_WAITING)
             return;
 
         assert downstream() != null;
 
-        waiting = -1;
+        waiting = NOT_WAITING;
 
         downstream().end();
```
alex-plekhanov, line 137: `requested` field check required here. And `requested = 0` before `downstream().end()`. See related problem https://issues.apache.org/jira/browse/IGNITE-28925

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
zstan: done

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java
zstan: done

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java
zstan: done

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/AbstractPlannerTest.java
zstan: missed, removed

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/AbstractPlannerTest.java
zstan: done

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/AbstractPlannerTest.java
zstan: done

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/AbstractPlannerTest.java
zstan: done

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/AbstractPlannerTest.java
zstan: done

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/HashJoinPlannerTest.java
zstan: otherwize it resolves as : org.apache.ignite.internal.processors.query.calcite.planner.AbstractPlannerTest#assertThrows

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java
zstan: no we can\`t add assert

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
zstan: done

zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
zstan: this issue is an adoption of related ignite-3 issue, let\`s move sequentially here and implement it during different activity, also i suppose that Rounding policy need to work equal for aggregates avg too, need additional tests and check that it work the same as in common DB\`s

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
zstan: partially revert, other part is a covered by DynamicParametersPlannerTest

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/AbstractNode.java
```diff
@@ -37,6 +37,9 @@
  * Abstract node of execution tree.
  */
 public abstract class AbstractNode<Row> implements Node<Row> {
+    /** Special value to highlight that all rows were received and we do not expect more. */
+    static final int NOT_WAITING = -1;
```
tkalkirill, line 41: Why is this constant declared in `AbstractNode` if the class does not have a corresponding waiting state? Should it be kept in the concrete nodes instead? We should also handle the similar constants declared in other subclasses consistently.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
```diff
@@ -17,66 +17,73 @@
...
+        long offset,
+        long fetch
     ) {
         super(ctx, rowType);
 
-        offset = offsetNode == null ? 0 : offsetNode.get();
-        fetch = fetchNode == null ? 0 : fetchNode.get();
-        this.fetchNode = fetchNode;
+        this.offset = offset;
+        rowsSummary = fetch == -1 ? Long.MAX_VALUE : IgniteMath.addExact(fetch, offset);
```
tkalkirill: Check `offset + fetch` for overflow in the constructor and remove the redundant `rowsSummary` field. After validation, use the immutable `offset` and `fetch` values directly in subsequent calculations.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
```diff
@@ -17,66 +17,73 @@
...
+     * @param offset How many rows need to be skipped.
+     * @param fetch How many rows need to be processed.
      */
     public LimitNode(
         ExecutionContext<Row> ctx,
         RelDataType rowType,
-        Supplier<Integer> offsetNode,
-        Supplier<Integer> fetchNode
+        long offset,
+        long fetch
```
tkalkirill, line 63: The documentation says nothing about what `-1` means.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
```diff
@@ -17,66 +17,73 @@
...
     ) {
         super(ctx, rowType);
 
-        offset = offsetNode == null ? 0 : offsetNode.get();
-        fetch = fetchNode == null ? 0 : fetchNode.get();
-        this.fetchNode = fetchNode;
+        this.offset = offset;
+        rowsSummary = fetch == -1 ? Long.MAX_VALUE : IgniteMath.addExact(fetch, offset);
+        this.fetch = fetch == -1 ? 0 : fetch;
```
tkalkirill: The documentation says nothing about what `0` means.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java
```diff
@@ -58,21 +57,21 @@ public class SortNode<Row> extends MemoryTrackingNode<Row> implements SingleNode
     public SortNode(
         ExecutionContext<Row> ctx, RelDataType rowType,
         Comparator<Row> comp,
-        @Nullable Supplier<Integer> offset,
-        @Nullable Supplier<Integer> fetch
+        long offset,
+        long fetch
```
tkalkirill, line 67: The documentation says nothing about what `-1` means.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java
```diff
@@ -58,21 +57,21 @@ public class SortNode<Row> extends MemoryTrackingNode<Row> implements SingleNode
...
     ) {
         super(ctx, rowType);
 
-        assert fetch == null || fetch.get() >= 0;
-        assert offset == null || offset.get() >= 0;
+        assert fetch == -1 || fetch > 0;
```
tkalkirill: You need to include at least the value in the error message.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java
```diff
@@ -58,21 +57,21 @@ public class SortNode<Row> extends MemoryTrackingNode<Row> implements SingleNode
...
+        assert offset >= 0;
```
tkalkirill: You need to include at least the value in the error message.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
```diff
@@ -1050,4 +1054,33 @@ private ScanStorageNode<Row> createStorageScan(
             otherColMapping
         );
     }
+
+    /** */
+    private long validateAndGetFetchOffsetParams(RexNode node, String op) {
```
tkalkirill, line 1069: Let’s have two separate methods for `fetch` and `offset`; validation can be shared, but each will handle nulls in its own way. This will make the code cleaner and easier to understand.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
```diff
@@ -1050,4 +1054,33 @@ private ScanStorageNode<Row> createStorageScan(
...
+        Supplier<Object> scalar = expressionFactory.execute(node);
+        Object param = scalar.get();
+
+        if (!(param instanceof Number)) {
+            String actual = param == null ? "null" : param.getClass().getSimpleName();
+            throw new IgniteSQLException(IgniteResource.INSTANCE.incorrectDynamicParameterType("BIGINT", actual).str(),
```
tkalkirill, line 1075: Why use a string literal when you can use the type name `SqlTypeName.BIGINT`?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -241,54 +241,80 @@ private void validateTableModify(SqlNode table) {
...
-        }
-
-        super.validateNamespace(namespace, targetRowType);
+        validateFetchOffset(select.getFetch(), "fetch / limit");
+        validateFetchOffset(select.getOffset(), "offset");
     }
 
     /**
-     * @param n Node to check limit.
+     * Invalidate fetch/offset params restrictions.
```
tkalkirill: Why "Invalidate"?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -241,54 +241,80 @@ private void validateTableModify(SqlNode table) {
...
+     *
+     * @param n        Node to check limit.
```
tkalkirill: Why just `LIMIT`? There’s `OFFSET`/`FETCH` and `LIMIT` here.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -241,54 +241,80 @@ private void validateTableModify(SqlNode table) {
...
+            if (res < 0)
+                throw newValidationError(n, IgniteResource.INSTANCE.illegalFetchLimit(nodeName));
+        }
+        catch (ArithmeticException e) {
+            throw newValidationError(n, IgniteResource.INSTANCE.illegalFetchLimit(nodeName));
+        }
+    }
+
+    /** {@inheritDoc} */
+    @Override protected void validateNamespace(SqlValidatorNamespace namespace, RelDataType targetRowType) {
```
tkalkirill, line 314: What does this change have to do with it?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java
```diff
@@ -78,10 +79,10 @@ private void checkLimitSort(int offset, int fetch) {
 
         RootNode<Object[]> rootNode = new RootNode<>(ctx, rowType);
 
-        SortNode<Object[]> sortNode = new SortNode<>(ctx, rowType, F::compareArrays, () -> offset,
-            fetch == 0 ? null : () -> fetch);
+        SortNode<Object[]> sortNode = new SortNode<>(ctx, rowType, F::compareArrays, offset,
+            fetch == 0 ? -1 : fetch);
```
tkalkirill: Magic values.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java
```diff
@@ -109,7 +110,7 @@ private void checkLimit(int offset, int fetch) {
         RelDataType rowType = TypeUtils.createRowType(tf, int.class);
 
         RootNode<Object[]> rootNode = new RootNode<>(ctx, rowType);
-        LimitNode<Object[]> limitNode = new LimitNode<>(ctx, rowType, () -> offset, fetch == 0 ? null : () -> fetch);
+        LimitNode<Object[]> limitNode = new LimitNode<>(ctx, rowType, offset, fetch == 0 ? -1 : fetch);
```
tkalkirill: Magic values.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DynamicParametersIntegrationTest.java
```diff
@@ -150,6 +150,9 @@ public void testDynamicParameters() {
         assertQuery("SELECT name LIKE '%' || ? || '%' FROM person where name is not null").withParams("go")
             .returns(true).returns(false).returns(false).returns(false).check();
 
+        assertQuery("SELECT id FROM person ORDER BY id LIMIT ?").withParams(1).returns(0).check();
+        assertQuery("SELECT id FROM person ORDER BY id LIMIT ?").withParams(new BigDecimal(1)).returns(0).check();
```
tkalkirill, line 161: Missing, Double, Float, Long.
Missing floating-point value, for example, 1.5.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
```diff
@@ -104,15 +104,15 @@ public void testNestedLimitOffsetWithUnion() {
     /** Tests correctness of fetch / offset params. */
     @Test
     public void testInvalidLimitOffset() {
```
tkalkirill, line 106: Missing, Double, Float, Long.
Missing floating-point value, for example, 1.5.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/DynamicParametersPlannerTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.ignite.internal.processors.query.calcite.schema.IgniteSchema;
+import org.junit.Test;
+
+import static org.apache.ignite.internal.processors.query.calcite.trait.IgniteDistributions.single;
+
+/** */
+public class DynamicParametersPlannerTest extends AbstractPlannerTest {
+    /** Dynamic parameters in LIMIT / OFFSET. */
+    @Test
+    public void testLimitOffset() throws Exception {
```
tkalkirill, line 31: Missing, Double, Float, Long.
Missing floating-point value, for example, 1.5.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
tkalkirill: > this issue is an adoption of related [ignite-3](https://issues.apache.org/jira/browse/IGNITE-3) issue### 

Which issue are you referring to?

> let`s move sequentially here and implement it during different activity

Why? In this ticket, you're checking how it handles BigDecimal, and it’s not obvious to me how it will behave with floating-point numbers.

> Rounding policy need to work equal for aggregates avg too

Rounding within a set of values ​​is one thing, while rounding during calculations is another. That is precisely why the interface is named `FetchOffsetRoundingPolicy` to avoid conflating the two types of logic.

> need additional tests and check that it work the same as in common DB`s

As a Calcite ticket investigation revealed, different DBMSs handle floating-point numbers differently for FETCH, OFFSET, and LIMIT. That is why this interface was introduced.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
tkalkirill: There also appears to be a logic error when passing the `BigDecimal`: it corresponds to the `DECIMAL` type, not `BIGINT`.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/AbstractNode.java
zstan: No, it shouldn\`t, cause it abstract, look this is not main goal of this issue, if you think that some constants need to be aligned - fill an issue and we align them

tkalkirill: However, we need to sort out this constant: should it be specific to each node, like similar constants or should it be more universal and used across other implementations.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
zstan: I miss your idea:
1. **Check for overflow** - checked here: IgniteMath.addExact
2. **use the immutable offset and fetch values** you suggest smth like : 

```
    private boolean hasMoreData() {
        return fetch + offset < rowsSummary;
    }
```
= evaluate it each call ?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java
zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java
zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
zstan: change doc for :

```
    /** How many rows need to be processed. */
    private final long fetch;
```
is it ok ?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
tkalkirill: No, what do you mean by "zero"?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
tkalkirill: Get rid of `rowsSummary` entirely so it isn't used, and add an overflow check once in the constructor.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
zstan: I miss you idea here too:
```
        long offset = rel.offset() == null ? 0 : validateAndGetFetchOffsetParams(rel.offset(), "offset");
        long fetch = rel.fetch() == null ? -1 : validateAndGetFetchOffsetParams(rel.fetch(), "fetch");
```
you suggest to handle null check logic into 2 different func which are call 3rd common validation ? I didn\`t see any simplification here (( 
```
        rel.offset() == null ? 0
        rel.fetch() == null ? -1
```

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
zstan: miss here, plz suggest your changes ? 
BIGINT is not about actual param, it\`s about inner expected check PG behavior:
[1] https://onecompiler.com/postgresql#draft-496f

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
tkalkirill: Let's look at an example: suppose you have two calls for `fetch`.
```
long offset = rel.offset() == null ? 0 : validateAndGetFetchOffsetParams(rel.offset(), "offset");
long offset = rel.offset == null ? 0 : validateAndGetFetchOffsetParams(rel.offset, "offset");
```
Issues:
1. Validation and value retrieval are combined in a single method, which isn't exactly clean from an architectural standpoint.
2. The logic for handling a null is duplicated; the result is the same in both cases, so this code duplication could be avoided.
3. The name implies that we are validating both FETCH and OFFSET simultaneously, rather than just one or the other.

We can make an improvement, even if it's not perfect:
```
long offset = validateAndGetOffset(rel.offset());
long offset = validateAndGetOffset(rel.offset);
```

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
zstan: brr, "Node to check limit." - is not about SqlNode, it\`s about node limitations

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
tkalkirill: To do it right, currently, you can pass any number as a parameter, yet for some reason, only the BIGINT is mentioned here; I don't understand that. What will the metadata contain for these parameters?

We could do it the way it is currently implemented in Calcite - DECIMAL, and update the metadata as well.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
zstan: i see it too, i wonder, seems some kind of UI bug ? compare this class with master and seems it the same = not my change.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java
zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/AbstractNode.java
zstan: It need to be used through all Nodes implementstions instead of  '-1'

tkalkirill: Why? Does every node have a field related to waiting?
"-1" doesn't necessarily mean that nothing else was expected.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
zstan: @tkalkirill you just re-phrase your upper message, i already reply to it, once more : 
1.  IgniteMath.addExact - overflow check
2.    
 ```
 private boolean hasMoreData() {
        return rowsProcessed < rowsSummary;
    }
```
rowsSummary - usage

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java (hunk 24 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java]: directly how it written in documentation: "How many rows need to be processed"

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java (hunk 22 above)
Comment (tkalkirill) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java]: Okay, we could just get rid of the extra field right away, but if it's absolutely necessary, we can keep it.

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java (hunk 30 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java]: done

File: modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java (hunk 34 above)
Comment (zstan) [reply modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java]: done

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java (hunk 26 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java]: done

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/AbstractNode.java (hunk 21 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/AbstractNode.java:41]: 1. Not every but i see no harm still to have this constant here.
2. ok, what else it mean in current implementation ? plz give appropriate code link

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java (hunk 22 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java]: extra field need to avoid repeated calculations like : fetch + offset - that was the case from previous reviewer

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java (hunk 28 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java:1069]: ok, done [plz resolve comments that are ok to you, otherwize we have already long history here]

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java (hunk 17 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java:88]: > Which issue are you referring to?
written in initial issue: 
https://issues.apache.org/jira/browse/IGNITE-24227

File: modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DynamicParametersIntegrationTest.java (hunk 35 above)
Comment (zstan) [reply modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DynamicParametersIntegrationTest.java:161]: done, plz check

File: modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java (hunk 36 above)
Comment (zstan) [reply modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java:106]: done but not here - check test_limit.test, we need to store such tests in script suite

File: modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/DynamicParametersPlannerTest.java (hunk 37 above)
Comment (zstan) [reply modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/DynamicParametersPlannerTest.java:31]: plz clarify - what kind of tests need to be appended here ?

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java (hunk 29 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java:1075]: I see changes in your branch, but seems correct decision will be in 1.43 (really we need more tests here) plz check my approach, i fix a bit

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java (hunk 17 above)
Comment (tkalkirill) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java:88]: Thanks for the link; I also left some questions.

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/AbstractNode.java (hunk 21 above)
Comment (tkalkirill) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/AbstractNode.java:41]: Here are implementations that contain exactly the same constant:
org.apache.ignite.internal.processors.query.calcite.exec.rel.MergeJoinNode#NOT_WAITING
org.apache.ignite.internal.processors.query.calcite.exec.rel.AbstractRightMaterializedJoinNode#NOT_WAITING

Of course, you don't have to consolidate this or modify the existing classes right now.
However, adding a shared constant for all subclasses isn't the right approach either; it would be better to follow the existing pattern—that is, adding a specific constant to each implementation:
org.apache.ignite.internal.processors.query.calcite.exec.rel.LimitNode#NOT_WAITING
org.apache.ignite.internal.processors.query.calcite.exec.rel.SortNode#NOT_WAITING

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java (hunk 24 above)
Comment (tkalkirill) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java]: This description is misleading because `fetch == 0` has two opposite meanings: either no rows should be returned when 1FETCH 01 is specified, or all remaining rows should be returned when `FETCH` is not specified. These two opposite cases should not be represented by the same value.

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java (hunk 27 above)
Comment (tkalkirill) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java]: I don't see any changes; I expected:
```
assert offset >= 0 : offset;
```

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java (hunk 31 above)
Comment (tkalkirill) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java]: Previously, the method name contained `Limit`, so this parameter description was at least consistent with it. Now that the method has been renamed to `validateFetchOffset`, update the Javadoc and parameter names accordingly: `n` may represent a FETCH/LIMIT or OFFSET expression, and `nodeName` is actually a clause name used in error messages.

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java (hunk 32 above)
Comment (tkalkirill) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java:314]: I don't know, try doing a rebase.

File: modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java (hunk 33 above)
Comment (tkalkirill) [reply modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java]: Replace these methods with constants declared separately in each corresponding class. `LimitNode` and `SortNode` should each own the constants that define their internal state instead of sharing them through a common accessor. This project does not use methods to expose fixed constant values, so keep the implementation consistent with the existing project style.

File: modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java (hunk 34 above)
Comment (tkalkirill) [reply modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java]: Replace these methods with constants declared separately in each corresponding class. `LimitNode` and `SortNode` should each own the constants that define their internal state instead of sharing them through a common accessor. This project does not use methods to expose fixed constant values, so keep the implementation consistent with the existing project style.

File: modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/DynamicParametersPlannerTest.java (hunk 37 above)
Comment (tkalkirill) [reply modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/DynamicParametersPlannerTest.java:31]: My mistake; nothing needs to be done.

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java (hunk 29 above)
Comment (tkalkirill) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java:1075]: The parameter contract is inconsistent. The validation accepts any `Number`, including fractional `Double` and `BigDecimal` values, and converts them to `long`, while the error message reports that `BIGINT` is required and the validated parameter type is set to `DECIMAL`. Define one supported SQL type and make the metadata, validation rules, and error messages consistent with it.

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/AbstractNode.java (hunk 21 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/AbstractNode.java:41]: aligned

File: modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java (hunk 33 above)
Comment (zstan) [reply modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java]: done

File: modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java (hunk 34 above)
Comment (zstan) [reply modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitExecutionTest.java]: done

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java (hunk 27 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java]: you right, missed due to huge numbed of comments, fixed

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java (hunk 24 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java]: ok, i tried to fix it ..

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java (hunk 31 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java]: fixed, if you disagree - plz suggest your variant

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java (hunk 29 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java:1075]: > The parameter contract is inconsistent. The validation accepts any Number, including fractional Double and BigDecimal values, and converts them to long, while the error message reports that BIGINT is required

I see no problem here, i already show you above that BIGINT is correlates with PG , also it not corresponds to input parameters at all.

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java (hunk 24 above)
Comment (tkalkirill) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java]: The documentation is fine, but replace the duplicated `-1` sentinel here:
```
rowsSummary = fetch == FETCH_DEFAULT ? Long.MAX_VALUE : IgniteMath.addExact(fetch, offset);
```
Also update the constructor parameter documentation to reference `{@link #FETCH_DEFAULT}` instead of mentioning the literal `-1` directly. This keeps the implementation and documentation tied to the declared constant.

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java (hunk 25 above)
Comment (tkalkirill) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java:67]: Use the declared default constants consistently throughout this constructor. Replace the literal `-1` in the `fetch` parameter documentation with `{@link #FETCH_DEFAULT}` and update the assertion accordingly:
```
assert fetch == FETCH_DEFAULT || fetch > 0 : "Unexpected fetch = " + fetch;
```
The delegating constructor should also pass `OFFSET_DEFAULT` and `FETCH_DEFAULT` instead of the literal `0` and `-1` values.

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java (hunk 24 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java]: done

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java (hunk 25 above)
Comment (zstan) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java:67]: done

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java (hunk 29 above)
Comment (tkalkirill) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java:1075]: Okay, let’s keep `BIGINT` to stay consistent with PG. However, a direct check of Ignite parameter metadata shows that `LIMIT ?`, `OFFSET ?`, and `FETCH FIRST ? ROWS ONLY` are all reported as `INTEGER (int, precision 10)`, not `BIGINT`. Therefore, the parameter metadata and validation are currently inconsistent. Please make them agree and report `BIGINT` in the metadata as well.

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java (hunk 24 above)
Comment (tkalkirill) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java]: You didn't fix the documentation.

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java (hunk 25 above)
Comment (tkalkirill) [reply modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java:67]: You didn't fix the documentation.
You didn't fix the second constructor.

File: modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java [hunk 38]
```diff
@@ -17,66 +17,79 @@
...
+    /** Upper requested rows. */
+    private int requested;
+
     /**
      * Constructor.
      *
      * @param ctx Execution context.
      * @param rowType Row type.
+     * @param offset How many rows need to be skipped.
+     * @param fetch How many rows need to be processed, {@code -1} if param is undefined.
```
tkalkirill:
```suggestion
     * @param fetch How many rows need to be processed, {@link #FETCH_DEFAULT} if param is undefined.
```

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java
```diff
@@ -53,26 +58,26 @@ public class SortNode<Row> extends MemoryTrackingNode<Row> implements SingleNode
      * @param ctx Execution context.
      * @param comp Rows comparator.
      * @param offset Offset.
-     * @param fetch Limit.
+     * @param fetch How many rows need to be processed, {@code -1} if param is undefined.
```
tkalkirill:
```suggestion
     * @param fetch How many rows need to be processed, {@link #FETCH_DEFAULT} if param is undefined.
```

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java
```diff
@@ -81,7 +86,7 @@ public SortNode(
      * @param comp Rows comparator.
      */
     public SortNode(ExecutionContext<Row> ctx, RelDataType rowType, Comparator<Row> comp) {
-        this(ctx, rowType, comp, null, null);
+        this(ctx, rowType, comp, 0, -1);
```
tkalkirill:
```suggestion
        this(ctx, rowType, comp, OFFSET_DEFAULT, FETCH_DEFAULT);
```

zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java
zstan: done thanks

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/SortNode.java
zstan: done

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/LimitNode.java
zstan: yep, missed, now it\`s ok

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
zstan: i think it\`s ok for now, append test

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
```diff
@@ -567,8 +571,8 @@ private boolean hasExchange(RelNode rel) {
                 ctx,
                 rowType,
                 idxBndRel.first() ? cmp : cmp.reversed(),
-                null,
-                () -> 1
+                0,
```
tkalkirill, line 574:
```suggestion
                SortNode.OFFSET_DEFAULT,
```

tkalkirill (approved):

PR #13319, opened 2026-07-02
IGNITE-28848 ZooKeeper discovery: stream marshalZip through DeflaterOutputStream to cut allocations and peak memory

anton-vinogradov:
JIRA: [IGNITE-28848](https://issues.apache.org/jira/browse/IGNITE-28848) (follow-up to IGNITE-28835)

**Why this matters.** `marshalZip()` sits on the two busiest control-plane paths of ZooKeeper discovery: the coordinator re-marshals the whole `ZkDiscoveryEventsData` on **every** discovery event (node joins/failures and every custom message — cache create/destroy, exchange, services, snapshots, ...), and join data / data-for-joined is megabytes in clusters with hundreds of caches (split across znodes by `jute.maxbuffer`). The current code materializes the uncompressed form before compressing: ~6x the uncompressed size in garbage and ~3x at peak **per call** — so a mass join, client reconnect storm or rolling restart hits the coordinator with GC pressure and heap spikes exactly when it is busiest, and the pauses land on the discovery critical path (events are processed sequentially, the whole cluster sees slower event propagation). `Deflater.end()` is never called either, so zlib native memory waits for the Cleaner.

**Change:** stream marshalling straight through `BufferedOutputStream -> DeflaterOutputStream`: the uncompressed form is never materialized, peak memory is bounded by the compressed output, and the deflater is released deterministically on close. The `BufferedOutputStream` is essential — `ObjectOutputStream` writes ~1 KB blocks, and unbuffered per-block JNI deflate calls cost +14–20% time. The private `zip()` helper is removed (`marshalZip` was its only caller); `unzip()` stays (still used for `ATTR_SECURITY_SUBJECT_V2`); the read side (`unmarshalZip`) already streams. Same pattern as `DiscoveryMessageParser.marshalZip` in the same package.

**Compatibility:** zlib format is unchanged (default `Deflater` settings in both versions) — old nodes inflate the bytes as is, rolling upgrade safe.

JMH (avgt, JDK 17, `-prof gc`), old vs new:

| payload (marshalled → zipped) | time old, us/op | time new, us/op | alloc old, B/op | alloc new, B/op |
|---|---|---|---|---|
| MAP_2K (1.6 KB → 0.7 KB) | 21.3 ± 11.5 | 15.4 ± 0.7 | 12,544 | 14,368 (+15%) |
| MAP_100K (132 KB → 40 KB) | 1,987 ± 512 | 1,931 ± 264 | 1,065,204 | 283,835 (−73%) |
| LIST_1M (1.66 MB → 318 KB) | 17,661 ± 3,433 | 16,825 ± 980 | 9,891,984 | 1,788,235 (−82%) |

Allocations drop 73–82% on realistic payloads; per-call time is on par or better. The small-payload trade-off (+15% B/op — bounded fixed overhead of the 8 KB buffer, break-even ~3 KB) is analyzed in the JIRA description.

### DIFF
--- a/modules/zookeeper/src/main/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryImpl.java
+++ b/modules/zookeeper/src/main/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryImpl.java
@@ -4074,7 +4074,8 @@ public class ZookeeperDiscoveryImpl {
 
         GridByteArrayOutputStream out = new GridByteArrayOutputStream();
 
-        // Buffer coalesces ~1KB ObjectOutputStream blocks into fewer JNI deflate calls.
+        // BufferedOutputStream's 8 KB buffer coalesces JdkMarshaller's ~1 KB ObjectOutputStream
+        // block-data writes into fewer Deflater JNI calls.
         try (BufferedOutputStream zipOut = new BufferedOutputStream(new DeflaterOutputStream(out))) {
             U.marshal(marsh, obj, zipOut);
         }
--- a/modules/zookeeper/src/main/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryImpl.java
+++ b/modules/zookeeper/src/main/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryImpl.java
@@ -17,7 +17,9 @@
 
 package org.apache.ignite.spi.discovery.zk.internal;
 
+import java.io.BufferedOutputStream;
 import java.io.ByteArrayInputStream;
+import java.io.IOException;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -39,7 +41,7 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.zip.DataFormatException;
-import java.util.zip.Deflater;
+import java.util.zip.DeflaterOutputStream;
 import java.util.zip.Inflater;
 import java.util.zip.InflaterInputStream;
 import org.apache.ignite.Ignite;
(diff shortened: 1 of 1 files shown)


modules/zookeeper/src/main/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryImpl.java
```diff
@@ -4064,33 +4066,20 @@ private <T> T unmarshalZip(byte[] zipBytes) throws Exception {
...
 
-        GridByteArrayOutputStream out = new GridByteArrayOutputStream(bytes.length);
-
-        final byte[] buf = new byte[bytes.length];
-
-        while (!deflater.finished()) {
-            int cnt = deflater.deflate(buf);
-
-            out.write(buf, 0, cnt);
+        // Buffer coalesces ~1KB ObjectOutputStream blocks into fewer JNI deflate calls.
```
alex-plekhanov: Why 1KB? Default BufferedOutputStream buffer size is 8KB

modules/zookeeper/src/main/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryImpl.java
```diff
@@ -4064,33 +4066,20 @@ private <T> T unmarshalZip(byte[] zipBytes) throws Exception {
...
-    /**
-     * @param bytes Bytes to compress.
-     * @return Zip-compressed bytes.
-     */
-    private static byte[] zip(byte[] bytes) {
-        Deflater deflater = new Deflater();
-
-        deflater.setInput(bytes);
-        deflater.finish();
+        GridByteArrayOutputStream out = new GridByteArrayOutputStream();
```
alex-plekhanov: Do we know average compression level for data, used here? Maybe it's worth to make reasonable buffer size assumption here (for example `bytes.length / 2` for 2:1 compression level) to avoid buffer reallocation and copy during resize?

modules/zookeeper/src/main/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryImpl.java
anton-vinogradov: The `~1 KB` referred to JdkMarshaller's ObjectOutputStream block-data granularity (MAX_BLOCK_SIZE = 1024), not the BufferedOutputStream buffer — the default 8 KB buffer is exactly what coalesces those ~1 KB blocks into fewer Deflater JNI calls. Reworded the comment to make that explicit.

modules/zookeeper/src/main/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoveryImpl.java
anton-vinogradov: We intentionally no longer materialize the uncompressed byte[] (that's the whole point of streaming here — cutting peak memory), so `bytes.length` isn't available for a ratio-based estimate. marshalZip is also called per-component for potentially many small payloads (joiningNodeData via `F.viewReadOnly`), so a large fixed initial capacity would over-allocate for the common small case and work against the allocation reduction this change targets. `GridByteArrayOutputStream` grows by doubling (amortized O(n)), and `toByteArray()` returns an exact-sized copy, so pre-sizing wouldn't reduce the retained/peak result memory — only the transient resize churn. If you'd still like to trim the initial doublings for the large join-data case, I'd rather seed a modest fixed capacity (e.g. 512 B, matching DeflaterOutputStream's flush chunk) than a ratio of the now-unknown input size. WDYT?

anton-vinogradov: Fair — agreed, pre-sizing is about cutting the resize garbage/copies, and that's worth doing for the large join-data case. One clarification on the ratio idea: we deliberately don't compute the uncompressed size anymore, so there's no input length to take 1/3 of — obtaining it would mean re-materializing the uncompressed `byte[]`, which is exactly the allocation this change removes. So I can't scale the seed to the input without giving the peak-memory win back. Instead I seed the output buffer with a fixed 8 KB initial capacity (same size as the BufferedOutputStream we already allocate, so no new cost tier for small messages), which drops the 32→…→N doubling cascade. Happy to bump the constant if you'd like it more aggressive for the megabyte case — I just can't make it proportional to the now-unknown input size.

anton-vinogradov: Per our discussion — reverted the seed, keeping the default `GridByteArrayOutputStream` sizing. Since the uncompressed size isn't available in the streaming path, the seed could only be a fixed constant rather than proportional to the input, so a fixed 8 KB isn't clearly worth it. Reverted in 3295ae266f2.

alex-plekhanov (approved):

PR #13274, opened 2026-06-24
IGNITE-28801 Calcite. Bump calcite dependency version to 1.42

zstan:
https://issues.apache.org/jira/browse/IGNITE-28801

PR #13324, opened 2026-07-03
IGNITE-28852 SQL Calcite: Add test for binary operator in UPDATE

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-28852


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -152,6 +152,13 @@ public IgniteSqlValidator(
         validateUpdateFields(call);
 
         super.validateUpdate(call);
+
+        SqlSelect srcSelect = call.getSourceSelect();
+        SqlValidatorScope scope = getWhereScope(srcSelect);
+
+        for (SqlNode expr : call.getSourceExpressionList()) {
+            deriveType(scope, expr);
```
alex-plekhanov: Redundant braces

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/OperatorsExtensionIntegrationTest.java
```diff
@@ -158,6 +167,56 @@ public void testCustomAggregateFunction() {
             .check();
     }
 
+    /** */
+    @Test
```
alex-plekhanov: Too complicated test. We can reproduce this problem much simplier without any extensions with test like:
```
    @Test
    public void testDmlIntervalArithmetic() {
        sql("CREATE TABLE test(ts TIMESTAMP)");
        sql("INSERT INTO test VALUES (?)", Timestamp.valueOf("2021-01-01 00:00:01"));
        sql("UPDATE test SET ts = ts - INTERVAL 1 SECOND");
        assertQuery("SELECT * FROM test").returns(Timestamp.valueOf("2021-01-01 00:00:00")).check();
    }
```
WIth this code we can prove that standard Calcite date arithmetic has the same problem (at least we can easily check it with pure Apache Calcite), report bug to Calcite, and remove our patch later.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -152,6 +152,13 @@ public IgniteSqlValidator(
         validateUpdateFields(call);
 
         super.validateUpdate(call);
+
+        SqlSelect srcSelect = call.getSourceSelect();
+        SqlValidatorScope scope = getWhereScope(srcSelect);
```
alex-plekhanov: Why `where` scope? Looks like `selectItems` are validated within `select` scope and `selectItems` derived from `sourceExpressionList`. Maybe `sourceExpressionList` also should be validated within `select` scope? (I'm not quite sure about this. Maybe there is a reason for `where` scope?)

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/OperatorsExtensionIntegrationTest.java
tkalkirill: I tried running this test using vanilla Apache Calcite and it didn't fail; it looks like this has already been fixed in [CALCITE-7276](https://issues.apache.org/jira/browse/).
I think I'll wait for https://github.com/apache/ignite/pull/13274 to be completed and then check again.
If it is fixed, then we can simply add the tests to this PR.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
tkalkirill: It seems more correct to use the `WHERE` clause, because we need to work only with aliases of the table being updated, rather than the entire `SELECT` statement, which might include auxiliary aliases.

alex-plekhanov (approved):

PR #13326, opened 2026-07-03
IGNITE-28854 CompressedMessageSerializer.readFrom() spins forever on a null chunk from a corrupted or incompatible peer

anton-vinogradov:
In `CompressedMessageSerializer.readFrom()`, state 2 reads the next compressed chunk. `DirectByteBufferStream.readByteArray()` returns `null` (with `lastFinished == true`) when the wire contains the null-array marker (length `-1`). A well-behaved sender never writes `null` at this position (`writeTo()` only enters state 2 with a non-null chunk), but a corrupted stream or a malicious/incompatible peer can produce it: the `-1` length is accepted, the null branch does not advance the reader state, and the `while (true)` loop re-executes state 2 forever — the grid-nio worker thread spins at 100% CPU and its stripe stops processing messages, effectively degrading the whole node.

The fix treats a null chunk at this protocol position as a protocol violation: `IgniteException` with a descriptive message is thrown instead of silently looping. The exception propagates out of message decoding and the connection is closed.

Covered by a unit test that feeds the reader a buffer with the null-array marker at the chunk position and asserts the exception is thrown.

Note: this file is also touched by #13325 (IGNITE-28853, direct-buffer accumulator rework of the same method). The overlap is trivial (`case 2` is structurally identical in both versions), so the PRs can be merged in any order; whichever lands second resolves a one-hunk conflict.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/managers/communication/CompressedMessageSerializer.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/managers/communication/CompressedMessageSerializer.java
@@ -18,6 +18,7 @@
 package org.apache.ignite.internal.managers.communication;
 
 import java.nio.ByteBuffer;
+import org.apache.ignite.IgniteException;
 import org.apache.ignite.plugin.extensions.communication.MessageReader;
 import org.apache.ignite.plugin.extensions.communication.MessageSerializer;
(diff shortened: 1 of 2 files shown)


wernerdv:
Implemented in https://github.com/apache/ignite/commit/fc33c9c6867d753ba726defd8973b4ec6acdad80

PR #13327, opened 2026-07-03
IGNITE-28855 Client node permanently loses discovery events delivered right after the reconnect response

anton-vinogradov:
https://issues.apache.org/jira/browse/IGNITE-28855

Regression of IGNITE-26111. Before it, `ClientImpl.SocketStream` owned a single `BufferedInputStream` shared by every reader of the discovery socket. After the refactoring, `Reconnector` and `SocketReader` each create their own `TcpDiscoveryIoSession`, each wrapping the raw socket into a new `BufferedInputStream`. When the router writes the reconnect response and a subsequent discovery message back-to-back (single TCP segment — typical during a topology storm), the Reconnector's session reads ahead and buffers the bytes of the following message; the socket is then handed over to `SocketReader`, which attaches a fresh session to the raw socket — the buffered bytes are silently discarded and the message is lost forever. Depending on whether more topology messages follow, the client either fails with `AssertionError: lastVer=N, newVer=N+2` in `updateTopologyHistory` or silently stays on a stale topology.

Fix: create `TcpDiscoveryIoSession` once per socket (in `sendJoinRequest`, the same session that has read the handshake response) and carry it inside `SocketStream`; `Reconnector` and `SocketReader` reuse it. This restores the pre-IGNITE-26111 invariant: a single buffered reader per socket.

Reproducer `TcpDiscoveryClientTopologyGapTest` mirrors the failing `IgniteCachePartitionLossPolicySelfTest` scenario (5 servers + 4 clients, sequential graceful stop of 3 servers including the clients' router and the coordinator): without the fix a client gets stuck one topology version short within ~150 iterations, with the fix 3x300 iterations pass clean. `TcpClientDiscoverySpiSelfTest` (52 tests), `TcpDiscoveryPendingMessageDeliveryTest` and `TcpClientDiscoverySpiCoordinatorChangeTest` pass as well.


anton-vinogradov:
Sequence of events leading to the message loss:

```mermaid
sequenceDiagram
    participant S as New router (server)
    participant RC as Client: Reconnector<br/>(own TcpDiscoveryIoSession)
    participant SR as Client: SocketReader<br/>(fresh TcpDiscoveryIoSession)

    Note over S: topology storm: NodeLeft(N+1) is enqueued<br/>to the client worker right behind the reconnect response
    S->>RC: [reconnect response][NodeLeft(N+1)] — one TCP segment
    Note over RC: BufferedInputStream reads ahead:<br/>BOTH messages land in the session buffer
    RC->>RC: parses reconnect response, returns success
    Note over RC: NodeLeft(N+1) bytes stay in the buffer<br/>of the abandoned session — lost forever
    RC-->>SR: socket handed over (raw socket, new session)
    S->>SR: NodeLeft(N+2) — later segment
    Note over SR: client applies N+2 while lastVer=N →<br/>AssertionError: lastVer=N, newVer=N+2<br/>(or stays on stale topology if nothing follows)
```

Before IGNITE-26111 both readers shared the single `BufferedInputStream` owned by `SocketStream`, so the read-ahead survived the handover — the fix restores exactly that invariant.

PR #13330, opened 2026-07-06
IGNITE-28863 Fix flaky GridFileDeploymentSelfTest

wernerdv:

### Notes

### DIFF
--- a/modules/extdata/uri/pom.xml
+++ b/modules/extdata/uri/pom.xml
@@ -253,9 +253,14 @@
 
                                 <sleep seconds="2" />
 
-                                <touch file="${basedir}/target/classes/org/apache/ignite/spi/deployment/uri/tasks/GridUriDeploymentTestWithNameTask11.class" />
-
-                                <zip destfile="${basedir}/target/file/${bad-signed.jar}" basedir="${basedir}/target/classes/" includes="org/apache/ignite/spi/deployment/uri/tasks/GridUriDeploymentTestWithNameTask11.class" update="yes" />
+                                <!-- Unpack, replace corrupted entry with correct class, repack to break signature -->
+                                <unzip src="${basedir}/target/file/${bad-signed.jar}" dest="${basedir}/target/file_tmp/zipfix" />
+                                <copy file="${basedir}/target/classes/org/apache/ignite/spi/deployment/uri/tasks/GridUriDeploymentTestWithNameTask11.class"
+                                      todir="${basedir}/target/file_tmp/zipfix/org/apache/ignite/spi/deployment/uri/tasks/"
+                                      overwrite="yes" />
+                                <delete file="${basedir}/target/file/${bad-signed.jar}" />
+                                <zip destfile="${basedir}/target/file/${bad-signed.jar}" basedir="${basedir}/target/file_tmp/zipfix" />
+                                <delete dir="${basedir}/target/file_tmp/zipfix" />
 
                                 <delete dir="${basedir}/target/file_tmp/"/>
                             </target>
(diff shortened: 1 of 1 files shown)


modules/extdata/uri/pom.xml
```diff
@@ -253,9 +253,14 @@
 
                                 <sleep seconds="2" />
```
anton-vinogradov: This `<sleep seconds="2"/>` (and the one in the `.gar` section below) only existed to make the old timestamp-based `<touch>` + `<zip update="yes">` trick work. With the unpack/replace/repack approach nothing depends on timestamps anymore, so both sleeps can be removed — saves 4 seconds on every build of this module.

modules/extdata/uri/pom.xml
```diff
@@ -253,9 +253,14 @@
...
-
-                                <zip destfile="${basedir}/target/file/${bad-signed.jar}" basedir="${basedir}/target/classes/" includes="org/apache/ignite/spi/deployment/uri/tasks/GridUriDeploymentTestWithNameTask11.class" update="yes" />
+                                <!-- Unpack, replace corrupted entry with correct class, repack to break signature -->
+                                <unzip src="${basedir}/target/file/${bad-signed.jar}" dest="${basedir}/target/file_tmp/zipfix" />
+                                <copy file="${basedir}/target/classes/org/apache/ignite/spi/deployment/uri/tasks/GridUriDeploymentTestWithNameTask11.class"
+                                      todir="${basedir}/target/file_tmp/zipfix/org/apache/ignite/spi/deployment/uri/tasks/"
+                                      overwrite="yes" />
+                                <delete file="${basedir}/target/file/${bad-signed.jar}" />
+                                <zip destfile="${basedir}/target/file/${bad-signed.jar}" basedir="${basedir}/target/file_tmp/zipfix" />
```
anton-vinogradov: `<zip>` writes entries in directory-scan order, so `META-INF/MANIFEST.MF` is no longer guaranteed to be the first entry. The test still passes because `GridUriDeploymentJarVerifier` uses `JarFile`, which looks the manifest up by name — but any `JarInputStream`-based reader (including the verifier's `verify(InputStream, ...)` overloads) would see such a jar as having no manifest at all and treat it as unsigned rather than badly signed.

Cheap insurance is to repack with the `<jar>` task, which always puts the manifest first:
```xml
<jar destfile="${basedir}/target/file/${bad-signed.jar}"
     basedir="${basedir}/target/file_tmp/zipfix"
     manifest="${basedir}/target/file_tmp/zipfix/META-INF/MANIFEST.MF" />
```
(Ant merges its own `Ant-Version`/`Created-By` attributes into the main section, which changes the manifest digest — but that only breaks the signature further, which is exactly what this artifact is for.)

anton-vinogradov (commented):
The approach looks correct to me: the old `<touch>` + `<zip update="yes">` trick relied on timestamp comparison (with zip's 2-second granularity), so the entry replacement could be silently skipped, leaving the jar validly signed — hence the flakiness. Unpack/replace/repack is deterministic, and since `GridUriDeploymentJarVerifier` goes through `JarFile` (name-based lookup, digest check on read), the repacked jar reliably fails verification. Also, the `<delete file=...>` before `<zip>` is load-bearing — it defeats Ant's up-to-date check, which could otherwise skip the repack.

A few minor points inline (each applies to both the `.jar` and the `.gar` sections).

modules/extdata/uri/pom.xml
```diff
@@ -253,9 +253,14 @@
...
+                                <delete dir="${basedir}/target/file_tmp/zipfix" />
```
anton-vinogradov: Nit: this `<delete dir=".../zipfix"/>` is redundant — the very next step deletes its parent `${basedir}/target/file_tmp/`.

modules/extdata/uri/pom.xml
wernerdv: Removed

modules/extdata/uri/pom.xml
wernerdv: Removed

modules/extdata/uri/pom.xml
wernerdv: Fixed

anton-vinogradov (approved):

PR #13375, opened 2026-07-15
IGNITE-28896 SQL Calcite: Adjust fractional value handling in FETCH, OFFSET, and LIMIT

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-28896


zstan:
So, it need to be merged somehow ? https://github.com/apache/ignite/pull/13311
I don\`t understand about rounding policy, this policy need to be applied globally for all rounding operations not only for limit\offset

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
```diff
@@ -94,27 +101,80 @@ public class LimitOffsetIntegrationTest extends AbstractBasicIntegrationTransact
...
+        cacheRepl.clear();
+
+        sql("INSERT into TEST_REPL VALUES (1, 'a'), (2, 'b'), (3, 'c'), (4, 'd')");
+
+        assertQuery("SELECT id FROM TEST_REPL ORDER BY id LIMIT 1.2")
+            .returns(1)
+            .returns(2)
+            .check();
+
+        assertQuery("SELECT id FROM TEST_REPL ORDER BY id OFFSET 1.1 ROWS FETCH FIRST 1.1 ROWS ONLY")
```
zstan: why 1.1 is rounded to 2 ? it need to be discussed

tkalkirill:
@zstan BigDecimal is already in Calcite's main branch, so I think it makes sense to use that.
Regarding the rounding policy: the name FetchOffsetRoundingPolicy implies it doesn't cover *all* rounding, but specifically OFFSET and FETCH, and the documentation describes it that way as well.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
```diff
@@ -571,7 +576,7 @@ private boolean hasExchange(RelNode rel) {
                 ctx,
                 rowType,
                 idxBndRel.first() ? cmp : cmp.reversed(),
-                0,
+                SortNode.OFFSET_DEFAULT,
```
zstan, line 576: I specially not touch this in my PR ) What more informative ? 
node with OFFSET_DEFAULT param - it need to lookup for reviewer or node with concreete "0" param ? 
Of course it\`s all kinda taste ... I just show you my logic here

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -262,36 +265,39 @@ private void validateFetchOffset(@Nullable SqlNode n, String clauseName) {
...
                 if (!(param instanceof Number)) {
                     Resources.ExInst<SqlValidatorException> err;
 
                     if (param == null)
-                        err = IgniteResource.INSTANCE.incorrectDynamicParameterType(SqlTypeName.BIGINT.toString(), "null");
+                        err = IgniteResource.INSTANCE.incorrectDynamicParameterType(SqlTypeName.DECIMAL.toString(), "null");
                     else {
                         SqlTypeName paramType = typeFactory().createType(param.getClass()).getSqlTypeName();
-                        err = IgniteResource.INSTANCE.incorrectDynamicParameterType(SqlTypeName.BIGINT.toString(), paramType.getName());
+                        err = IgniteResource.INSTANCE.incorrectDynamicParameterType(SqlTypeName.DECIMAL.toString(), paramType.getName());
```
zstan: say once more: this is wrong fix !

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
```diff
@@ -121,10 +128,62 @@ public void testInvalidLimitOffset() {
...
+        assertQuery("SELECT id FROM TEST_REPL ORDER BY id OFFSET 2.3 ROWS").returns(2).returns(3).check();
+        assertQuery("SELECT id FROM TEST_REPL ORDER BY id OFFSET 2.6 ROWS").returns(3).check();
+    }
+
+    /** */
+    @Test
+    public void testFetchOffsetRoundingPolicy() throws Exception {
+        fillCache(cacheRepl, 4);
+
+        FetchOffsetRoundingPolicy floorPlc = value -> value.setScale(0, RoundingMode.FLOOR);
```
zstan: I think we do not need to test Calcite configuration setting, we need to fix it in our code and use over all functionality and tests, otherwize we need to give an ability to configure it through Ignite setting which is discussable.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
tkalkirill: Why? That’s how it is in the CALCITE main branch right now.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
tkalkirill: I see your point. My intention is to express the semantics of this positional argument: no offset is specified here. The reader does not need to look up the constant value unless implementation details matter, while 0 shows the value but not the intent. It also keeps this call consistent with the defaults defined by `SortNode`, so I’d prefer to keep it.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
tkalkirill: I don’t think this needs to become a public Ignite setting. Rounding semantics may depend on the required SQL compatibility mode, so having one hardcoded behavior for every use case is unnecessarily restrictive.
This policy is an internal Calcite extension point intended for advanced integrations, not a part of the public Ignite API. The default behavior is still used everywhere unless a custom policy is explicitly provided. Keeping it internal also allows us to evolve or replace the mechanism later without introducing a public API compatibility burden.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
```diff
@@ -1079,7 +1081,8 @@ private long validateAndGetFetchOffsetParams(RexNode node, String op) {
         long paramAsLong;
 
         try {
-            paramAsLong = IgniteMath.convertToLongExact((Number)param);
+            BigDecimal val = IgniteMath.convertToBigDecimal((Number)param);
```
zstan: My point : 
seems we need to change : public static final RoundingMode NUMERIC_ROUNDING_MODE = RoundingMode.HALF_UP;
like : 
    /** Decimal rounding mode. */
    public static final RoundingMode ROUNDING_MODE = IgniteTypeSystem.INSTANCE.roundingMode();

plz check how it made here :

https://github.com/apache/ignite-3/blob/main/modules/sql-engine/src/main/java/org/apache/ignite/internal/sql/engine/util/IgniteMath.java#L52

and calls through IgniteSqlFunctions#toBigDecimal(java.lang.Number, int, int) will use correct rounding

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
tkalkirill: I got rid of that.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
tkalkirill: I got rid of that.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
```diff
@@ -1076,21 +1077,18 @@ private long validateAndGetFetchOffsetParams(RexNode node, String op) {
...
         }
 
-        long paramAsLong;
-
         try {
-            paramAsLong = IgniteMath.convertToLongExact((Number)param);
+            BigDecimal paramAsDecimal = IgniteMath.convertToBigDecimal((Number)param);
+
+            if (paramAsDecimal.signum() < 0) {
+                throw new IllegalArgumentException("Negative value of " + op);
```
zstan: We already have correct message for such a case, why do you use additional one ? 

```
throw new IgniteSQLException(IgniteResource.INSTANCE.illegalFetchLimit(op).str(),
                IgniteQueryErrorCode.UNEXPECTED_ELEMENT_TYPE);
```

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DataTypesTest.java
```diff
@@ -973,49 +971,49 @@ public void testCastDecimalOverflows() {
         // BIGINT
         assertQuery("SELECT CAST(9223372036854775807.1 AS BIGINT)").returns(9223372036854775807L).check();
         assertQuery("SELECT CAST(9223372036854775807.4 AS BIGINT)").returns(9223372036854775807L).check();
-        assertQuery("SELECT CAST(9223372036854775806.9 AS BIGINT)").returns(9223372036854775807L).check();
-        assertThrows("SELECT CAST(9223372036854775807.5 AS BIGINT)", IgniteSQLException.class, "BIGINT overflow");
-        assertThrows("SELECT CAST(9223372036854775807.9 AS BIGINT)", IgniteSQLException.class, "BIGINT overflow");
-        assertQuery("SELECT CAST(9223372036854775807.9 - 1 AS BIGINT)").returns(9223372036854775807L).check();
+        assertQuery("SELECT CAST(9223372036854775806.9 AS BIGINT)").returns(9223372036854775806L).check();
```
zstan: with such approach seems we miss huge number of corner-case tests i.e.
now we need to check : 
assertThrows("SELECT CAST(922337203685477580[8].5 AS BIGINT)", IgniteSQLException.class, "BIGINT overflow");
and all below, wdyt ?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
tkalkirill: The user-facing error message remains unchanged - we still use `illegalFetchLimit(op)`. The additional exception is only attached as the cause and provides context about why validation failed. It also allows us to keep a single error-handling path, which makes the code simpler and cleaner.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DataTypesTest.java
tkalkirill: I've added some tests; if any are missing, please make suggestions.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
```diff
@@ -121,10 +122,31 @@ public void testInvalidLimitOffset() {
         assertThrows("SELECT * FROM TEST_REPL OFFSET -1 ROWS",
             IgniteSQLException.class, null);
 
+        assertThrows("SELECT * FROM TEST_REPL OFFSET -1.5 ROWS",
```
zstan: plz fix all such a cases for:
```suggestion
        assertThrowsSqlException("SELECT * FROM TEST_REPL OFFSET -1.5 ROWS",
            "Incorrect syntax near the keyword 'OFFSET'");
```

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DataTypesTest.java
```diff
@@ -73,60 +73,68 @@ public void testRoundingOfDynamicNumericsPrecasted() {
 
     /** @return input type, input value, target type, expected result. */
     private static List<List<Object>> numericsToRound() {
-        List<List<Object>> lst = new ArrayList<>(50);
+        List<List<Object>> lst = new ArrayList<>(90);
```
zstan: pedantic )))

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
```diff
@@ -121,10 +122,29 @@ public void testInvalidLimitOffset() {
...
+
+        assertThrowsSqlException("SELECT * FROM TEST_REPL OFFSET -0.5 ROWS", null);
+
         assertThrows("SELECT * FROM TEST_REPL OFFSET 2+1 ROWS",
             IgniteSQLException.class, null);
     }
 
+    /** */
+    @Test
+    public void testFractionalLimitOffset() throws Exception {
```
zstan, line 134: why we need this case if already have : testFractionalLimitOffset?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/IgniteMath.java
```diff
@@ -411,7 +412,7 @@ else if (x instanceof Float) {
 
     /** */
     private static double extendToRound(double x) {
-        return x < 0.0d ? x - 0.5d : x + 0.5d;
+        return round(x).doubleValue();
```
zstan, line 420: I remove all four usages of this func and PlannerTestSuite and IntegrationTestSuite are passed locally, 
this are failed:
```
test_floor_ceil.test
test_trigo.test
```
but seems it\`s expected and need to be also fixed, wdyt ?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
tkalkirill: These tests cover different execution paths. This test checks fractional literals and the `SqlLiteral` validation path, while `DynamicParametersIntegrationTest` checks dynamic parameters and their runtime conversion.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/IgniteMath.java
tkalkirill: Direct Java casts work correctly for the current 'RoundingMode.DOWN', but they implicitly hard-code this behavior. If 'IgniteTypeSystem.roundingMode()' is changed later, floating-point-to-integral conversions will no longer follow the configured rounding policy. Keeping the conversion through 'BigDecimal' makes the behavior consistent with the type system. However, changing the rounding mode would also require reviewing the overflow boundaries.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DataTypesTest.java
```diff
@@ -86,13 +86,17 @@ private static List<List<Object>> numericsToRound() {
 
         lst.add(F.asList("DECIMAL(20,1)", new BigDecimal("-9223372036854775808.4"), "BIGINT", -9223372036854775808L));
         lst.add(F.asList("DECIMAL(20,1)", new BigDecimal("-9223372036854775808.5"), "BIGINT", overflowErr));
+        lst.add(F.asList("DECIMAL(20,1)", new BigDecimal("-9223372036854775809.5"), "BIGINT", overflowErr));
```
zstan: redundant this and all such below, because -9223372036854775808.5 already checked

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
```diff
@@ -54,7 +54,8 @@ public class LimitOffsetIntegrationTest extends AbstractBasicIntegrationTransact
 
     /** {@inheritDoc} */
     @Override protected void afterTest() throws Exception {
-        // Override method to keep caches after tests.
+        // Keep caches between tests but do not leak an active transaction into the next test.
+        clearTransaction();
```
zstan: I wonder, why we need such code just here ? If this is some kind of problem - fill an issue and revert this code. IntegrationTestSuite passed correctly for me without this change.

zstan:
@tkalkirill plz fix minor changes, overall looks good, call me to review after final fix plz

zstan (changes requested):

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
```diff
@@ -121,10 +121,33 @@ public void testInvalidLimitOffset() {
...
+
+        assertThrowsSqlException("SELECT * FROM TEST_REPL OFFSET -0.5 ROWS", null);
+
         assertThrows("SELECT * FROM TEST_REPL OFFSET 2+1 ROWS",
             IgniteSQLException.class, null);
     }
 
+    /** */
+    @Test
+    public void testFractionalLimitOffset() throws Exception {
```
zstan, line 134: we already have the same tests in script ones: *limit*.test do you think we also need it here ?

zstan (approved):
Looks good if TC is ok, also i think that we have duplicated tests functionality through *Integration* and script tests, don\`t forget to change initial issue description cause now it\`s not about decimal but about rounding as i can see

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
tkalkirill: I would keep it.

PR #13390, opened 2026-07-21
IGNITE-28913 GitHub Actions: run fork PR checks without checking out untrusted code in a privileged context

anton-vinogradov:
### Problem

Since **actions/checkout v4.4.0** (released 2026-07-20, a backport of the v7.0.0 "pwn request" protection announced in the [GitHub Changelog on 2026-06-18](https://github.blog/changelog/2026-06-18-safer-pull_request_target-defaults-for-github-actions-checkout/)), `actions/checkout` refuses **by default** to check out fork PR code under `pull_request_target`. The floating `v4`/`v6` tags were moved onto this backport, so `actions/checkout@v4`/`@v6` picked it up automatically. Every fork PR now fails at the checkout step in `commit-check.yml` (all 5 jobs) and `check-protected-classes.yml`.

### Fix

Following [Securely using pull_request_target](https://docs.github.com/en/actions/reference/security/securely-using-pull_request_target), neither workflow checks out untrusted PR code under a privileged token any more:

- **`commit-check.yml` → back to `pull_request`.** These jobs check out and *execute* untrusted PR code (`mvn`/`tox`/`dotnet`) but need no secrets and no write token, so they don't need the privileged context. On `pull_request` the fork code runs with an unprivileged, fork-scoped token, and the checkout protection doesn't apply — no opt-in flag. This reverts the trigger change of IGNITE-28827; the deliberate cost is that PRs conflicting with the base branch are no longer checked (recorded in a comment).

- **`check-protected-classes.yml` → no checkout at all.** This job needs `issues`/`pull-requests`/`checks: write` to comment on and label the PR, so it stays on `pull_request_target`. Instead of checking out the fork and running `git`, it now reads the PR diff **through the GitHub API as data** (`pulls.listFiles` + `repos.getContent` at the appropriate revision) inside `github-script`. The fork's code never lands on the runner, so there's nothing to opt into and nothing to execute — the cleanest form of "treat PR code as untrusted data". Detection logic is preserved exactly (added → head revision, deleted/modified → base, renames expanded to delete+add, `@Order` = protected class), verified against the previous git-based logic on real PRs.

  Hardening carried in the rewrite: file names (fork-controlled) are rendered as inert inline code so they can't inject markdown/`@mentions` into the bot's comment; the existing comment is located by a stable hidden marker instead of a substring match.

### Verification

The API-based detection was cross-checked against the original `git diff`/`git show` logic on real PRs (e.g. PR #13095: both report the same 92 `@Order` files; PRs with no protected changes report none).

Checks on this PR still run the **base-branch** version of the workflows (old `pull_request_target` without the flag), so they keep failing at checkout here — that demonstrates the bug. The fix takes effect once merged to master.

### DIFF
--- a/.github/workflows/check-protected-classes.yml
+++ b/.github/workflows/check-protected-classes.yml
@@ -15,6 +15,9 @@
 
 name: Protected Classes
 
+# pull_request_target for the write token needed to comment/label the PR. The fork's changes are
+# read through the API as data and never checked out or executed, so no allow-unsafe-pr-checkout and
+# no working copy of the PR code on the runner. Do not add a step that builds or runs the PR code.
 on: pull_request_target
 
 concurrency:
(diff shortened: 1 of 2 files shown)


zstan:
https://docs.github.com/en/actions/reference/security/securely-using-pull_request_target

zstan (approved):

Issue #13286, opened 2026-06-26
Hibernate 7 compatibility (Ignite 2.x + records problem)


wortkarg:
In Hibernate 7 (at least in 7.3.x+) the query cache internally stores instances of QueryResultsCacheImpl.CacheItem and this class is now a Java record.
Ignite 2 throws an UnsupportedOperationException for records, see the following issue:
https://github.com/apache/ignite/issues/11853

As a result, Ignite 2 is not compatible with Hibernate 7, at least not with the query cache enabled.

**Stacktrace**

...
Caused by: java.lang.UnsupportedOperationException: can't get field offset on a record class: private final long org.hibernate.cache.internal.QueryResultsCacheImpl$CacheItem.timestamp
	at jdk.unsupported/sun.misc.Unsafe.objectFieldOffset(Unsafe.java:655)
	at org.apache.ignite.internal.util.GridUnsafe.objectFieldOffset(GridUnsafe.java:1216)
	at org.apache.ignite.internal.binary.BinariesFactoryImpl.create(BinariesFactoryImpl.java:110)
	at org.apache.ignite.internal.binary.BinaryClassDescriptor.<init>(BinaryClassDescriptor.java:366)
	at org.apache.ignite.internal.binary.BinaryClassDescriptor.<init>(BinaryClassDescriptor.java:152)
	at org.apache.ignite.internal.binary.BinaryContext.createDescriptorForClass(BinaryContext.java:669)
	at org.apache.ignite.internal.binary.BinaryContext.descriptorForClass(BinaryContext.java:643)
	at org.apache.ignite.internal.binary.BinaryWriterExImpl.marshal0(BinaryWriterExImpl.java:156)
	at org.apache.ignite.internal.binary.BinaryWriterExImpl.marshal(BinaryWriterExImpl.java:139)
	at org.apache.ignite.internal.binary.BinaryWriterExImpl.marshal(BinaryWriterExImpl.java:126)
	at org.apache.ignite.internal.binary.GridBinaryMarshaller.marshal(GridBinaryMarshaller.java:254)
	at org.apache.ignite.internal.processors.cache.binary.CacheObjectBinaryProcessorImpl.marshalToBinary(CacheObjectBinaryProcessorImpl.java:487)
	at org.apache.ignite.internal.processors.cache.binary.CacheObjectBinaryProcessorImpl.toBinary(CacheObjectBinaryProcessorImpl.java:1383)
[...]

PR #13391, opened 2026-07-22
IGNITE-13286 Support Java records in binary marshaller

kush2439p:
## Summary`n`nRoutes Java records through Ignite's optimized Java-serialization path, avoiding `Unsafe.objectFieldOffset()` on immutable record fields.`n`n## Root cause`n`nHibernate 7 uses a serializable record for `QueryResultsCacheImpl.CacheItem`. Ignite's reflective binary marshaller attempts to obtain unsafe offsets for record fields, which modern JDKs reject.`n`n## Changes`n`n- Detect records without introducing a Java 16 compile-time dependency.`n- Use `OptimizedMarshaller` for records.`n- Make the developer warning accurately include records.`n- Add a runtime-compiled serializable-record round-trip regression test; it runs on JDK 16+ and skips on older JDKs.`n`n## Verification`n`n- Rebased on current `ignite-2.17` and `git diff --check` passes.`n- All 4,531 `ignite-core` production sources compile successfully on JDK 25.`n- Focused test execution is blocked during unrelated full test compilation because JDK 25 removed `Thread.suspend()` and `Thread.resume()`, which legacy Ignite tests still reference. A supported JDK 17 is not installed locally.`n- Existing GitHub check failures are fork checkout-security refusals before code execution, not test failures.`n`nFixes #13286

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryClassDescriptor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryClassDescriptor.java
@@ -202,7 +202,7 @@ public class BinaryClassDescriptor {
         initialSerializer = serializer;
 
         // If serializer is not defined at this point, then we have to use OptimizedMarshaller.
-        useOptMarshaller = serializer == null || isGeometryClass(cls);
+        useOptMarshaller = serializer == null || isGeometryClass(cls) || isRecord(cls);
 
         // Reset reflective serializer so that we rely on existing reflection-based serialization.
         if (serializer instanceof BinaryReflectiveSerializer)
@@ -1094,6 +1094,21 @@ public class BinaryClassDescriptor {
     /** */
     Constructor<?> ctor() {
         return ctor;
+    }
+
+    /**
+     * Checks whether a class is a Java record without requiring a Java 16 runtime at compile time.
+     *
+     * Records are immutable and the binary marshaller cannot restore their final fields. Java serialization,
+     * used by the optimized marshaller, supports serializable records starting with Java 16.
+     *
+     * @param cls Class to check.
+     * @return {@code True} if the class is a record.
+     */
+    private static boolean isRecord(Class<?> cls) {
+        Class<?> superCls = cls.getSuperclass();
+
+        return superCls != null && "java.lang.Record".equals(superCls.getName());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


kush2439p:
Fixes the Hibernate 7 query-cache incompatibility caused by Java records.  
Hibernate’s QueryResultsCacheImpl.CacheItem is now a record, while Ignite’s reflective binary marshaller calls sun.misc.Unsafe.objectFieldOffset(), which is unsupported for record fields on modern JDKs.  
This change detects records without adding a Java 16 compile-time dependency and routes them through Ignite’s optimized Java-serialization path, which supports serializable records.

kush2439p:
Hi maintainers, following up on this PR since it has been open for several weeks without review. The record-handling fix and regression coverage are ready for review, and I am happy to rebase or revise the approach based on feedback. Thank you.

modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryClassDescriptor.java
```diff
@@ -1096,6 +1096,21 @@ Constructor<?> ctor() {
...
+     * Records are immutable and the binary marshaller cannot restore their final fields. Java serialization,
+     * used by the optimized marshaller, supports serializable records starting with Java 16.
+     *
+     * @param cls Class to check.
+     * @return {@code True} if the class is a record.
+     */
+    private static boolean isRecord(Class<?> cls) {
+        Class<?> superCls = cls.getSuperclass();
+
+        return superCls != null && "java.lang.Record".equals(superCls.getName());
```
zstan, line 1111: why you avoid to runtime check like : cls.isRecord() here ?

modules/core/src/test/java/org/apache/ignite/internal/binary/BinaryMarshallerSelfTest.java
```diff
@@ -126,6 +134,46 @@ public void testNull() throws Exception {
         assertNull(marshalUnmarshal(null));
     }
 
+    /**
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testSerializableRecord() throws Exception {
```
zstan, line 141: apache ignite master currently support only 17 java - thus i suppose you can simplify this test and rewrite it like other presented here ? Also you need this PR agains master branch, not against ignite-2.17

zstan:
IGNITE-13286 - is different issue and not related to this activity

kush2439p:
Superseded by #13503, which is rebased onto current `master`, uses `Class.isRecord()`, and adds a Java 17 static-record regression test. This PR still targets the historical `ignite-2.17` file layout and uses an unrelated JIRA key, so I am closing it to avoid duplicate review.

PR #13408, opened 2026-07-28
IGNITE-28356 Get rid of redundant wire companion fields and hand-written marshalling in messages

anton-vinogradov:
Many messages keep both a logical field and a hand-written `marshal()`/`unmarshal()` pair that fills its
wire companion, even where the codegen can do it and where the companion describes data that is never
sent. Each item below is a problem found and the change that closes it.

**A map whose values are always null.** All three producers of `GridDistributedTxPrepareRequest#dhtVers`
put `null` as the value, and the consumer's `ver.getValue() == null || !equals` is therefore always true.
The field is now `Collection<IgniteTxKey> dhtVerKeys`, which drops the value companion and `4 + 2N` bytes.

**A key travelling next to its own value.** `GridNearTxPrepareResponse#ownedVals` was a map plus two
companions. The key moves into the value as `KeyedVersionedValue`, so the field becomes one collection.
`CacheVersionedValue` already carries `cacheId`, which now stops travelling twice: `-(4 + 6N)` bytes.

**An array companion for an ordered set.** `TxLocksRequest#txKeys` and `TxLocksResponse#txKeys` kept a
`Set` and an array to write it. `writeCollection` and `writeObjectArray` emit the same bytes, so the
companion is gone and the runtime type stays `HashSet`, preserving the sender-side dedup.

**A hand-written zip of two parallel fields.** `GridNearGetRequest#keyMap` was built by pairing `keys` and
`readersFlags` by index. The map stays, the conversion moves into `@Marshalled(keys = ..., values = ...)`.

**A map copied entry by entry into an empty one.** `GridDhtTxPrepareRequest#owned` was drained into
`GridNearTxRemote` with `putAll` after the request was done with it. `ownedVersions` now takes the map
over, which drops a `GridLeanMap` and `N` nodes per prepare.

**Blob marshalling that the codegen already expresses.** `LazyServiceConfigurationMessage#affKey`,
`GenericValueMessage#val`, `QueryStartRequest#params` and `QueryEntityMessage#dfltFieldValues` are now
`@Marshalled`; the generated code matches the hand-written one call for call. `QueryEntityMessage` keeps
its `!F.isEmpty` gate, moved into the constructor, because `QueryEntity` initializes the defaults with an
empty map and the generated gate is `!= null`.

**Collection marshalling in an atomic request.** `GridNearAtomicFullUpdateRequest#entryProcessors` moves
to `@Marshalled`; the field is retyped to `List<Object>` because the generated code calls `add(Object)`.
The `operation() == TRANSFORM` gates stay on `invokeArgs` in all three methods: dropping them only in
`marshal` would fill `invokeArgsBytes` without `deployInvokeArguments`, and a receiver with peer class
loading on would get a `NoClassDefFoundError` instead of today's silent degradation. The `deploy()` gate
becomes `entryProcessors != null`, because the codegen nulls the companion after restoring it.

**A job payload unmarshalled with the wrong class loader.** `GridJobExecuteRequest` had two hand-written
methods because its five payload fields are user classes that only the deployment class loader can read,
while `GridIoManager.unmarshalPayload` would use the configuration one. The fields move to `@Marshalled`
and the class becomes a `DeferredUnmarshalMessage`, the marker that exists for this: `GridJobProcessor`
unmarshals the request itself, with the same loader as before. A job of a continuous task also stops
sending its siblings - the receiver discarded them and `GridJobSessionImpl#getJobSiblings` requests them
from the task node when they are missing.

**Wire companions standing in for a conditional field (IGNITE-28922, closed as a duplicate).**
`CacheContinuousQueryEntry` held `keyWire`/`newValWire`/`oldValWire` so that `marshal()` could hide the
data of a filtered entry. The three pairs collapse and `key`/`newVal`/`oldVal` go back to plain `@Order`.
The invariant becomes structural instead: a filtered entry gives its data up at
`CacheContinuousQueryEventBuffer#processEntry`, the only door into the buffer, so nothing downstream has
to remember it. Two copies that this makes redundant are removed, and a filtered `EXPIRED` entry with
`updateCntr == -1` is no longer sent at all - `handleEvent` already discarded it on arrival.

Wire layout is unchanged except where noted, and no path allocates more than before, with one exception
worth stating: a filtered entry taking the out-of-range counter path in the continuous query buffer now
costs one object, where the data used to be masked at serialization time instead.

Verified by the full 43-module build with `-Pcheckstyle`, and by
`CacheContinuousQueryFailoverAtomicSelfTest`, `GridCacheAtomicNearEnabledFullApiSelfTest`,
`CacheContinuousQueryEventBufferTest`, `CacheContinuousQueryBufferLimitTest`,
`CacheContinuousQueryEntriesExpireTest`, `IgniteCacheContinuousQueryBackupQueueTest`,
`IgniteCacheContinuousQueryImmutableEntryTest`, `QueryEntityMessageSerializationTest`,
`IgniteCacheAtomicInvokeTest`, `IgniteCacheAtomicNearEnabledInvokeTest`, `MessageProcessorTest`,
`DirectMarshallingMessagesTest`, `GridCachePartitionedMultiNodeFullApiSelfTest`, the deadlock detection
tests, the event storage tests, 17 P2P tests and 22 compute tests.


anton-vinogradov:
/runall

---
🚀 **RunAll queued** — [build 9236765](https://ci2.ignite.apache.org/build/9236765) · live progress & verdict: [Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13408). The verdict lands here when the run finishes.

anton-vinogradov:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13408)** verdict · RunAll build [9236765](https://ci2.ignite.apache.org/build/9236765) · 147 suites ran, 0 reused

❌ **1 blocker(s) in 1 suite(s):**
- Cache 2: `org.apache.ignite.testsuites.IgniteCacheTestSuite2: org.apache.ignite.internal.processors.cache.distributed.near.GridNearTxPrepareDestroyedCacheTest.testPrepareOfDestroyedCacheAnsweredWithErrorResponse`


_♻️ Settled after 2 auto re-run wave(s): #1 — 2 blocker suite(s); #2 — 1 blocker suite(s)._

anton-vinogradov:
/run-all

---
🚀 **RunAll queued** — [build 9237865](https://ci2.ignite.apache.org/build/9237865) · live progress & verdict: [Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13408). The verdict lands here when the run finishes.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/atomic/GridNearAtomicFullUpdateRequest.java
```diff
@@ -387,7 +379,7 @@ else if (conflictVers != null)
 
             forceDeploymentInfo(ctx);
 
-            if (entryProcessorsBytes == null)
+            if (entryProcessors != null)
```
wernerdv, line 385: Why the change? For example, in `GridDhtAtomicUpdateRequest#deploy` bytes are still checked for null before deployCollection is called.

anton-vinogradov:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13408)** verdict · RunAll build [9237865](https://ci2.ignite.apache.org/build/9237865) · 147 suites ran, 0 reused

✅ **No blockers** — nothing in this run looks caused by this PR. 31 pre-existing/flaky tests filtered out.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearGetRequest.java
```diff
@@ -147,19 +145,7 @@ public GridNearGetRequest(
...
-            readersFlags = new ArrayList<>(keys.size());
-
-        for (Map.Entry<KeyCacheObject, Boolean> entry : keys.entrySet()) {
-            this.keys.add(entry.getKey());
-
-            if (addReader)
-                readersFlags.add(entry.getValue());
-        }
-
+        this.keyMap = keys;
```
wernerdv: Remove unnecessary 'this' qualifier.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearGetRequest.java
```diff
@@ -67,15 +64,16 @@ public class GridNearGetRequest extends GridCacheIdMessage implements GridCacheD
 
     /** */
     @GridToStringInclude
-    private LinkedHashMap<KeyCacheObject, Boolean> keyMap;
+    @Marshalled(keys = "keys", values = "readersFlags")
+    Map<KeyCacheObject, Boolean> keyMap;
```
wernerdv, line 70: After the changes, HashMap will be created instead of LinkedHashMap in GridNearGetRequestMarshaller#unmarshal:
`msg.keyMap = U.newHashMap(msg.keys.size());`

Is there a problem with that?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/GridDistributedTxPrepareRequest.java
```diff
@@ -235,23 +226,20 @@ public byte policy() {
...
     }
 
     /**
-     * @return Map of versions to be verified.
+     * @return Keys whose DHT version is verified.
      */
-    public Map<IgniteTxKey, GridCacheVersion> dhtVersions() {
-        return dhtVers == null ? Collections.emptyMap() : dhtVers;
+    public Collection<IgniteTxKey> dhtVersionKeys() {
+        return dhtVerKeys == null ? Collections.emptyList() : dhtVerKeys;
```
wernerdv:
```suggestion
        return F.emptyIfNull(dhtVerKeys);
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/TxLocksRequest.java
```diff
@@ -35,14 +34,9 @@ public class TxLocksRequest extends GridCacheMessage {
...
-    @GridToStringInclude
-    @Marshalled("txKeysArr")
-    Set<IgniteTxKey> txKeys;
-
-    /** Array of txKeys from {@link #txKeys}. Used during marshalling and unmarshalling. */
-    @GridToStringExclude
     @Order(1)
-    IgniteTxKey[] txKeysArr;
+    @GridToStringInclude
+    Collection<IgniteTxKey> txKeys;
```
wernerdv, line 44: Why can't we keep Set?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearTxPrepareDestroyedCacheTest.java
```diff
@@ -163,13 +163,13 @@ public void testPrepareOfDestroyedCacheAnsweredWithErrorResponse() throws Except
 
         Throwable err = ((GridNearTxPrepareResponse)resps.get(0)).error();
 
-        assertNotNull("The response must carry the unmarshalling error.", err);
+        assertNotNull("The response must carry the error of the destroyed cache.", err);
 
-        CacheObjectNotResolvedException cause = X.cause(err, CacheObjectNotResolvedException.class);
+        CacheInvalidStateException cause = X.cause(err, CacheInvalidStateException.class);
```
wernerdv, line 168: Need to update javadoc for GridNearTxPrepareDestroyedCacheTest regarding `throws {@link CacheObjectNotResolvedException}`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/TxLocksResponse.java
```diff
@@ -43,21 +43,16 @@ public class TxLocksResponse extends GridCacheMessage {
     @Marshalled(keys = "nearTxKeysArr", values = "locksArr")
     final Map<IgniteTxKey, List<TxLock>> nearTxKeyLocks = new HashMap<>();
 
-    /** Remote keys involved into transactions. Doesn't include near keys. */
+    /** Remote keys involved into transactions. Doesn't include near keys. A set: the sender dedups them. */
+    @Order(2)
     @GridToStringInclude
-    @Marshalled("txKeysArr")
-    Set<IgniteTxKey> txKeys;
+    Collection<IgniteTxKey> txKeys;
```
wernerdv, line 55: Why can't we keep Set?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/atomic/GridNearAtomicFullUpdateRequest.java
anton-vinogradov: The two cases are inverted, which is why the check moved.

`entryProcessors` here is now `@Marshalled("entryProcessorsBytes")`, and the generated unmarshal nulls the companion once it has inflated it — `GridNearAtomicFullUpdateRequestMarshaller:138`, `msg.entryProcessorsBytes = null;`. So under `@Marshalled` `entryProcessorsBytes == null` means "already unmarshalled", the opposite of what it used to mean.

In `GridDhtAtomicUpdateRequest` the field is not `@Marshalled` (untouched by this PR): its hand-written unmarshal keeps the bytes, so there `entryProcessorsBytes == null` still means "not marshalled yet" and the check is right. Same for the `invokeArgsBytes == null` check one line below in this very method — `invokeArgs` stayed on hand-written marshalling, so its check stayed too. The bytes check was dropped exactly where the field moved to codegen, not everywhere.

Behaviour is identical either way: `deploy()` runs only on the send path (`GridCacheIoManager#onSend:1125`, plus two pre-send spots in `IgniteTxManager`), and on the sending node the companion is always null at that point, because the only writer is the generated marshal, which runs later in `GridIoManager#marshal`. Both spellings pick the same branch; the new one just doesn't become vacuously true on a received message.

Added a one-line comment stating that reason.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearGetRequest.java
anton-vinogradov: No problem in practice, though the change is real — thanks for spotting it.

`keyMap()` has exactly one consumer: `GridDhtCacheAdapter#processNearGetRequest:1292` → `getDhtAsync` → `GridDhtGetFuture`. None of the steps there needs the order — `init()` passes `keys.keySet()` to the preloader, reader registration is per-entry, and `getDhtAllAsync` takes a key set. And the order dies at the next step anyway, already in master: the local get result goes into `U.newHashMap` (`GridDhtCacheAdapter:730`), `misses` into `new HashMap<>()` (:816), and `GridDhtGetFuture#toEntryInfos` iterates that map to build the response. So the entry order in `GridNearGetResponse` never matched the request order. The near side matches by key, not by position (`CacheDistributedGetFutureAdapter#onResult` → `createResultMap`, `GridPartitionedGetFuture:600-620`).

`partition()` is unaffected: it runs on the NIO thread before unmarshal, when `keyMap` is still null and `keys` is the wire-order `ArrayList`, so it returns the same partition as master.

I also looked for the reason behind `LinkedHashMap`: `git log -S` traces it to the very first import (`7016f154b20`), then it was carried over mechanically in `832f114e8eb` when `keyBytes` was split into `keys`/`flags`. No ticket, no comment, no test pins it.

One thing worth a follow-up rather than this PR: the `@Order` path *does* derive the implementation from the declared type — `MessageSerializerGenerator:542` emits `linked=true` for a `LinkedHashMap` field and `readMap` then uses `U.newLinkedHashMap`. `MessageMarshallerGenerator` hardcodes `U.newHashMap`/`U.newHashSet` instead, so declaring the field `LinkedHashMap` here does not even compile (I tried: `incompatible types: HashMap cannot be converted to LinkedHashMap`). Teaching the marshaller generator to derive it the same way would remove the trap for future `@Marshalled` fields where the order does matter. Happy to file that separately if you agree.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/TxLocksRequest.java
anton-vinogradov: We can't — and it turned out to be the more interesting answer, so let me lay it out.

The codegen does honour the declared type on the `@Order` path: `MessageSerializerGenerator:551` emits `set=true` for a `Set` field, and `DirectByteBufferStream:1659` then builds `U.newHashSet` instead of an `ArrayList`. The catch is *when*: the set is filled inside the read loop (`col.add(item)`, :1667), and at that moment `IgniteTxKey#hashCode()` → `KeyCacheObjectImpl#hashCode()` throws `CacheObjectNotResolvedException`, because the reader created the key as `new KeyCacheObjectImpl(null, bytes, -1)` (`CacheObjectBinaryProcessorImpl:1299`). The value is resolved later, in the generated marshaller's `unmarshal`.

I verified it instead of reasoning about it: declaring the field `Set<IgniteTxKey>` compiles, but `TxDeadlockDetectionMessageMarshallingTest` then fails with `GridDirectParser: Failed to read message ... CacheObjectNotResolvedException` and the NIO session closed. With `Collection` it passes.

That is exactly what the companion array in master was for — the generated ELEMENTS code does `for (e : arr) { unmarshal(e); col.add(e); }`, i.e. hashing deferred until the cache objects are resolved. The same pattern is still visible in `TxLocksResponse` for `nearTxKeyLocks`: `put(k, v)` happens only after `MessageMarshalling.unmarshal(k, ...)`.

So keeping `Set` in the declaration means keeping the companion array. The wire is byte-identical either way (`writeCollection` and `writeObjectArray` both emit size + elements, and the `set` flag never reaches the wire), so the only thing it buys is the type-level guarantee — and nothing depends on it. The sole consumers are `IgniteTxManager#hasKeys` (plain iteration) and `TxDeadlockDetection#mapTxKeys` (iteration into per-node `HashSet`s and a `UniqueDeque`): no `contains()` over these keys, no collection `equals`, no receive-side dedup needed. The sender-side dedup stays and does real work — `TxLocksRequest`'s constructor still takes a `Set`, and `TxLocksResponse#addKey` builds one, which matters because the same key can be added from two different transactions.

Expanded the field comment to state the reason rather than just assert "a set", so the next reader doesn't have to ask.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/TxLocksResponse.java
anton-vinogradov: We can't — and it turned out to be the more interesting answer, so let me lay it out.

The codegen does honour the declared type on the `@Order` path: `MessageSerializerGenerator:551` emits `set=true` for a `Set` field, and `DirectByteBufferStream:1659` then builds `U.newHashSet` instead of an `ArrayList`. The catch is *when*: the set is filled inside the read loop (`col.add(item)`, :1667), and at that moment `IgniteTxKey#hashCode()` → `KeyCacheObjectImpl#hashCode()` throws `CacheObjectNotResolvedException`, because the reader created the key as `new KeyCacheObjectImpl(null, bytes, -1)` (`CacheObjectBinaryProcessorImpl:1299`). The value is resolved later, in the generated marshaller's `unmarshal`.

I verified it instead of reasoning about it: declaring the field `Set<IgniteTxKey>` compiles, but `TxDeadlockDetectionMessageMarshallingTest` then fails with `GridDirectParser: Failed to read message ... CacheObjectNotResolvedException` and the NIO session closed. With `Collection` it passes.

That is exactly what the companion array in master was for — the generated ELEMENTS code does `for (e : arr) { unmarshal(e); col.add(e); }`, i.e. hashing deferred until the cache objects are resolved. The same pattern is still visible in `TxLocksResponse` for `nearTxKeyLocks`: `put(k, v)` happens only after `MessageMarshalling.unmarshal(k, ...)`.

So keeping `Set` in the declaration means keeping the companion array. The wire is byte-identical either way (`writeCollection` and `writeObjectArray` both emit size + elements, and the `set` flag never reaches the wire), so the only thing it buys is the type-level guarantee — and nothing depends on it. The sole consumers are `IgniteTxManager#hasKeys` (plain iteration) and `TxDeadlockDetection#mapTxKeys` (iteration into per-node `HashSet`s and a `UniqueDeque`): no `contains()` over these keys, no collection `equals`, no receive-side dedup needed. The sender-side dedup stays and does real work — `TxLocksRequest`'s constructor still takes a `Set`, and `TxLocksResponse#addKey` builds one, which matters because the same key can be added from two different transactions.

Expanded the field comment to state the reason rather than just assert "a set", so the next reader doesn't have to ask.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearGetRequest.java
anton-vinogradov: Done — the parameter there is `keys`, so the qualifier distinguished nothing (unlike the neighbouring lines, where the names do collide).

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/GridDistributedTxPrepareRequest.java
anton-vinogradov: Applied, thanks — `Collections` import dropped with it.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearTxPrepareDestroyedCacheTest.java
anton-vinogradov: Good catch — the class javadoc described exactly the mechanism this PR removes. Rewrote it around the actual behaviour: the entry has no cache context to initialize against, so the prepare fails with `CacheInvalidStateException`. The `{@link}` and the import now point there.

wernerdv (approved):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearGetRequest.java
wernerdv: > Happy to file that separately if you agree.

Yes, that makes sense!

anton-vinogradov:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13408)** verdict · RunAll build [9240744](https://ci2.ignite.apache.org/build/9240744) · 147 suites ran, 0 reused

✅ **No blockers** — nothing in this run looks caused by this PR. 28 pre-existing/flaky tests filtered out.

_♻️ Settled after 1 auto re-run wave(s): #1 — 1 broken suite(s)._

PR #13414, opened 2026-07-29
IGNITE-28928 Use MessageSerializer for SecurityPermissionSet

shishkovilja:


modules/codegen/src/main/java/org/apache/ignite/internal/MessageSerializerGenerator.java
```diff
@@ -109,6 +110,9 @@ public class MessageSerializerGenerator extends MessageCompanionGenerator {
     /** */
     private final List<String> read = new ArrayList<>();
 
+    /** Class fields, which should come before ordinary class fields. */
```
sergey-chugunov-1985:
```suggestion
    /** Static class fields, which should come before ordinary class fields. */
```

modules/core/src/main/java/org/apache/ignite/internal/processors/security/SecurityUtils.java
```diff
@@ -372,4 +373,59 @@ private static void authorizeAll(IgniteSecurity security, Map<String, Collection
...
+            return false;
+
+        for (Map.Entry<String, Collection<SecurityPermission>> e : m1.entrySet()) {
+            Collection<SecurityPermission> v1 = e.getValue();
+            Collection<SecurityPermission> v2 = m2.get(e.getKey());
+
+            if (v1 == v2)
+                continue;
+
+            if (v1 == null || v2 == null)
```
sergey-chugunov-1985: I suspect we don't need to do null checks here separately because F.eqNotOrdered already contains null checks - could you take a look and decide if we indeed can drop this?

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
```diff
@@ -120,6 +123,9 @@ public class MessageProcessor extends AbstractProcessor {
     /** */
     private final Map<String, IgniteBiTuple<String, String>> enumMappersInUse = new HashMap<>();
 
+    /** */
+    private final Map<Element, String> enumsPerField = new HashMap<>();
```
sergey-chugunov-1985: I think it should be possible to turn this field into a parameter of validateEnumType method and initialize it inside the method when necessage.

By doing so we would eliminate a MessageProcessor's level field which is relevant only for validation process of an isolated message field and get rid of a bit confusing clearing `enumsPerField` in `validateEnumbFieldMapping` method.

modules/core/src/main/java/org/apache/ignite/internal/processors/security/SecurityUtils.java
shishkovilja: Fixed and moved to `SecurityBasicPermissionSetSerializationTest`.

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
```diff
@@ -294,11 +300,27 @@ private List<List<VariableElement>> hierarchicalOrderedFields(TypeElement type)
...
+        }
+
+        enumsPerField.clear();
+    }
+
+    /**
+     * @param msgClsName Message class name.
+     * @param field Field.
+     * @param type Type.
+     * @param custMappAnn Cust mapp ann.
```
sergey-chugunov-1985:
```suggestion
     * @param custMappAnn Custom mapper annotation declared for an enum type.
```

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
```diff
@@ -294,11 +300,27 @@ private List<List<VariableElement>> hierarchicalOrderedFields(TypeElement type)
...
+                "Annotation @CustomMapper must only be used for enum fields or enum collections and maps, including nested ones.",
+                el);
+        }
+
+        enumsPerField.clear();
+    }
+
+    /**
+     * @param msgClsName Message class name.
+     * @param field Field.
```
sergey-chugunov-1985:
```suggestion
     * @param field Field being inspected.
```

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
```diff
@@ -294,11 +300,27 @@ private List<List<VariableElement>> hierarchicalOrderedFields(TypeElement type)
...
+     * @param type Type.
+     * @param custMappAnn Cust mapp ann.
+     */
+    private boolean validateEnumType(String msgClsName, Element field, TypeMirror type, CustomMapper custMappAnn) {
```
sergey-chugunov-1985:
```suggestion
    private boolean inspectFieldForEnumTypes(String msgClsName, Element field, TypeMirror type, CustomMapper custMappAnn) {
```

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
```diff
@@ -294,11 +300,27 @@ private List<List<VariableElement>> hierarchicalOrderedFields(TypeElement type)
...
+                el);
+        }
+
+        enumsPerField.clear();
+    }
+
+    /**
+     * @param msgClsName Message class name.
+     * @param field Field.
+     * @param type Type.
```
sergey-chugunov-1985:
```suggestion
     * @param type Type that should be inpected for enum type (direct type or type parameter).
```

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
```diff
@@ -294,11 +300,27 @@ private List<List<VariableElement>> hierarchicalOrderedFields(TypeElement type)
...
+                Diagnostic.Kind.ERROR,
+                "Annotation @CustomMapper must only be used for enum fields or enum collections and maps, including nested ones.",
+                el);
+        }
+
+        enumsPerField.clear();
+    }
+
+    /**
+     * @param msgClsName Message class name.
```
sergey-chugunov-1985:
```suggestion
     * @param msgClsName Message class name currently being inspected.
```

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
```diff
@@ -294,11 +300,27 @@ private List<List<VariableElement>> hierarchicalOrderedFields(TypeElement type)
...
+     * @param field Field.
+     * @param type Type.
+     * @param custMappAnn Cust mapp ann.
+     */
+    private boolean validateEnumType(String msgClsName, Element field, TypeMirror type, CustomMapper custMappAnn) {
+        String enumClsFullName = type.toString();
+        String enumMapperClsName = custMappAnn != null ? custMappAnn.value() : DLFT_ENUM_MAPPER_CLS;
 
+        if (enumType(processingEnv, type)) {
             IgniteBiTuple<String, String> otherMsgAndMapperClassesNames =
```
sergey-chugunov-1985: I suggest to move these two pieces of code into separate methods: this first one is checking that this enum type has only one mapper, either default or custom. And another one should perform a check that given type has only enum field associated with it.

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
```diff
@@ -313,16 +335,45 @@ private void validateEnumFieldMapping(TypeElement type, Element el) {
                             otherEnumMapperClsName + " in " + otherMsgClsName + " and " +
                             enumMapperClsName + " in " + msgClsName +
                             ". Only one mapper is allowed per enum type.",
-                        el);
+                        field);
                 }
             }
+
+            String otherEnum = type.toString();
```
sergey-chugunov-1985: This code could be moved to a separate method as the previous comment suggests.

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
shishkovilja: Field removed.

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
```diff
@@ -294,34 +297,99 @@ private List<List<VariableElement>> hierarchicalOrderedFields(TypeElement type)
...
+                "Annotation @CustomMapper must only be used for enum fields or enum collections and maps, including nested ones.",
+                el);
+        }
+    }
+
+    /**
+     * @param msgClsName Message class name currently being inspected.
+     * @param field Field being inspected.
+     * @param type Type that should be inpected for enum type (direct type or type parameter).
+     * @param custMappAnn Custom mapper annotation declared for an enum type.
```
sergey-chugunov-1985:
```suggestion
     * @param custMappAnn Custom mapper annotation declared for the enum type.
```

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
```diff
@@ -294,34 +297,99 @@ private List<List<VariableElement>> hierarchicalOrderedFields(TypeElement type)
...
+     * @param enumsPerField Map for collecting enum types related to a partiular field.
```
sergey-chugunov-1985:
```suggestion
     * @param enumsPerField Map for collecting enum types related to a particular field.
```

modules/codegen/src/main/java/org/apache/ignite/internal/MessageProcessor.java
```diff
@@ -294,34 +297,99 @@ private List<List<VariableElement>> hierarchicalOrderedFields(TypeElement type)
...
+
+            return inspectFieldForEnumTypes(msgClsName, field, keyType, custMappAnn, enumsPerField) |
+                inspectFieldForEnumTypes(msgClsName, field, valType, custMappAnn, enumsPerField);
+        }
+
+        return false;
+    }
+
+    /**
+     * Checks, that only single type of mapper is used for an enum type.
```
sergey-chugunov-1985:
```suggestion
     * Checks, that only single type of mapper is used for the enum type.
```

modules/core/src/test/java/org/apache/ignite/internal/codegen/MessageProcessorTest.java
```diff
@@ -343,6 +390,25 @@ public void testDifferentMappersForTheSameEnumAreProhibited() {
         assertThat(compilation).hadErrorContaining(errMsg);
     }
 
+    /**
+     * Negative test for a coflict situation when two enum mappers are used for the same enum in different messages.
```
sergey-chugunov-1985:
```suggestion
     * Negative test for a conflict situation when two enum mappers are used for the same enum in different messages.
```

sergey-chugunov-1985 (approved):

PR #13437, opened 2026-08-05
IGNITE-26630 Separate the stripe key from the cache partition

anton-vinogradov:
Step 1 of IGNITE-26630. No wire change: only method names move, no message field is touched.

### Why

`GridCacheMessage#partition()` sounds like a data partition, but it is the key that picks a stripe. The name has already misled: the ticket asked to remove the method, and removing it would send every cache message from the striped pool to the plain system pool, because the base class returns `-1` while "do not stripe" is `Integer.MIN_VALUE`. No functional test would notice.

### What changed

A new `StripedMessage` interface, next to `ExecutorAwareMessage` and shaped like it:

```java
public interface StripedMessage extends Message {
    public static final int ANY_STRIPE = -1;
    public static final int NO_STRIPE = Integer.MIN_VALUE;

    public int stripeIdx();
}
```

`GridIoMessage` loses its chain:

```java
-    public int partition() {
-        if (msg instanceof GridCacheMessage)
-            return ((GridCacheMessage)msg).partition();
-        if (msg instanceof DataStreamerRequest)
-            return ((DataStreamerRequest)msg).partition();
-        else
-            return STRIPE_DISABLED_PART;
-    }
+    @Override public int stripeIdx() {
+        return msg instanceof StripedMessage ? ((StripedMessage)msg).stripeIdx() : NO_STRIPE;
+    }
```

`STRIPE_DISABLED_PART` moves into the interface as `NO_STRIPE` and stops being a boxed `Integer`, and the bare `-1` literals become `ANY_STRIPE`.

Two classes where the name and the meaning did not match are now explicit:

* `GridCacheQueryRequest` keeps `partition()` — that one really is the partition to scan, read in `GridCacheDistributedQueryManager` — and gets `stripeIdx()` alongside it.
* `GridNearAtomicCheckUpdateRequest` loses `partition()`. The value it carries is the stripe of the update request, which the primary copies into the stripe of the response; the handler says so itself: "Message is processed in the same stripe".

### What is deliberately left alone

The fields are still named `part` and `partId`, and so are the constructor parameters they feed. Step 2 of the ticket deletes those fields — six of them travel the wire only to pick a stripe and nothing reads them — so renaming them now would churn the generated serializers and their reference files for nothing.

### Checks

* the set of striped messages is unchanged: `StripedMessage` is implemented by `GridCacheMessage` and `DataStreamerRequest`, exactly the two arms of the old chain;
* no message class was left with an orphaned `partition()` that silently stopped overriding anything — the only such method left is the intentional one in `GridCacheQueryRequest`;
* `IgniteCacheAtomicProtocolTest` 26/26 with assertions enabled — this is the stripe-sensitive path, where the deferred response buffer is a `ThreadLocal` flushed back into its own stripe;
* `DataStreamerImplSelfTest`, `CacheScanQueryFailoverTest`, `GridCacheQueryTransformerSelfTest`: 61 run, 0 failed;
* full `test-compile` of all modules, and checkstyle under `-Pcheckstyle`.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/StripedMessage.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/StripedMessage.java
@@ -19,9 +19,7 @@ package org.apache.ignite.internal;
 
 import org.apache.ignite.plugin.extensions.communication.Message;
 
-/**
- * Message that chooses the stripe it is processed in.
- */
+/** Message that chooses the stripe it is processed in. */
 public interface StripedMessage extends Message {
     /** Process in any stripe. */
     public static final int ANY_STRIPE = -1;
--- a/modules/core/src/main/java/org/apache/ignite/internal/StripedMessage.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/StripedMessage.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal;
+
+import org.apache.ignite.plugin.extensions.communication.Message;
+
+/**
+ * Message that chooses the stripe it is processed in.
+ */
+public interface StripedMessage extends Message {
+    /** Process in any stripe. */
+    public static final int ANY_STRIPE = -1;
+
+    /** Process in the pool itself, not in a stripe. */
+    public static final int NO_STRIPE = Integer.MIN_VALUE;
+
+    /**
+     * The value is an index, not a cache partition: messages sharing it are processed one after another
+     * by the same thread.
+     *
+     * @return Stripe index, {@link #ANY_STRIPE} or {@link #NO_STRIPE}.
+     */
+    public int stripeIdx();
+}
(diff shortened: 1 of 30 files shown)


anton-vinogradov:
/runall

---
🚀 **RunAll queued** — [build 9258046](https://ci2.ignite.apache.org/build/9258046) · live progress & verdict: [Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13437). The verdict lands here when the run finishes.

anton-vinogradov:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13437)** verdict · RunAll build [9258046](https://ci2.ignite.apache.org/build/9258046) · 147 suites ran, 0 reused

- 1 suite(s) have no reliable result (compilation error, timeout, crash)

- Snapshots 6: JVM crash / out of memory · 1 failed test detected

🔎 **No blockers found — but the run above can't prove the PR is clean.** 18 pre-existing/flaky tests filtered out. Re-run once the above is sorted out.

_♻️ Settled after 2 auto re-run wave(s): #1 — 3 blocker + 2 broken suite(s); #2 — 1 broken suite(s)._

modules/core/src/main/java/org/apache/ignite/internal/StripedMessage.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal;
+
+import org.apache.ignite.plugin.extensions.communication.Message;
+
+/**
+ * Message that chooses the stripe it is processed in.
```
wernerdv: Let's do it in one line.

modules/core/src/main/java/org/apache/ignite/internal/processors/datastreamer/DataStreamerImpl.java
```diff
@@ -2003,7 +2003,7 @@ private void submit(
                     dep == null,
                     topVer,
                     (rcvr == ISOLATED_UPDATER) ?
-                        partId : GridIoMessage.STRIPE_DISABLED_PART);
+                        partId : NO_STRIPE);
```
wernerdv: Let's combine with the previous line.

wernerdv (approved):

anton-vinogradov:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13437)** verdict · RunAll build [9259924](https://ci2.ignite.apache.org/build/9259924) · 147 suites ran, 0 reused

- 1 suite(s) have no reliable result (compilation error, timeout, crash)

- [Check Code Style Ducktests]: non-zero exit code

🔎 **No blockers found — but the run above can't prove the PR is clean.** 22 pre-existing/flaky tests filtered out. Re-run once the above is sorted out.

PR #13439, opened 2026-08-05
IGNITE-28528 Move the start request conversion out of the message

anton-vinogradov:
`StartRequestData` marshalled itself: `marshal(GridKernalContext)` and `unmarshal(GridKernalContext, UUID)` picked the marshaller, resolved the peer deployment and ran the p2p hooks — all from inside the message. This moves that work to `GridContinuousProcessor`, the only place that has the sender id, the deployment and a thread where marshalling is allowed. The message keeps its fields and nothing else.

Nothing changes on the wire: the same eight `@Order` fields in the same order, the same marshaller, the same call order. What changes is where the marshaller is taken from — which is what IGNITE-28940 needs to switch it per transport.

Two details worth a look:

- The restored handler goes into the message right after it is read, before `p2pUnmarshal`. That order matters: `p2pUnmarshal` throws when the deployment is unavailable, and the caller still reads the handler afterwards to report the failure and stop the routine.
- `nodeFilter` and `hnd` became package-private, like the other fields of the class; the processor lives in the same package and reads them directly, as it already did with `nodeFilterBytes` and `hndBytes`.

Tested: `CacheContinuousQueryFilterDeploymentFailedTest`, `GridCacheContinuousQueryNodesFilteringTest`, `CacheContinuousQueryDeploymentToClientTest`, `CacheContinuousQueryAsyncFilterListenerTest`, `CacheContinuousQueryEventBufferTest`, `GridEventConsumeSelfTest` — green, checkstyle clean.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/continuous/GridContinuousProcessor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/continuous/GridContinuousProcessor.java
@@ -981,7 +981,7 @@ public class GridContinuousProcessor extends GridProcessorAdapter {
             reqData.deploymentInfo(dep);
         }
 
-        reqData.marshal(ctx);
+        marshalStartRequestData(reqData);
 
         if (!immutableDiscoCustomMsg) {
             StartRoutineDiscoveryMessage msg = new StartRoutineDiscoveryMessage(routineId, reqData, Mode.MUTABLE);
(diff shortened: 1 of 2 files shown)


anton-vinogradov:
Superseded by #13440, which revises the marshalling instead of moving the conversion code around: the message now states the deployment it carries, and the class loader is resolved once, in the deployment manager.

PR #13440, opened 2026-08-06
IGNITE-28528 Revise marshalling of the node filter in StartRequestData

anton-vinogradov:
The node filter in `StartRequestData` was marshalled by hand, and the receiver
had to build the deployment class loader itself before reading the filter.

### Why the code looked like that

A message that carries peer-deployed classes said nothing about it. It held the
deployment as four or five flat fields, so every receiver rebuilt the deployment
from those fields and passed the resulting loader into the unmarshal call. Eight
places did exactly that. `StartRequestData` also repeated a "peer class loading
is off" branch, although `GridDeploymentManager` already covers that case: with
peer class loading off it looks the deployment up locally.

### Change

* `DeploymentAware` lets a message state that its classes come with a deployment,
  the same way `CacheIdAware` states the need for a cache object context.
* `GridDeploymentManager.classLoader(DeploymentAware)` is now the single place
  that turns a carried deployment into a class loader.
* The generated marshaller asks for that loader when the caller passes none, so
  receivers no longer resolve it by hand.
* Messages carry `GridDeploymentInfoBean` as a whole instead of the flat fields.
  This changes the wire format.

`DeploymentAware` is independent of `DeferredUnmarshalMessage`, which says who
starts the read: a receiver that reports a missing deployment back to the sender
has to start the read itself, inside its own error handling.

### Reading a discovery message off the reading thread

The socket-reading threads of both discovery SPIs unmarshalled every message
they read. A continuous-query start request cannot be read there: obtaining the
deployment of its node filter may block on a network request, and a missing user
class has to be tolerated so the acknowledgement is still sent. Both readers now
honour `DeferredUnmarshalMessage`, as `GridIoManager` already did, and the
continuous processor reads the request where that work is allowed.

`StartRequestData` is then left as plain data. Its `marshal`/`unmarshal` methods
are gone: the fields are marshalled by the generated marshaller, and preparing
the handler moved to the processor that builds and consumes the message.

### Verified

`MessageProcessorTest` (a new case covers the resolved loader and `keepBytes`),
`GridEventStorageCheckAllEventsSelfTest`, `GridEventStorageManagerSelfTest`,
`GridEventConsumeSelfTest`, `GridMessageListenSelfTest`,
`GridP2PContinuousDeploymentSelfTest`, `P2PClassLoadingFailureHandlingTest`,
`IgniteNoCustomEventsOnNodeStart`, `DataStreamerImplSelfTest`,
`GridP2PContinuousDeploymentClientDisconnectTest`, and the continuous-query
suites `CacheContinuousQueryOperationP2PTest`,
`CacheContinuousQueryOperationFromCallbackTest`,
`CacheContinuousQueryFailoverAtomicSelfTest`,
`GridCacheContinuousQueryNodesFilteringTest`,
`IgniteCacheContinuousQueryClientTest`,
`IgniteCacheContinuousQueryReconnectTest`, `ContinuousQueryBuffersCleanupTest`.


anton-vinogradov:
/runall

---
🎨 _Checkstyle autofix: fixed 2 of 15 violation(s) in 2 file(s) — commit 7454525e. 13 remain (javadoc/naming/wrapping need a human) — the Checkstyle suite may still fail._
🚀 **RunAll queued** — [build 9266790](https://ci2.ignite.apache.org/build/9266790) · live progress & verdict: [Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13440). The verdict lands here when the run finishes.
🏁 _Run finished — analysing; the verdict comment follows._

anton-vinogradov:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13440)** verdict · RunAll build [9266790](https://ci2.ignite.apache.org/build/9266790) · 147 suites ran, 0 reused

- a newer run is still going — its unfinished suites can still fail

👀 **6 test(s) started failing on this code** — not proven blockers yet: too few runs of this revision to tell a break from a flake, so a re-run of the suite decides it.
- PDS 10: `org.apache.ignite.testsuites.IgnitePdsTestSuite10: org.apache.ignite.internal.processors.cache.distributed.dht.preloader.HistoricalRebalanceCheckpointTest.testDelayedToBackupsRequests1BackupMorePuts`
- Disk Page Compressions 3: `org.apache.ignite.testsuites.IgnitePdsCompressionTestSuite3: org.apache.ignite.internal.processors.localtask.DurableBackgroundTasksProcessorSelfTest.testDontDeleteTaskIfItsRestart`
- Cache 19: `org.apache.ignite.testsuites.IgniteCacheTestSuite17: org.apache.ignite.internal.processors.cache.transactions.TxPartitionCounterStateOnePrimaryTwoBackupsHistoryRebalanceTest.testPartialCommit_2TX_4`
- Client Nodes: `org.apache.ignite.testsuites.IgniteClientReconnectTestSuite: org.apache.ignite.internal.IgniteClientReconnectAtomicsWithLostPartitionsTest.testAtomicSequenceAddAngGet`
- Client Nodes: `org.apache.ignite.testsuites.IgniteClientReconnectTestSuite: org.apache.ignite.internal.IgniteClientReconnectAtomicsWithLostPartitionsTest.testAtomicStampedCompareAndSet`
- Client Nodes: `org.apache.ignite.testsuites.IgniteClientReconnectTestSuite: org.apache.ignite.internal.IgniteClientReconnectAtomicsWithLostPartitionsTest.testAtomicLongGet`

⚠️ **No proven blocker yet** — this is not an all-clear: see the tests above. 51 pre-existing/flaky filtered out.

⏳ _Auto re-run **#2** in progress — 2 blocker suite(s) re-queued (attempt 2/2), **≈ settled by 09:13 MSK**. This comment updates when they settle._
_Earlier re-runs: #1 — 3 blocker suite(s)._

anton-vinogradov:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13440)** verdict · RunAll build [9267428](https://ci2.ignite.apache.org/build/9267428) · 147 suites ran, 0 reused

- 1 commit(s) pushed since this run — it tested older code

🔎 **No blockers found — but the run above can't prove the PR is clean.** 25 pre-existing/flaky tests filtered out. Re-run once the above is sorted out.

_♻️ Settled after 1 auto re-run wave(s): #1 — 4 watch suite(s)._

anton-vinogradov:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13440)** verdict · RunAll build [9268108](https://ci2.ignite.apache.org/build/9268108) · 147 suites ran, 0 reused

✅ **No blockers** — nothing in this run looks caused by this PR. 18 pre-existing/flaky tests filtered out.

_♻️ Settled after 1 auto re-run wave(s): #1 — 3 watch suite(s)._

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentManager.java
```diff
@@ -402,6 +404,76 @@ private GridDeployment checkDeployment(GridDeployment deployment, String store)
...
+     * @param rsrcName Name the classes are deployed under.
+     * @param clsName Name of a class the deployment must be able to load.
+     * @return The deployment, or {@code null} when there is none.
+     */
+    @Nullable public GridDeployment globalDeployment(GridDeploymentInfo depInfo, String rsrcName, String clsName) {
+        return getGlobalDeployment(depInfo.deployMode(),
+            rsrcName,
+            clsName,
+            depInfo.userVersion(),
+            depInfo.classLoaderId().globalId(),
```
petrov-mg: It's confusing that `depInfo.classLoaderId().globalId()` is used as `sndNodeId`. Why is that?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryDeployableObject.java
```diff
@@ -88,11 +88,7 @@ protected CacheContinuousQueryDeployableObject(Object obj, GridKernalContext ctx
     <T> T unmarshal(UUID nodeId, GridKernalContext ctx) throws IgniteCheckedException {
         assert ctx != null;
 
-        GridDeployment dep = ctx.deploy().getGlobalDeployment(depInfo.deployMode(), clsName, clsName,
-            depInfo.userVersion(), nodeId, depInfo.classLoaderId(), depInfo.participants(), null);
-
-        if (dep == null)
-            throw new IgniteDeploymentCheckedException("Failed to obtain deployment for class: " + clsName);
+        GridDeployment dep = ctx.deploy().globalDeployment(depInfo, clsName);
```
petrov-mg: Was the `nodeId` method parameter lost during refactoring, or is it no longer required and should be deleted?

petrov-mg:
Generated code for `GridEventStorageRequestMarshaller` looks confusing:

```
     /** */
    @Override public void unmarshal(GridEventStorageRequest msg, GridKernalContext kctx, CacheObjectContext cacheObjCtx, ClassLoader clsLdr) throws IgniteCheckedException {
        CacheObjectContext ctx = cacheObjCtx;

        clsLdr = kctx.deploy().classLoader(msg);
```

Why is `clsLdr` overwritten here?

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentManager.java
anton-vinogradov: The two are the same node by construction, so this was not a shortcut - but you are right that nothing at the call site said so.

A class loader id carries the id of the node that created it, and `GridDeploymentClassLoader` asserts it in both places where such a loader is built:

```java
assert nodeId.equals(clsLdrId.globalId());   // :184, constructor
assert nodeId.equals(ldrId.globalId());      // :324, register(nodeId, ldrId), for a participant
```

So a message describing a deployment cannot name a sender that disagrees with the loader it describes: the old code, which passed the sender separately, blew up on that assert if the two ever diverged.

I have made it readable instead of implied: `GridDeploymentInfo#nodeId()` now returns it, with the invariant written in its javadoc, and the call site reads `depInfo.nodeId()`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryDeployableObject.java
anton-vinogradov: No longer required - and it turned out to be true well past this method.

The deployment is now resolved from the descriptor the message carries, so the node id has nothing to do here. Following it upwards, it had nothing to do anywhere else either: it went through `GridContinuousHandler#p2pUnmarshal`, its three implementations and the private helper in `CacheContinuousQueryHandler`, and every one of them only fed it to an `assert nodeId != null`.

Deleted along the whole chain, so the signature is `p2pUnmarshal(GridKernalContext)` now.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentManager.java
```diff
@@ -402,6 +404,76 @@ private GridDeployment checkDeployment(GridDeployment deployment, String store)
...
+     * @param depInfo Deployment of the classes, as it came with the message carrying them.
+     * @param rsrcName Name the classes are deployed under.
+     * @param clsName Name of a class the deployment must be able to load.
+     * @return The deployment, or {@code null} when there is none.
+     */
+    @Nullable public GridDeployment globalDeployment(GridDeploymentInfo depInfo, String rsrcName, String clsName) {
+        return getGlobalDeployment(depInfo.deployMode(),
+            rsrcName,
+            clsName,
+            depInfo.userVersion(),
```
petrov-mg, line 467: Is `depInfo` guaranteed to be non-null here? In some places it seems that it can be null, for example in `org.apache.ignite.internal.managers.communication.GridIoUserMessage#deploymentInfo`.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentManager.java
anton-vinogradov: I was wrong here, and your question is what led me to it - thanks.

The two nodes are not the same in general. `GridCacheDeploymentManager` states it directly:

```java
assert sndId.equals(ldrId.globalId()) || participants != null;
```

and `GridDeploymentManager#deploy` hands out the deployment of the *original* owner when the class arrived by peer loading - the comment there calls it a nested execution:

```java
// Check for nested execution. In that case, if task
// is available locally by name, then we should ignore class loader ID.
dep = checkDeployment(ldrStore.getDeployment(ldr.classLoaderId()), "perLoader");
```

So a node that received classes from another one passes them further as a participant of the same deployment, while being the sender itself.

And the receiving side needs the sender, not the owner: `GridDeploymentPerVersionStore` asks it for the classes, records it with `addParticipant`, and refuses the deployment when `discovery().node(senderNodeId) == null`. Deriving the node from the loader id would break exactly the case where the owner has left and the sender still holds the classes.

Fixed: the resolution takes the node as a parameter again. Since generated code has no way to know the sender, the `DeploymentAware` interface is gone with it, and the callers - which do know the sender - pass it in.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryDeployableObject.java
anton-vinogradov: Correction to my previous reply: the parameter is back.

The deployment resolution needs the sending node after all - it is not derivable from the loader id, see the other thread. So `p2pUnmarshal(UUID nodeId, GridKernalContext ctx)` keeps its signature, and the node is passed down to the resolution instead of only feeding an assert.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentManager.java
anton-vinogradov: It cannot be null today, but only by an invariant that is nowhere stated - you are right not to trust it.

On the sending side both fields come from the same deployment, and the send fails outright when there is none:

```java
dep = ctx.deploy().deploy(cls0, U.detectClassLoader(cls0));

if (dep == null)
    throw new IgniteDeploymentCheckedException(...);

depClsName = cls0.getName();
```

So `deploymentClassName() != null` implies `deploymentInfo() != null`, and the guard on the receiving side happens to check the first while the code dereferences the second.

Rather than document that, I made it structural: the resolution now accepts a null deployment and returns null for it, so nothing to resolve gives nothing back, and the caller reports it with the message it already has. The throwing overload turns the same case into its usual `IgniteDeploymentCheckedException`.

anton-vinogradov:
> Generated code for `GridEventStorageRequestMarshaller` looks confusing:
>
> ```java
> @Override public void unmarshal(GridEventStorageRequest msg, GridKernalContext kctx, CacheObjectContext cacheObjCtx, ClassLoader clsLdr) throws IgniteCheckedException {
>     CacheObjectContext ctx = cacheObjCtx;
>
>     clsLdr = kctx.deploy().classLoader(msg);
> ```
>
> Why is `clsLdr` overwritten here?

Good catch, and it was a symptom of the same thing your inline question uncovered - that generated code cannot pick the class loader for a message.

Overwriting the parameter was deliberate but wrong. It came from `DeploymentAware`: a message declared that it carries its own deployment, and the generated marshaller resolved the loader from it. I made the assignment unconditional because the convenience overloads pass a loader rather than `null`:

```java
default void unmarshal(M msg, GridKernalContext kctx) {
    unmarshal(msg, kctx, null, U.resolveClassLoader(kctx.config()));
}
```

so a caller using them would silently read peer-deployed classes with the local loader. Ignoring the argument fixed that symptom and hid a worse problem: the resolution needs the sending node, which the generated code has no way to know, and the sender is not always the owner of the class loader.

So the whole thing is gone now. `DeploymentAware`, the resolution in the generator and its test are removed - `modules/codegen` is no longer part of this PR at all - and the resolution takes the node as a parameter, passed by the callers that know it.

What is left is what the ticket asked for: the deployment travels as one descriptor instead of four flat fields, and the eight places that used to assemble it by hand now make a single call. The diff went from 35 files to 25, and from +577/-497 to +248/-475.

Since the change is materially different from what the last RunAll saw, a fresh run is needed.

anton-vinogradov:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13440)** verdict · RunAll build [9269815](https://ci2.ignite.apache.org/build/9269815) · 147 suites ran, 0 reused

- 1 suite(s) ran far fewer tests than the same suites on master

- Platform C++ CMake (Linux Clang): 728 tests vs 1047 on master (**−30%**)

🔎 **No blockers found — but the run above can't prove the PR is clean.** 20 pre-existing/flaky tests filtered out. Re-run once the above is sorted out.

shishkovilja (approved):
There are some messages, that have `GridDeploymentInfoMessage` field, but doesn't implement `DeploymentAware`:
- `GridJobExecuteRequest`
- `GridIoUserMessage`
- `GridCacheMessage`
- `DataStreamerRequest`

Is it planned, that these messages (or any other) become `DeploymentAware` in the future?

anton-vinogradov:
The wire format change is intentional and is the point of this ticket, so here is what actually changes and what does not.

**Changed** - four messages carry the deployment as one descriptor instead of four flat fields (`deployMode`, `userVersion`, `classLoaderId`, `participants`):

| message | `@Order` fields |
|---|---|
| `GridJobExecuteRequest` | 25 → 22 |
| `DataStreamerRequest` | 16 → 13 |
| `GridIoUserMessage` | 7 → 4 |
| `GridEventStorageRequest` | 7 → 4 |

**Not changed**, though the check lists them:

- `StartRequestData` - 8 fields before and after; only field visibility and the code reading them moved.
- `GridCacheMessage` - the type of an existing field was renamed, the field itself stays.
- `GridDeploymentInfoBean` / `GridDeploymentInfoMessage` - one class renamed under IGNITE-28967. Its own fields and its `directType` are untouched: the registration keeps its place in `CoreMessagesProvider`, and `META-INF/classnames.properties` is updated.

Rolling upgrade: this lands before 2.19 is released, which is the window such changes are meant for - the same reasoning as in IGNITE-28940. Between 2.18 and 2.19 these four messages are not compatible, deliberately.

One note on the check itself: `GridJobExecuteRequest` changes the most here and is **missing** from the list above, while three classes that do not change the format are listed. That matches the detection defects filed as IGNITE-28966 - the check is blind to messages in its own package and reports files whose change is not on the wire.

anton-vinogradov:
> There are some messages, that have `GridDeploymentInfoMessage` field, but doesn't implement `DeploymentAware`:
> - `GridJobExecuteRequest`
> - `GridIoUserMessage`
> - `GridCacheMessage`
> - `DataStreamerRequest`
>
> Is it planned, that these messages (or any other) become `DeploymentAware` in the future?

Thanks for the review. The answer turned out to be no, and `DeploymentAware` is gone from this PR entirely - it was removed a few hours before your review, so you were looking at the earlier version.

It let a message declare "my classes come with this deployment", and the generated marshaller resolved the class loader from it. @petrov-mg asked why `depInfo.classLoaderId().globalId()` was used as the sender node, and digging into that showed the interface was promising something a message cannot deliver.

Resolving a deployment needs the **sending** node, and it is not the owner of the class loader in general:

```java
assert sndId.equals(ldrId.globalId()) || participants != null;   // GridCacheDeploymentManager
```

A node that received classes by peer loading passes them on as a participant of the same deployment - `GridDeploymentManager#deploy` hands it the deployment of the original owner, which the comment there calls a nested execution. The receiving side then asks the sender for the classes, records it with `addParticipant`, and drops the deployment when that node has left. Deriving it from the loader id breaks exactly the case participants exist for.

Generated code cannot supply the sender - `unmarshal` only gets a `GridKernalContext` - so the interface had no way to be correct. It is deleted, together with the resolution in the generator and its test; `modules/codegen` is no longer part of this PR.

What all these messages do share is the descriptor itself: they carry `GridDeploymentInfoMessage` instead of four flat fields, and each receiver makes one call - `globalDeployment(depInfo, clsName, sndNodeId)` - instead of assembling the deployment by hand. That was the point of the ticket, and it holds without the marker interface.

anton-vinogradov:
/runall

---
🎨 _Checkstyle autofix: fixed 2 of 15 violation(s) in 2 file(s) — commit 7454525e. 13 remain (javadoc/naming/wrapping need a human) — the Checkstyle suite may still fail._
🚀 **RunAll queued** — [build 9266790](https://ci2.ignite.apache.org/build/9266790) · live progress & verdict: [Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13440). The verdict lands here when the run finishes.
🏁 _Run finished — analysing; the verdict comment follows._

⚠️ _GitHub rejected your personal access token, so this is narrated from the checker's own account. Save a fresh PAT in the checker's settings to get your own back._

anton-vinogradov:
@anton-vinogradov 🚀 **RunAll queued** — [build 9275828](https://ci2.ignite.apache.org/build/9275828) · live progress & verdict: [Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13440). The verdict lands here when the run finishes.
⏱ _~15m remaining — **≈ 05:35 MSK** (updates every minute)._

PR #13447, opened 2026-08-07
IGNITE-28920 Reduce MarshallableMessage scope: GridCacheEntryInfo uses marshal hooks only to rebase expireTime

Vladsz83:

### Notes


anton-vinogradov:
An entry from a cache without an expiry policy now arrives already expired.

`initTime == 0` is the "no expiration" mark, but `initTime` has no `@Order`, so it does not go on the wire, and the empty constructor the message factory uses always sets it:

```java
public GridCacheEntryInfo() {
    initTime = System.currentTimeMillis();
}
```

So on the receiver `initTime` is never zero and `expireTime()` returns `initTime + expireTimeDelta`, the moment the message object was created. Every expiration check in the code reads that as expired, `expireTime > 0 && expireTime <= U.currentTimeMillis()`.

* sender, no expiry policy: `initTime = 0`, `expireTimeDelta = 0`, and only the delta is transferred
* receiver: `initTime = <read time>`, `expireTimeDelta = 0`, `expireTime() = <read time>`

That is every entry of every cache that has no expiry policy, on rebalance and on every get response. On rebalance `IgniteCacheOffheapManagerImpl.storeEntries` puts the value into the data row, so the entries land with a pending expiration and the TTL thread removes them.

`GridCacheRebalancingSyncCheckDataTest` catches it. Green on master, red on bc38d2b2:

```
[ERROR] Tests run: 1, Failures: 1 -- GridCacheRebalancingSyncCheckDataTest
GridCacheRebalancingSyncCheckDataTest.access$100:37->JUnitAssertAware.assertNotNull:180
```

Same thing at the unit level:

```java
assertEquals(0, writeAndReadBack(entryInfo(0)).expireTime());

java.lang.AssertionError: expected:<0> but was:<1786287823668>
```

The mark has to live in a transferred field. `-1` in the delta itself does that, and in the varint wire format it costs the same single byte as `0`, so it is free. That, together with the clamp that keeps the negative range unambiguous, is what https://github.com/Vladsz83/ignite/pull/1 does.

anton-vinogradov:
`assert initTime > 0` in `expireTime()` fires for a cache without an expiry policy.

The constructor sets `initTime` only when there is an expiration:

```java
if (expireTime != 0) {
    initTime = U.currentTimeMillis();
    ...
}
```

so for a never expiring entry it stays 0, and the assert runs before the `-1` check:

```java
public long expireTime() {
    assert initTime > 0;            // <- here
    assert expireTimeDelta >= -1L;

    return expireTimeDelta == -1L ? 0L : initTime + expireTimeDelta;
}
```

That hits any locally built info of a cache with no expiry policy, `GridCacheMapEntry.info()` read through `GridNearCacheEntry.initializeFromDht:156`. Assertions are on in a normal test run, I did not have to pass `-ea`:

```
java.lang.AssertionError
	at org.apache.ignite.internal.processors.cache.GridCacheEntryInfo.expireTime(GridCacheEntryInfo.java:131)
	at ...GridCacheEntryInfoSerializationTest.testLocalNeverExpiringEntry
```

with

```java
assertEquals(0, new GridCacheEntryInfo(cacheId, key, val, ver, 0, 0).expireTime());
```

Setting `initTime` unconditionally fixes it and costs one volatile read.

Thanks for taking the `-1` sentinel, the clamp and `U.currentTimeMillis()`, the never expiring entry is a single byte on the wire again.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
```diff
@@ -64,58 +68,70 @@ public class GridCacheEntryInfo implements SelfMarshallingMessage, CacheIdAware
...
-    }
-
     /**
-     * @param cacheId Cache ID.
+     * Empty constructor for serialization purposes.
+     * see {@link #expireTimeDelta}.
      */
-    public void cacheId(int cacheId) {
+    public GridCacheEntryInfo() {
+        initTime = U.currentTimeMillis();
```
anton-vinogradov, line 77: Do we need initTime always, even when expireTime == 0?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
```diff
@@ -64,58 +68,70 @@ public class GridCacheEntryInfo implements SelfMarshallingMessage, CacheIdAware
...
+
+        if (expireTime != 0) {
+            initTime = U.currentTimeMillis();
+
+            expireTimeDelta = expireTime - initTime;
+
+            // In theory, here we can get the thread paused causing a negative delta value. Possible negative values
+            // shouldn't be treated as disabled expiration. Correct behavior is expired timeout.
+            if (expireTimeDelta < 0)
+                expireTimeDelta = 0;
```
anton-vinogradov: Do we need a warning here?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -396,23 +396,14 @@ protected GridDhtLocalPartition localPartition() {
...
             if (!obsolete()) {
-                info = new GridCacheEntryInfo();
-
-                info.key(key);
-                info.cacheId(cctx.cacheId());
-
                 long expireTime = expireTimeExtras();
 
-                boolean expired = expireTime != 0 && expireTime <= U.currentTimeMillis();
+                CacheObject val0 = expireTime == 0 || expireTime > U.currentTimeMillis() ? val : null;
```
anton-vinogradov: Should expireTime be transferred to GridCacheEntryInfo to make a clear decision once?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
```diff
@@ -64,58 +68,70 @@ public class GridCacheEntryInfo implements SelfMarshallingMessage, CacheIdAware
...
-
-    /**
-     * @return Expire time.
+     * @return Expire time >= 0. 0 means no expiration is set.
      */
     public long expireTime() {
-        return expireTime;
-    }
+        assert initTime > 0;
+        assert expireTimeDelta >= -1L;
```
anton-vinogradov: Asserts give us nothing in prod.
Should be solved by logic, not by asserts.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
```diff
@@ -64,58 +68,70 @@ public class GridCacheEntryInfo implements SelfMarshallingMessage, CacheIdAware
...
 
-    /**
-     * @param expireTime Expiration time.
-     */
-    public void expireTime(long expireTime) {
-        this.expireTime = expireTime;
+        return expireTimeDelta == -1L ? 0L : initTime + expireTimeDelta;
```
anton-vinogradov: Possible negative number without handling.

anton-vinogradov (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
```diff
@@ -64,58 +68,70 @@ public class GridCacheEntryInfo implements SelfMarshallingMessage, CacheIdAware
...
-    }
-
-    /**
-     * @return Expire time.
+     * @return Expire time >= 0. 0 means no expiration is set.
      */
     public long expireTime() {
-        return expireTime;
-    }
+        assert initTime > 0;
```
anton-vinogradov: Will it fail on caches without expiration?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
```diff
@@ -64,58 +68,70 @@ public class GridCacheEntryInfo implements SelfMarshallingMessage, CacheIdAware
...
+    /** */
+    public GridCacheEntryInfo(int cacheId, KeyCacheObject key, @Nullable CacheObject val, GridCacheVersion ver, long expireTime, long ttl) {
+        assert expireTime >= 0;
+
+        if (expireTime != 0) {
+            initTime = U.currentTimeMillis();
+
+            expireTimeDelta = expireTime - initTime;
+
+            // In theory, here we can get the thread paused causing a negative delta value. Possible negative values
```
anton-vinogradov: Wrong definition, see my comment about code relocation.
GridCacheMapEntry.info() can now supply an obsolete expireTime (less than the current time)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
```diff
@@ -64,58 +68,70 @@ public class GridCacheEntryInfo implements SelfMarshallingMessage, CacheIdAware
...
 
-    /** {@inheritDoc} */
-    @Override public int cacheId() {
-        return cacheId;
-    }
-
     /**
-     * @param cacheId Cache ID.
+     * Empty constructor for serialization purposes.
+     * see {@link #expireTimeDelta}.
```
anton-vinogradov: Not a clear see. See for what?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
```diff
@@ -19,40 +19,44 @@
...
-    CacheObject val;
+    @Nullable CacheObject val;
 
     /** Time to live. */
     @Order(3)
     long ttl;
 
-    /** Expiration time. */
+    /** Base time to calculate {@link #expireTime()}. */
+    long initTime;
```
anton-vinogradov: private?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
```diff
@@ -190,30 +192,6 @@ public int marshalledSize(CacheObjectContext ctx) throws IgniteCheckedException
...
-
-        expireTime = remaining < 0 ? 0 : U.currentTimeMillis() + remaining;
-
-        // Account for overflow.
-        if (expireTime < 0)
-            expireTime = 0;
-    }
-
     /** {@inheritDoc} */
     @Override public String toString() {
```
anton-vinogradov, line 208: Missed new fields

anton-vinogradov (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
Vladsz83: We can't determine. We have no Message lifecycle available except deprecated `SelfMarshallingMessage`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
Vladsz83: There is no expireTime any more

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
Vladsz83: Explain pls

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
Vladsz83: Should not be missed. Or explain pls.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
Vladsz83: Didn't get you. We estimate a value here. Needed or not. Messages should not be smart. Left as is before. Looks not exploding at least.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
anton-vinogradov: initTime + expireTimeDelta has no overflow guard, while the removed selfUnmarshal clamped a negative result back to 0 with the comment Account for overflow, and GridDhtPartitionDemander.preloadEntry asserts row.expireTime() >= 0. Please clamp the sum in expireTime() the same way, and compare the delta with < 0 rather than == -1L, since selfUnmarshal read the whole negative range as no expiration.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
anton-vinogradov: The expired decision is split across two clock reads, expireTime > U.currentTimeMillis() here and expireTime - U.currentTimeMillis() in the constructor, so on different ticks the entry goes out with its value while its delta is clamped to 0. Please read the time once and pass it into the constructor, so the value and the delta come from the same decision.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
anton-vinogradov: Correct, checked that it autofilled.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
anton-vinogradov: The idea is to have it unset at all. 
Any reason to set it here?

Vladsz83: > The idea is to have it unset at all. Any reason to set it here?

Reason to work cache entries expiration of course. We need to calculate `expireTime()`. See the javacomment pls

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
Vladsz83: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
```diff
@@ -64,58 +68,80 @@ public class GridCacheEntryInfo implements SelfMarshallingMessage, CacheIdAware
...
+        long initTime,
+        long expireTime,
+        long ttl
+    ) {
+        assert expireTime >= 0;
+        assert initTime > 0;
+
+        this.initTime = initTime;
+
+        expireTimeDelta = expireTime - initTime;
```
anton-vinogradov: Incorrect logic for expireTime == 0 (CU.EXPIRE_TIME_ETERNAL) will cause a timeout for such entries.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
```diff
@@ -64,58 +68,80 @@ public class GridCacheEntryInfo implements SelfMarshallingMessage, CacheIdAware
...
+
+        // Timeouted mark.
```
anton-vinogradov: Timed out/Timeout?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
```diff
@@ -64,58 +68,80 @@ public class GridCacheEntryInfo implements SelfMarshallingMessage, CacheIdAware
...
-    }
-
-    /**
-     * @return Expire time.
+     * @return Expire time >= 0. 0 means no expiration is set.
      */
     public long expireTime() {
-        return expireTime;
-    }
+        if (expireTimeDelta == -1L)
```
anton-vinogradov, line 140: Seems to be an impossible case, since -1 is an initial value replaced by real?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
```diff
@@ -19,40 +19,44 @@
...
+import org.apache.ignite.plugin.extensions.communication.Message;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Entry information that gets passed over wire.
  */
-public class GridCacheEntryInfo implements SelfMarshallingMessage, CacheIdAware {
+public class GridCacheEntryInfo implements CacheIdAware, Message {
     /** */
     private static final int SIZE_OVERHEAD = 3 * 8 /* reference */ + 4 /* int */ + 2 * 8 /* long */ + 32 /* version */;
```
anton-vinogradov, line 35: Please check correctness.

anton-vinogradov (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
Vladsz83: Pls explain. Didnt touch. Unrelated to this ticket.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
Vladsz83: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
Vladsz83: -1 is if the expiration isnt set at all. Then -1

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryInfo.java
anton-vinogradov: The question was: Is it correct, BTW %)

anton-vinogradov (approved):

PR #13461, opened 2026-08-10
IGNITE-28972 Generation must fail Message to byte[] marshalling

nizhikov:

### Notes


modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
```diff
@@ -1147,6 +1147,8 @@ private enum MarshalledKind {
         if (ann == null)
             return null;
 
+        ensureCorrectlyAnnotated(field, ann);
```
anton-vinogradov: The check runs before the kind is computed, so it judges the field type instead of the companion and rejects a working pattern (Set<GridTopicMessage> + GridTopicMessage[] companion) that the protocol writes itself. Please move the check after the kind is known and apply it only to BLOB and ELEMENT_BLOBS.

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
```diff
@@ -1170,6 +1172,41 @@ private enum MarshalledKind {
         return MarshalledKind.ELEMENT_BLOBS;
     }
 
+    /** */
```
anton-vinogradov: The contract here is not obvious and the name promises a general check while one rule is implemented. Please describe what is rejected and rename after that rule.

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
```diff
@@ -1170,6 +1172,41 @@ private enum MarshalledKind {
...
+    private void ensureCorrectlyAnnotated(VariableElement field, Marshalled ann) {
+        boolean msgToBytes = false;
+
+        if (assignableFrom(field.asType(), msgType)) {
```
anton-vinogradov: Single-statement branch. Please drop the braces.

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
```diff
@@ -1170,6 +1172,41 @@ private enum MarshalledKind {
...
+            msgToBytes = true;
+        }
+        else if (isCollection(field.asType())) {
+            DeclaredType type = (DeclaredType)field.asType();
+
+            List<? extends TypeMirror> typeArgs = type.getTypeArguments();
+
+            assert typeArgs.size() == 1;
```
anton-vinogradov: A raw Collection field has no type arguments, and assert is off in a normal build while its Error escapes catch (Exception) in MessageProcessor. Please replace both asserts (here and at line 1199) with a size check and report the failure through the Messager.

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
```diff
@@ -1170,6 +1172,41 @@ private enum MarshalledKind {
...
+
+            msgToBytes = assignableFrom(typeArgs.get(0), msgType);
```
anton-vinogradov: Only one generic level is inspected, so Collection<List<Message>> and Map<K, List<Message>> still reach the marshaller silently. Please walk the type recursively, the way needsCtxType() above does.

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
```diff
@@ -1170,6 +1172,41 @@ private enum MarshalledKind {
...
+        }
+        else if (field.asType() instanceof ArrayType) {
```
anton-vinogradov: Line 1166 checks the same thing as getKind() == TypeKind.ARRAY. Please use that form here too.

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
```diff
@@ -1170,6 +1172,41 @@ private enum MarshalledKind {
...
+
+            List<? extends TypeMirror> typeArgs = type.getTypeArguments();
+
+            assert typeArgs.size() == 2;
+
+            msgToBytes = assignableFrom(typeArgs.get(0), msgType) || assignableFrom(typeArgs.get(1), msgType);
+        }
+
+        if (msgToBytes) {
+            throw new IllegalArgumentException("Field \"" + field + "\" is Message. Must be written by communication protocol" +
```
anton-vinogradov: Throwing attaches the diagnostic to the class instead of the field, adds a misleading "Failed to generate" prefix, and repeats per subclass for fields from an abstract parent. Please report via env.getMessager().printMessage(ERROR, msg, field), as this method already does six lines above.

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
```diff
@@ -1170,6 +1172,41 @@ private enum MarshalledKind {
...
+                ". Remove @" + Marshalled.class.getSimpleName() + " annotation and remove corresponding byte[] field.");
```
anton-vinogradov: "Field X is Message" is inaccurate for collection, array and map fields, and the companion is Collection<byte[]> for ELEMENT_BLOBS. Please reword to cover the container cases and drop the byte[] wording.

modules/core/src/test/java/org/apache/ignite/internal/codegen/MessageProcessorTest.java
```diff
@@ -683,6 +683,46 @@ public void testNonMarshallableSelfMarshallingFailed() {
             "or SelfMarshallingMessage, nor declare @Marshalled fields");
     }
 
+    /** Test that {@code @Marshalled} annotation on {@link Message} field will fail generation. */
+    @Test
+    public void testMarshallableOnMessageFieldFailGeneration() {
```
anton-vinogradov: "Marshallable" collides with MarshallableMessage, already covered by testMarshallableMessage in this file. Please rename all four to testMarshalledOn..., after the @Marshalled annotation.

modules/core/src/test/java/org/apache/ignite/internal/codegen/MessageProcessorTest.java
```diff
@@ -683,6 +683,46 @@ public void testNonMarshallableSelfMarshallingFailed() {
...
+    @Test
+    public void testMarshallableOnMessageMapFieldFailGeneration() {
+        Compilation compilation = compile("InternalMessage.java", "InternalMessageMapContainer.java");
+
+        assertThat(compilation).failed();
+
+        assertThat(compilation).hadErrorContaining("Field \"msgMap\" is Message. Must be written by communication protocol");
+    }
+
+    /** Test that {@code @Marshalled} annotation on {@link Message} map field will fail generation. */
```
anton-vinogradov: The javadoc says "map field" while the test covers the array case. Please fix the copy-paste.

anton-vinogradov (commented):
Thanks for picking this up, the gap is real.

One blocker: the check is applied before `MarshalledKind` is computed, so it keys off the field type while the rule actually depends on the companion wire field. `ELEMENTS` (an `X[]` companion) and `MAP` are written by the communication protocol, not by a marshaller, so `Message` elements there are correct and must stay allowed. As written, `Set<SomeMessage>` with a `SomeMessage[]` companion is rejected even though generation emits `writeObjectArray(..., MessageCollectionItemType.MSG)` and produces no marshaller blob at all. The current tests do not catch it: `testMarshalledCollectionMessage` uses `GridCacheVersion`, which is not a `Message`, and the symmetric map case is allowed only because map form is excluded by hand.

Computing the kind first and checking only `BLOB` and `ELEMENT_BLOBS` fixes the false positive, lets the same walk cover nested generics, and keeps all 49 `MessageProcessorTest` tests green including the four new ones (verified locally, the error text can stay as is).

The rest are smaller notes inline.

modules/core/src/test/java/org/apache/ignite/internal/codegen/MessageProcessorTest.java
```diff
@@ -683,6 +683,46 @@ public void testNonMarshallableSelfMarshallingFailed() {
...
+    @Test
+    public void testMarshallableOnMessageArrayFieldFailGeneration() {
+        Compilation compilation = compile("InternalMessage.java", "InternalMessageArrayContainer.java");
+
+        assertThat(compilation).failed();
+
+        assertThat(compilation).hadErrorContaining("Field \"msgArr\" is Message. Must be written by communication protocol");
+    }
```
anton-vinogradov, line 714: ELEMENT_BLOBS is the second rejected kind and has no test. Please add one for Collection<Message> with a Collection<byte[]> companion.

modules/core/src/test/resources/codegen/InternalMessage.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal;
+
+import org.apache.ignite.plugin.extensions.communication.Message;
+
+/** */
+public class InternalMessage implements Message {
```
anton-vinogradov: Resources in this folder are named Test*Message / Wrong*Message, and these read like production classes. Please rename to match.

modules/core/src/test/resources/codegen/InternalMessageContainer.java
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.ignite.plugin.extensions.communication.Message;
+
+/** */
+public class InternalMessageContainer implements Message {
+    /** */
+    @Marshalled("msgBytes")
+    InternalMessage msg;
+
+    /** */
+    @Order(value = 0)
```
anton-vinogradov: Three resources use @Order(value = 0) while InternalMessageMapContainer in this same PR uses @Order(0). Please use the short form everywhere.

modules/core/src/test/resources/codegen/InternalMessageArrayContainer.java
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal;
+
+import java.util.Collection;
```
anton-vinogradov, line 20: Unused import. Please remove it.

modules/core/src/test/resources/codegen/InternalMessageMapContainer.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal;
+
+import java.util.Map;
```
anton-vinogradov, line 20: Collection is unused and the two imports are out of alphabetical order. Please remove it and swap the remaining ones.

modules/core/src/test/resources/codegen/MarshalledOnMessageArray2.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal;
+
+import java.util.Collection;
+import org.apache.ignite.plugin.extensions.communication.Message;
+
+/** */
+public class MarshalledOnMessageArray2 implements Message {
```
Vladsz83: Suggestion: `MarshalledOnMessageCollectionArray`

modules/core/src/test/resources/codegen/IncorrectMarshalledOnMessageCollection.java
```diff
@@ -0,0 +1,32 @@
...
+package org.apache.ignite.internal;
+
+import java.util.Collection;
+import org.apache.ignite.plugin.extensions.communication.Message;
+
+/** */
+public class MarshalledOnMessageCollection implements Message {
+    /** */
+    @Marshalled("msgBytes")
+    Collection<TestMessage> msgColl;
```
Vladsz83, line 27: Do we need to check also `List` and `List` in the same way?

modules/core/src/test/resources/codegen/IncorrectMarshalledOnMessageMap.java
```diff
@@ -0,0 +1,32 @@
...
+
+package org.apache.ignite.internal;
+
+import org.apache.ignite.plugin.extensions.communication.Message;
+import java.util.Map;
+
+/** */
+public class MarshalledOnMessageMap implements Message {
+    /** */
+    @Marshalled("bytes")
```
Vladsz83, line 26: Do we need to check also `@Marshalled` with `keys()` and `values()`?

modules/core/src/test/java/org/apache/ignite/internal/codegen/MessageProcessorTest.java
```diff
@@ -683,6 +684,43 @@ public void testNonMarshallableSelfMarshallingFailed() {
...
+            "MarshalledOnMessageMap3.java",
+            "MarshalledOnMessageArray.java",
+            "MarshalledOnMessageArray2.java"
+        );
+
+        for (String file : cases) {
+            Compilation compilation = compile("TestMessage.java", file);
+
+            assertThat(compilation).failed();
+            assertThat(compilation).hadErrorContaining("Message must be written by communication protocol");
```
Vladsz83: `by communication protocol` is a bit unclear. I suggest smth. like `by dedicated message serializers`

modules/core/src/test/resources/codegen/MarshalledOnMessageSet.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal;
+
+import java.util.Set;
+import org.apache.ignite.plugin.extensions.communication.Message;
+
+/** */
+public class MarshalledOnMessageSet implements Message {
```
Vladsz83: Looks wierd. Why should we allow `@Marshalled` not on `byte[]` at all?

modules/core/src/test/java/org/apache/ignite/internal/codegen/MessageProcessorTest.java
```diff
@@ -683,6 +684,43 @@ public void testNonMarshallableSelfMarshallingFailed() {
...
+    /** Test that {@code @Marshalled} annotation on {@link Message} field will fail generation. */
+    @Test
+    public void testMarshalledOnMessageSet() {
+        assertThat(compile("TestMessage.java", "MarshalledOnMessageSet.java")).succeeded();
+    }
+
+    /** Test that {@code @Marshalled} annotation on {@link Message} field will fail generation. */
+    @Test
+    public void testMarshalledOnSetToArraySucceed() {
+        Compilation compilation = compile("TestMessage.java", "MarshalledOnMessageSet.java");
```
Vladsz83: The same. if `will fail generation.` why `.succeeded()`?

Vladsz83: Why not a single line btw like in the previous test?

modules/core/src/test/java/org/apache/ignite/internal/codegen/MessageProcessorTest.java
```diff
@@ -683,6 +684,43 @@ public void testNonMarshallableSelfMarshallingFailed() {
...
+
+    /** Test that {@code @Marshalled} annotation on {@link Message} field will fail generation. */
+    @Test
+    public void testMarshalledOnMessageSet() {
+        assertThat(compile("TestMessage.java", "MarshalledOnMessageSet.java")).succeeded();
+    }
+
+    /** Test that {@code @Marshalled} annotation on {@link Message} field will fail generation. */
+    @Test
+    public void testMarshalledOnSetToArraySucceed() {
```
Vladsz83: How differs from `testMarshalledOnMessageSet()`?

modules/core/src/test/resources/codegen/IncorrectMarshalledOnMessageMap.java
nizhikov: Checked in `TestMarshalledMapMessage`

modules/core/src/test/resources/codegen/IncorrectMarshalledOnMessageCollection.java
nizhikov: Generator works for all Collections, don't think we want to check all of the `Collection` implementations in tests.

modules/core/src/test/resources/codegen/MarshalledOnMessageSet.java
nizhikov: I have the same question.

LLM review posted by @anton-vinogradov stays that we should support this.

@anton-vinogradov lets' remove possibility to have `Set<Message>` -> `Message[]` transformation during serialization. Moreover, there are no such cases in codebase for now.

Vladsz83: I believe we should not consider this feature in this ticket.

modules/core/src/test/resources/codegen/IncorrectMarshalledOnMessageMap.java
Vladsz83: There are `Collection<GridCacheVersion>` and `Collection<GridCacheVersion> mapVals;` Expected: `byte[]` arrays

modules/core/src/test/java/org/apache/ignite/internal/codegen/MessageProcessorTest.java
```diff
@@ -683,6 +685,57 @@ public void testNonMarshallableSelfMarshallingFailed() {
...
+    /** Test that {@code @Marshalled} annotation on raw {@link Collection} fail generation. */
+    @Test
+    public void testRawCollectionFailGeneration() {
+        List<String> cases = Arrays.asList(
+            "TestRawListMessage.java",
+            "TestRawCollectionMessage.java"
+        );
+
+        for (String file : cases) {
+            Compilation compilation = compile("TestMessage.java", file);
```
Vladsz83, line 722: Do we need "TestMessage.java"?

modules/core/src/test/java/org/apache/ignite/internal/codegen/MessageProcessorTest.java
```diff
@@ -683,6 +685,57 @@ public void testNonMarshallableSelfMarshallingFailed() {
...
+
+            assertThat(compilation).failed();
+            assertThat(compilation).hadErrorContaining("Raw collection not supported");
+        }
+    }
+
+    /** Test that {@code @Marshalled} annotation on raw {@link Map} fail generation. */
+    @Test
+    public void testRawMapFailGeneration() {
```
Vladsz83: Join with `testRawCollectionFailGeneration`? + The same, do we need "TestMessage.java"

modules/core/src/test/java/org/apache/ignite/internal/codegen/MessageProcessorTest.java
```diff
@@ -683,6 +685,57 @@ public void testNonMarshallableSelfMarshallingFailed() {
...
+            "MarshalledOnMessageMap3.java",
+            "MarshalledOnMessageArray.java",
+            "MarshalledOnMessageCollectionArray.java",
+            "MarshalledOnMessageSet.java",
+            "MarshalledOnMessageList.java",
+            "MarshalledOnMessageList2.java"
+        );
+
+        for (String file : cases) {
+            Compilation compilation = compile("TestMessage.java", file);
```
Vladsz83, line 709: Same. Let's ensure we need "TestMessage.java". here.

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
```diff
@@ -1156,18 +1156,78 @@ private enum MarshalledKind {
             return null;
         }
 
+        MarshalledKind res;
+
         if (map)
-            return MarshalledKind.MAP;
+            res = MarshalledKind.MAP;
+        else {
+            TypeMirror wire = requireEnclosed(enclosed, ann.value(), "@Marshalled").asType();
```
Vladsz83, line 1165: Just a notise. Annotation here look as something important. Important to search. But is onle for the error.

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
```diff
@@ -1156,18 +1156,78 @@ private enum MarshalledKind {
...
+        }
+
+        return res;
+    }
+
+    /**
+     * Recursively checks no {@code Message} -> {@code byte[]} transformation.
+     * @return {@code True} in case error transformation found.
+     */
+    private boolean foundMessageToBytesTransformation(TypeMirror type, VariableElement field, Marshalled ann) {
```
Vladsz83: Suggestion: Not `found...`. `search...`, `find...`, `ceck...` or even simplier `messageToBytesTransformation()`

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
```diff
@@ -1156,18 +1156,78 @@ private enum MarshalledKind {
...
+        if (assignableFrom(type, msgType))
+            return true;
+        else if (isCollection(type)) {
```
Vladsz83: Nexy line and not `else`

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
```diff
@@ -1156,18 +1156,78 @@ private enum MarshalledKind {
...
-                : MarshalledKind.ELEMENTS;
+            if (typeArgs.size() != 1) {
+                env.getMessager().printMessage(Diagnostic.Kind.ERROR, "Raw collection not supported.", field);
+
+                return false;
+            }
+
+            return foundMessageToBytesTransformation(typeArgs.get(0), field, ann);
         }
+        else if (type.getKind() == TypeKind.ARRAY)
```
Vladsz83: Same. No need of else`s.

modules/core/src/test/resources/codegen/MarshalledOnMessageSet.java
anton-vinogradov: Let's add such check and skip?

modules/core/src/test/java/org/apache/ignite/internal/codegen/MessageProcessorTest.java
nizhikov: yes, it used in test classes as dependency

modules/core/src/test/resources/codegen/MarshalledOnMessageArray.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal;
+
+import org.apache.ignite.plugin.extensions.communication.Message;
+
+/** */
+public class MarshalledOnMessageArray implements Message {
```
anton-vinogradov: Let's call such messages as IncorrectXXX

modules/core/src/test/resources/codegen/MarshalledOnMessageSet.java
nizhikov: Done.

modules/core/src/test/java/org/apache/ignite/internal/codegen/MessageProcessorTest.java
nizhikov: We need it because it used inside test classes in field type

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
nizhikov: Renamed

modules/core/src/test/resources/codegen/IncorrectMarshalledOnMessageCollection2.java
```diff
@@ -0,0 +1,32 @@
...
+package org.apache.ignite.internal;
+
+import java.util.Collection;
+import org.apache.ignite.plugin.extensions.communication.Message;
+
+/** */
+public class MarshalledOnMessageCollection2 implements Message {
+    /** */
+    @Marshalled("msgBytes")
+    Collection<TestMessage> msgColl;
```
anton-vinogradov, line 27: Should we check wildcards as well, like Collection<? extends TestMessage>?

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
```diff
@@ -1156,18 +1156,78 @@ private enum MarshalledKind {
...
+        }
+
+        /*
+         * Ensures that field annotated with {@link Marshalled} doesn't perform {@code Message} -> {@code byte[]} transformation
+         * which escapes {@link Order} and other rules implemented on top of communication {@code MessageWriter, MessageReader} logic.
+         */
+        if (foundMessageToBytesTransformation(field.asType(), field, ann)) {
+            env.getMessager().printMessage(Diagnostic.Kind.ERROR,
+                "Message must be written by dedicated message serializers. " +
+                    "Remove @" + Marshalled.class.getSimpleName() + " annotation and remove companion field.", field);
```
anton-vinogradov: And set an @Order annotation?

Vladsz83 (approved):

anton-vinogradov (changes requested):

modules/core/src/test/resources/codegen/MarshalledOnMessageArray.java
nizhikov: Done

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
nizhikov: Done.

modules/core/src/test/resources/codegen/IncorrectMarshalledOnMessageCollection2.java
nizhikov: I don't think we ever want to support any of WildcardType in communication, anyway, I added checks to patch.

anton-vinogradov (approved):

PR #13462, opened 2026-08-11
IGNITE-28940 Choose the marshaller by transport, not by message class

anton-vinogradov:
The marshaller used for a message's `@Marshalled` fields was a property of the
message class: `@UseBinaryMarshaller` decided it, and the factory bound the
chosen marshaller into the generated companion at registration time.

### Why that is the wrong place

Binary cannot be used where marshalling cannot afford a cluster-wide class
registration, and that is a property of the call site. The registration is
`MarshallerContextImpl#registerClassName` -> `proposeMapping` -> `fut.get()`,
which waits for discovery, so it must never happen on a discovery thread. Ignite
classes usually skip it thanks to `META-INF/classnames.properties`, but that list
is not closed under nesting - a plain `CacheConfiguration` needs types that are
not on it.

The call sites are already separated by transport, and the hand-written half of
the API works this way: `MarshallableMessage#marshal(Marshaller)` takes the
marshaller as a parameter. Only generated code kept it in a field.

### Change

* `Marshaller` is now a parameter of `MessageMarshaller` and `MessageMarshalling`;
  the generator stops storing it and passes it down to nested messages.
* Communication call sites pass `ctx.marshaller()`, discovery call sites pass
  `marshallerContext().jdkMarshaller()`.
* `@UseBinaryMarshaller` (60 classes), `AbstractMessageFactoryProvider#init` and
  both marshaller fields are gone, as is `initProvider` in `IgniteKernal`.

### Wire format

The format is now a function of the transport, not of the class, so a class that
travels both transports is marshalled differently on each. Deliberate changes:

| class | change |
|---|---|
| `ErrorMessage`, `PartitionHashRecord`, `TransactionsHashRecord` | jdk -> binary, on the communication leg |
| `BinaryMetadataVersionInfo` | binary -> jdk, on the discovery leg |
| `StoredCacheData` | jdk -> binary, on the communication leg (snapshot restore) |

`StoredCacheData` is not in the ticket's list: it was found while reviewing this
change. It carries `CacheConfiguration` and `QueryEntity` blobs and reaches
communication through `SnapshotRestoreOperationResponse` -> `SingleNodeMessage`,
so the transport rule moves it to binary. Registration is allowed there, and
binary is the smaller of the two for a `CacheConfiguration`.

`BinaryMetadataVersionInfo` needed one more fix: the same long-lived instance
from the local metadata cache travels discovery in the data bag and communication
in `MetadataResponseMessage`, and a marshalled instance keeps its serialized
form. `MetadataResponseMessage#metadataVersionInfo` now stores a copy, so bytes
of one transport cannot leak into the other.

### Verified

`MessageProcessorTest` (codegen goldens regenerated), `IgniteCoreMessagesSerializationTest`,
`DirectMarshallingMessagesTest`, `MessageMarshalOnceTest`, `CompressedMessageTest`,
`GridIoManagerOrderedUnmarshalFailureTest`, `DiscoveryUnmarshalVulnerabilityTest`,
`QueryEntityMessageSerializationTest`, `LazyServiceConfigurationMessageSerializationTest`,
`SecurityBasicPermissionSetSerializationTest`, `TxDeadlockDetectionMessageMarshallingTest`,
`GridCacheQueryResponseUnmarshalTest`, `BinaryMetadataRegistrationInsideEntryProcessorTest`,
`GridCacheAtomicFullApiSelfTest`, the continuous-query and p2p suites, and the
calcite `PlanExecutionTest` / `ContinuousExecutionTest` /
`CalciteCommunicationMessageSerializationTest`.

Full build of all modules and the strict checkstyle profile are clean.


anton-vinogradov:
/runall


---
🚀 **RunAll queued** — [build 9275680](https://ci2.ignite.apache.org/build/9275680) · live progress & verdict: [Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13462). The verdict lands here when the run finishes.
🛑 _Superseded by a newer /run-all._

anton-vinogradov:
@anton-vinogradov 🚀 **RunAll queued** — [build 9275976](https://ci2.ignite.apache.org/build/9275976) · live progress & verdict: [Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13462). Your previous run was cancelled — this one supersedes it. The verdict lands here when the run finishes.
🏁 _Run finished._ ♻️ _Auto re-run **#3** — 1 suite that failed mid-run, **≈ settled by 05:57 MSK** — details in the verdict comment._

anton-vinogradov:
@anton-vinogradov that looks like an [Ignite PR Checker](https://ignite-pr-checker.is-a.dev) command — but the checker doesn't know your accounts yet, so **nothing was triggered**. Everything it does runs under your own accounts (there is no bot); setting that up takes about two minutes:

1. **Log in** at https://ignite-pr-checker.is-a.dev with a TeamCity ([ci2](https://ci2.ignite.apache.org)) access token — create one at [ci2 → Profile → Access Tokens](https://ci2.ignite.apache.org/profile.html?item=accessTokens).
2. In settings (⚙) switch on at least one option — **Auto re-run blocker suites** needs nothing extra — and save your **GitHub login** in the PR-commands field. That's enough: commands work, the checker acks and narrates from its own account.
3. **The full experience** — switch on **Comment my runs' verdicts on the GitHub PR** with a GitHub personal access token ([create one here](https://github.com/settings/tokens/new?scopes=public_repo&description=Ignite+PR+Checker), classic, `public_repo` scope): acks and the live run status then come from your own account, plus checkstyle autofix becomes available. **Auto-visa all my runs** posts the verdict to the IGNITE ticket (needs a [JIRA PAT](https://issues.apache.org/jira/secure/ViewProfile.jspa?selectedTab=com.atlassian.pats.pats-plugin:jira-user-personal-access-tokens)).

Then comment here:

- `/run-all` — queue the whole RunAll chain under your TeamCity account (`/run-all top` — at the top of the build queue);
- `/top` — move the run your command started to the top of the queue while it still waits.

Your command comment gets a 🚀 and narrates the run — live ETA, finish, auto re-run waves — and the verdict lands as one comment that updates in place until everything settles. Tokens are stored encrypted, and only while the options are on.

anton-vinogradov:
/runall

---
🚀 **RunAll queued** — [build 9278217](https://ci2.ignite.apache.org/build/9278217) · live progress & verdict: [Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13462). Your previous run was cancelled — this one supersedes it. The verdict lands here when the run finishes.
🏁 _Run finished — analysing; the verdict comment follows._

modules/codegen/src/main/java/org/apache/ignite/internal/JdkMarshalled.java
```diff
@@ -14,25 +14,22 @@
...
+ * Marshals the message with the JDK marshaller of the local node, whatever the transport speaks. Put it on a message
+ * that travels both transports: a {@link Marshalled} field caches its wire form, and only the JDK marshaller is read
+ * by both. It also asks no cluster-wide type registration, which never completes on a discovery thread.
+ *
+ * @see Marshalled
  */
-@Documented
-@Target(value = TYPE)
-@Retention(RUNTIME)
-public @interface UseBinaryMarshaller {
```
Vladsz83, line 36: Why binary changed on opposite JDK?

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/CommunicationMarshalling.java
```diff
@@ -0,0 +1,49 @@
...
+
+package org.apache.ignite.internal.managers.communication;
+
+import org.apache.ignite.IgniteCheckedException;
+import org.apache.ignite.internal.GridKernalContext;
+import org.apache.ignite.internal.processors.cache.CacheObjectContext;
+import org.apache.ignite.plugin.extensions.communication.Message;
+import org.jetbrains.annotations.Nullable;
+
+/** Marshalling of the communication transport, which speaks the schema-aware marshaller. */
```
Vladsz83: Let's reclarify `speaks`

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/DiscoveryMarshalling.java
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.ignite.internal.GridKernalContext;
+import org.apache.ignite.internal.processors.cache.CacheObjectContext;
+import org.apache.ignite.plugin.extensions.communication.Message;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Marshalling of the discovery transport, which speaks the JDK marshaller: a discovery message is marshalled on a
+ * discovery thread, where waiting for a cluster-wide type registration would never finish.
+ */
+public final class DiscoveryMarshalling {
```
Vladsz83, line 35: Why it is in `communication` package? Also package `managers` might be revised

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/DiscoveryMarshalling.java
```diff
@@ -0,0 +1,47 @@
...
+package org.apache.ignite.internal.managers.communication;
+
+import org.apache.ignite.IgniteCheckedException;
+import org.apache.ignite.internal.GridKernalContext;
+import org.apache.ignite.internal.processors.cache.CacheObjectContext;
+import org.apache.ignite.plugin.extensions.communication.Message;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Marshalling of the discovery transport, which speaks the JDK marshaller: a discovery message is marshalled on a
```
Vladsz83: Let's rephrase `speaks`

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/CommunicationMarshalling.java
```diff
@@ -0,0 +1,49 @@
...
+public final class CommunicationMarshalling {
```
Vladsz83, line 31: Only minority of the messages requires BinaryMarshaller. Why we use it for all of them? It is able to marshal non-serializable and won't fail where it should. Is it faster than `JDKMarshaller`? Any proofs? It processes schema. Won't we get any perf. drop on marshalling with it?

Vladsz83, line 31: Maybe package `managers` is not the best place for it.

modules/core/src/main/java/org/apache/ignite/internal/managers/eventstorage/GridEventStorageManager.java
```diff
@@ -1242,7 +1242,7 @@ private class RequestListener implements GridMessageListener {
                 Collection<Event> evts;
 
                 try {
-                    MessageMarshalling.unmarshal(req, ctx, null,
+                    CommunicationMarshalling.unmarshal(req, ctx, null,
```
Vladsz83, line 1245: My idea was to bind thread an marshaller, not places where called. WDYT?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -3821,7 +3821,7 @@ else if (forceAffReassignment)
 
             // Marshal eagerly: the heavy partition-map copy lands in the "Full message preparing" stage, and the
             // message cached in FinishState is sent to late joiners as is (the send-path marshal-once turns no-op).
-            MessageMarshalling.marshal(msg, cctx.kernalContext(), null);
+            CommunicationMarshalling.marshal(msg, cctx.kernalContext(), null);
```
Vladsz83, line 3824: Is it so obvious so it Communication-related? Why would I suppose it? Why not discovery? If Communication proccesses the marshalling, lets let it decide and choose.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryHandler.java
```diff
@@ -1072,7 +1072,7 @@ private void notifyCallback0(UUID nodeId,
                     }
                 }
 
-                MessageMarshalling.unmarshal(e, ctx, cctx.cacheObjectContext(), ldr);
+                CommunicationMarshalling.unmarshal(e, ctx, cctx.cacheObjectContext(), ldr);
```
Vladsz83, line 1094: The same. Why Communication marshaling? How do I know it? To me, `MessageMarshalling` look much better. It will decide what and how to do. Now I need to decide. Let's reconsider.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
```diff
@@ -3507,7 +3507,7 @@ private void unmarshall(UUID nodeId, GridCacheMessage cacheMsg) {
                 return;
 
             try {
-                MessageMarshalling.unmarshal(cacheMsg, cctx.kernalContext(), null, cctx.deploy().globalLoader());
+                CommunicationMarshalling.unmarshal(cacheMsg, cctx.kernalContext(), null, cctx.deploy().globalLoader());
```
Vladsz83, line 3510: Same. Why `CommunicationMarshalling` ? Why if someone puts `DiscoveryMarshalling`? Is it deprecated, protected?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/StoredCacheData.java
```diff
@@ -43,6 +44,8 @@
  * <p>
  * All changes must be made with the respect of RU rules.
  */
+// Travels both transports: by discovery when a cache starts, by communication when a snapshot is restored.
+@JdkMarshalled
```
Vladsz83, line 49: Even so, why not marshall with the marshaller choosen? Why discovery ort Communication should change marshaller?

modules/core/src/main/java/org/apache/ignite/internal/processors/datastreamer/DataStreamProcessor.java
```diff
@@ -230,7 +230,7 @@ private void processRequest(final UUID nodeId, final DataStreamerRequest req) {
             StreamReceiver<?, ?> updater;
 
             try {
-                MessageMarshalling.unmarshal(req, ctx, null, U.resolveClassLoader(clsLdr, ctx.config()));
+                CommunicationMarshalling.unmarshal(req, ctx, null, U.resolveClassLoader(clsLdr, ctx.config()));
```
Vladsz83, line 233: Same. `MessageMarshalling` looks better. Noone cares which marshalling to call. Just pick up `MessageMarshalling`. Now we need to know.

modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskWorker.java
```diff
@@ -824,7 +824,7 @@ void onResponse(GridJobExecuteResponse msg) {
                         boolean loc = ctx.localNodeId().equals(res.nodeId()) && !ctx.config().isMarshalLocalJobs();
 
                         if (!loc)
-                            MessageMarshalling.unmarshal(res, ctx, null, U.resolveClassLoader(dep.classLoader(), ctx.config()));
```
Vladsz83, line 827: All the others places too. `MessageMarshalling` looks better. Gets the work of marsahaling type choise incapsulated. Not, exposed to a developer.

Vladsz83 (changes requested):

modules/codegen/src/main/java/org/apache/ignite/internal/JdkMarshalled.java
anton-vinogradov: Now Discovery uses jdk, Communication - binary, with some exceptions.

modules/core/src/main/java/org/apache/ignite/internal/managers/eventstorage/GridEventStorageManager.java
anton-vinogradov: Interesting idea, maybe next time?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
anton-vinogradov: Let's refactor these issues later.
Currently, we should make the smallest possible change to close the wire-format gap before release.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryHandler.java
anton-vinogradov: Let's refactor naming, linking to the thread in a separate issue.
This one has another goal.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
anton-vinogradov: Tests will fail :)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/StoredCacheData.java
anton-vinogradov: They are not changing marshallers, they keep it as is with minor fixes, which was luckily passed before.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/CommunicationMarshalling.java
anton-vinogradov: Reworded: "It uses the schema-aware marshaller."

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/DiscoveryMarshalling.java
anton-vinogradov: Reworded the same way: "It uses the JDK marshaller, because a discovery message is marshalled on a discovery thread, where waiting for a cluster-wide type registration would never finish."

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/DiscoveryMarshalling.java
anton-vinogradov: Agree, the package does not fit them. Moving these two means moving MessageMarshalling as well, since they sit next to it and it comes from master, so I would rather do all three in a separate issue than widen this one.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/CommunicationMarshalling.java
anton-vinogradov: Same here, see the answer in DiscoveryMarshalling.

anton-vinogradov: Measured it. Binary is faster on every payload a message field actually carries, except exceptions, where the two are equal.

Round trip (marshal + unmarshal), corretto-17, 2000 warmup + 20000 iterations, three runs, numbers from the last one:

| payload | jdk, us/op | binary, us/op | speedup | jdk, bytes | binary, bytes |
|---|---|---|---|---|---|
| small POJO | 6.21 | 3.43 | 1.8x | 121 | 40 |
| list of 100 POJOs | 57.72 | 22.24 | 2.6x | 2048 | 4196 |
| map of 50 String to UUID | 14.95 | 5.38 | 2.8x | 1626 | 1346 |
| exception with a cause | 20.59 | 24.79 | 0.83x | 2138 | 2144 |
| CacheConfiguration | 27.60 | 7.60 | 3.6x | 3432 | 636 |
| QueryEntity | 7.74 | 1.48 | 5.2x | 783 | 161 |

Two results are worth reading carefully.

**Exceptions are a tie, and that is expected.** `Throwable` declares `writeObject`, so `BinaryUtils.isCustomJavaSerialization` sends it to `OptimizedMarshaller` - the binary format never runs. That is also why `ErrorMessage` pins jdk in this PR: binary buys nothing there and would only register the class name of every new exception type cluster-wide.

**Binary is bigger on a list of equal objects.** It writes the schema per object, while jdk writes the class descriptor once and back-references it. On everything else binary is smaller, up to 5x on `CacheConfiguration`.

The schema processing you asked about is what pays for itself here: it is written per object, but reading does not walk the whole graph.

The benchmark is a throwaway test class, not part of the PR. It builds both marshallers the way tests do - `Marshallers.jdk()` and `createStandaloneBinaryMarshaller()` - and times `U.unmarshal(marsh, U.marshal(marsh, obj), loader)` in a loop. I can attach it if you want to re-run it.

anton-vinogradov:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13462)** verdict · RunAll build [9278815](https://ci2.ignite.apache.org/build/9278815) · 147 suites ran, 0 reused

- 1 commit(s) pushed since this run — it tested older code

🔎 **No blockers found — but the run above can't prove the PR is clean.** 22 pre-existing/flaky tests filtered out. Re-run once the above is sorted out.

_♻️ Settled after 2 auto re-run wave(s): #1 — early: Control Utility 1; #2 — 1 broken suite(s)._

anton-vinogradov:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13462)** verdict · RunAll build [9279430](https://ci2.ignite.apache.org/build/9279430) · 147 suites ran, 0 reused

✅ **No blockers** — nothing in this run looks caused by this PR. 20 pre-existing/flaky tests filtered out.

_♻️ Settled after 1 auto re-run wave(s): #1 — 2 watch suite(s)._

modules/core/src/test/java/org/apache/ignite/spi/MessagesPluginProvider.java
```diff
@@ -43,13 +46,23 @@ public MessagesPluginProvider(Class<? extends Message>... msgs) {
...
             }
         };
     }
 
+    /** @return Generated marshaller of the message, or {@code null} when the message has nothing to marshal. */
+    private static <T extends Message> @Nullable MessageMarshaller<T> marshaller(Class<? extends Message> msg) {
+        try {
+            return loadMarshaller(msg);
+        }
+        catch (RuntimeException ignored) {
```
Vladsz83, line 66: Looks too tough. Maybe we need a comment here?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
Vladsz83: Refactoring this issue might set us to revert or rewrite this ticket again I'm afraid. We split marshalling type. This is the nase of this ticket

modules/core/src/test/java/org/apache/ignite/internal/util/distributed/DistributedProcessResultMarshallingTest.java
```diff
@@ -0,0 +1,236 @@
...
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static org.apache.ignite.internal.util.distributed.DistributedProcess.DistributedProcessType.TEST_PROCESS;
+import static org.apache.ignite.testframework.GridTestUtils.loadSerializer;
+
+/**
+ * Tests that the result of a {@link DistributedProcess} is marshalled with the JDK marshaller on both transports: it
+ * goes to the coordinator by communication and comes back in the {@link FullMessage} by discovery, while a marshalled
+ * field caches its wire form for the second leg.
+ */
+public class DistributedProcessResultMarshallingTest extends GridCommonAbstractTest {
```
Vladsz83, line 70: One Test for one `@Test`? Can we put it in some existing Test?

modules/core/src/main/java/org/apache/ignite/internal/plugin/AbstractMessageFactoryProvider.java
```diff
@@ -87,28 +58,27 @@ private static <T extends Message> void register(IgniteMessageFactory factory, C
...
-     * companion ever takes a {@code Marshaller}, and only when the message has fields to marshal with one, so
-     * {@code marsh} is {@code null} for the other two. Constructor lookups, including missing companions, are cached
-     * per message class in {@link #COMPANIONS}.
+     * Instantiates the generated companion class {@code <message>Serializer/Marshaller/Deployer}. Companions are
+     * stateless: the marshaller, when one is needed, is passed per call by the transport. Constructor lookups,
+     * including missing companions, are cached per message class in {@link #COMPANIONS}.
      *
      * @return the companion, or {@code null} when it is not generated and {@code required} is {@code false}.
      */
     @SuppressWarnings("unchecked")
```
shishkovilja, line 80:
```suggestion
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/BinaryMetadataVersionInfo.java
```diff
@@ -29,7 +29,8 @@
  * The version refers solely to the internal protocol for updating BinaryMetadata and is unknown externally.
  * It can be updated dynamically from different nodes and threads on the same node.
  */
-@UseBinaryMarshaller
+// Travels both transports: by discovery in the data bag, by communication in the MetadataResponseMessage.
```
shishkovilja: Comment should be placed into the Javadoc.

shishkovilja (changes requested):
Compilation currently fails:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.15.0:compile (default-compile) on project ignite-core: Compilation failure: Compilation failure: 
[ERROR] /home/shish/IdeaProjects/ignite/modules/core/src/main/java/org/apache/ignite/internal/processors/continuous/ContinousRoutineLocalInfo.java:[24,34] cannot find symbol
[ERROR]   symbol:   class UseBinaryMarshaller
[ERROR]   location: package org.apache.ignite.internal
[ERROR] /home/shish/IdeaProjects/ignite/modules/core/src/main/java/org/apache/ignite/internal/processors/continuous/ContinousRoutineLocalInfo.java:[31,2] cannot find symbol
[ERROR]   symbol: class UseBinaryMarshaller
```

modules/core/src/test/java/org/apache/ignite/internal/util/distributed/DistributedProcessResultMarshallingTest.java
```diff
@@ -0,0 +1,236 @@
...
+            return new PayloadMessage();
+        }
+    }
+
+    /** Marshals the payload with whatever marshaller the caller hands over - the test asserts which one that is. */
+    private static class PayloadMarshaller implements MessageMarshaller<PayloadMessage> {
+        /** {@inheritDoc} */
+        @Override public void marshal(PayloadMessage msg, Marshaller marsh, GridKernalContext kctx,
+            CacheObjectContext cacheObjCtx) throws IgniteCheckedException {
+            if (msg.val != null && msg.valBytes == null)
```
shishkovilja, line 303: We can assert type of `Marshaller` here. It seems more clear, than checking magic bytes.

modules/core/src/test/java/org/apache/ignite/internal/util/distributed/DistributedProcessResultMarshallingTest.java
```diff
@@ -0,0 +1,236 @@
...
+                msg.valBytes = U.marshal(marsh, msg.val);
+        }
+
+        /** {@inheritDoc} */
+        @Override public void unmarshal(PayloadMessage msg, Marshaller marsh, GridKernalContext kctx,
+            CacheObjectContext cacheObjCtx, ClassLoader clsLdr) throws IgniteCheckedException {
+            if (msg.valBytes != null) {
```
shishkovilja, line 316: We can assert type of `Marshaller` here. It seems more clear, than checking magic bytes.

modules/core/src/main/java/org/apache/ignite/internal/util/ErrorMessage.java
```diff
@@ -31,6 +32,8 @@
  */
 // TODO IGNITE-28912: move to a common package.
 @SuppressWarnings({"NullableProblems", "unused"})
+// Travels both transports, and an error is an arbitrary user class the cluster may not know yet.
```
shishkovilja: Comment should be moved into the Javadoc.

modules/core/src/test/java/org/apache/ignite/internal/util/distributed/DistributedProcessResultMarshallingTest.java
anton-vinogradov: Good point, done: the test now captures the marshaller handed to the payload and asserts `instanceof JdkMarshaller`. Reads better and fails with the actual marshaller in the message.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/BinaryMetadataVersionInfo.java
anton-vinogradov: Moved into the class javadoc, here and in the five other pinned messages.

modules/core/src/main/java/org/apache/ignite/internal/util/ErrorMessage.java
anton-vinogradov: Moved into the class javadoc.

modules/core/src/test/java/org/apache/ignite/spi/MessagesPluginProvider.java
anton-vinogradov: Added a javadoc saying why: a marshaller companion exists only for a message that has something to marshal, and the loader throws when there is none - which is the normal case for the plain test messages.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/DiscoveryMarshalling.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public final class DiscoveryMarshalling {
+    /** */
+    public static <M extends Message> void marshal(M msg, GridKernalContext kctx,
+        @Nullable CacheObjectContext cacheObjCtx) throws IgniteCheckedException {
+        MessageMarshalling.marshal(msg, kctx.marshallerContext().jdkMarshaller(), kctx, cacheObjCtx);
+    }
+
+    /** */
+    public static <M extends Message> void unmarshal(M msg, GridKernalContext kctx,
```
shishkovilja: Used only in tests, can be safely removed.

modules/core/src/test/java/org/apache/ignite/internal/util/distributed/DistributedProcessResultMarshallingTest.java
anton-vinogradov: Let's keep it easy and separate.

modules/core/src/main/java/org/apache/ignite/internal/plugin/AbstractMessageFactoryProvider.java
anton-vinogradov: AbstractMessageFactoryProvider.java:94: warning: [unchecked] unchecked cast

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/DiscoveryMarshalling.java
anton-vinogradov: Right, no production caller for it - removed. The two tests that used it moved to the cache-free overload.

anton-vinogradov:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13462)** verdict · RunAll build [9281035](https://ci2.ignite.apache.org/build/9281035) · 147 suites ran, 0 reused

✅ **No blockers** — nothing in this run looks caused by this PR. 22 pre-existing/flaky tests filtered out.

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
```diff
@@ -115,6 +115,12 @@ public class MessageMarshallerGenerator extends MessageCompanionGenerator {
...
     private boolean selfMarshalling;
 
+    /** Whether the message pins the JDK marshaller, see {@link JdkMarshalled}. */
+    private boolean jdkMarshalled;
+
+    /** Name of the marshaller the generated body uses: the transport one, or the pinned JDK one. */
+    private String marshVar;
+
     /** */
     private boolean hasMarshalled;
```
Vladsz83: Never used any more?

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/CommunicationMarshalling.java
```diff
@@ -0,0 +1,49 @@
...
+
+package org.apache.ignite.internal.managers.communication;
+
+import org.apache.ignite.IgniteCheckedException;
+import org.apache.ignite.internal.GridKernalContext;
+import org.apache.ignite.internal.processors.cache.CacheObjectContext;
+import org.apache.ignite.plugin.extensions.communication.Message;
+import org.jetbrains.annotations.Nullable;
+
+/** Marshalling of the communication transport. It uses the schema-aware marshaller. */
```
Vladsz83: I'd say `schema-aware marshaller by default until onter marshaller isn't set for message by...`. Let's say in short what the schema-arare marshalling is or put link to it.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/CommunicationMarshalling.java
```diff
@@ -0,0 +1,49 @@
...
+public final class CommunicationMarshalling {
+    /** */
+    public static <M extends Message> void marshal(M msg, GridKernalContext kctx,
+        @Nullable CacheObjectContext cacheObjCtx) throws IgniteCheckedException {
+        MessageMarshalling.marshal(msg, kctx.marshaller(), kctx, cacheObjCtx);
+    }
+
+    /** */
+    public static <M extends Message> void unmarshal(M msg, GridKernalContext kctx,
```
Vladsz83: Let's codestyle this: split params lines.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/CommunicationMarshalling.java
```diff
@@ -0,0 +1,49 @@
...
+package org.apache.ignite.internal.managers.communication;
+
+import org.apache.ignite.IgniteCheckedException;
+import org.apache.ignite.internal.GridKernalContext;
+import org.apache.ignite.internal.processors.cache.CacheObjectContext;
+import org.apache.ignite.plugin.extensions.communication.Message;
+import org.jetbrains.annotations.Nullable;
+
+/** Marshalling of the communication transport. It uses the schema-aware marshaller. */
+public final class CommunicationMarshalling {
```
Vladsz83, line 31: Private constr maybe?

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/DiscoveryMarshalling.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Marshalling of the discovery transport. It uses the JDK marshaller, which asks for no type registration at all.
+ * <p>
+ * A discovery message is marshalled by a thread that has to move discovery forward itself: the ring worker writing
+ * it to the socket, a custom event listener marshalling inline, or a node that is not in the ring yet. The binary
+ * marshaller would make such a thread wait for a cluster-wide round - a class name mapping, or a new metadata
+ * version - and only a discovery thread delivers the answer, so the wait never ends and the ring stops with it.
+ * A type the cluster has already accepted marshals fine, but a message cannot rely on that.
+ */
+public final class DiscoveryMarshalling {
```
Vladsz83, line 35: Private constr maybe?

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/DiscoveryMarshalling.java
```diff
@@ -0,0 +1,46 @@
...
+    /** */
+    public static <M extends Message> void marshal(M msg, GridKernalContext kctx,
```
Vladsz83: Let's codestyle this: one line or split params lines.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/MessageMarshalling.java
```diff
@@ -52,28 +49,30 @@ public static <M extends Message> void marshal(M msg, GridKernalContext kctx, @N
      *
      * @param msgFactory Message factory to resolve the marshaller from.
      * @param msg Message to marshal.
+     * @param marsh Marshaller of the transport.
      * @param kctx Kernal context.
      * @param cacheObjCtx Cache object context of the enclosing message, or {@code null} at the top level.
      */
-    public static <M extends Message> void marshal(IgniteMessageFactory msgFactory, M msg, GridKernalContext kctx,
-        @Nullable CacheObjectContext cacheObjCtx) throws IgniteCheckedException {
+    public static <M extends Message> void marshal(IgniteMessageFactory msgFactory, M msg, Marshaller marsh,
```
Vladsz83, line 61: Suggestion: let's codestyle this, similar and related places. One line or split-for-params lines.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/BinaryMetadataVersionInfo.java
```diff
@@ -28,8 +28,10 @@
  * Used internally to track version counters (see javadoc for {@link MetadataUpdateProposedMessage} for more details).
  * The version refers solely to the internal protocol for updating BinaryMetadata and is unknown externally.
  * It can be updated dynamically from different nodes and threads on the same node.
+ * <p>
+ * Travels both transports: by discovery in the data bag, by communication in the {@link MetadataResponseMessage}.
```
Vladsz83: Suggestion: upper-cased `Discovery` and `Communication`. Better search for analogues in other places..

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/StoredCacheData.java
```diff
@@ -42,7 +43,10 @@
  * in order to be serialization wise agnostic to further additions or removals of fields.
  * <p>
  * All changes must be made with the respect of RU rules.
+ * <p>
+ * Travels both transports: by discovery when a cache starts, by communication when a snapshot is restored.
```
Vladsz83: Same. Upper-cased Discovery and Connunication.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/SingleNodeMessage.java
```diff
@@ -27,12 +28,16 @@
 
 /**
  * Single node result message.
+ * <p>
+ * The process result travels both transports: this message carries it to the coordinator by communication, and the
```
Vladsz83: Upper-cased Communication and Discovery

modules/core/src/main/java/org/apache/ignite/internal/util/ErrorMessage.java
```diff
@@ -28,9 +29,12 @@
 
 /**
  * Message used to transfer {@link Throwable} objects.
+ * <p>
+ * Travels both transports, and an error is an arbitrary user class the cluster may not know yet.
```
Vladsz83: Suggestion `both transports (Communication and Discovery)`. Let's keep such a note in another similar places.

modules/core/src/test/java/org/apache/ignite/internal/managers/communication/GridIoManagerOrderedUnmarshalFailureTest.java
```diff
@@ -242,13 +243,13 @@ private static class Serializer implements MessageSerializer<FailingUnmarshalMes
     /** Fails the unmarshal of the flagged messages. */
     private static class FailingMarshaller implements MessageMarshaller<FailingUnmarshalMessage> {
         /** {@inheritDoc} */
-        @Override public void marshal(FailingUnmarshalMessage msg, GridKernalContext kctx,
+        @Override public void marshal(FailingUnmarshalMessage msg, Marshaller marsh, GridKernalContext kctx,
```
Vladsz83: Let's codestyle this. One line or several for-param lines. Same below.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/transactions/TxDeadlockDetectionMessageMarshallingTest.java
```diff
@@ -67,8 +67,8 @@ public void testMessageUnmarshallWithoutCacheContext() throws Exception {
                 @Override public void onMessage(UUID nodeId, Object msg, byte plc) {
                     if (msg instanceof TxLocksResponse) {
                         try {
-                            MessageMarshalling.unmarshal((TxLocksResponse)msg,
-                                clientCtx.kernalContext(), null, clientCtx.deploy().globalLoader());
+                            CommunicationMarshalling.unmarshal((TxLocksResponse)msg, clientCtx.kernalContext(), null,
```
Vladsz83: Let's codestyle this. One line or several for-param lines.

modules/core/src/test/java/org/apache/ignite/internal/util/distributed/DistributedProcessResultMarshallingTest.java
```diff
@@ -0,0 +1,320 @@
...
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.junit.Test;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static org.apache.ignite.internal.util.distributed.DistributedProcess.DistributedProcessType.TEST_PROCESS;
+import static org.apache.ignite.testframework.GridTestUtils.loadSerializer;
+
+/**
+ * Tests that the result of a {@link DistributedProcess} is marshalled with the JDK marshaller on both transports: it
+ * goes to the coordinator by communication and comes back in the {@link FullMessage} by discovery, while a marshalled
```
Vladsz83: Let's clarify a bit. The coordinatot sends `InitMessage` via `Discovery` at the process start. Nodes send `SingleNodeMessage` via `Communication` as theirs work results. Coordinator collects nodes results containing within `SingleNodesMessage` (not `SingleNodesMessage` itself) and send `FullNodeMessage`. But, what do we check here?

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/SingleNodeMessage.java
```diff
@@ -27,12 +28,16 @@
...
+ * {@link FullMessage} of the coordinator carries it back to every node by discovery.
  *
  * @param <R> Result type.
  * @see DistributedProcess
  * @see FullMessage
  * @see InitMessage
  */
+@JdkMarshalled
```
Vladsz83, line 40: `FullMessage` doesnt contain exaclty `SingleMessage`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/verify/TransactionsHashRecord.java
```diff
@@ -18,13 +18,19 @@
...
 import org.apache.ignite.internal.Order;
 import org.apache.ignite.internal.util.tostring.GridToStringInclude;
 import org.apache.ignite.internal.util.typedef.internal.S;
 import org.apache.ignite.plugin.extensions.communication.Message;
 
-/** Represents committed transactions hash for a pair of nodes. */
+/**
+ * Represents committed transactions hash for a pair of nodes.
+ * <p>
+ * Travels both transports inside the result of a snapshot check, and carries a user consistent id.
```
Vladsz83: Let's say what the transports

shishkovilja (commented):
# Code Review: Message Marshalling Refactoring

## Summary

A marshalling refactoring: the `Marshaller` is now threaded as an explicit parameter through `MessageMarshaller.marshal/unmarshal` and the message-marshalling helpers, instead of being stored/initialized inside message factories. The `MessageMarshalling` helper was split into `CommunicationMarshalling` (uses binary/schema-aware marshaller) and `DiscoveryMarshalling` (uses the JDK marshaller), and the `@UseBinaryMarshaller` annotation was renamed to `@JdkMarshalled` and moved from `core` to `codegen`.

Overall this is a **well-reasoned, internally consistent refactor** with correct transport routing and no critical defects. The separation of communication vs. discovery marshalling is a genuine improvement: it removes the risk of a discovery-ring thread deadlocking on a binary-marshaller class-name round.

Scope: 134 files, +967 / −611 lines (staged).

## Findings

### Suggestion

1. **No automated guard for the cross-transport invariant**
   - **Files:** `JdkMarshalled.java`, `CommunicationMarshalling.java:31`, `DiscoveryMarshalling.java:38-44`
   - **What's wrong:** After this change, a non-`@JdkMarshalled` message with a marshaller-dependent blob gets *binary* on the communication leg and *JDK* on the discovery leg. Any both-transport message that's marshaller-dependent but not annotated `@JdkMarshalled` would break under cross-transport deserialization.
   - **Why it matters:** The six confirmed cases (`SingleNodeMessage`, `ErrorMessage`, `StoredCacheData`, `BinaryMetadataVersionInfo`, `PartitionHashRecord`, `TransactionsHashRecord`) are correctly pinned, but this invariant is only enforced by manual analysis.
   - **Suggested fix:** Add a test that enumerates registered messages reachable from both the discovery and communication legs and asserts any `@Marshalled`/`MarshallableMessage` member shared by both is `@JdkMarshalled` (or make it a codegen error).

2. **Static utility classes lost their private constructors**
   - **Files:** `MessageMarshalling.java:41`, `CommunicationMarshalling.java`, `DiscoveryMarshalling.java`
   - **What's wrong:** The diff removed the existing `private MessageMarshalling() {}` and the new classes never added one, leaving implicit public default constructors.
   - **Why it matters:** This violates Ignite's convention for static-utility holders.
   - **Suggested fix:** Restore a `/** */ private <Class>() { }` in all three.

### Nice to have

3. **Codegen `prependPinnedMarshaller` uses a substring heuristic**
   - **File:** `MessageMarshallerGenerator.java:1013-1022`
   - **What's wrong:** The `jdkMarsh` declaration is emitted when no generated line *contains* `"jdkMarsh"`. A field whose accessor happened to contain that literal could falsely suppress/trigger the declaration.
   - **Why it matters:** Fragile; today works for all generated companions but is not robust to future field names.
   - **Suggested fix:** Track a boolean "marsh var referenced" during codegen instead of re-scanning lines.

4. **Generated reference files are out of sync with the generator**
   - **Files:** `TestJdkMarshalledMessageMarshaller.java`, `TestJdkMarshalledChildMessageMarshaller.java`
   - **What's wrong:** These end with a trailing `\n`, but `MessageMarshallerGenerator` does not emit one. The other updated resources in this diff had theirs stripped to match.
   - **Why it matters:** Compile-testing normalizes whitespace so tests pass, but the files are not byte-identical to generator output.
   - **Suggested fix:** Regenerate these via the actual `MessageProcessor`.

5. **`MessagesPluginProvider.marshaller(...)` swallows all failures**
   - **File:** `modules/core/src/test/.../spi/MessagesPluginProvider.java:62`
   - **What's wrong:** The catch-all for `loadMarshaller`'s wrapped `RuntimeException` returns `null`, hiding real generation bugs.
   - **Why it matters:** A genuine defect in marshaller generation would be silently hidden.
   - **Suggested fix:** Restrict swallowing to the "companion absent" case (e.g., inspect cause for `ClassNotFoundException`).

6. **`JdkMarshalled` Javadoc implies runtime behavior**
   - **File:** `JdkMarshalled.java:~26-30`
   - **What's wrong:** With `@Retention(CLASS)` it's a compile-time hint consumed only by the generator, but the Javadoc reads like it does runtime work.
   - **Suggested fix:** Word it as a compile-time hint, e.g., "instructs the generated marshaller to pin the JDK marshaller".

7. **Unused marshaller resolved for `@JdkMarshalled` messages on the comm leg**
   - **File:** `CommunicationMarshalling.java`
   - **What's wrong:** `kctx.marshaller()` (binary) is computed and passed in, but the generated `@JdkMarshalled` body ignores it and re-resolves `jdkMarshaller()`.
   - **Why it matters:** One trivial final-field getter per message; cleanup only.

8. **`ServiceSingleNodeDeploymentResult` keeps a now-redundant `Serializable` marker**
   - **File:** `ServiceSingleNodeDeploymentResult.java`
   - **What's wrong:** Implements `Serializable` only for legacy reasons while being a `Message`; harmless now that the annotation is gone.
   - **Suggested fix:** Optionally revisit.

## Verdict

**Approve** — no critical issues; the refactor is correct and well-structured.

Most valuable follow-up: **finding #1** (automating the cross-transport `@JdkMarshalled` invariant), which is currently only checked manually.

---

*Review date: 2026-08-14*

modules/codegen/src/main/java/org/apache/ignite/internal/MessageMarshallerGenerator.java
anton-vinogradov: Right, it stopped being read when the generated companion lost its marshaller constructor - removed.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/CommunicationMarshalling.java
anton-vinogradov: Reworded: "It uses the schema-aware marshaller, {@link BinaryMarshaller}, which writes a type once and refers to it by id afterwards, so a type unknown to the cluster has to be registered first."

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/binary/BinaryMetadataVersionInfo.java
anton-vinogradov: Done, here and in the other five notes: Communication and Discovery are now spelled as the subsystems are.

modules/core/src/main/java/org/apache/ignite/internal/util/distributed/SingleNodeMessage.java
anton-vinogradov: True, fixed: the FullMessage carries the results collected from the single node messages, not the messages themselves.

modules/core/src/test/java/org/apache/ignite/internal/util/distributed/DistributedProcessResultMarshallingTest.java
anton-vinogradov: Rewrote the class javadoc along those lines: InitMessage by Discovery, SingleNodeMessage by Communication, then the collected results in a FullMessage by Discovery.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/CommunicationMarshalling.java
anton-vinogradov: Split one parameter per line.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/DiscoveryMarshalling.java
anton-vinogradov: Split one parameter per line.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/MessageMarshalling.java
anton-vinogradov: Split one parameter per line.

modules/core/src/test/java/org/apache/ignite/internal/managers/communication/GridIoManagerOrderedUnmarshalFailureTest.java
anton-vinogradov: Split one parameter per line.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/transactions/TxDeadlockDetectionMessageMarshallingTest.java
anton-vinogradov: Split one parameter per line.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/CommunicationMarshalling.java
anton-vinogradov: Dropped these on purpose earlier in the review: the class is final and holds static methods only, so nothing can instantiate it anyway. Happy to bring them back if you prefer them for consistency with the rest of the code.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/DiscoveryMarshalling.java
anton-vinogradov: Same as in CommunicationMarshalling.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/verify/TransactionsHashRecord.java
anton-vinogradov: Named them: "Travels both transports, Communication and Discovery, inside the result of a snapshot check".

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/CommunicationMarshalling.java
anton-vinogradov: Added back, here and in MessageMarshalling and DiscoveryMarshalling - the intent to keep anyone from instantiating these is worth stating explicitly.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/DiscoveryMarshalling.java
anton-vinogradov: Added back as well.

anton-vinogradov:
/runall

---
🚀 **RunAll queued** — [build 9283013](https://ci2.ignite.apache.org/build/9283013) · live progress & verdict: [Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13462). The verdict lands here when the run finishes.

Vladsz83 (approved):

anton-vinogradov:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13462)** verdict · RunAll build [9283013](https://ci2.ignite.apache.org/build/9283013) · 147 suites ran, 0 reused

✅ **No blockers** — nothing in this run looks caused by this PR. 26 pre-existing/flaky tests filtered out.

_♻️ Settled after 1 auto re-run wave(s): #1 — 2 watch + 2 broken suite(s)._

PR #13464, opened 2026-08-11
IGNITE-28739 SQL Calcite: Add expression support for FETCH

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-28739


modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/LimitOffsetPlannerTest.java
```diff
@@ -61,6 +66,51 @@ public void testLimit() throws Exception {
...
+        assertPlan("SELECT * FROM TEST ORDER BY ID OFFSET 1 ROWS "
+                + "FETCH FIRST (ABS(0.5)) ROWS ONLY", publicSchema,
+            isInstanceOf(IgniteLimit.class)
+                .and(limit -> limit.offset() != null && limit.fetch() != null)
+                .and(input(isInstanceOf(IgniteExchange.class)
+                    .and(input(isInstanceOf(IgniteSort.class)
+                        .and(sort -> sort.offset == null && sort.fetch == null))))));
+
+        assertPlan("SELECT * FROM TEST ORDER BY ID OFFSET 1 ROWS "
+                + "FETCH FIRST (RAND_INTEGER(1) + 2) ROWS ONLY", publicSchema,
```
zstan, line 99: plz explain why '(RAND_INTEGER(1) + 2)' is not push down into IgniteSort, white '(ABS(100))' pushes ?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
```diff
@@ -283,4 +403,9 @@ private String createSql(int lim, int off, boolean param, boolean sorted) {
 
         return sb.toString();
     }
+
+    /** */
+    private static BigDecimal bigValue() {
```
zstan:
```suggestion
    private static BigDecimal moreThanMaxLong() {
```

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
```diff
@@ -199,6 +197,128 @@ public void testOffsetOutOfRange() throws Exception {
...
+        assertQuery("SELECT id FROM TEST_REPL ORDER BY id FETCH FIRST (1 + NVL(?, 10000)) ROWS ONLY")
+            .withParams(1)
+            .returns(0)
+            .returns(1)
+            .check();
+    }
+
+    /** */
+    @Test
+    public void testFetchExpressionWithoutPushDown() throws Exception {
```
zstan, line 350: Why it called "WithoutPushDown" ? i didn\`t see any push down here, if it relates to appropriate planner test - it can be changed after some time...

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
```diff
@@ -199,6 +197,128 @@ public void testOffsetOutOfRange() throws Exception {
...
+    /** */
+    @Test
+    public void testInvalidFetchExpression() {
+        assertThrowsSqlException("SELECT * FROM TEST_REPL FETCH FIRST (" + bigValue() + ") ROWS ONLY", null);
+        assertThrowsSqlException("SELECT * FROM TEST_REPL FETCH FIRST (" + bigValue() + " + 1) ROWS ONLY", null);
+        assertThrowsSqlException("SELECT * FROM TEST_REPL FETCH FIRST (1 + " + bigValue() + ") ROWS ONLY", null);
+
+        assertThrowsSqlException("SELECT * FROM TEST_REPL FETCH FIRST (-2) ROWS ONLY", null);
+        assertThrowsSqlException("SELECT * FROM TEST_REPL FETCH FIRST (2 - 3) ROWS ONLY", null);
+        assertThrowsSqlException("SELECT * FROM TEST_REPL FETCH FIRST (1 + 2 - 4) ROWS ONLY", null);
```
zstan: redundant previous test will be enough

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
```diff
@@ -199,6 +197,128 @@ public void testOffsetOutOfRange() throws Exception {
         assertQuery("SELECT (SELECT id FROM TEST_PART ORDER BY id LIMIT 1 OFFSET 10)").returns(NULL_RESULT).check();
     }
 
+    /** */
+    @Test
+    public void testInvalidFetchExpression() {
```
zstan, line 219: test cases are throws different exceptions, why you don\`t want to fix them ?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DynamicParametersIntegrationTest.java
```diff
@@ -391,4 +524,9 @@ public void testWrongParametersNumberInUpdate() {
     private void assertUnexpectedNumberOfParameters(String qry, Object... params) {
         assertThrows(qry, IgniteSQLException.class, "Wrong number of query parameters", params);
     }
+
+    /** */
+    private static BigDecimal bigValue() {
```
zstan:
```suggestion
    private static BigDecimal moreThanMaxLong() {
```

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
```diff
@@ -199,6 +197,128 @@ public void testOffsetOutOfRange() throws Exception {
...
+        assertQuery("SELECT id FROM (SELECT id from TEST_REPL ORDER BY id FETCH FIRST (2 + 3) ROWS ONLY) " +
+            "ORDER BY id FETCH NEXT (1 + 1) ROWS ONLY")
+            .returns(0)
+            .returns(1)
+            .check();
+    }
+
+    /** */
+    @Test
+    public void testFetchExpressionWithRewrite() throws Exception {
```
zstan: need to belongs to DynamicParametersIntegrationTest

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DynamicParametersIntegrationTest.java
```diff
@@ -216,6 +216,139 @@ public void testFractionalLimitOffset() {
...
+            "ORDER BY id FETCH NEXT (1 + ?) ROWS ONLY")
+            .withParams(2, 1)
+            .returns(0)
+            .returns(1)
+            .check();
+    }
+
+    /** */
+    @Test
+    public void testFetchExpressionWithRewrite() {
```
zstan: what 'rewrite' is means here ?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DynamicParametersIntegrationTest.java
```diff
@@ -216,6 +216,139 @@ public void testFractionalLimitOffset() {
...
+
+        assertThrowsSqlException("SELECT * FROM PERSON FETCH FIRST (1 + ? - 4) ROWS ONLY", null, 1);
+        assertThrowsSqlException("SELECT * FROM PERSON FETCH FIRST (? - (50 - 20)) ROWS ONLY", null, 2);
+
+        assertThrowsSqlException("SELECT * FROM PERSON FETCH FIRST SQRT(?) ROWS ONLY", null, 4);
+    }
+
+    /** */
+    @Test
+    public void testFetchExpression() {
```
zstan, line 261: I think all such kind of tests need to be implemented in scope of : test_limit.test
Integrations test need to contain only rare\concreete\hard script-tests implemented cases

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteRelFieldTrimmer.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.calcite.util.mapping.Mapping;
+import org.apache.calcite.util.mapping.Mappings;
+import org.jetbrains.annotations.Nullable;
+
+import static java.util.Collections.emptySet;
+
+/** Field trimmer that preserves expression-based FETCH nodes. */
+// TODO: https://issues.apache.org/jira/browse/CALCITE-7592
+//  Remove this class and IgniteSqlToRelConvertor.newFieldTrimmer() after upgrading to Calcite 1.43.
+public class IgniteRelFieldTrimmer extends RelFieldTrimmer {
```
zstan, line 43: May it\'s better to fix it in IgnitePlanner ? with appropriate TODO: CALCITE-7592 remove after ... mention ?

```
    public RelRoot trimUnusedFields(RelRoot root) {
        final SqlToRelConverter.Config cfg = sqlToRelConverterCfg
            .withExpand(false)
            .withTrimUnusedFields(true);
        SqlToRelConverter converter = sqlToRelConverter(validator(), catalogReader, cfg);
        boolean ordered = !root.collation.getFieldCollations().isEmpty();
        boolean dml = SqlKind.DML.contains(root.kind);

        if (root.rel instanceof Sort) {
            if (!supportedByRelBuilder(((Sort)root.rel).fetch))
                return root;
        }

        return root.withRel(converter.trimUnusedFields(dml || ordered, root.rel));
    }
```

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgnitePlanner.java
```diff
@@ -343,7 +343,14 @@ private static boolean isAsCall(SqlNode node) {
 
     /** {@inheritDoc} */
     @Override public RelRoot rel(SqlNode sql) {
-        SqlToRelConverter sqlToRelConverter = sqlToRelConverter(validator(), catalogReader, sqlToRelConverterCfg);
+        // TODO: https://issues.apache.org/jira/browse/CALCITE-7592
+        //  Restore the original implementation after upgrading to Calcite 1.43: use validator() directly and remove
+        //  deriveLimitDynamicParameterTypes().
+        IgniteSqlValidator validator = (IgniteSqlValidator)validator();
```
zstan: misunderstand this change, plz highlight what test will fail without this change ?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
tkalkirill: I'd prefer to leave it here, just one line.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
tkalkirill: For simplicity's sake, I think it'll be helpful to ensure that incorrect expressions won't work. 
What exactly are you suggesting? 
What do you mean by "fix"?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
tkalkirill: Yes, I fix the test, a similar one already exists.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DynamicParametersIntegrationTest.java
tkalkirill: It refers to the unconditional SQL-node rewrite performed by `IgniteSqlCallRewriteTable`: the `NVL(?, 10000)` call is rewritten to an equivalent `CASE` expression during validation. The test verifies that a dynamic parameter inside `FETCH` is still typed and converted correctly after this rewrite.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DynamicParametersIntegrationTest.java
tkalkirill: I'm not familiar with the rules for writing such tests.
Is there somewhere I can read about this and learn more?
It's not clear to me why they should be separate from Java tests. 
What's the advantage of this approach?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/SortConverterRule.java
```diff
@@ -74,13 +84,15 @@ public interface Config extends RelRule.Config {
...
+                boolean pushLimit = canPushLimit(sort);
+
                 RelNode igniteSort = new IgniteSort(
                     cluster,
                     cluster.traitSetOf(IgniteConvention.INSTANCE).replace(sort.getCollation()),
                     convert(sort.getInput(), cluster.traitSetOf(IgniteConvention.INSTANCE)),
                     sort.getCollation(),
-                    sort.offset,
-                    sort.fetch,
+                    pushLimit ? sort.offset : null,
```
zstan, line 94: In case where canPushLimit(sort) return 'false' how offset and fetch will pass into final execution Nodes ?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgnitePlanner.java
tkalkirill: This change was carried over from the previous Calcite 1.40-based implementation. At that time, the issue was reproducible for `FETCH` expressions containing dynamic parameters after an SQL rewrite, particularly `NVL(?, ...)`.  
I rechecked the scenario with the current code, including `testFetchExpressionWithRewrite`, and the issue is no longer reproducible: `validateFetch()` already derives the parameter types from the rewritten `FETCH` expression. Therefore, this additional traversal and the cast are no longer necessary, and I will remove them.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/LimitOffsetPlannerTest.java
tkalkirill: `FETCH` pushdown causes the expression to be evaluated more than once: by every local `IgniteSort` and again by the global `IgniteLimit`. `ABS(100)` is deterministic and always produces the same value, so the pushdown is safe. `RAND_INTEGER(...)` is marked non-deterministic, therefore different nodes/evaluations may produce different limits. If a local sort produces a smaller value, it may discard rows required by the global limit and change the query result. `RAND_INTEGER(1)` is used to keep the test result stable while still exercising the non-deterministic-expression path.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteRelFieldTrimmer.java
tkalkirill: The check in `IgnitePlanner` would only handle a root `Sort`, while an expression-based `FETCH` may belong to a nested `Sort`, for example under a `Project` or inside a subquery. It would also disable field trimming for the whole tree. Overriding `SqlToRelConverter.newFieldTrimmer()` localizes the workaround to every affected `Sort` and keeps trimming enabled for the rest of the plan. The `TODO` already mentions removing both the custom trimmer and the override after upgrading to Calcite 1.43.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
tkalkirill: You’re right. I overdid it a bit with this test: it does not verify pushdown directly, and the same behavior is already covered by the planner test. I’ll just remove it.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/SortConverterRule.java
tkalkirill: `pushLimit` only controls the optional local Top-N optimization in `IgniteSort`. The original offset and fetch are always preserved in the outer `IgniteLimit`. When pushdown is disabled, `IgniteSort` sorts the complete input, and the outer single-distribution `IgniteLimit` is converted to `LimitNode`, which evaluates and applies the original expressions.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
zstan: fix = determine in test, 
```
assertThrowsSqlException("SELECT * FROM TEST_REPL FETCH FIRST (" + bigValue() + ") ROWS ONLY", 
IgniteResource.INSTANCE.illegalFetchLimit("fetch \\ limit").str());
```
and so on ?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DynamicParametersIntegrationTest.java
zstan: If it rewrites to expression and we expect it - seems we need to check rewritten is applied ? Because it can be skipped ...

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DynamicParametersIntegrationTest.java
zstan: Yep you are right and a can\`t force you to make it as i suggest, but it just like a non documented rule that tests for common cases\operands\invariants need to be placed there, it just more quick to check expected behavior for all cases in script tests than search appropriate *Integration test

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/SortConverterRule.java
zstan: got it

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
tkalkirill: I’d prefer not to assert specific error messages in these tests. Their purpose is to verify that invalid `FETCH` expressions are rejected, while the actual message depends on the stage where the error is detected (parsing, validation, or execution) and may change after the Calcite upgrade. Checking it would make the tests unnecessarily brittle. However, if you strongly prefer checking the messages here, I can add them.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DynamicParametersIntegrationTest.java
tkalkirill: I would leave it as is, unless you insist.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
zstan: Error messages is a part of public API, thus it better to be deterministic

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DynamicParametersIntegrationTest.java
tkalkirill: Good point. I renamed the original test so that it no longer implies that it explicitly verifies rewriting. I also added dedicated tests to `OperatorsExtensionIntegrationTest` that explicitly verify custom query rewriting from a `ROWNUM` predicate to `FETCH`, both with a literal and a dynamic parameter.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
tkalkirill: I agree that error messages are part of the public API and should be deterministic. However, asserting the exact message for every invalid case would significantly increase the size and verbosity of these tests. Do you insist on adding message assertions here? If so, should they be added for every negative check or only for representative cases covering each error category?

zstan: Just explain - that error message you expect in your java code will changed ? Or returned error class at all ?  I think it need to be covered by tests !

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/DynamicParametersIntegrationTest.java
zstan: your right ... it can\`t be a blocker for merge of cource )

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/SortConverterRule.java
```diff
@@ -101,4 +113,44 @@ public interface Config extends RelRule.Config {
...
+
+    /** Returns {@code true} if FETCH can be safely pushed to the sort. */
+    private static boolean canPushLimit(Sort sort) {
+        RexNode fetch = sort.fetch;
+
+        // A non-deterministic FETCH must be evaluated only once by the outer IgniteLimit.
+        if (!RexUtil.isDeterministic(fetch))
+            return false;
+
+        if (!RexUtil.isConstant(fetch) || containsDynamicParameter(fetch))
```
zstan: wht you wan\'`t  push down dyn param ? it differs from previous behavior

tkalkirill: Fix it.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
tkalkirill: Fix it.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
```diff
@@ -655,6 +656,10 @@ private boolean hasExchange(RelNode rel) {
         long offset = validateAndGetOffset(rel.offset, SortNode.OFFSET_DEFAULT);
         long fetch = validateAndGetFetch(rel.fetch, SortNode.FETCH_DEFAULT);
 
+        // Zero FETCH is enforced by the outer IgniteLimit, while SortNode accepts only positive FETCH values.
+        if (fetch == 0)
```
zstan, line 660: why do we need this change ? I run limited set of tests and it\`s all ok, plz clarify ? 
Also if this fix is helpful for smth. it looks hacky 
1. run validateAndGetFetch with SortNode.FETCH_DEFAULT
2. additional check and additional set : fetch = SortNode.FETCH_DEFAULT;

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
```diff
@@ -1072,6 +1077,11 @@ private long validateAndGetFetchOffsetParams(RexNode node, String op) {
         Supplier<Object> scalar = expressionFactory.execute(node);
         Object param = scalar.get();
 
+        if (param == null && !(node instanceof RexDynamicParam)) {
```
zstan, line 1080: 'node instanceof RexDynamicParam' - looks like redundant check ? Or plz show me appropriate test ?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -267,10 +268,65 @@ private void validateTableModify(SqlNode table) {
     @Override protected void validateSelect(SqlSelect select, RelDataType targetRowType) {
         super.validateSelect(select, targetRowType);
 
-        validateFetchOffset(select.getFetch(), "fetch / limit");
+        validateFetch(select, "fetch / limit");
```
zstan, line 272: now we have 2 functions: one - for 'validateFetch' and other for 'validateFetchOffset' ? looks weird

zstan (changes requested):

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -267,10 +268,65 @@ private void validateTableModify(SqlNode table) {
...
+
+        if (node instanceof SqlNodeList) {
+            for (SqlNode child : (SqlNodeList)node)
+                validateFetchExpression(child, clauseName);
+        }
+        else if (node instanceof SqlCall call) {
+            if (call.isA(SqlKind.QUERY))
+                throw newValidationError(call, IgniteResource.INSTANCE.illegalFetchLimit(clauseName));
+
+            if (call.getKind() == SqlKind.OVER)
```
zstan: I don\`t think that this is a good approach: 
1. comment this line : if (call.getKind() == SqlKind.OVER)
and run 'testInvalidFetchExpression' - you obtain aggregate related error, i mean - if you miss some operands comparison here - you will obtain confusing exception.
2. I don\`t like an idea to handle each stupid oparation here, you will always miss smth, check : 
'SELECT * FROM TEST_REPL FETCH FIRST (1=1) ROWS ONLY' but if you want - you can do it with extended tests for all such a cases

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
tkalkirill: You're right, I overdid it a bit, I got rid of it.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
tkalkirill: If this check is commented out, the `FETCH FIRST (NULLIF(?, 1))` assertion starts failing because it gets an incorrect dynamic parameter type error instead of an illegal FETCH value. The `RexDynamicParam` exclusion preserves the existing error for a bare `FETCH FIRST ?` with a `NULL` parameter.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
tkalkirill: Agree, it looks weird. This is temporary because the current ticket adds expression support only for FETCH/LIMIT, while OFFSET still uses the existing validation path. We will align OFFSET behavior and consolidate these methods in the follow-up ticket.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
tkalkirill: I partially agree that covering every possible invalid expression and keeping all related errors consistent may be quite cumbersome. For now, I’ll remove these specific checks and rely on the general validation. If a concrete need arises, we can restore them together with focused tests.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/IgniteSqlValidator.java
```diff
@@ -408,6 +460,21 @@ else if (call.getKind() == SqlKind.OVER && call.operand(1) instanceof SqlWindow)
 
     /** {@inheritDoc} */
     @Override protected SqlNode performUnconditionalRewrites(SqlNode node, boolean underFrom) {
+        if (node instanceof SqlOrderBy) {
```
zstan: in such a case this need to be passed too ?  (now it fails)

`SELECT * FROM PERSON FETCH FIRST (COALESCE(3, 5)) ROWS ONLY`

seems we need to check all logical operators , unfortunately i don\`t know how to get full supported list of such a functions (

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
```diff
@@ -199,6 +207,132 @@ public void testOffsetOutOfRange() throws Exception {
...
+        assertThrowsSqlException("SELECT * FROM TEST_REPL FETCH FIRST (id) ROWS ONLY", ILLEGAL_FETCH_VAL_ERR_MSG);
+        assertThrowsSqlException("SELECT * FROM TEST_REPL FETCH FIRST (id + 1) ROWS ONLY", ILLEGAL_FETCH_VAL_ERR_MSG);
+        assertThrowsSqlException("SELECT * FROM TEST_REPL FETCH FIRST (ABS(id)) ROWS ONLY", ILLEGAL_FETCH_VAL_ERR_MSG);
+
+        assertThrowsSqlException("SELECT * FROM TEST_REPL FETCH FIRST SQRT(4) ROWS ONLY", ENCOUNTERED_ERR_MSG);
+    }
+
+    /** */
+    @Test
+    public void testFetchExpression() throws Exception {
```
zstan, line 246: let simplify these tests a bit, plz append "assumeNoTransactions();"

    /** */
    private void assumeNoTransactions() {
        assumeTrue("Test use queries that doesn't touch any data. Skip for tx modes", sqlTxMode == SqlTransactionMode.NONE);
    }

zstan (approved):
one minor, plz append this check - help TC with time economy plz )

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/LimitOffsetIntegrationTest.java
tkalkirill: I tried it locally and didn't see any reduction in test execution.
I'll leave it as is for now; I might improve this later with a separate task.

zstan: you right this tests are fast , but i still see no problem to add this check

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/metadata/IgniteMdMinRowCount.java
```diff
@@ -0,0 +1,75 @@
...
+        rowCnt = Math.max(rowCnt - offset, 0D);
+
+        double limit = literalValueApproximatedByDouble(rel.fetch,
+            rel.fetch == null ? rowCnt : 0D);
+
+        return limit < rowCnt ? limit : rowCnt;
+    }
+
+    /** {@inheritDoc} */
+    @Override public Double getMinRowCount(EnumerableLimit rel, RelMetadataQuery mq) {
```
zstan: EnumerableLimit - we not play with enumerables !!

zstan (changes requested):

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/metadata/IgniteMdMinRowCount.java
tkalkirill: Fix it.

PR #13479, opened 2026-08-13
IGNITE-27822 Support recursive query for Calcite engine

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-27822


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/logical/IgniteLogicalRecursiveStaticSpool.java
```diff
@@ -0,0 +1,36 @@
...
+    /** */
+    public IgniteLogicalRecursiveStaticSpool(RelNode input) {
+        super(input.getCluster(), input.getCluster().traitSet(), input);
+    }
+
+    /** {@inheritDoc} */
+    @Override public RelNode copy(RelTraitSet traitSet, List<RelNode> inputs) {
+        return new IgniteLogicalRecursiveStaticSpool(sole(inputs));
+    }
+}
```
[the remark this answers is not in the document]
zstan: The same comment from me - seems you can merge RecursiveTableSpoolNode and existing TableSpoolNode, also i can\`t find different implementations in calcite thus i think - this is correct approach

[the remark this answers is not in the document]
vldpyatkov: 1) IgniteLogicalRecursiveStaticSpool#copy is definitely ignoring traitSet, which formally is a contract violation. (fixed)
2) I believe that the traitSet has not to be inherited for input. Look at the Calcite class LogicalTableSpool.
3) RecursiveTableSpoolNode and TableSpoolNode are different classes. TableSpoolNode accumulates rows one time and replays them on rewind(), but RecursiveTableSpoolNode makes a new row set in each iteration. We can use TableSpoolNode for only static datasets. Look at RecursiveStaticSpoolConverterRule. It creates IgniteTableSpool, which uses TableSpoolNode.

zstan:
ok, looks good for me, let\`s wait additional review

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/RecursiveCteState.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Collections;
+import java.util.List;
+import org.apache.ignite.internal.processors.query.calcite.exec.ExecutionContext;
+import org.apache.ignite.internal.processors.query.calcite.exec.tracker.RowTracker;
+import org.apache.ignite.internal.util.GridUnsafe;
+
+/** Query-local current and next deltas of a recursive CTE. */
+public class RecursiveCteState<Row> {
+    /** Rows visible to the recursive table scan. */
+    private List<Row> current = Collections.emptyList();
```
alex-plekhanov: Abbreviation should be used for `current`

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/RecursiveCteState.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /** Rows produced by the active seed or recursive term. */
+    private List<Row> next;
+
+    /** Memory tracker for rows in the current delta. */
+    private RowTracker<Row> currentMemoryTracker;
```
alex-plekhanov: Abbreviation should be used for `current`

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/RecursiveCteIntegrationTest.java
```diff
@@ -43,18 +99,149 @@ public void testHierarchicalQueryIsNotSupported() {
...
-        Throwable err = GridTestUtils.assertThrows(
-            log,
-            () -> sql(qry),
+        assertTrue(plan, plan.contains("IgniteRepeatUnion"));
+        assertTrue(plan, plan.contains("IgniteRecursiveTableSpool"));
+    }
+
+    /** */
+    @Test
+    public void testRecursiveDeltaIsAccountedForMemoryQuota() {
```
alex-plekhanov: Let's move quota test to `MemoryQuotasIntegrationTest`

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/RepeatUnionNode.java
```diff
@@ -0,0 +1,163 @@
...
+    private static final int SEED_SOURCE = 0;
+
+    /** Index of the recursive-term input. */
+    private static final int RECURSIVE_SOURCE = 1;
+
+    /** Query-local recursive state. */
+    private final RecursiveCteState<Row> state;
+
+    /** Maximum number of recursive iterations, or a negative value for no limit. */
+    private final int iterationLimit;
```
alex-plekhanov, line 40: As far as I understand there is no way to now to set iterationLimit to value other than -1. Maybe we should limit it in execution node?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/RecursiveCteUtils.java
```diff
@@ -0,0 +1,195 @@
...
+    /** Returns whether the subtree produces the same result on every recursive iteration. */
+    private static boolean isInvariant(RelNode rel) {
+        rel = original(rel);
+
+        if (!RelOptUtil.getVariablesUsed(rel).isEmpty())
+            return false;
+
+        DeterminismChecker checker = new DeterminismChecker();
+
+        rel.accept(checker);
```
alex-plekhanov, line 129: Looks like ProjectableFilterableTableScan, IgniteSortedIndexSpool, IgniteHashIndexSpool are not implement `accept` method as required (not sure if this can affect this ticket).

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteRepeatUnion.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.calcite.plan.RelTraitSet;
+import org.apache.calcite.rel.BiRel;
+import org.apache.calcite.rel.RelInput;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.RelWriter;
+
+import static java.util.Objects.requireNonNull;
+
+/** Coordinator-side iterative UNION ALL for a recursive CTE. */
+public class IgniteRepeatUnion extends BiRel implements IgniteRel {
```
alex-plekhanov: Why not `extends RepeatUnion`?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/RecursiveCteUtils.java
vldpyatkov: This comment is definitely relevant. We use _RexShuttle_ to determine whether a subtree of the recursive term is invariant and can be materialized only once. Without an accept implementation, expressions owned by these nodes are not visited, so a subtree containing non-deterministic expressions could be incorrectly treated as invariant and not re-evaluated on subsequent iterations.
I implemented accept for _ProjectableFilterableTableScan_, _IgniteHashIndexSpool_, and _IgniteSortedIndexSpool_, covering projections, conditions, and search rows.
This behavior is covered by the following tests in RecursiveCteIntegrationTest:
- testIndependentNonDeterministicSubtreeIsEvaluatedForEveryIteration
- testNonDeterministicTableScanIsNotMaterialized
- testNonDeterministicExpressionsInIndexSpoolsAreDetected — covers both hash and sorted index spools.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rel/IgniteRepeatUnion.java
vldpyatkov: Yes, it makes code simpler there.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/RepeatUnionNode.java
vldpyatkov: I understand the intention to protect the query from infinite recursion. However, I would prefer not to hard-code an arbitrary positive limit, such as 100, because the SQL standard does not define such a limit and valid queries may require more iterations. Moreover, we always can create an explicit recursive parameter (like LEVEL in Oracle) to limit recursion depth through the query.
Calcite represents (RepeatUnion#iterationLimit) the iteration limit explicitly and defines the following semantics: -1 means unlimited recursion, 0 returns only the seed, and N allows no more than N recursive iterations. We currently preserve this value throughout the execution path:
LogicalRepeatUnion → RecursiveCteConverterRule → IgniteRepeatUnion → RepeatUnionNode.
If we want a safety limit, I think it should be configurable rather than hard-coded. We can create a separate ticket to expose it through a query-level hint, for example MAX_RECURSION(100). Do you think we should create such a ticket?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/RepeatUnionConverterRule.java
```diff
@@ -0,0 +1,107 @@
...
+    /** */
+    private static RelNode unwrapSpool(RelNode rel, RelOptTable table, String term) {
+        rel = RecursiveCteUtils.original(rel);
+
+        if (!(rel instanceof LogicalTableSpool))
+            throw unsupported("the " + term + " must use a transient table spool");
+
+        LogicalTableSpool spool = (LogicalTableSpool)rel;
+
+        if (!RecursiveCteUtils.sameTransientTable(spool.getTable(), table)) {
```
alex-plekhanov: Redundant braces

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/RepeatUnionNode.java
alex-plekhanov: In case of wrong query user can fail the whole cluster. I think there should be at least some configuration (maybe distributed property) with reasonable default limit (and error message with a hint that property can be changed  if limit exceeded).

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/RecursiveCtePlannerTest.java
```diff
@@ -0,0 +1,184 @@
...
+            schema
+        );
+
+        assertRecursivePlan(plan);
+        IgniteRepeatUnion repeatUnion = findFirstNode(plan, byClass(IgniteRepeatUnion.class));
+
+        assertTrue(planDescription(plan), repeatUnion.getLeft() instanceof IgniteValues);
+        assertEquals(1, findNodes(plan, byClass(IgniteRecursiveTableScan.class)).size());
+
+        checkSplitAndSerialization(plan, schema);
```
alex-plekhanov: Let's avoid manual plans check and use assertPlan instead:
```
        assertPlan(sql, schema, isInstanceOf(IgniteRepeatUnion.class)
            .and(hasDistribution(IgniteDistributions.single()))
            .and(input(0, isInstanceOf(IgniteValues.class)))
            .and(input(1, hasChildThat(isInstanceOf(IgniteRecursiveTableScan.class)))
        ));
```

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/RecursiveCtePlannerTest.java
```diff
@@ -0,0 +1,184 @@
...
+        IgniteValues seedValues = findFirstNode(repeatUnion.getLeft(), byClass(IgniteValues.class));
+
+        assertNotNull(planDescription(plan), seedValues);
+        assertEquals(planDescription(plan), 2, seedValues.getTuples().size());
+
+        checkSplitAndSerialization(plan, schema);
+    }
+
+    /** Creates an employee table with the requested distribution and optional manager index. */
+    private static IgniteSchema hierarchySchema(IgniteDistribution distribution, boolean withManagerIndex) {
```
alex-plekhanov: Abbreviation should be used for "index"

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/RecursiveCtePlannerTest.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /** Calcite places multiple non-recursive branches into the seed input of RepeatUnion. */
+    @Test
+    public void testRecursiveCteWithMultipleSeedBranches() throws Exception {
+        IgniteSchema schema = new IgniteSchema(DEFAULT_SCHEMA);
+
+        IgniteRel plan = physicalPlan(
+            "WITH RECURSIVE numbers(n) AS (" +
+                "SELECT 1 " +
```
alex-plekhanov: Values are merged on union and this is not quite informative test. Lets check that "union" for seed remains and nested into left hand of RepeatUnion, for example, something like:
```
        IgniteSchema schema = createSchema(createTable("T", IgniteDistributions.single(), "ID", SqlTypeName.INTEGER));

        String sql = "WITH RECURSIVE numbers(n) AS (" +
            "SELECT ID FROM T WHERE ID = 1 " +
            "UNION ALL " +
            "SELECT ID FROM T WHERE ID = 2 " +
            "UNION ALL " +
            "SELECT n + 1 FROM numbers WHERE n < 3" +
            ") " +
            "SELECT n FROM numbers";

        assertPlan(sql, schema, isInstanceOf(IgniteRepeatUnion.class)
            .and(input(0, isInstanceOf(IgniteUnionAll.class)
                .and(input(0, isTableScan("T")))
                .and(input(1, isTableScan("T")))
            ))
        );
```

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/RecursiveCteIntegrationTest.java
```diff
@@ -43,18 +96,209 @@ public void testHierarchicalQueryIsNotSupported() {
...
             "JOIN employee_hierarchy h ON e.manager_id = h.id" +
             ") " +
-            "SELECT id, manager_id, name, depth FROM employee_hierarchy ORDER BY depth, id";
+            "SELECT id, manager_id, name, depth FROM employee_hierarchy";
+
+        String plan = (String)sql("EXPLAIN PLAN FOR " + qry).get(0).get(0);
+
+        assertTrue(plan, plan.contains("IgniteRepeatUnion"));
+        assertTrue(plan, plan.contains("IgniteRecursiveTableScan"));
+        assertFalse(plan, plan.contains("IgniteRecursiveTableSpool"));
```
alex-plekhanov: There is no such node now

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/RecursiveCteIntegrationTest.java
```diff
@@ -43,18 +96,209 @@ public void testHierarchicalQueryIsNotSupported() {
...
+                "SELECT 1 " +
+                "UNION ALL " +
+                "SELECT left_numbers.n + 1 " +
+                "FROM numbers left_numbers " +
+                "JOIN numbers right_numbers ON left_numbers.n = right_numbers.n " +
+                "WHERE left_numbers.n < 3" +
+            ") " +
+            "SELECT n FROM numbers",
+            IgniteSQLException.class,
+            "the recursive term must contain no more than one self-reference"
```
alex-plekhanov: Do we really need this limitation? What kind of problems are we face without check for single recursive table usage?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/RecursiveCteIntegrationTest.java
```diff
@@ -43,18 +96,209 @@ public void testHierarchicalQueryIsNotSupported() {
...
+        String plan = (String)sql("EXPLAIN PLAN FOR " + qry).get(0).get(0);
+
+        assertTrue(plan, plan.contains("IgniteTableScan"));
+        assertTrue(plan, plan.contains("NEXTRECURSIVEVALUE"));
+        assertFalse(plan, plan.contains("IgniteTableSpool"));
+    }
+
+    /** */
+    @Test
+    public void testNonDeterministicIndexScanIsNotMaterialized() {
```
alex-plekhanov: Result is not checked. Only plan. Maybe move it to the planner test?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/RecursiveCteIntegrationTest.java
```diff
@@ -43,18 +96,209 @@ public void testHierarchicalQueryIsNotSupported() {
...
+            .returns(1, 0)
+            .returns(2, 1)
+            .returns(3, 2)
+            .returns(4, 3)
+            .check();
+    }
+
+    /** */
+    @Test
+    public void testNonDeterministicTableScanIsNotMaterialized() {
```
alex-plekhanov: Result is not checked. Only plan. Maybe move it to the planner test?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/RecursiveCteConverterRule.java
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.calcite.rel.metadata.RelMetadataQuery;
+import org.apache.ignite.internal.processors.cache.query.IgniteQueryErrorCode;
+import org.apache.ignite.internal.processors.query.IgniteSQLException;
+import org.apache.ignite.internal.processors.query.calcite.rel.IgniteConvention;
+import org.apache.ignite.internal.processors.query.calcite.rel.IgniteRepeatUnion;
+
+import static org.apache.ignite.internal.processors.query.calcite.trait.IgniteDistributions.single;
+
+/** Converts Calcite's logical recursive union to coordinator-side execution. */
+public class RecursiveCteConverterRule extends AbstractIgniteConverterRule<LogicalRepeatUnion> {
```
alex-plekhanov: RepeatUnionConverterRule?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
```diff
@@ -162,6 +168,9 @@ public class LogicalRelImplementor<Row> implements IgniteRelVisitor<Node<Row>> {
     /** */
     private final ExpressionFactory<Row> expressionFactory;
 
+    /** Query-local recursive CTE states, keyed by transient table identifier. */
+    private final Map<String, RecursiveCteState<Row>> recursiveStates = new HashMap<>();
```
alex-plekhanov: Maybe it's worth to store cte state inside execution context, not inside rel implementor?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/BaseQueryContext.java
```diff
@@ -312,6 +322,18 @@ public int[] partitions() {
         return null;
     }
 
+    /** Returns an identifier unique for the given recursive CTE within this planning context. */
+    public synchronized String recursiveCteStateId(RelOptTable table) {
```
alex-plekhanov: State id only generated on planning phase, looks like planning context is more suitable place for this method.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/RecursiveCteUtils.java
```diff
@@ -0,0 +1,214 @@
...
+        List<RelNode> inputs = rel.getInputs();
+
+        if (inputs.isEmpty())
+            return rel;
+
+        List<RelNode> newInputs = new ArrayList<>(inputs.size());
+
+        for (RelNode input : inputs) {
+            if (referenceCount(input, table) == 0 && isInvariant(input))
+                newInputs.add(new StaticInput(input));
```
alex-plekhanov: Can we convert traits in markStaticInputs instead of creating new rel node and new rule for this relationship node?

alex-plekhanov (approved):

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/RepeatUnionNode.java
vldpyatkov: Added a distributed property `sql.calcite.recursiveCteIterationLimit` with default value 100.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/RecursiveCteIntegrationTest.java
vldpyatkov: There is no strong technical reason for this limitation anymore.
Removed this limitation, and tests finish without exception now:
testRecursiveCteWithMultipleRecursiveBranchesIsRejected;
testRecursiveTermWithMultipleSelfReferencesIsRejected

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/RecursiveCteConverterRule.java
vldpyatkov: Ok.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
vldpyatkov: I moved the state map and lookup method to ExecutionContext.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/BaseQueryContext.java
vldpyatkov: Agreed. The identifiers are generated and consumed only during planning, so PlanningContext is a better owner. I moved the identity map and sequence there.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/rule/RecursiveCteUtils.java
vldpyatkov: Yes. StaticInput is currently only a marker used to defer a special trait conversion.
I converted inputs immediately in RepeatUnionConverterRule and removed both the marker node and RecursiveStaticInputConverterRule.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/RecursiveCtePlannerTest.java
vldpyatkov: I fixed it.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/RecursiveCtePlannerTest.java
vldpyatkov: The current test proves the resulting seed values but does not prove that a multi-branch seed is preserved. I used table scans as suggested and asserted that IgniteUnionAll is the left input of IgniteRepeatUnion.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/RecursiveCteIntegrationTest.java
vldpyatkov: Moved.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/RecursiveCteIntegrationTest.java
vldpyatkov: Moved.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/RecursiveCteIntegrationTest.java
vldpyatkov: I removed testExplainRecursiveCte because it is duplicated in the RecursiveCtePlannerTest.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/planner/RecursiveCtePlannerTest.java
vldpyatkov: Right, I fixed it.

PR #13483, opened 2026-08-15
IGNITE-28986 Local and Partitions queries are broken if plan contains Exchange

zstan:


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/metadata/ColocationGroup.java
```diff
@@ -134,16 +134,21 @@ public List<UUID> nodeIds() {
      * @return List of partitions (index) and nodes (items) having an appropriate partition in
      * {@link GridDhtPartitionState#OWNING} state, calculated for distributed tables, involved in query execution.
      */
-    public List<List<UUID>> assignments() {
+    public @Nullable List<List<UUID>> assignments() {
```
alex-plekhanov: Looks like return value can't be null

vldpyatkov, line 137: The `@Nullable` annotation is redundant in this method.

vldpyatkov (approved):

PR #13535, opened 2026-08-28
IGNITE-29021 SQL Calcite: UDF with BigDecimal argument fails

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-29021

### DIFF
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
@@ -495,9 +495,9 @@ public class ConverterUtils {
      * Handles decimal type specifically with explicit type conversion.
      */
     private static Expression convertAssignableType(Expression argument, Type targetType) {
-        if (targetType != BigDecimal.class)
+        if (targetType != BigDecimal.class || !Types.needTypeCast(argument.getType(), targetType))
             return argument;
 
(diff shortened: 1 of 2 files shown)


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
```diff
@@ -495,9 +495,9 @@ static List<Expression> convertAssignableTypes(Class<?>[] targetTypes,
      * Handles decimal type specifically with explicit type conversion.
      */
     private static Expression convertAssignableType(Expression argument, Type targetType) {
-        if (targetType != BigDecimal.class)
+        if (targetType != BigDecimal.class || !Types.needTypeCast(argument.getType(), targetType))
```
zstan, line 498: plz explain why do we need this branch ? '|| !Types.needTypeCast(argument.getType(), targetType)'

tkalkirill: It is not required for correctness. It preserves the previous `convert(...)` behavior and avoids generating a redundant `toBigDecimal(...)` call when the argument is already a `BigDecimal`.

zstan: got it !

zstan (approved):

PR #13540, opened 2026-08-31
IGNITE-29026 Introduce extension points in compatibility testcontainers

wernerdv:

### Notes


modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteContainer.java
```diff
@@ -439,13 +467,71 @@ private static File testClassesJar() throws IOException {
...
+     */
+    private static Collection<String> classResources(String clsPath, String nestedPrefix) throws IOException {
+        URL url = IgniteContainer.class.getClassLoader().getResource(clsPath);
+
+        if (url == null)
+            throw new IOException("Class not found on classpath: " + clsPath);
+
+        List<String> res = new ArrayList<>();
+
+        res.add(clsPath);
```
maksaska: classResources() returns the top-level class twice on the file: path, so building the test-classes jar always fails.

Line 486 unconditionally seeds the list with res.add(clsPath), and then the file: branch's filter on line 496 explicitly matches name.equals(simple + ".class") and re-adds the very same resource as pkg + f.getName(). testClassesJar() calls out.putNextEntry(new JarEntry(resName)) per element, and ZipOutputStream.putNextEntry throws ZipException: duplicate entry on the second occurrence.

This is the default execution path: all four classes in testClasses() live in modules/compatibility/src/test/java, so under surefire they resolve from target/test-classes (protocol file). Since testClassesJar() is called from the constructor (line 200), IgniteRebalanceOnUpgradeTest fails at new IgniteContainer(...) before any container starts.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteContainer.java
```diff
@@ -142,29 +125,41 @@ public class IgniteContainer extends GenericContainer<IgniteContainer> {
...
-     * param idx Node index.
+     * @param idx Node index.
      */
     public IgniteContainer(String imageName, Network net, String hostname, String consistentId, int idx) throws Exception {
         super(DockerImageName.parse(imageName));
 
         this.hostname = hostname;
         this.consistentId = consistentId;
-        workDirPath = WORK_DIR_PATH + "/" + hostname;
+        rootDir = rootDirPath();
```
maksaska: The constructor calls the new overridable hooks before a subclass can initialize its own state: rootDirPath() (line 149), commonConfigResource()/sourceConfigResource() (lines 198-200), testClasses() via testClassesJar() (line 200) and waitStrategy() (line 215).

Since extensibility is the point of the PR, could these be resolved lazily on first use (or moved into configure() / containerIsCreating(), which Testcontainers already provides for exactly this)? At minimum the javadoc should state that overrides must not depend on subclass instance state.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteContainer.java
```diff
@@ -116,18 +107,10 @@ public class IgniteContainer extends GenericContainer<IgniteContainer> {
...
-        DisabledRollingUpgradeProcessor.class.getName(),
-        DisabledValidationProcessor.class.getName()
-    );
-
     /** Seconds to wait after SIGTERM before SIGKILL. */
     private static final int SHUTDOWN_TIMEOUT_SEC = 30;
 
-    /** Jar holding {@link #TEST_CLASSES}, injected so the old image can load it. */
+    /** Jar holding the {@link #testClasses() test classes}, injected so the old image can load it. */
     private static volatile File testClassesJar;
```
maksaska: testClasses() is now overridable per subclass, but the resulting jar is still memoized in private static volatile File testClassesJar (line 114) with no key. targetLibsArchive (line 117) / libsArchive() (line 534) have the same shape, and the latter embeds that jar.

Within one surefire fork, whichever container is constructed first wins: if the base test builds the jar with the four default classes and a subclass overriding testClasses() runs afterwards, the subclass silently reuses the stale jar and its node dies with ClassNotFoundException — with the outcome depending on test execution order.

Suggest keying the cache on the effective class list (e.g. ConcurrentHashMap<List<String>, File>) or making it instance-level.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteClusterContainer.java
```diff
@@ -30,21 +31,52 @@ public class IgniteClusterContainer implements Startable {
...
+     * @param consistentIds Consistent ID's.
+     * @param idx Node index.
+     * @return The node container.
+     */
+    protected IgniteContainer container(String imageName, Network net, List<String> consistentIds, int idx) throws Exception {
+        return new IgniteContainer(imageName, net, "node" + (1 + idx), consistentIds.get(idx), idx);
+    }
+
+    /** Builds the node containers. */
+    protected void initContainers() throws Exception {
```
maksaska, line 68: Moving container creation from the constructor into initContainers() makes start() non-idempotent: the method always appends to the pre-existing containers list, so a second start() — or a retry after Startables.deepStart(...).join() fails — yields 2N containers with duplicate hostnames, duplicate consistent IDs and duplicate fixed host ports, after which activateCluster(containers.size()) waits for a baseline size that can never be reached.

Also, containers() used to be populated right after construction and now returns an empty list until start() runs. Nothing in the current test depends on that, but it is a silent contract change for subclasses that might want to inspect containers before starting.

A if (!containers.isEmpty()) return; guard (or a started flag) would cover both.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteClusterContainer.java
```diff
@@ -30,21 +31,52 @@ public class IgniteClusterContainer implements Startable {
...
+    }
 
+    /**
+     * @param imageName Image name.
+     * @param net Shared test network the container must be attached to.
+     * @param consistentIds Consistent ID's.
+     * @param idx Node index.
+     * @return The node container.
+     */
+    protected IgniteContainer container(String imageName, Network net, List<String> consistentIds, int idx) throws Exception {
```
maksaska: container(String imageName, Network net, List<String> consistentIds, int idx) re-passes imageName, net and consistentIds, all of which are already fields of the instance, and hands the override the whole list plus an index rather than the id it needs. protected IgniteContainer container(int idx) would be a cleaner hook; if the parameters are kept for readability, passing consistentIds.get(idx) instead of the list would still be an improvement.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteContainer.java
```diff
@@ -439,13 +467,71 @@ private static File testClassesJar() throws IOException {
...
 
+    /**
+     * Resolves the fully qualified resources (the top-level class plus its nested classes, e.g. {@code Outer$1})
+     * for a class located either on the file system or inside a jar on the classpath.
+     *
+     * @param clsPath Resource path of the top-level class (package separator replaced with '/', ending in {@code .class}).
+     * @param nestedPrefix Package-based prefix of the nested classes, e.g. {@code org/apache/foo/Simple$}.
+     * @return Resource names of the class and its nested classes.
+     */
+    private static Collection<String> classResources(String clsPath, String nestedPrefix) throws IOException {
```
maksaska: Two small redundancies in classResources():
- nestedPrefix is a parameter but only the jar: branch uses it — the file: branch recomputes pkg/simple from clsPath itself. It could just be derived inside the method and the parameter dropped.
- The method resolves getResource(clsPath) and the caller then resolves getResource(resName) again for the same top-level class.

maksaska (changes requested):

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteClusterContainer.java
```diff
@@ -30,21 +31,52 @@ public class IgniteClusterContainer implements Startable {
...
      * @param consistentIds Consistent ID's.
      */
-    public IgniteClusterContainer(String imageName, List<String> consistentIds) throws Exception {
+    public IgniteClusterContainer(String imageName, List<String> consistentIds) {
+        this.imageName = imageName;
+        this.consistentIds = consistentIds;
+
+        net = Network.newNetwork();
         containers = new ArrayList<>(consistentIds.size());
+    }
```
chesnokoff: Can we keep `net` initialization at field declaration? It does not depend on constructor arguments

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteContainer.java
wernerdv: Fixed

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteContainer.java
wernerdv: Moved into configure().

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteContainer.java
wernerdv: Added caching to ConcurrentHashMap<List, File>.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteClusterContainer.java
wernerdv: Fixed

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteClusterContainer.java
wernerdv: Fixed

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteContainer.java
wernerdv: Fixed

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteClusterContainer.java
wernerdv: Fixed

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteContainer.java
```diff
@@ -383,77 +398,194 @@ private String execControl(String... cmd) {
...
+     */
+    protected String[] command(String... cmd) {
+        String[] fullCmd = new String[cmd.length + 1];
+
+        fullCmd[0] = rootDir + "bin/control.sh";
+
+        System.arraycopy(cmd, 0, fullCmd, 1, cmd.length);
+
+        return fullCmd;
+    }
```
chesnokoff: Do we need both extension points? Overriding command() seems enough to customize the command while keeping execution and error handling in the base class. What use case requires execControl() to be protected?

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteClusterContainer.java
```diff
@@ -30,24 +31,66 @@ public class IgniteClusterContainer implements Startable {
...
+
+    /**
+     * Factory hook for the node container. Overrides only receive {@code idx}; the image name, network and
+     * consistent IDs are instance fields (see {@link #imageName}, {@link #net}, {@link #consistentIds}).
+     *
+     * @param idx Node index.
+     * @return The node container.
+     */
+    protected IgniteContainer container(int idx) throws Exception {
+        return new IgniteContainer(imageName, net, "node" + (1 + idx), consistentIds.get(idx), idx);
```
chesnokoff, line 64: Why do we do 1+idx here?

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteClusterContainer.java
```diff
@@ -30,24 +31,66 @@ public class IgniteClusterContainer implements Startable {
...
     }
 
     /** {@inheritDoc} */
     @Override public void start() {
+        // Idempotent: either the cluster already started successfully, or container creation succeeded
+        // but startup (deepStart/activateCluster) failed on a previous attempt — in both cases the
+        // containers list is already populated and must not be built a second time (duplicate hostnames,
+        // consistent IDs and fixed host ports would make the baseline unreachable).
+        if (started || !containers.isEmpty())
+            return;
```
chesnokoff: Question: I agree that we should not call initContainers() twice. But should we just return when `containers` is not empty and `started` is still false?

I'm not sure what is better here: silently skip it or throw an error

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteContainer.java
```diff
@@ -383,77 +398,194 @@ private String execControl(String... cmd) {
...
+
+    /**
+     * @return Jar with the {@link #testClasses() test classes}, built per distinct class list and reused.
+     *      The cache is keyed on the effective {@link #testClasses()} result so subclasses overriding it get
+     *      their own jar instead of silently reusing the one built for the base class.
+     */
+    protected File testClassesJar() throws IOException {
+        // List.copyOf makes an immutable, value-comparable key.
+        List<String> classes = List.copyOf(testClasses());
```
chesnokoff: I'm not sure we need the comment. Looks like it just explains semantics of `List.copyOf`

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteContainer.java
```diff
@@ -383,77 +398,194 @@ private String execControl(String... cmd) {
...
+
+        /** Resolved URL of the resource. */
+        final URL url;
+
+        /** @param name Resource name on the classpath. */
+        ClassResource(String name, URL url) {
+            this.name = name;
+            this.url = url;
         }
     }
```
chesnokoff: Probably we could use `record` to reduce LOC, but it's up to you

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteClusterContainer.java
wernerdv: 1 + idx maps the 0-based internal node index idx to the 1-based node naming scheme (node1..nodeN). The hostname doubles as the Docker network alias, and nodes discover each other through the static IP finder in common-test-config.xml, which hardcodes node1:47500, node2:47500, node3:47500. Since that list starts at node1, idx == 0 must produce hostname node1, hence the +1.

The consistent IDs also follow the same 1-based convention (node1, node2, …) in the compatibility tests, so keeping the hostname in sync avoids mismatches in the baseline. The raw 0-based idx is still passed through unchanged for the fixed published ports and work-directory indexing, which are 0-based internally — only the hostname needs the shift.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteClusterContainer.java
wernerdv: Good catch, fixed.

Now idempotency only applies to a fully started cluster, and a half-started cluster fails loudly instead of being silently skipped.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/testframework/testcontainers/IgniteContainer.java
wernerdv: You're right — there's no current need for execControl() to be a separate extension point.
I've reverted execControl() back to private.

chesnokoff (approved):
LGTM

maksaska (approved):

PR #13543, opened 2026-09-02
IGNITE-29031 SQL Calcite: Support byte[] in UDF and UDTF parameters and results

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-29031


zstan:
First of all you need to follow the common process [1] i.e. make a PR from your own ignite mirror, not from origin = https://github.com/apache but from: https://github.com/tkalkirill/ignite

[1] https://cwiki.apache.org/confluence/spaces/IGNITE/pages/177047163/How+to+Contribute#HowtoContribute-GITworkflow

tkalkirill:
@zstan Okay, subsequent tickets will do as described.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/TableFunctionScan.java
```diff
@@ -66,6 +68,22 @@ private Row convertToRow(Object rowContainer) {
...
-        return rowFactory.create(rowArr);
+        return rowFactory.create(convertBinaryColumns(rowArr));
+    }
+
+    /** Converts binary column values to the internal representation. */
+    private Object[] convertBinaryColumns(Object[] row) {
+        Object[] convertedRow = row;
+
+        for (int i = 0; i < row.length; i++) {
+            if (row[i] instanceof byte[] && SqlTypeUtil.isBinary(rowType.getFieldList().get(i).getType())) {
```
zstan: 1. we already have common converter : TypeUtils#toInternal - it need to be used
2. explain why do you need row.clone() ? .toArray() - already produced a new array.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
```diff
@@ -309,7 +309,7 @@ public static <Row> Function<Row, Row> resultTypeConverter(ExecutionContext<Row>
 
     /** */
     private static Function<Object, Object> fieldConverter(ExecutionContext<?> ectx, RelDataType fieldType) {
-        Type storageType = ectx.getTypeFactory().getJavaClass(fieldType);
+        Type storageType = SqlTypeUtil.isBinary(fieldType) ? byte[].class : ectx.getTypeFactory().getJavaClass(fieldType);
```
zstan: plz explain why do we need changes in this class ? All changes need to be covered by tests, i miss them.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/IgniteTableFunction.java
```diff
@@ -132,4 +145,43 @@ private static void raiseValidationError(Method method, String errPostfix) {
...
+        for (int i = 0; i < method.getParameterTypes().length; i++) {
+            if (method.getParameterTypes()[i] == byte[].class)
+                res.set(i, sqlBinaryParameter(res.get(i)));
+        }
+
+        return Collections.unmodifiableList(res);
+    }
+
+    /** Prevents Calcite from evaluating binary literals while deriving a table function row type. */
+    private static FunctionParameter sqlBinaryParameter(FunctionParameter delegate) {
```
zstan: looks like a hack as for me

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/UserDefinedFunctionsIntegrationTest.java
```diff
@@ -464,6 +466,43 @@ public void testBigDecimalFunctionArgument() {
         assertQuery("SELECT udf.decimalToInt(5.3)").returns(5).check();
     }
 
+    /** */
+    @Test
+    public void testBinaryFunctions() {
```
zstan, line 479: I disagree that this PR need to cover only byte[], what about temporal types ? BO, appropriate tests need to be appended.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/IgniteTableFunction.java
```diff
@@ -132,4 +145,43 @@ private static void raiseValidationError(Method method, String errPostfix) {
 
         throw new IgniteSQLException("Unable to create table function for method '" + mtdSign + "'. " + errPostfix);
     }
+
+    /** Returns function parameters represented as SQL types where required. */
+    private static List<FunctionParameter> sqlParameters(Method method, List<FunctionParameter> functionParameters) {
+        var res = new ArrayList<>(functionParameters);
+
+        for (int i = 0; i < method.getParameterTypes().length; i++) {
```
zstan: I\`d prefer not to relay on implicit size alignment
```suggestion
        for (int i = 0; i < functionParameters.size(); i++) {
```

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/TableFunctionScan.java
tkalkirill: Tried to fix them.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
tkalkirill: This was needed because the previous implementation was only for a `byte[]`, now it is not needed, I got rid of it.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/IgniteTableFunction.java
tkalkirill: This change is only needed when a `byte[]` table-function argument is a binary literal, for example `binaryTableLength(x'010203')`. Calcite tries to materialize the literal as byte[] during row-type inference, before the runtime converter is involved. Dynamic parameters and other non-literal expressions work without this workaround.

Would you prefer that we drop binary-literal support from this fix and address it separately if needed, or keep the workaround with an explanatory comment and a dedicated test?

zstan: I prefer to revise this approach, if it\`s hard to implement now - let\`s fill an additional issue, thanks !

tkalkirill: When testing with temporal types, it turned out that this is currently necessary and does not work without it.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/UserDefinedFunctionsIntegrationTest.java
tkalkirill: Added quite a lot of tests.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ReflectiveCallNotNullImplementor.java
```diff
@@ -66,6 +68,17 @@ public ReflectiveCallNotNullImplementor(Method method) {
 
             callExpr = Expressions.call(target, method, translatedOperands);
         }
+
+        if (TypeUtils.isConvertableType(method.getReturnType())) {
```
zstan, line 72:
```
        @QuerySqlFunction
        public static java.util.Date kill1() {
            return java.sql.Date.valueOf("2020-01-01");
        }
```

and further call:

```
assertQuery("SELECT kill1()")
            .resultSize(1)
            .check();
```

will raise ClassCastException, while

```
        @QuerySqlFunction
        public static java.sql.Date kill2() {
            return java.sql.Date.valueOf("2020-01-01");
        }
``` 

passes, wdyt ?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/TableFunctionScan.java
```diff
@@ -58,14 +65,19 @@ private Row convertToRow(Object rowContainer) {
...
             : ((Collection<?>)rowContainer).toArray();
 
         if (rowArr.length != rowType.getFieldCount()) {
             throw new IgniteSQLException("Unable to process table function data: row length [" + rowArr.length
                 + "] doesn't match defined columns number [" + rowType.getFieldCount() + "].");
         }
 
+        for (int i = 0; i < rowArr.length; i++) {
+            if (!(rowType.getFieldList().get(i).getType() instanceof OtherType))
+                rowArr[i] = TypeUtils.toInternal(ctx, rowArr[i]);
```
zstan, line 78: failed:
```
        assertQuery("SELECT EXTRACT(YEAR FROM D) FROM f()")
            .resultSize(1)
            .check();

        @QuerySqlTableFunction(columnTypes =
            {java.util.Date.class}, columnNames = {"D"})
        public static Iterable<Object[]> f() { // kill
            return Collections.singletonList(new Object[] {java.sql.Date.valueOf("2020-01-01")});
        }
```

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
```diff
@@ -94,6 +94,8 @@ public class TypeUtils {
         LocalTime.class,
         Duration.class,
         Period.class,
+        char.class,
+        Character.class,
```
zstan, line 98: UDF like : static boolean f(char c)
and further `SELECT f('')` -> 
literal CHAR(0) passes SqlTypeAssignmentRule (CHAR assignable from CHAR, 
precision ignored) so no CAST is added; 
TypeUtils.fromInternal(root, \"\", char.class) -> 
StringIndexOutOfBoundsException

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
```diff
@@ -74,6 +76,9 @@ else if (fromType == java.sql.Timestamp.class) {
             else if (targetType == Long.class)
                 return Expressions.call(BuiltInMethod.TIMESTAMP_TO_LONG_OPTIONAL.method, operand);
         }
+        else if (fromType == byte[].class && targetType == ByteString.class)
```
zstan, line 79: dead code

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
```diff
@@ -111,6 +116,9 @@ else if (targetType == java.sql.Timestamp.class) {
             if (isA(fromType, Primitive.LONG))
                 return Expressions.call(BuiltInMethod.INTERNAL_TO_TIMESTAMP.method, operand);
         }
+        else if (targetType == byte[].class && fromType == ByteString.class)
```
zstan, line 119: dead code

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/IgniteTableFunction.java
```diff
@@ -100,6 +106,11 @@ public static IgniteTableFunction create(Method method, Class<?>[] colTypes, Str
         return Iterable.class;
     }
 
+    /** {@inheritDoc} */
+    @Override public List<FunctionParameter> getParameters() {
```
zstan, line 110: need to be moved into abstract class cause `IgniteScalarFunction#getParameters` do the same as i can see

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/TableFunctionScan.java
```diff
@@ -58,14 +65,19 @@ private Row convertToRow(Object rowContainer) {
...
+            ? ((Object[])rowContainer).clone()
             : ((Collection<?>)rowContainer).toArray();
 
         if (rowArr.length != rowType.getFieldCount()) {
             throw new IgniteSQLException("Unable to process table function data: row length [" + rowArr.length
                 + "] doesn't match defined columns number [" + rowType.getFieldCount() + "].");
         }
 
+        for (int i = 0; i < rowArr.length; i++) {
+            if (!(rowType.getFieldList().get(i).getType() instanceof OtherType))
```
zstan, line 77: Also near passed on master and failed in current branch, seems regression: 

```
    @Test
    public void testCollectionColumnTableFunctions() {
        client.getOrCreateCache(new
            CacheConfiguration<>("collection-column-table-functions")
            .setSqlSchema("PUBLIC")
            .setSqlFunctionClasses(CollectionColumnFunctionsLibrary.class));

        // ArrayList column -> ARRAY.
        assertQuery("SELECT ID, CARDINALITY(VALS), VALS[1] FROM " +
            "listColumn()")
            .returns(1, 2, 10)
            .returns(2, 3, 30)
            .check();

        assertQuery("SELECT t.ID, v.VAL FROM listColumn() t, " +
            " UNNEST(t.VALS) v(VAL) ORDER BY t.ID, v.VAL")
            .returns(1, 10)
            .returns(1, 20)
            .returns(2, 30)
            .returns(2, 40)
            .returns(2, 50)
            .check();

        // String[] column -> MULTISET.
        assertQuery("SELECT ID, CARDINALITY(NAMES) FROM arrayColumn()")
            .returns(1, 1)
            .returns(2, 2)
            .check();

        assertQuery("SELECT t.ID, n.NAME FROM arrayColumn() t, " +
            " UNNEST(t.NAMES) n(NAME) ORDER BY t.ID, n.NAME")
            .returns(1, "a")
            .returns(2, "b")
            .returns(2, "c")
            .check();
    }

    /** */
    public static final class CollectionColumnFunctionsLibrary {
        /** */
        private CollectionColumnFunctionsLibrary() {
            // No-op.
        }

        /** Returns rows with a {@link List}-typed column. */
        @QuerySqlTableFunction(columnTypes = {int.class, ArrayList.class},
            columnNames = {"ID", "VALS"})
        public static Iterable<Collection<?>> listColumn() {
            return Arrays.asList(
                Arrays.asList(1, new ArrayList<>(Arrays.asList(10, 20))),
                Arrays.asList(2, new ArrayList<>(Arrays.asList(30, 40,
                    50)))
            );
        }

        /** Returns rows with an array-typed column. */
        @QuerySqlTableFunction(columnTypes = {int.class, String[].class},
            columnNames = {"ID", "NAMES"})
        public static Iterable<Collection<?>> arrayColumn() {
            return Arrays.asList(
                Arrays.asList(1, Arrays.asList("a")),
                Arrays.asList(2, Arrays.asList("b", "c"))
            );
        }
    }
```

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ReflectiveCallNotNullImplementor.java
tkalkirill: Try to fix it.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/TableFunctionScan.java
tkalkirill: Try to fix it.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
tkalkirill: You're right, I went too far with `char/Character` support. I've removed the related conversions and tests from this PR.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
tkalkirill: The reverse branch in fromInternal is reachable through `RexImpTable.defineReflective`. I verified this with an operator backed by `binaryLength(byte[])`: with the conversion, a binary literal works; without it, generated code fails to compile because it passes `ByteString` to a method expecting `byte[]`. I suggest keeping this branch.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/TableFunctionScan.java
tkalkirill: I tried it, now it works.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/ExecutionTest.java
```diff
@@ -63,6 +79,51 @@ public class ExecutionTest extends AbstractExecutionTest {
...
+            for (Object container : new Object[] {row, Collections.singletonList(val)}) {
+                TableFunctionScan<Object[]> scan = new TableFunctionScan<>(ctx, rowType,
+                    () -> Collections.singletonList(container), ctx.rowHandler().factory(ctx.getTypeFactory(), rowType));
+
+                Object[] res = scan.iterator().next();
+
+                assertNotSame(row, res);
+                assertEquals(1, res.length);
+                assertSame(val, res[0]);
+                assertSame(val, row[0]);
```
zstan, line 122: redundant

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/LogicalRelImplementor.java
```diff
@@ -788,7 +788,7 @@ else if (rel instanceof Intersect)
 
         RowFactory<Row> rowFactory = ctx.rowHandler().factory(ctx.getTypeFactory(), rowType);
 
-        return new ScanNode<>(ctx, rowType, new TableFunctionScan<>(rowType, dataSupplier, rowFactory));
+        return new ScanNode<>(ctx, rowType, new TableFunctionScan<>(ctx, rowType, dataSupplier, rowFactory));
```
zstan: rowFactory = ctx.rowHandler().factory(ctx.getTypeFactory(), rowType); - gathered from `ctx` and `rowType` that already passed, so - no need to pass `rowFactory` as param? Now - it can be derived inside ScanNode constructor.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/IgniteFunctionParameter.java
```diff
@@ -0,0 +1,79 @@
...
+        return delegate.getName();
+    }
+
+    /** {@inheritDoc} */
+    @Override public RelDataType getType(RelDataTypeFactory typeFactory) {
+        JavaTypeFactory tf = (JavaTypeFactory)typeFactory;
+        RelDataType type = tf.toSql(delegate.getType(typeFactory));
+
+        // Prevent the validator from replacing OTHER with a structured type derived from a dynamic parameter value.
+        return type.getSqlTypeName() == SqlTypeName.OTHER ? new OtherType(type.isNullable()) : type;
```
zstan, line 68: optimization: if type.getSqlTypeName() == SqlTypeName.OTHER - above derived `RelDataType type` will be dropped

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
```diff
@@ -145,6 +150,66 @@ static List<Expression> fromInternal(Class<?>[] targetTypes,
         return list;
     }
 
+    /** */
+    static List<Expression> fromInternal(RexToLixTranslator translator,
+        Class<?>[] targetTypes,
+        List<Expression> expressions
+    ) {
+        final List<Expression> list = new ArrayList<>();
```
zstan:
```suggestion
        final List<Expression> list = new ArrayList<>(expressions.size());
```

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
zstan: sorry, but still miss it ( do we have a test for it ? suggest it plz ?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
```diff
@@ -145,6 +150,66 @@ static List<Expression> fromInternal(Class<?>[] targetTypes,
         return list;
     }
 
+    /** */
+    static List<Expression> fromInternal(RexToLixTranslator translator,
```
zstan: Look now you have two identical code, differs only with  "RexToLixTranslator translator" param
i mean:
ConverterUtils#fromInternal(java.lang.Class<?>[], List<Expression>)
and 
ConverterUtils#fromInternal(RexToLixTranslator, java.lang.Class<?>[], java.util.List<Expression>)

seems you can just rewrite your code like :

```
    private static Expression fromInternal(@Nullable Expression root, Expression operand, Type targetType) {
        if (Types.isAssignableFrom(targetType, operand.getType())
            || Types.isAssignableFrom(targetType, Primitive.box(operand.getType())))
            return operand;

        if (!TypeUtils.isConvertableType(targetType))
            return targetType == BigDecimal.class ? fromInternal(operand, operand.getType(), targetType) :
                convert(operand, operand.getType(), targetType);

        Primitive primitive = Primitive.of(targetType);

        if (Primitive.is(operand.getType()))
            operand = Expressions.box(operand);

        Expression converted = Expressions.call(
            TypeUtils.class,
            "fromInternal",
            root,//translator.getRoot(),
            operand,
            Expressions.constant(targetType)
        );

        return primitive == null
            ? Expressions.convert_(converted, targetType)
            : Expressions.unbox(Expressions.convert_(converted, primitive.boxClass), primitive);
    }
```

probably some assertions check are helpful, it\`s just a prototype
I run this approach through all calcite tests and it`s ok
wdyt ? props: more readable code, less code base

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
tkalkirill: Added a `org.apache.ignite.internal.processors.query.calcite.integration.OperatorsExtensionIntegrationTest#testByteArrayFunctions` that reproduces the issue if this is removed.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/exec/rel/ExecutionTest.java
tkalkirill: Why?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/IgniteFunctionParameter.java
tkalkirill: We need to call `toSql` first: Java types such as `LocalDateTime` initially report `OTHER`. Moving the check earlier changes the parameter type to `ANY` and skips required implicit casts. I verified that passing a `DATE` literal to a `LocalDateTime` parameter then fails with `ClassCastException` in both `UDF` and `UDTF`. I suggest keeping the current implementation.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
tkalkirill: Thanks, I did it a bit differently, and the duplication went away.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
```diff
@@ -230,6 +276,12 @@ public static Expression convert(Expression operand, Type fromType, Type toType)
         if (toType == BigDecimal.class)
             throw new AssertionError("For conversion to decimal, ConverterUtils#convertToDecimal method should be used instead.");
 
+        if (fromType == byte[].class && toType == ByteString.class)
+            return Expressions.call(BuiltInMethod.BYTE_ARRAY_TO_BYTE_STRING.method, operand);
+
+        if (fromType == ByteString.class && toType == byte[].class)
+            return Expressions.call(BuiltInMethod.BYTE_STRING_TO_BYTE_ARRAY.method, operand);
```
zstan, line 283: these conversion still has no test coverage, isn\`t it ?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/IgniteFunctionParameter.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.ignite.internal.processors.query.calcite.type.OtherType;
+
+/**
+ * Reflective Java function parameter represented with a SQL type.
+ *
+ * <p>The SQL representation is required to validate user-defined function arguments and to convert literal arguments
+ * while deriving a table function row type.
+ */
+final class IgniteFunctionParameter implements FunctionParameter {
```
zstan, line 33: seems all tests are passed if you completelly remote this class, am i right ?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
```diff
@@ -123,28 +124,73 @@ else if (targetType == java.sql.Timestamp.class) {
...
+            root,
+            operand,
+            Expressions.constant(targetType)
+        );
+
+        Primitive primitive = Primitive.of(targetType);
+
+        return primitive == null
+            ? Expressions.convert_(converted, targetType)
+            : Expressions.unbox(Expressions.convert_(converted, primitive.boxClass), primitive);
```
zstan, line 191: redundant ?

PR #13544, opened 2026-09-02
IGNITE-29033 SQL Calcite: Support UDF and UDTF Overloading

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-29033

### DIFF
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/schema/IgniteSchema.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/schema/IgniteSchema.java
@@ -24,6 +24,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
+import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.schema.Function;
 import org.apache.calcite.schema.FunctionParameter;
 import org.apache.calcite.schema.SchemaPlus;
@@ -31,6 +32,7 @@ import org.apache.calcite.schema.Table;
 import org.apache.calcite.schema.impl.AbstractSchema;
 import org.apache.calcite.tools.FrameworkConfig;
 import org.apache.ignite.IgniteException;
+import org.apache.ignite.internal.processors.query.calcite.type.IgniteTypeFactory;
 import org.apache.ignite.internal.processors.query.calcite.util.Commons;
 
 /**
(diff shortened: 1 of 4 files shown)


modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/UserDefinedFunctionsIntegrationTest.java
```diff
@@ -94,6 +94,27 @@ public void testSameSignatureNotRegistered() throws Exception {
...
+    public void testOverloadedFunctions() {
+        client.getOrCreateCache(new CacheConfiguration<Integer, Object>("overloaded-functions")
+            .setSqlSchema("UDF")
+            .setSqlFunctionClasses(OverloadedFunctionsLibrary.class));
+
+        SchemaPlus schema = queryProcessor(client).schemaHolder().schema("UDF");
+
+        assertEquals(2, schema.getFunctions("OVERLOADED").size());
+        assertEquals(2, schema.getFunctions("OVERLOADED_TABLE").size());
+        assertEquals(1, schema.getFunctions("SQL_EQUIVALENT").size());
```
zstan, line 108: I wonder - why ? Behavior is different from java - is it ok ?  It can confuse and brings, as minimal, unexpected behavior ? If it calcite related design - plz show me the related links ? And as minimal - i expect it need to be documented somehow ?

tkalkirill: This behavior intentionally differs from Java because overload resolution uses SQL types. Both int and Integer correspond to SQL INTEGER, so Calcite cannot reliably distinguish these overloads and their resolution would depend on registration order.
Therefore, overloads must have different SQL parameter types or parameter order. I’ve also documented this behavior in QuerySqlFunction and QuerySqlTableFunction.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/schema/IgniteSchema.java
```diff
@@ -94,23 +96,13 @@ public void removeTable(String tblName) {
...
-
-            for (int i = 0; i < params.size(); ++i) {
-                FunctionParameter p = params.get(i);
-                FunctionParameter existingP = existingParams.get(i);
+        IgniteTypeFactory typeFactory = Commons.typeFactory();
 
-                if (!p.getType(Commons.typeFactory()).equalsSansFieldNames(existingP.getType(Commons.typeFactory())))
-                    break;
+        for (Function existingFun : getFunctions(name)) {
+            if (sameParameters(func.getParameters(), existingFun.getParameters(), typeFactory)) {
```
zstan, line 102: If you will use non ignoring nullability comparison:
```
            if (!paramType.equalsSansFieldNames(existingParamType))
                return false;
```
Seems you can change behavior as it was in java and reduce further overloading missing errors

zstan (changes requested):

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/schema/IgniteSchema.java
tkalkirill: I tried this change and checked `SqlUtil.lookupRoutine` directly. Both overloads remain candidates for both `INTEGER` and `INTEGER NOT NULL`, and reversing the candidate order changes which overload is selected.
The earlier integration test was misleading: both methods had the same Java name, so compilation of the generated Java code could resolve the overload again. It did not prove that SQL resolution distinguishes nullability.
I reverted the change and kept the comparison ignoring nullability. Treating int and Integer as the same SQL signature avoids ambiguous selection and additional rules users would need to remember. This limitation is documented in both annotations.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/UserDefinedFunctionsIntegrationTest.java
```diff
@@ -94,6 +94,27 @@ public void testSameSignatureNotRegistered() throws Exception {
         assertEquals(1, schema.getFunctions("SAMESIGN").size());
     }
 
+    /** */
+    @Test
+    public void testOverloadedFunctions() {
```
zstan: ok, lets align this test with already existing, i mean :  append LogListener - check plz ?

```
    public void testSameSignatureNotRegistered() throws Exception {
        LogListener logChecker = LogListener.matches("Unable to register function 'SAMESIGN'. Other function " +
            "with the same name and parameters is already registered").build();
```

zstan (approved):

PR #13545, opened 2026-09-02
IGNITE-29034 SQL Calcite: Add support for FROM DUAL queries

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-29034

### DIFF
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/schema/IgniteSchema.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/schema/IgniteSchema.java
@@ -28,6 +28,7 @@ import org.apache.calcite.schema.Function;
 import org.apache.calcite.schema.FunctionParameter;
 import org.apache.calcite.schema.SchemaPlus;
 import org.apache.calcite.schema.Table;
+import org.apache.calcite.schema.TableMacro;
 import org.apache.calcite.schema.impl.AbstractSchema;
 import org.apache.calcite.tools.FrameworkConfig;
 import org.apache.ignite.IgniteException;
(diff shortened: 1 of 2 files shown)


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/schema/IgniteSchema.java
```diff
@@ -143,6 +150,17 @@ public SchemaPlus register(SchemaPlus parent, FrameworkConfig frameworkCfg) {
...
+
         return newSchema;
     }
+
+    /** */
+    private static void registerDualTableIfSupported(SchemaPlus schema, FrameworkConfig frameworkCfg) {
+        if (frameworkCfg != null && frameworkCfg.getParserConfig().conformance().isSupportedDualTable()
+            && schema.tables().get(DUAL_TBL_NAME) == null
+            && schema.getFunctions(DUAL_TBL_NAME).stream().noneMatch(TableMacro.class::isInstance)) {
+            schema.add(DUAL_TBL_NAME, new ViewTableMacroImpl(DUAL_TBL_VIEW, schema, frameworkCfg));
```
vldpyatkov, line 163: I’m not sure whether we really need this, but Calcite supports this property. If we decide to support it, let’s implement it the same way as Calcite does, because its implementation also works with grammars that do not allow SELECT without a FROM clause.
`VALUES ('X')`

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/schema/IgniteSchema.java
```diff
@@ -143,6 +150,17 @@ public SchemaPlus register(SchemaPlus parent, FrameworkConfig frameworkCfg) {
...
         viewMap.forEach((name, sql) -> newSchema.add(name, new ViewTableMacroImpl(sql, newSchema, frameworkCfg)));
 
+        registerDualTableIfSupported(newSchema, frameworkCfg);
+
         return newSchema;
     }
+
+    /** */
+    private static void registerDualTableIfSupported(SchemaPlus schema, FrameworkConfig frameworkCfg) {
+        if (frameworkCfg != null && frameworkCfg.getParserConfig().conformance().isSupportedDualTable()
```
vldpyatkov, line 160: Wouldn't it be more appropriate to use the validator conformance here?

vldpyatkov (approved):

PR #13554, opened 2026-09-04
IGNITE-29032 Improvement of the page eviction mechanism for in-memory

wernerdv:

### Notes


wernerdv:
JmhPageEvictionBenchmark results on my local machine:

master:
```
Benchmark                          (evictionMode)  (scenario)   Mode  Cnt    Score     Error   Units
JmhPageEvictionBenchmark.putLarge      RANDOM_LRU       SMALL  thrpt    5    0,733 ±   0,008  ops/ms
JmhPageEvictionBenchmark.putLarge      RANDOM_LRU       LARGE  thrpt    5    fail with IgniteOutOfMemoryException: Out of memory in data region
JmhPageEvictionBenchmark.putSmall      RANDOM_LRU       SMALL  thrpt    5  938,566 ±  70,609  ops/ms
JmhPageEvictionBenchmark.putSmall      RANDOM_LRU       LARGE  thrpt    5  820,833 ± 145,542  ops/ms
```

branch:
```
Benchmark                          (evictionMode)  (scenario)   Mode  Cnt    Score    Error   Units
JmhPageEvictionBenchmark.putLarge      RANDOM_LRU       SMALL  thrpt    5    0,789 ±  0,035  ops/ms
JmhPageEvictionBenchmark.putLarge      RANDOM_LRU       LARGE  thrpt    5    0,674 ±  0,018  ops/ms
JmhPageEvictionBenchmark.putSmall      RANDOM_LRU       SMALL  thrpt    5  944,594 ± 54,536  ops/ms
JmhPageEvictionBenchmark.putSmall      RANDOM_LRU       LARGE  thrpt    5  858,830 ± 37,303  ops/ms
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -1172,32 +1196,62 @@ public WALPointer latestWalPointerReservedForPreloading() {
...
      * @param region Data region to be checked.
      * @param dataRowSize Size of data row to be inserted.
-     * @throws IgniteOutOfMemoryException In case of the given data region does not have enough free space
-     * for putting a new entry.
+     * @throws IgniteOutOfMemoryException In case the given data region does not have enough free space
+     * for putting a new entry, even after eviction.
+     * @throws IgniteCheckedException If failed to evict data pages.
      */
-    public void ensureFreeSpaceForInsert(DataRegion region, int dataRowSize) throws IgniteOutOfMemoryException {
+    public void ensureFreeSpaceForInsert(DataRegion region, int dataRowSize)
```
alex-plekhanov: Codestyle violation: In case of multi-line method declaration, each parameter should be on it's own line.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -1172,32 +1196,62 @@ public WALPointer latestWalPointerReservedForPreloading() {
...
+    /**
+     * Checks that a non-persistent region with disabled page eviction has enough pages for a new row, taking into
+     * account the pages required to support the free list.
+     *
+     * @param region Data region.
+     * @param regCfg Data region configuration.
+     * @param dataRowSize Size of data row to be inserted.
+     * @throws IgniteOutOfMemoryException If the region does not have enough free space for the new entry.
+     */
+    private void checkOomThreshold(DataRegion region, DataRegionConfiguration regCfg, int dataRowSize)
```
alex-plekhanov: Codestyle violation

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -1216,24 +1270,170 @@ public void ensureFreeSpaceForInsert(DataRegion region, int dataRowSize) throws
...
+     * threshold.
+     *
+     * @param region Data region.
+     * @param regCfg Data region configuration.
+     * @param dataRowSize Size of data row to be inserted.
+     * @throws IgniteOutOfMemoryException If the target cannot be reached (row too large for the region or eviction
+     * makes no progress).
+     * @throws IgniteCheckedException If failed to evict data pages.
+     */
+    private void ensureFreeSpaceForEviction(DataRegion region, DataRegionConfiguration regCfg, int dataRowSize)
```
alex-plekhanov: Codestyle violation

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -1216,24 +1270,170 @@ public void ensureFreeSpaceForInsert(DataRegion region, int dataRowSize) throws
...
+        // eviction into a premature OOM. On each stalled iteration the thread backs off (rather than busy-spinning)
+        // both to save CPU and to let a lock holder run and release it.
+        long bestEmptyPages = emptyPages;
+
+        long lastProgressNanos = System.nanoTime();
+
+        long backoffNanos = EVICTION_BACKOFF_START_NANOS;
+
+        while (bestEmptyPages < requiredPages) {
+            if (region.metrics().onPageEvictionsStarted())
```
alex-plekhanov: Braces required for multi-line statements

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/RowStore.java
```diff
@@ -132,8 +133,25 @@ public void addRow(CacheDataRow row, IoStatisticsHolder statHolder) throws Ignit
...
+            // re-runs the reserve on the row remainder when a fragmented write cannot take a page (a raw OOM there
+            // would otherwise be wrapped by insertDataRows into CorruptedFreeListException and reported as corruption).
+            //
+            // The reserve evicts non-blockingly even though the batch path holds no entry locks (so blocking would be
+            // deadlock-safe and more effective here): the same reserve path is shared with single-row insertion,
+            // which runs under an entry lock and must not block.
+            for (CacheDataRow row : rows) {
+                int rowSize = row.size();
+
+                if (rowSize > 0)
```
alex-plekhanov: How it can be <= 0?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -1216,24 +1270,170 @@ public void ensureFreeSpaceForInsert(DataRegion region, int dataRowSize) throws
...
+        // Pages the row will actually occupy once written, and which the free list must hand out on demand during
+        // the fragmented write.
+        long requiredPages = (dataRowSize + pagePayload - 1) / pagePayload;
 
-            throw oom;
+        // The row fundamentally cannot fit into the whole region.
+        if (requiredPages > totalPages)
+            throw outOfMemory(regCfg);
+
+        // The reserve must guarantee that the free list holds `requiredPages` REAL empty pages, not merely that the
```
alex-plekhanov, line 1324: > Apparent headroom (totalPages - loadedPages) is shared and non-exclusive

But empty pages are also shared and non-exclusive

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -1216,24 +1270,170 @@ public void ensureFreeSpaceForInsert(DataRegion region, int dataRowSize) throws
...
+        // can safely grow the region, so a row that fits into the combined spare space is satisfied without eviction
+        // (which would otherwise destroy evictable, e.g. short-TTL, entries just to accumulate empty pages the slack
+        // could have absorbed).
+        long pagesThreshold = (long)(totalPages * regCfg.getEvictionThreshold());
+
+        boolean underPressure = pageMem.loadedPages() >= pagesThreshold;
+
+        // Fast path: the row is satisfiable without eviction when (a) the free list already holds enough real empty
+        // pages, or (b) the region is not under pressure and has enough spare space to grow into.
+        if (emptyPages >= requiredPages || (!underPressure && emptyPages + headroom >= requiredPages))
```
alex-plekhanov: Eviction always started when pagesThreshold is exceeded and there is not enough empty pages. So page memory will never exceeds pagesThreshold (by default 10% of page memory will be unusable). Looks strange.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -1216,24 +1270,170 @@ public void ensureFreeSpaceForInsert(DataRegion region, int dataRowSize) throws
...
+     * Invokes a single page eviction, acquiring entry locks non-blockingly so that contended entries are skipped.
+     * This is required when eviction runs while the current thread already holds entry locks (size-aware eviction
+     * from a single-row insertion) to avoid a lock-ordering deadlock. {@link NoOpPageEvictionTracker}
+     * (disabled eviction, never reaching this path) falls back to the plain {@code evictDataPage()}.
+     *
+     * @param evictionTracker Page eviction tracker.
+     * @throws IgniteCheckedException If failed to evict a data page.
+     */
+    private void evictDataPageNonBlocking(PageEvictionTracker evictionTracker) throws IgniteCheckedException {
+        if (evictionTracker instanceof PageAbstractEvictionTracker)
```
alex-plekhanov: Why this method not in PageEvictionTracker?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/evict/PageAbstractEvictionTracker.java
```diff
@@ -41,6 +41,13 @@ public abstract class PageAbstractEvictionTracker implements PageEvictionTracker
     /** Millis in day. */
     private static final int DAY = 24 * 60 * 60 * 1000;
 
+    /**
+     * Thread-local marker that the current eviction is requested by size-aware eviction, which may run
+     * while the calling thread already holds entry locks. When set, entries whose locks are contended are skipped
+     * (via a non-blocking {@code evictInternal}) instead of blocking, avoiding a lock-ordering deadlock.
+     */
+    private static final ThreadLocal<Boolean> EVICT_NON_BLOCKING = new ThreadLocal<>();
```
alex-plekhanov: It's strange to use thread-local to pass parameter to next stack level. Let's use
evictDataPage(boolean blocking) -> evictDataPage(int pageIdx, boolean blocking)
Instead of
evictDataPageNonBlocking -> set thread-local -> evictDataPage() -> evictDataPage(int pageIdx) -> get thread local.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/AbstractFreeList.java
```diff
@@ -701,10 +713,29 @@ private int writeWholePages(T row, IoStatisticsHolder statHolder) throws IgniteC
...
      */
     private int writeSinglePage(T row, int written, IoStatisticsHolder statHolder) throws IgniteCheckedException {
+        // TOCTOU closure: the size-aware reserve (ensureFreeSpaceForInsert, invoked from RowStore.addRow/addRows
+        // before this write) accumulates enough real empty pages but does not pin them to this thread - a concurrent
+        // writer can consume them between the reserve and this allocation. When the free list cannot hand out a page,
+        // re-reserve on the remaining size and retry before allocating a brand-new page; otherwise the race surfaces
+        // as a raw IgniteOutOfMemoryException (wrapped into CorruptedFreeListException in the batch path).
+        //
+        // The re-reserve is an inline demand-eviction: reached from the BPlusTree.invoke row-creation closure, it may
+        // re-entrantly remove other entries from the same data tree. That is safe because the closure runs with no
```
alex-plekhanov: > That is safe because the closure runs with no data-tree page locks held.

Are you sure about this statement?
As far as I know BPlusTree.invoke holds the write lock on leaf page while execution the closure, so it's not deadlock safe. Deadlock is possible between data-tree leaf pages since page can contain more than one entry (acquired entry lock is not enough protection). 
Also, as far as I understand there can be deadlock on expiration: Expiration thread holds write lock on pending tree leaf page and reads data tree (with read lock). While addRow thread holds write lock on data tree leaf page and can concurrently evict data (remove rows) which remove ttl entries and requires write lock on pending tree leaf page.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/RowStore.java
```diff
@@ -132,8 +133,25 @@ public void addRow(CacheDataRow row, IoStatisticsHolder statHolder) throws Ignit
      * @param statHolder Statistics holder to track IO operations.
      * @throws IgniteCheckedException If failed.
      */
-    public void addRows(Collection<? extends CacheDataRow> rows,
-        IoStatisticsHolder statHolder) throws IgniteCheckedException {
+    public void addRows(Collection<? extends CacheDataRow> rows, IoStatisticsHolder statHolder) throws IgniteCheckedException {
```
alex-plekhanov, line 136: addRows is only executed on rebalance. Why do we need this new code for rebalance but not for regular put?

> reserving only the largest is insufficient

But you do exactly the same. Ensure free space for each row before any row is inserted, so only free space for largest row will be ensured. Other rows reservation will go through writeSinglePage.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/AbstractFreeList.java
```diff
@@ -701,10 +713,29 @@ private int writeWholePages(T row, IoStatisticsHolder statHolder) throws IgniteC
...
 
+        if (pageId == 0L) {
+            if (dbMgr != null)
+                dbMgr.ensureFreeSpaceForInsert(dataRegion, row.size() - written);
+
+            pageId = takePage(row.size() - written, row, statHolder);
+        }
+
         if (pageId == 0L) {
             pageId = allocateDataPage(row.partition());
```
alex-plekhanov, line 775: There is also insertDataRows method where page can be requested on rows insert and oom can be fired.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/RowStore.java
wernerdv: The check was unnecessary.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
wernerdv: Correct. `emptyDataPages()` is a snapshot of a shared AtomicLong counter; any writer can consume it. 
The real distinction isn't exclusivity but reachability at the capacity limit: at loadedPages == totalPages headroom can't grow the region (fresh allocateDataPage → raw OOM) whereas reuse-bucket empty pages remain reachable via takePage(). 
Correctness rests on the lazy re-reserve (now AbstractFreeList#takePageWithReserve), not on the reserve being exclusive.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
wernerdv: Not a bug. 
The gate reuses evictionThreshold only as a regime boundary, not as "when to start eviction" (that's evictionRequired(), which stops on emptyPages >= poolSize). (1) loadedPages() counts every allocated page including empty reuse pages, so on a filling region it routinely exceeds the threshold — no last 10% is unusable; (2) below the threshold headroom is trusted and the region grows, at/above it only real empty pages are counted. 
Renamed to evictionRegime; no behavior change.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
wernerdv: After the latest changes, the method is no longer needed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/evict/PageAbstractEvictionTracker.java
wernerdv: Agreed.
Moved the capability into the interface as evictDataPage(boolean tryLock) (default evictDataPage() delegates to false), removed the ThreadLocal.
As a follow-up the method now returns boolean so the size-aware progress guard can count an actual eviction as progress. Parameter named tryLock (matching evictInternal(..., tryLock) / lockEntry(boolean tryLock)) rather than blocking.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/AbstractFreeList.java
wernerdv: Partially correct. 
The premise — "BPlusTree.invoke holds the leaf write lock while running the closure" — is not supported by the code: BPlusTree.invokeDown releases the leaf read lock when read(...) returns and takes the leaf write lock only afterwards (in tryInsert/tryReplace/tryRemoveFromLeaf). Your broader point stands: entry tryLock covers only entry-level ordering and does not make eviction page-level lock-free. 
The cross-tree (data→pending vs pending→data) residual risk with the TTL worker is real and documented in the code comment; a full fix is out of scope.
Please correct me if I'm wrong.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/RowStore.java
wernerdv: Two points correct, one not. 
(1) Rebalance-only: correct — regular puts use addRow (per-put reserve), both paths covered.
(2) Per-row loop == reserve for the largest: correct — a single reserve for the max row is equivalent.
(3) "Other rows go through writeSinglePage" — actually the batch path uses insertDataRows, whose per-page protection now comes from the shared takePageWithReserve re-reserve; the pre-reserve bounds the empty-pages counter up front.
Updated comment.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/AbstractFreeList.java
wernerdv: Valid.
insertDataRows had the same allocateDataPage OOM-firing fallback without the lazy re-reserve, and since IgniteOutOfMemoryException is a RuntimeException, both methods mislabelled it as CorruptedFreeListException.
Made the batch trailing-fragment path symmetric via the shared takePageWithReserve, and both catch clauses rethrow OOM as-is so it is reported as OOM/critical failure, not corruption.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -4182,6 +4199,23 @@ private int extrasSize() {
...
+     * Acquires the entry lock either blocking ({@code tryLock == false}) or non-blockingly with an immediate
+     * {@code tryLock(0)} ({@code tryLock == true}). Used by {@link #evictInternal} to let size-aware
+     * eviction skip contended entries instead of blocking, avoiding a lock-ordering deadlock.
+     *
+     * @param tryLock {@code true} to acquire the lock non-blockingly.
+     * @return {@code true} if the lock was acquired (always {@code true} when {@code tryLock == false}).
+     */
+    private boolean lockEntry(boolean tryLock) {
+        if (tryLock)
+            return !lock.isHeldByCurrentThread() && tryLockEntry(0);
```
alex-plekhanov, line 4212: As far as I understand  we forbid lock reentry for tryLock to avoid self-eviction (not to avoid deadlocks). This behavior looks strange in general lockEntry method, someone can reuse this method and get unexpected results, when instead of reentry lock fails. Let's move `f (tryLock && lock.isHeldByCurrentThread()) return false; ` to the evictInternal method, with comment, why we forbid reentry.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/AbstractFreeList.java
```diff
@@ -591,6 +610,44 @@ private long allocateDataPage(int part) throws IgniteCheckedException {
         return pageMem.allocatePage(grpId, part, FLAG_DATA);
     }
 
+    /**
+     * @return {@code true} when the region has effectively no headroom left (allocated pages reached the configured
+     * max), so a fresh {@code allocateDataPage} could no longer grow it.
+     */
+    private boolean regionEffectivelyFull() {
+        return pageMem.loadedPages() >= dataRegion.config().getMaxSize() / pageMem.systemPageSize();
```
alex-plekhanov, line 618: In real case scenarios this condition almost never reachable. All page memory divided to segments (up to 16), most probably there will be some unusable space at the end of each segment, so total pages that can be allocated inside each segment is less than maxSize/pageSize.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/AbstractFreeList.java
```diff
@@ -591,6 +610,44 @@ private long allocateDataPage(int part) throws IgniteCheckedException {
...
+     * @param size Free space required on the page.
+     * @param row Row to write.
+     * @param statHolder Statistics holder to track IO operations.
+     * @return Page identifier or 0 if no page could be obtained after re-reserving.
+     * @throws IgniteCheckedException If failed.
+     */
+    private long takePageWithReserve(int size, T row, IoStatisticsHolder statHolder) throws IgniteCheckedException {
+        long pageId = takePage(size, row, statHolder);
+
+        if (pageId == 0L && dbMgr != null) {
```
alex-plekhanov, line 638: `&& !dataRegion.config().isPersistenceEnabled()` 
We can't evict pages for persistent region, so 1-4 cycles with ensureFreeSpace and takePage are redundant.
The same for disabled eviction. takePage is not free, there is a cycle over buckets inside. It Worth to check preconditions before next attempt to takePage. For disabled eviction checking oom threshold. For disabled eviction check oom threshold  on each page is also redundant. We still will fail if there is an error during allocation, so it's no matter if it happens sooner or later.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/AbstractFreeList.java
```diff
@@ -607,6 +664,9 @@ private long allocateDataPage(int part) throws IgniteCheckedException {
         catch (IgniteCheckedException | Error e) {
             throw e;
         }
+        catch (IgniteOutOfMemoryException e) {
```
alex-plekhanov, line 667: catch (IgniteCheckedException | Error | IgniteOutOfMemoryException e)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/RowStore.java
alex-plekhanov: But why can't we prereserve the whole size and not rely to per-page reservation? What the profit of max entry size reservation?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/AbstractFreeList.java
alex-plekhanov: Looks like you are right about the cache data tree lock. But if lock is not held during invokeClosure, then pendingTree lock is also can't lead to deadlocks, so new comment is not correct. New eviction mechanism  removes entries from cache data tree (with write lock-unlock), and after that removes pending tree entry (under write lock-unlock), and removes data entry (under write lock-unlock). If invokeClosure is not holds the lock, there are no nested locks and TTL path is safe too.

alex-plekhanov: Also statement `removes entries with no data-tree page locks held` can be read as `locks held by removing entries`, but here it means `locks held by row-creation closure`, please rephrase.

alex-plekhanov: Also statement `removes entries with no data-tree page locks held` can be read as `locks held by removing entries`, but here it means `locks held by row-creation closure`, please rephrase.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/AbstractFreeList.java
```diff
@@ -693,6 +756,15 @@ private int writeWholePages(T row, IoStatisticsHolder statHolder) throws IgniteC
 
     /**
      * Take a page and write row on it.
+     * <p>
+     * The page is acquired via {@link #takePageWithReserve}: the size-aware reserve (RowStore.addRow/addRows) only
```
alex-plekhanov, line 760: Comment partially intersects with takePageWithReserve javadoc

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -1213,27 +1267,159 @@ public void ensureFreeSpaceForInsert(DataRegion region, int dataRowSize) throws
...
+
+        // Maximum payload bytes that a single data page can hold for a fragmented row.
+        long pagePayload = pageMem.pageSize() - AbstractDataPageIO.MIN_DATA_PAGE_OVERHEAD;
 
-            if (cctx.kernalContext() != null)
-                cctx.kernalContext().failure().process(new FailureContext(FailureType.CRITICAL_ERROR, oom));
+        // A row that fits into the steady-state empty-pages pool is satisfied by normal threshold eviction, so the
+        // fast path is a single comparison (no page computation, free-list lookup or page-memory reads on the hot
+        // small-put path).
+        if (dataRowSize <= regCfg.getEmptyPagesPoolSize() * pagePayload)
```
alex-plekhanov, line 1305: What if there are two concurrent puts with 60 pages each and only 100 pages left?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -1213,27 +1267,159 @@ public void ensureFreeSpaceForInsert(DataRegion region, int dataRowSize) throws
...
+        // The reserve must guarantee `requiredPages` REAL empty pages, not just apparent headroom. Both are shared and
+        // non-exclusive (emptyDataPages() is a snapshot; any writer can consume them), but once the region is full
+        // (loadedPages == totalPages) headroom can no longer grow it (fresh allocateDataPage -> raw OOM), while empty
+        // pages in the reuse bucket stay reachable via takePage(). So empty pages are the only resource the fragmented
+        // write can consume on a full region. The TOCTOU between this reserve and the actual write is closed by the
+        // lazy re-reserve in AbstractFreeList#writeSinglePage.
+        long emptyPages = freeList.emptyDataPages();
+
+        // The gate reuses evictionThreshold as a regime boundary, not as "when to start eviction" (evictionRequired()
+        // does that, stopping on emptyPages >= poolSize; no last 10% of page memory is left unusable). Below the
```
alex-plekhanov, line 1332: Hardcoded 10%

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -1213,27 +1267,159 @@ public void ensureFreeSpaceForInsert(DataRegion region, int dataRowSize) throws
...
+        // threshold the region has real slack, so a row fitting into the combined spare space is satisfied without
+        // eviction (live, e.g. short-TTL, entries are not evicted just to accumulate empty pages). At/above it headroom
+        // is no longer trustworthy (concurrent writers could commit the same headroom - TOCTOU), so only real empty
+        // pages are counted and eviction is driven below.
+        boolean evictionRegime = pageMem.loadedPages() >= (long)(totalPages * regCfg.getEvictionThreshold());
```
alex-plekhanov, line 1337: Still don't understand how it works.
Suppose we have 10Gb total page memory, 1Gb not allocated and no empty pages. We are trying to insert a 500Mb row. Here we have evictionRegime = true, in this case we only check emptyPages >= requiredPages (false) before eviction and stop to evict only when emptyPages >= requiredPages. So 500Mb will be evicted and we came to the same state: 1Gb is not allocated and no empty pages. How can we reuse top 10% of page memory in this case?
Perhaps we can use regCfg.getEvictionThreshold as fast path (never use eviction if we are below this point), but not as "rely only on empty pages after this point" flag.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionConcurrentWritesAbstractTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+                for (StackTraceElement frame : t.getStackTrace())
+                    log.error("  at " + frame);
+            }
+        }
+
+        for (Thread t : threads)
+            assertFalse("Writer thread " + t.getName() + " did not finish (possible deadlock)", t.isAlive());
+
+        assertEquals("Writer threads reported errors, reason: " + firstErr.get(), 0, errors.get());
```
alex-plekhanov, line 174: Why all this can't be replaced with:
```
        IgniteInternalFuture<?> fut = GridTestUtils.runMultiThreadedAsync(() -> {
                U.awaitQuiet(startLatch);

                for (int k = 0; k < LARGE_ROWS_PER_THREAD; k++)
                    cache.put(..., largeVal);
            },
            THREADS, "paged-writer");
        startLatch.countDown();
        fut.get(DEADLINE);
```
?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionConcurrentWritesAbstractTest.java
```diff
@@ -0,0 +1,176 @@
...
+    /** Empty pages pool size. */
+    private static final int POOL_SIZE = 100;
+
+    /** Number of small pre-fill entries, leaving a buffer that is exceeded by the total of the large writes, so that
+     * the last of them can only be stored by freeing pages via size-aware eviction. The large records are small
+     * enough that concurrent size-aware eviction reliably frees the required pages (no spurious guard OOM). */
+    private static final int SMALL_ENTRIES = 48_000;
+
+    /** Number of writer threads. */
+    private static final int THREADS = 2;
```
alex-plekhanov, line 68: It's not too concurrent, maybe increase threads count to increase probability of catching concurrent problems.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionGuardOomTest.java
```diff
@@ -0,0 +1,176 @@
...
+            keys.add(i);
+
+        CountDownLatch ready = new CountDownLatch(1);
+
+        CountDownLatch release = new CountDownLatch(1);
+
+        AtomicReference<Throwable> lockerErr = new AtomicReference<>();
+
+        // Hold entry locks on every resident key from a background thread so that eviction has no evictable page.
+        Thread locker = new Thread(() -> {
```
alex-plekhanov: Can be simplified using GridTestUtils.runAsync (we don't need lockerErr in this case, and thread managing)

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionSizeAwareAbstractTest.java
```diff
@@ -0,0 +1,264 @@
...
+
+    /** {@inheritDoc} */
+    @Override protected IgniteConfiguration getConfiguration(String gridName) throws Exception {
+        return super.getConfiguration(gridName)
+            .setDataStorageConfiguration(new DataStorageConfiguration()
+                .setDefaultDataRegionConfiguration(new DataRegionConfiguration()
+                    .setInitialSize(SIZE)
+                    .setMaxSize(SIZE)
+                    .setEmptyPagesPoolSize(POOL_SIZE))
+                .setPageSize(DFLT_PAGE_SIZE));
```
alex-plekhanov: 4 tests duplicate the same block, redefining these properties, some of them are default values. Maybe it's worth to move these blocks to some abstract class? Why do we need to redefine properties to default value?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionWithExpiryPolicyAbstractTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+        // Short-TTL entries keep the TTL worker actively freeing pages while eviction runs.
+        IgniteCache<Integer, Object> cache = createCache(ignite, DEFAULT_CACHE_NAME, TTL);
+
+        Object val = new byte[RECORD_SIZE];
+
+        // Writing more data than the region can hold forces eviction; concurrent expiry of short-TTL entries must not
+        // deadlock with it. The test itself is protected against a hang by the framework test timeout.
+        for (int i = 0; i < 30; i++)
+            cache.put(i, val);
```
alex-plekhanov, line 72: There can't be concurrent expire with 8 seconds TTL. Also chance to face with concurrency problem is miserable with only 30 entries.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionWithExpiryPolicyAbstractTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+        // Verify the TTL entries are present before expiry.
+        assertNotNull("TTL entry must be present before expiry", ttlCache.get(0));
+
+        // Wait for the TTL worker to expire and free the short-TTL entries. Polled instead of a fixed sleep so that a
+        // slow CI machine does not proceed before the entries have actually expired.
+        long expiryDeadline = System.currentTimeMillis() + 30_000;
+
+        while (ttlCache.get(0) != null && System.currentTimeMillis() < expiryDeadline)
+            U.sleep(200);
```
alex-plekhanov: We have waitForCondition, please use it. 8 seconds for til and 30 seconds for wait looks too time consuming

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionWithExpiryPolicyAbstractTest.java
```diff
@@ -0,0 +1,199 @@
...
+    private static final int SMALL_RECORD_SIZE = 1024;
+
+    /** Number of non-expiring pre-fill records. Deliberately conservative: the pre-fill keeps a comfortable margin to
+     * the region capacity so that the large short-TTL records are guaranteed not to be evicted before they expire. */
+    private static final int SMALL_ENTRIES = 6_000;
+
+    /** Fresh large records written after TTL expiry. Together with the pre-fill ({@link #SMALL_ENTRIES} small entries,
+     * each occupying roughly one data page) plus the index structures, they exceed the region capacity (~32768 pages
+     * for a 128 MiB region with 4 KiB pages), forcing size-aware eviction that accounts for the TTL-freed space. */
+    private static final int FRESH_RECORDS = 12;
```
alex-plekhanov: The test class contains two tests, but these global constants used only by one. Maybe it's worth to move it to local variables, since they are related only to one test

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionWithExpiryPolicyAbstractTest.java
```diff
@@ -0,0 +1,199 @@
...
+        // slow CI machine does not proceed before the entries have actually expired.
+        long expiryDeadline = System.currentTimeMillis() + 30_000;
+
+        while (ttlCache.get(0) != null && System.currentTimeMillis() < expiryDeadline)
+            U.sleep(200);
+
+        // Verify the TTL entries have expired.
+        assertNull("TTL entry must be expired", ttlCache.get(0));
+
+        // Now write large records to the non-expiring cache, totalling more than the space freed by TTL (the prefill
```
alex-plekhanov: > totalling more than the space freed by TTL 

Inserted 12 records * 1 Kb (12 Kb total), expired 3 records * 32Mb (96 Mb total)

Moreover entries in plain cache are replaced (keys used by preloaded entries 0-6000, keys for fresh inserted entries 100 - 111). 

Please validate comments after AI agents or decrease amount of such comments, since comments are also should be validated by reviewer, and sometimes they don't add any value, but add work to do.

Constants are used for all values, but 100 is hardcoded (and looks like hardcoded with incorrect value)

IMO the more relevant test scenario: after expiring, put the same large values to plain cache and check that preloaded SMALL_ENTRIES are not evicted.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionSizeAwareAbstractTest.java
```diff
@@ -0,0 +1,264 @@
...
+    private IgniteCache<Integer, Object> createCache(IgniteEx ignite) {
+        return ignite.createCache(new CacheConfiguration<Integer, Object>(DEFAULT_CACHE_NAME)
+            .setAffinity(new RendezvousAffinityFunction(false, PARTITIONS)));
+    }
+
+    /**
+     * A large record (larger than the empty-pages pool) must be stored without OOM when there is evictable data,
+     * by evicting previously stored records to free enough space.
+     * <p>
+     * This scenario intentionally overlaps with
```
alex-plekhanov: Additional coverage is gained by other tests, why do we need to duplicate this test?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionSizeAwareAbstractTest.java
```diff
@@ -0,0 +1,264 @@
...
+            cache.put(key, val);
+
+        // Eviction must have bounded the number of resident entries.
+        assertTrue("Expected some entries to be evicted, but cache.size()=" + cache.size(),
+            cache.size() > 0 && cache.size() < ENTRIES);
+    }
+
+    /**
+     * A large record written must be readable right away (the just-written entry is the most recently used and is not
+     * a candidate for eviction before the write completes).
```
alex-plekhanov: But there is no eviction at all in this test.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionSizeAwareAbstractTest.java
```diff
@@ -0,0 +1,264 @@
...
+        assertNotNull("Updated large value must be readable", read);
+
+        assertTrue("Updated value must equal the stored value", Arrays.equals(big, read));
+    }
+
+    /**
+     * @param t Throwable.
+     * @return {@code True} if {@code t} or any of its causes is an out-of-memory.
+     */
+    static boolean isOutOfMemory(Throwable t) {
```
alex-plekhanov: X.hasCause()

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionSizeAwareAbstractTest.java
```diff
@@ -0,0 +1,264 @@
...
+        byte[] read = (byte[])cache.get(1);
+
+        assertNotNull("Large value must be readable after put", read);
+
+        assertTrue("Value read back must equal the stored value", Arrays.equals(val, read));
+    }
+
+    /**
+     * A record larger than the whole region must fail (not hang) even when size-aware eviction is enabled.
+     * Uses a transactional cache so that the size-aware OOM propagates directly (in an atomic cache it is wrapped in
```
alex-plekhanov: Use `X.hasCause(e, IgniteOutOfMemoryException.class)` and test will pass for atomic cache too.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionSizeAwareAbstractTest.java
```diff
@@ -0,0 +1,264 @@
...
+        IgniteEx ignite = startGrid(1);
+
+        IgniteCache<Integer, Object> cache = ignite.createCache(new CacheConfiguration<Integer, Object>(DEFAULT_CACHE_NAME)
+            .setAffinity(new RendezvousAffinityFunction(false, PARTITIONS))
+            .setAtomicityMode(CacheAtomicityMode.TRANSACTIONAL));
+
+        boolean rejected = false;
+
+        try {
+            cache.put(1, new byte[SIZE * 2]);
```
alex-plekhanov: Lets also check putAll with 4 rows of RECORD_SIZE

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionSizeAwareAbstractTest.java
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * Small pre-fill entries count. Chosen to fill the 128 MiB region close to capacity so that less than one large
+     * record ({@link #RECORD_SIZE}) remains available, forcing {@code ensureFreeSpaceForInsert} to actually evict
+     * prefilled entries rather than taking its fast path.
+     */
+    private static final int SMALL_ENTRIES = 28_000;
+
+    /** Large rows written via putAll. */
+    private static final int PUT_ALL_LARGE_ROWS = 3;
```
alex-plekhanov: Used only by 1 test

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
alex-plekhanov: Any test for this behavior (that we can grow beyond EvictionThreshold)?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/RowStore.java
wernerdv: The primary advantage is that only the minimum required for the largest single row is evicted, not the entire batch upfront.
The reserve is non-exclusive (shared emptyDataPages counter), so even sum-reserve does not guarantee pages for this batch — concurrent writers can consume them, and the re-reserve in takePageWithReserve is still needed. WriteRowsHandler packs multiple small rows into one page, so sum(rowSizes) / pagePayload is a pessimistic upper bound: live entries are evicted for pages that packing would never use.
Max-row reserve evicts only what the largest row requires; the remainder is picked up per-page by the re-reserve in insertDataRows → takePageWithReserve — exactly at the point of need.
Under contention both approaches degrade to the same per-page re-reserve.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
wernerdv: With emptyPagesPoolSize=100, both puts hit the fast path (60 × pagePayload ≤ 100 × pagePayload → return without eviction). The first put consumes 60 of the 100 available pages. The second put finds only 40 left → takePage returns 0L → takePageWithReserve re-runs ensureFreeSpaceForInsert for the remaining 60 pages. If eviction can free 20 pages (evictable entries exist), the put succeeds. If not (region full, all candidates locked), ensureFreeSpaceForEviction throws OOM after the no-progress timeout.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
wernerdv: The eviction in your scenario (10GB total, 1GB headroom, 500MB row) is indeed excessive - 500MB of live entries are evicted even though 1GB of headroom is available.
Part 1 of your suggestion (never use eviction below threshold) is already implemented: the !evictionRegime branch in the fast path (line 1305) trusts headroom below the threshold.
Part 2 (still consider headroom above threshold) is unsafe under contention: when headroom is trusted in the reserve but later exhausted by a concurrent writer, the re-reserve runs inside writeSinglePage → BPlusTree.invoke under an entry lock. Eviction with tryLock=true skips locked entries, and if all eviction candidates are locked by concurrent writers, eviction makes no progress → OOM (RandomLruPageEvictionConcurrentWritesTest fails).

Two alternatives were tried:
1) removing the !evictionRegime gate;
2) partial eviction (toEvict = requiredPages - emptyPages - headroom) + RE_RESERVE_ATTEMPTS=4.

Both fail with the same OOM.

The excessive eviction is the intentional cost of correctness under contention. Eliminating it requires moving the size-aware reserve before lockEntry(). But because this requires quite a few changes, I wanted to know your opinion first - is it worth doing?

About the test: added `testLargeRowAboveThresholdEvictsAndSucceeds` verifies the above-threshold path — the region is pre-filled near capacity, a 32 MiB row triggers the size-aware eviction loop, isEvictionsStarted() becomes true, and the write succeeds.
The region does not grow beyond the threshold in practice: evictionRequired() in insertDataRows triggers normal threshold eviction when loadedPages crosses the threshold, holding the region at the boundary.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/AbstractFreeList.java
```diff
@@ -615,7 +616,15 @@ private long allocateDataPage(int part) throws IgniteCheckedException {
      * max), so a fresh {@code allocateDataPage} could no longer grow it.
      */
     private boolean regionEffectivelyFull() {
-        return pageMem.loadedPages() >= dataRegion.config().getMaxSize() / pageMem.systemPageSize();
+        long maxPages = dataRegion.config().getMaxSize() / pageMem.systemPageSize();
+
+        // Each of up to 16 segments loses up to one page to allocation overhead (lastAllocatedIdxPtr + alignment),
+        // so the theoretical max is never reached in practice. Subtract the worst-case segment loss to get an
+        // effective limit that reflects real product scenarios.
+        if (maxPages > 16)
```
alex-plekhanov: Let's use SEG_CNT instead of hardcoded value

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
alex-plekhanov: > The second put finds only 40 left → takePage returns 0L → takePageWithReserve re-runs ensureFreeSpaceForInsert for the remaining 60 pages. 

But ensureFreeSpaceForInsert for remaining 60 don't even try to evict something, since it's quited on this condition. Did I miss something?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
alex-plekhanov: For me it's still no clear why there is a difference between empty pages and headroom in therms of safety under contention. Both are consumed concurrently and looks like only reason why concurrent tests are passed, is because 10% of memory remain unused. Maybe real reason for failures is something else (like condition above with `getEmptyPagesPoolSize` check).

About test: I mean it should be checked that we can consume, for example, 95% of memory.
Naive check shows that we can't:
```
        for (int i = 0; i < SMALL_ENTRIES * 2; i++)
            cache.put(i, small);

        DataRegion dr = ignite.context().cache().context().database().dataRegion(null);

        System.out.println(dr.pageMemory().loadedPages()); // 29320
        System.out.println(SIZE / dr.pageMemory().systemPageSize()); // 32577
```

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionConcurrentWritesAbstractTest.java
```diff
@@ -19,86 +19,28 @@
...
     /** {@inheritDoc} */
     @Override protected IgniteConfiguration getConfiguration(String gridName) throws Exception {
-        return super.getConfiguration(gridName)
-            .setDataStorageConfiguration(new DataStorageConfiguration()
-                .setDefaultDataRegionConfiguration(new DataRegionConfiguration()
-                    .setInitialSize(SIZE)
-                    .setMaxSize(SIZE)
-                    .setEmptyPagesPoolSize(POOL_SIZE))
-                .setPageSize(DFLT_PAGE_SIZE));
+        return super.getConfiguration(gridName).setDataStorageConfiguration(new DataStorageConfiguration());
```
alex-plekhanov: Now default data region size is used, eviction not even started.

Let's add `assertTrue(metrics.isEvictionsStarted());` to ensure that we check something in this test.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/paged/PageEvictionWithExpiryPolicyAbstractTest.java
```diff
@@ -115,85 +64,54 @@ private IgniteCache<Integer, Object> createCache(IgniteEx ignite, String cacheNa
     public void testLargePutWithExpiryNoDeadlock() throws Exception {
```
alex-plekhanov: We can't check if there is no deadlocks if there is only one thread and no concurrent eviction

PR #13562, opened 2026-09-08
IGNITE-29043 SQL Calcite: Support recursive CTEs without the RECURSIVE keyword

vldpyatkov:
https://issues.apache.org/jira/browse/IGNITE-29043

### DIFF
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/RecursiveCteRewriter.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/RecursiveCteRewriter.java
@@ -32,14 +32,14 @@ import org.apache.calcite.sql.SqlWith;
 import org.apache.calcite.sql.SqlWithItem;
 
 /** Infers an omitted RECURSIVE keyword before the validator registers WITH scopes. */
-final class RecursiveCteRewriter {
+class RecursiveCteRewriter {
     /** FROM operators whose first operand is a table reference. */
     private static final Set<SqlKind> FROM_WRAPPERS = EnumSet.of(
         SqlKind.AS, SqlKind.TABLE_REF, SqlKind.EXTEND, SqlKind.SNAPSHOT, SqlKind.TABLESAMPLE,
         SqlKind.LATERAL, SqlKind.PIVOT, SqlKind.UNPIVOT, SqlKind.MATCH_RECOGNIZE
     );
 
-    /** Utility class. */
+    /** */
     private RecursiveCteRewriter() {
         // No-op.
     }
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/RecursiveCteRewriter.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/RecursiveCteRewriter.java
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.processors.query.calcite.prepare;
+
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Set;
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlIdentifier;
+import org.apache.calcite.sql.SqlJoin;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlLiteral;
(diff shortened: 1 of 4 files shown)


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/RecursiveCteRewriter.java
```diff
@@ -0,0 +1,118 @@
...
+/** Infers an omitted RECURSIVE keyword before the validator registers WITH scopes. */
+final class RecursiveCteRewriter {
+    /** FROM operators whose first operand is a table reference. */
+    private static final Set<SqlKind> FROM_WRAPPERS = EnumSet.of(
+        SqlKind.AS, SqlKind.TABLE_REF, SqlKind.EXTEND, SqlKind.SNAPSHOT, SqlKind.TABLESAMPLE,
+        SqlKind.LATERAL, SqlKind.PIVOT, SqlKind.UNPIVOT, SqlKind.MATCH_RECOGNIZE
+    );
+
+    /** Utility class. */
+    private RecursiveCteRewriter() {
```
tkalkirill, line 43: Please get rid of it, we don’t do that in the project.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/RecursiveCteRewriter.java
```diff
@@ -0,0 +1,118 @@
...
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlLiteral;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlNodeList;
+import org.apache.calcite.sql.SqlSelect;
+import org.apache.calcite.sql.SqlWith;
+import org.apache.calcite.sql.SqlWithItem;
+
+/** Infers an omitted RECURSIVE keyword before the validator registers WITH scopes. */
+final class RecursiveCteRewriter {
```
tkalkirill:
```suggestion
class RecursiveCteRewriter {
```

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/RecursiveCteIntegrationTest.java
```diff
@@ -31,6 +31,32 @@ public class RecursiveCteIntegrationTest extends AbstractBasicIntegrationTest {
...
+
+            assertQuery("SELECT * FROM (WITH " + keyword + "\"Numbers\"(n) AS (" +
+                "SELECT 1 UNION ALL SELECT x.n + 1 FROM \"Numbers\" x WHERE x.n < 3) " +
+                "SELECT * FROM \"Numbers\")")
+                .returns(1)
+                .returns(2)
+                .returns(3)
+                .check();
+
+            assertThrows("WITH " + keyword + "numbers(n) AS (" +
```
tkalkirill, line 54: Please use `assertThrowsSqlException`.

tkalkirill (approved):

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/RecursiveCteRewriter.java
vldpyatkov: Done.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/RecursiveCteRewriter.java
vldpyatkov: The private constructor prevents creating an instance of the utility class. This approach is used in our project (for example org.apache.ignite.internal.processors.query.calcite.util.Commons)

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/RecursiveCteIntegrationTest.java
vldpyatkov: assertThrowsSqlException fails here because Calcite wraps the _IgniteSQLException_ in a _RuntimeException_.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/RecursiveCteRewriter.java
tkalkirill: I will ask for statistics to be gathered on how many utility classes have private constructors and how many do not.

Come on, who would try to create a utility class? That’s absurd—and trying to guard against it is, too; something like that would be spotted instantly during a code review.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/RecursiveCteIntegrationTest.java
tkalkirill: This method is a wrapper that does the same thing yours does; it's better to use it.

vldpyatkov: I tried assertThrowsSqlException, but the test fails with it. It calls sql() through the internal query API and expects an IgniteSQLException directly, whereas Calcite wraps it in a RuntimeException here. The current assertion checks the cause chain, so the two are not equivalent in this case.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/prepare/RecursiveCteRewriter.java
vldpyatkov: This pattern is already used in the project and has a clear purpose. I’d prefer adding private constructors where they are missing rather than removing this one.

PR #13566, opened 2026-09-09
Fix #13549: treat metadata result sets without a statement as open

### LINKED ISSUE #13549: SQLLine 1.12.0: !tables no longer works due to JDBCThinResultSet#isClosed() reporting metadata ResultSet as closed
After the upgrade to **SQLLine 1.12.0** introduced by commit 25ab91529c9f0b327ad677af90cdfa4e5a0c17e8 @zstan , SQLLine commands that rely on JDBC metadata result sets, such as !tables, no longer work correctly with the Ignite thin JDBC driver.

### Root Cause
Starting with SQLLine 1.12.0, SQLLine performs an explicit `ResultSet.isClosed()` check before reading from the result set.
For metadata queries, Ignite creates a JdbcThinResultSet with a null statement reference (stmt == null).
However, `JdbcThinResultSet.isClosed()` currently returns:
```
@Override public boolean isClosed() throws SQLException {
    return closed || stmt == null || stmt.connection().isClosed();
}
```
As a result, any metadata ResultSet created without an associated statement is immediately reported as closed, even though it is perfectly usable.

Consequently, SQLLine 1.12.0 refuses to process the metadata result set and commands such as: `!tables` returns empty result

and this is the change in sqlline that introduced the check on the ResultSet isClosed
https://github.com/julianhyde/sqlline/commit/e0f9e9a1eb714a05988ae1323ff5e3bd463f64fb

Adarsh-Me:
Fixes #13549

### DIFF
--- a/modules/clients/src/test/java/org/apache/ignite/jdbc/thin/JdbcThinResultSetSelfTest.java
+++ b/modules/clients/src/test/java/org/apache/ignite/jdbc/thin/JdbcThinResultSetSelfTest.java
@@ -26,6 +26,7 @@ import java.net.URL;
 import java.sql.Blob;
 import java.sql.Clob;
 import java.sql.Connection;
+import java.sql.DatabaseMetaData;
 import java.sql.Date;
 import java.sql.DriverManager;
 import java.sql.NClob;
@@ -1880,6 +1881,31 @@ public class JdbcThinResultSetSelfTest extends JdbcThinAbstractSelfTest {
                 rs.getRow();
             }
         });
+    }
+
+    /**
+     * Tests that metadata result sets, which are created without an associated statement (stmt == null),
+     * report a usable (not closed) state, in line with {@link ResultSet#isClosed()} semantics.
+     *
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testMetadataResultSetIsClosed() throws Exception {
+        DatabaseMetaData meta = stmt.getConnection().getMetaData();
+
+        ResultSet rs = meta.getTables(null, null, "%", null);
+
+        // Metadata result sets are created without a statement and must be usable, not reported as closed.
+        assertFalse("Metadata result set must not be reported as closed", rs.isClosed());
+
+        // The result set must be iterable/usable.
+        while (rs.next())
+            rs.getString("TABLE_NAME");
+
+        // Explicitly closing the metadata result set must mark it as closed.
+        rs.close();
+
+        assertTrue("Explicitly closed metadata result set must be reported as closed", rs.isClosed());
     }
 
     /**
--- a/modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinResultSet.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinResultSet.java
@@ -1487,7 +1487,7 @@ public class JdbcThinResultSet implements ResultSet {
 
     /** {@inheritDoc} */
     @Override public boolean isClosed() throws SQLException {
-        return closed || stmt == null || stmt.connection().isClosed();
+        return closed || (stmt != null && stmt.connection().isClosed());
     }
 
     /** {@inheritDoc} */


Adarsh-Me:
Gentle ping — this is a one-behaviour fix for #13549 (`!tables` breaks in SQLLine 1.12.0 because
`JDBCThinResultSet#isClosed()` reports a metadata result set as closed when no statement ever opened
it). There are no outstanding comments and it has sat unreviewed since it was opened on Sep 09.

On CI, to be accurate rather than optimistic: the only job that has reported against the current head
is "Rolling Upgrade check", which passed. Nothing else has reported at all, so I would not describe the
rest as green — just silent.

If this needs a named reviewer or a dev-list thread before anyone can look at it, say which and I will
get it there.

Adarsh-Me:
@zstan @wernerdv could one of you take a look? You are the two most recent people working in the thin-client area (`zstan` assigned #13549 and asked for this PR, and merged #13578; `wernerdv` reviewed and approved it), and this has been sitting without a review since Sep 09.

The description now carries the full analysis. Three things worth pulling out:

1. **The behaviour change is two lines.** `isClosed()` stops reporting a metadata result set (`stmt == null`, which is every result set `JdbcThinDatabaseMetadata` builds) as closed, which is what unbreaks `!tables` in SQLLine 1.12.0.

2. **The first commit here was unsafe and I have since fixed it.** `close0()` was using that same `isClosed()` as its guard, so once a metadata result set reported itself open it fell through to `stmt.conn.sendRequest(...)` with `stmt == null` — `rs.close()` threw NPE, on the very result set this PR's own test closes. The second commit rewrites the predicate to `stmt != null && !stmt.isCancelled() && (…)`, which is truth-functionally identical wherever a statement exists. If you only look at the diff, `5bb60777` is the commit worth reading.

3. **I have not built or tested this**, so there is no local green run behind any of it. What I do have is an exhaustive comparison of `close0()` over the 96 reachable field combinations the two methods read, for master vs the first commit vs this head: the first commit adds 10 cursor-close requests that all dereference a null statement, this head adds none and diverges from master nowhere a statement exists. I would like that checked by a real build rather than trusted.

Two asks:

- **CI**: apart from "Rolling Upgrade check", nothing has ever reported on this PR — including on the first commit, where the NPE would have shown up on the first real build of `JdbcThinResultSetSelfTest`. Could a build be scheduled against `5bb60777` on `ci.ignite.apache.org`?
- **Process**: `CONTRIBUTING.md` points code contributions at JIRA (`IGNITE-NNNNN`) and this change has no ticket, only the GitHub issue. I left the title as `Fix #13549` rather than invent a key. Tell me which way this project wants it and I will retitle and reword the commits before anyone spends review time on it.

zstan:
Seems your patch isn\`t working for me ( 

```
sqlline version 1.12.0
0: jdbc:ignite:thin://127.0.0.1/> !tables
+-----------+-------------+------------+------------+---------+----------+------------+-----------+---------------------------+----------------+
| TABLE_CAT | TABLE_SCHEM | TABLE_NAME | TABLE_TYPE | REMARKS | TYPE_CAT | TYPE_SCHEM | TYPE_NAME | SELF_REFERENCING_COL_NAME | REF_GENERATION |
+-----------+-------------+------------+------------+---------+----------+------------+-----------+---------------------------+----------------+
0: jdbc:ignite:thin://127.0.0.1/> create table tbl(id int primary key, val int);
Error: Table already exists: TBL (state=42000,code=3007)
java.sql.SQLException: Table already exists: TBL
```

and it\`s all ok with prev sqlline ver 

```
sqlline version 1.9.0
0: jdbc:ignite:thin://127.0.0.1/> !tables
+-----------+-------------+------------------------------+------------+---------+----------+------------+-----------+---------------------------+----------------+
| TABLE_CAT | TABLE_SCHEM |          TABLE_NAME          | TABLE_TYPE | REMARKS | TYPE_CAT | TYPE_SCHEM | TYPE_NAME | SELF_REFERENCING_COL_NAME | REF_GENERATION |
+-----------+-------------+------------------------------+------------+---------+----------+------------+-----------+---------------------------+----------------+
| IGNITE    | PUBLIC      | TBL 
```

Can you give some additional info ? Show output for : `ls -al ./sqlline` ?

Adarsh-Me:
@zstan Thanks for actually running it — that is the first execution data this PR has ever had, and it says the change is not sufficient as written. To be blunt about what I can and cannot stand behind here: the description says I never built or tested this, and I still cannot. Only GitHub's compile jobs have run (`Check java code on JDK 17` passes, so it builds); nothing that starts a node has touched it. So I have no green result to argue with yours, and I am not going to push a second speculative commit against a symptom I cannot reproduce.

On the layout question, what I can answer factually: there is no `./sqlline` directory in an Ignite source checkout, so `ls -al ./sqlline` is empty here. The pieces are:

- `modules/sqlline` — artifact `ignite-sqlline`, and it pins `sqlline` to **1.12.0** (`modules/sqlline/pom.xml:41-42`). So 1.12.0 is not a tool you opted into, it is what the distribution ships; that also means 1.9.0-vs-1.12.0 is a real behaviour difference inside the supported path, not a misconfigured client.
- The class this PR changes is in `modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinResultSet.java`, i.e. it ships inside **ignite-core**. So for a patch to be visible to sqlline, the `ignite-core.jar` on that classpath has to be the one you just rebuilt — an `ignite-core` from a released zip sitting earlier on the classpath would silently win and reproduce exactly your "no change". That is the one thing I cannot see from here, and it is the only part of what you asked for that would actually discriminate.

What I did check in source, rather than guess: `getTables` can return a zero-row result set for reasons this patch cannot touch — `JdbcThinDatabaseMetadata.java:745` returns `new JdbcThinResultSet(emptyList, meta)` before sending any request when the catalog is not accepted or `tblTypes` names neither TABLE nor VIEW. A headers-only table is exactly what that looks like. But the catalog branch does not fire from a plain connection: `JdbcThinConnection.getCatalog()` returns `null` (`:725-729`) and `isValidCatalog` accepts null (`:1461-1463`). So that path needs SQLLine 1.12.0 to supply a non-null catalog of its own, which I cannot confirm from this host — there is no `sqlline` jar here for me to inspect, and I would rather say that than pretend I checked.

Two things that would settle it, whichever is cheaper for you:

1. The classpath/timestamps behind your run, so we know which `ignite-core` answered.
2. Does `!tables` return rows against your patched driver under SQLLine 1.9.0? That splits "driver still wrong" from "1.12.0 reads metadata a different way", which is where my analysis is currently stuck.

And if you can schedule a real build of this head on `ci.ignite.apache.org`, `JdbcThinResultSetSelfTest` is the test that needs a live node — I cannot start one here. If it turns out 1.12.0 is asking something the driver genuinely does not answer, the right fix is likely elsewhere in `JdbcThinDatabaseMetadata` and I would rather close this PR than widen it on a guess.

zstan:
@Adarsh-Me thank you for reply, look, can you just run instance of ignite with your fix and check it with modern sqlline 1.12.0 ? Very fast and simple way is : run in IDE a bit changed test, for example take:
org.apache.ignite.internal.processors.cache.index.BasicIndexTest#testNoIndexesNoPersistence
and write 

```
            startGridsMultiThreaded(gridCount());

            System.in.read(); // <--  append this hack right after node starts ^
```

connect and check:
./sqlline.sh --verbose=true -u jdbc:ignite:thin://127.0.0.1/

or how you use sqlline ?  I mean this : https://ignite.apache.org/docs/ignite2/latest/tools/sqlline.html#connecting-to-ignite-cluster

PR #11822, opened 2025-01-19
Ignite-23772 - old dates

Vladsz83:
Thank you for submitting the pull request to the Apache Ignite.

In order to streamline the review of the contribution 
we ask you to ensure the following steps have been taken:

### The Contribution Checklist
- [ ] There is a single JIRA ticket related to the pull request. 
- [ ] The web-link to the pull request is attached to the JIRA ticket.
- [ ] The JIRA ticket has the _Patch Available_ state.
- [ ] The pull request body describes changes that have been made. 
The description explains _WHAT_ and _WHY_ was made instead of _HOW_.
- [ ] The pull request title is treated as the final commit message. 
The following pattern must be used: `IGNITE-XXXX Change summary` where `XXXX` - number of JIRA issue.
- [ ] A reviewer has been mentioned through the JIRA comments 
(see [the Maintainers list](https://cwiki.apache.org/confluence/display/IGNITE/How+to+Contribute#HowtoContribute-ReviewProcessandMaintainers)) 
- [ ] The pull request has been checked by the Teamcity Bot and 
the `green visa` attached to the JIRA ticket (see [TC.Bot: Check PR](https://mtcga.gridgain.com/prs.html))

### Notes
- [How to Contribute](https://cwiki.apache.org/confluence/display/IGNITE/How+to+Contribute)
- [Coding abbreviation rules](https://cwiki.apache.org/confluence/display/IGNITE/Abbreviation+Rules)
- [Coding Guidelines](https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines)
- [Apache Ignite Teamcity Bot](https://cwiki.apache.org/confluence/display/IGNITE/Apache+Ignite+Teamcity+Bot)

If you need any help, please email dev@ignite.apache.org or ask anу advice on http://asf.slack.com _#ignite_ channel.

PR #13569, opened 2026-09-09
IGNITE-29045 SQL Calcite: Support Java temporal types in UDF and UDTF parameters and results

tkalkirill:
https://issues.apache.org/jira/browse/IGNITE-29045

### DIFF
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/TableFunctionScan.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/TableFunctionScan.java
@@ -17,16 +17,26 @@
 
 package org.apache.ignite.internal.processors.query.calcite.exec;
 
+import java.lang.reflect.Type;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.function.Supplier;
+import org.apache.calcite.linq4j.tree.Primitive;
+import org.apache.calcite.linq4j.tree.Types;
 import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.ignite.internal.processors.query.IgniteSQLException;
 import org.apache.ignite.internal.processors.query.calcite.exec.RowHandler.RowFactory;
+import org.apache.ignite.internal.processors.query.calcite.type.OtherType;
+import org.apache.ignite.internal.processors.query.calcite.util.TypeUtils;
 import org.apache.ignite.internal.util.typedef.F;
+import org.jetbrains.annotations.Nullable;
 
 /** */
 public class TableFunctionScan<Row> implements Iterable<Row> {
+    /** */
+    private final ExecutionContext<Row> ctx;
+
     /** */
     private final RelDataType rowType;
 
@@ -38,13 +48,15 @@ public class TableFunctionScan<Row> implements Iterable<Row> {
 
     /** */
     public TableFunctionScan(
+        ExecutionContext<Row> ctx,
         RelDataType rowType,
-        Supplier<Iterable<?>> dataSupplier,
-        RowFactory<Row> rowFactory
+        Supplier<Iterable<?>> dataSupplier
     ) {
+        this.ctx = ctx;
         this.rowType = rowType;
         this.dataSupplier = dataSupplier;
-        this.rowFactory = rowFactory;
+
+        rowFactory = ctx.rowHandler().factory(ctx.getTypeFactory(), rowType);
     }
 
     /** {@inheritDoc} */
@@ -58,7 +70,7 @@ public class TableFunctionScan<Row> implements Iterable<Row> {
             throw new IgniteSQLException("Unable to process table function data: row type is neither Collection or Object[].");
 
         Object[] rowArr = rowContainer.getClass() == Object[].class
-            ? (Object[])rowContainer
+            ? ((Object[])rowContainer).clone()
             : ((Collection<?>)rowContainer).toArray();
 
         if (rowArr.length != rowType.getFieldCount()) {
@@ -66,6 +78,27 @@ public class TableFunctionScan<Row> implements Iterable<Row> {
                 + "] doesn't match defined columns number [" + rowType.getFieldCount() + "].");
         }
 
+        for (int i = 0; i < rowArr.length; i++)
+            rowArr[i] = convertToInternal(rowArr[i], rowType.getFieldList().get(i).getType());
+
         return rowFactory.create(rowArr);
+    }
+
+    /** */
+    private @Nullable Object convertToInternal(@Nullable Object val, RelDataType type) {
+        // Preserve objects for both Ignite's custom OTHER type and Calcite's SQL OTHER type.
+        if (val == null || type instanceof OtherType || type.getSqlTypeName() == SqlTypeName.OTHER)
+            return val;
+
+        Type storageType = ctx.getTypeFactory().getResultClass(type);
+
+        if (!TypeUtils.isConvertableType(storageType))
+            return val;
+
+        // SQL table functions can already return values in the internal representation.
+        if (Types.isAssignableFrom(Primitive.box(ctx.getTypeFactory().getJavaClass(type)), val.getClass()))
+            return val;
+
+        return TypeUtils.toInternal(ctx, val, storageType);
     }
 }
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
@@ -38,6 +38,8 @@ import org.apache.calcite.sql.type.SqlTypeUtil;
 import org.apache.calcite.util.BuiltInMethod;
 import org.apache.calcite.util.Util;
 import org.apache.ignite.internal.processors.query.calcite.util.Commons;
+import org.apache.ignite.internal.processors.query.calcite.util.TypeUtils;
+import org.jetbrains.annotations.Nullable;
 
 /** */
 public class ConverterUtils {
@@ -123,10 +125,18 @@ public class ConverterUtils {
     /** */
     static List<Expression> fromInternal(Class<?>[] targetTypes,
         List<Expression> expressions) {
-        final List<Expression> list = new ArrayList<>();
+        return fromInternal(null, targetTypes, expressions);
+    }
+
+    /** Converts user-defined function arguments using the execution context when available. */
+    static List<Expression> fromInternal(@Nullable Expression root,
+        Class<?>[] targetTypes,
+        List<Expression> expressions
+    ) {
+        final List<Expression> list = new ArrayList<>(expressions.size());
         if (targetTypes.length == expressions.size()) {
             for (int i = 0; i < expressions.size(); i++)
-                list.add(fromInternal(expressions.get(i), targetTypes[i]));
+                list.add(fromInternal(root, expressions.get(i), targetTypes[i]));
         }
         else {
             int j = 0;
@@ -139,10 +149,30 @@ public class ConverterUtils {
                 else
                     type = targetTypes[j].getComponentType();
 
-                list.add(fromInternal(expressions.get(i), type));
+                list.add(fromInternal(root, expressions.get(i), type));
             }
         }
         return list;
+    }
+
+    /** */
+    private static Expression fromInternal(@Nullable Expression root, Expression operand, Type targetType) {
+        // Preserve Calcite's calendar conversion for JDBC dates and timestamps.
+        Expression converted = fromInternal(operand, targetType);
+
+        if (root == null || converted != operand || !TypeUtils.isConvertableType(targetType))
+            return converted;
+
+        if (Types.isAssignableFrom(targetType, operand.getType()))
+            return operand;
+
+        if (Primitive.is(operand.getType()))
+            operand = Expressions.box(operand);
+
+        return Expressions.convert_(
+            Expressions.call(TypeUtils.class, "fromInternal", root, operand, Expressions.constant(targetType)),
+            targetType
+        );
     }
 
     /** */
--- a/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ReflectiveCallNotNullImplementor.java
+++ b/modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ReflectiveCallNotNullImplementor.java
@@ -18,11 +18,13 @@ package org.apache.ignite.internal.processors.query.calcite.exec.exp;
 
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.lang.reflect.Type;
 import java.util.List;
 
 import org.apache.calcite.linq4j.tree.Expression;
 import org.apache.calcite.linq4j.tree.Expressions;
 import org.apache.calcite.rex.RexCall;
+import org.apache.ignite.internal.processors.query.calcite.util.TypeUtils;
 
 import static org.apache.ignite.internal.processors.query.calcite.util.IgniteMethod.UDF_INSTANCE;
 
@@ -49,10 +51,10 @@ public class ReflectiveCallNotNullImplementor implements NotNullImplementor {
     @Override public Expression implement(RexToLixTranslator translator,
         RexCall call, List<Expression> translatedOperands) {
         translatedOperands =
-            ConverterUtils.fromInternal(method.getParameterTypes(), translatedOperands);
+            ConverterUtils.fromInternal(translator.getRoot(), method.getParameterTypes(), translatedOperands);
         translatedOperands =
             ConverterUtils.convertAssignableTypes(method.getParameterTypes(), translatedOperands);
-        final Expression callExpr;
+        Expression callExpr;
         if ((method.getModifiers() & Modifier.STATIC) != 0)
             callExpr = Expressions.call(method, translatedOperands);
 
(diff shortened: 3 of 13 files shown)


modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ReflectiveCallNotNullImplementor.java
```diff
@@ -66,6 +68,24 @@ public ReflectiveCallNotNullImplementor(Method method) {
...
         }
+
+        if (TypeUtils.isConvertableType(method.getReturnType())) {
+            Type targetType = translator.typeFactory.getJavaClass(call.getType());
+            Expression converted = ConverterUtils.toInternal(callExpr, targetType);
+
+            if (converted != callExpr)
+                callExpr = converted;
+            else {
+                Expression result = method.getReturnType().isPrimitive() ? Expressions.box(callExpr) : callExpr;
```
zstan: dead code ? CONVERTABLE_TYPES has no primitives

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ConverterUtils.java
```diff
@@ -139,12 +149,32 @@ static List<Expression> fromInternal(Class<?>[] targetTypes,
...
+    /** */
+    private static Expression fromInternal(@Nullable Expression root, Expression operand, Type targetType) {
+        // Preserve Calcite's calendar conversion for JDBC dates and timestamps.
+        Expression converted = fromInternal(operand, targetType);
+
+        if (root == null || converted != operand || !TypeUtils.isConvertableType(targetType))
+            return converted;
+
+        if (Types.isAssignableFrom(targetType, operand.getType()))
+            return operand;
```
zstan:
```suggestion
        if (Types.isAssignableFrom(targetType, operand.getType()))
            return operand;

        Expression converted = fromInternal(operand, targetType);

        if (root == null || converted != operand || !TypeUtils.isConvertableType(targetType))
            return converted;
```

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/type/OtherType.java
```diff
@@ -29,21 +28,12 @@ public OtherType(boolean nullable) {
 
     /** {@inheritDoc} */
     @Override protected void generateTypeString(StringBuilder sb, boolean withDetail) {
-        sb.append("OTHER");
+        // The digest must differ from Calcite's OTHER to keep the types distinct in its shared type cache.
+        sb.append(withDetail ? "IGNITE_OTHER" : "OTHER");
```
zstan, line 32: can we use simple : sb.append("IGNITE_OTHER"); here ?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtilsTest.java
```diff
@@ -0,0 +1,143 @@
...
+import org.apache.calcite.DataContexts;
+import org.apache.calcite.util.DateString;
+import org.apache.calcite.util.TimeString;
+import org.apache.calcite.util.TimestampString;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertSame;
+
+/** */
```
zstan, line 42: Plz append minimal description, it\`s not clear a mission of such a test class.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
```diff
@@ -83,6 +85,10 @@
 
 /** */
 public class TypeUtils {
+    /** Start of the Gregorian part of the calendar used by JDBC temporal types. */
+    private static final long GREGORIAN_CUTOVER =
```
zstan: I wonder, i can\`t see any changes in jdbc tests but they are mentioned, plz clarify - why do we need such a change ?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
```diff
@@ -409,23 +415,39 @@ else if (val instanceof Number && storageType != val.getClass()) {
...
-        return toLong((java.util.Date)val, DataContext.Variable.TIME_ZONE.get(ctx));
+        return toLong((java.util.Date)val, timeZone(ctx));
     }
 
     /** */
     private static long toLong(java.util.Date val, TimeZone tz) {
         long time = val.getTime();
+        long locTs = time + tz.getOffset(time);
+
+        if (locTs >= GREGORIAN_CUTOVER)
```
zstan: I just try to know - why do we need it , comment this code and all tests from JdbcTestSuite are passed, plz clarify why we need it ?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtilsTest.java
tkalkirill: Will it be useful? Quite a few classes in this module don't have it, and I’d prefer not to include information that isn't useful.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
tkalkirill: By “JDBC types” I meant `java.sql.Date` and `Timestamp`, which are also used for embedded SQL results. I’ve clarified the comments.
These types and Calcite use different calendars before October 15, 1582. Without this conversion, a UDF returning `LocalDate.of(1500, 1, 2)` produces 1499-12-24 in a plain SELECT.
This is covered by `TypeUtilsTest` and `UserDefinedFunctionsIntegrationTest.testJavaTimeFunctionResultsAsJdbcValues`; no JDBC driver tests were changed.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
tkalkirill: Added a comment explaining why the conversion is needed, with the 1500-01-02 → 1500-01-11 example. Without it, three tests in `TypeUtilsTest` fail; these tests are not part of `JdbcTestSuite`.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/type/OtherType.java
tkalkirill: I’d keep the conditional. `withDetail=true` builds the unique digest, where `IGNITE_OTHER` prevents the cache collision. With false, it produces the regular type name, which `IgniteTypeCoercion` also uses when constructing `CAST` expressions. Keeping `OTHER` there preserves the existing behavior.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
zstan: we have different branches ?) 

Image

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/type/OtherType.java
zstan: Sounds strange - like : we need cases with and without cache collisions ?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtilsTest.java
zstan: It\`s up to u

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
tkalkirill: I meant about the changes in general, I updated the comment for this section.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/type/OtherType.java
tkalkirill: I thought the answer might provide some clarity.
I added `DataTypesTest#testOtherTypeImplicitCast` that demonstrating that following your suggestion results in an error, because `IgniteTypeCoercion` uses `toString()` to build implicit casts, so returning `IGNITE_OTHER` unconditionally breaks inserts into `OTHER` columns with `Unknown identifier 'IGNITE_OTHER'`.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
```diff
@@ -435,15 +464,15 @@ public static Object fromInternal(DataContext ctx, Object val, Type storageType)
         else if (storageType == java.sql.Date.class && val instanceof Integer)
             return new java.sql.Date(fromLocalTs(ctx, (Integer)val * DateTimeUtils.MILLIS_PER_DAY));
         else if (storageType == LocalDate.class && val instanceof Integer)
-            return new java.sql.Date(fromLocalTs(ctx, (Integer)val * DateTimeUtils.MILLIS_PER_DAY)).toLocalDate();
+            return LocalDate.ofEpochDay((Integer)val);
```
zstan, line 467: all these GRIGORIAN cut over is very hard for understanding and i think you in wrong way here, look - i revert all changes in this class and store only line: 467 and 475 i.e. : 
return LocalDate.ofEpochDay((Integer)val);
and
return LocalDateTime.ofInstant(Instant.ofEpochMilli((Long)val), ZoneOffset.UTC);
and **testJavaTimeFunctionsDuringDstTransition** passes for literals and fails for dyn params -thus problem is in dyn params treating, i also know that AI2 work not perfectly well with Local* historically by design .. probably you can also check appropriate AI3 code

tkalkirill: I checked both variants. Keeping direct Local* conversions in both directions makes the DST tests pass without cutover handling. Fixing only fromInternal still leaves failures in dynamic parameters and UDF/UDTF results.
However, without cutover handling, converting SQL DATE 1500-01-02 to java.sql.Date produces 1499-12-24. I reproduced this with the base-branch implementation too.
Should we limit this change to direct Local* conversions and address historical java.sql.* dates separately, or preserve those calendar dates in this PR as well?
Or what suggestions do you have?

zstan: Plz fill additional issue (probably with related code, or intermediate pr link ?) and remove GREGORIAN related stuff from this PR, thank you !

tkalkirill: Created a ticket and removed GREGORIAN related stuff from this PR.

zstan (approved):
thanks ! now looks good, i\`m ok with it if viza will pass.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/TableFunctionScan.java
```diff
@@ -58,14 +70,35 @@ private Row convertToRow(Object rowContainer) {
             throw new IgniteSQLException("Unable to process table function data: row type is neither Collection or Object[].");
 
         Object[] rowArr = rowContainer.getClass() == Object[].class
-            ? (Object[])rowContainer
+            ? ((Object[])rowContainer).clone()
```
alex-plekhanov: Minor: We can check if any type is convertible (for example in constructor) and clone array only if required.

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/UserDefinedFunctionsIntegrationTest.java
```diff
@@ -492,6 +504,383 @@ public void testBigDecimalFunctionArgument() {
...
+        assertQuery("SELECT * FROM objectTableValues()")
+            .withResultChecker(rows -> {
+                assertEquals(1, rows.size());
+                assertEquals(exp.length, rows.get(0).size());
+
+                for (int i = 0; i < exp.length; i++) {
+                    Object actual = rows.get(0).get(i);
+
+                    assertEquals("Unexpected value type at index " + i, exp[i].getClass(), actual.getClass());
+                    assertEqualsArraysAware("Unexpected value at index " + i, exp[i], actual);
```
alex-plekhanov: Minor: There are no arrays in result, assertEqualsArraysAware is confusing.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
```diff
@@ -528,4 +570,11 @@ public static RexNode toRexLiteral(Object dfltVal, RelDataType type, DataContext
 
         return rexBuilder.makeLiteral(dfltVal, type, true);
     }
+
+    /** */
+    private static TimeZone timeZone(DataContext ctx) {
+        TimeZone tz = DataContext.Variable.TIME_ZONE.get(ctx);
+
+        return tz != null ? tz : TimeZone.getDefault();
```
alex-plekhanov: Looks like DataContext.Variable.TIME_ZONE.get(ctx) is never can be null, why do we need this new check?

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtilsTest.java
```diff
@@ -0,0 +1,143 @@
...
+                Timestamp.class);
+        }
+    }
+
+    /** */
+    @Test
+    public void testSqlDateConversion() {
+        for (String date : new String[] {
+            "0001-01-01", "1500-01-02", "1582-10-04", "1582-10-15", "1969-12-31", "1970-01-01", "9999-12-31"
+        }) {
```
alex-plekhanov: Redundant braces

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtilsTest.java
```diff
@@ -0,0 +1,143 @@
...
+import static org.junit.Assert.assertSame;
+
+/** */
+public class TypeUtilsTest {
+    /** */
+    @Test
+    public void testLocalDateConversion() {
+        for (String date : new String[] {
+            "0001-01-01", "1500-01-02", "1582-10-10", "1969-12-31", "1970-01-01", "2011-12-30", "9999-12-31"
+        }) {
```
alex-plekhanov: Redundant braces

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
```diff
@@ -514,7 +543,20 @@ else if (UUID.class.equals(storageType)) {
 
     /** */
     private static long fromLocalTs(DataContext ctx, long ts) {
-        TimeZone tz = DataContext.Variable.TIME_ZONE.get(ctx);
+        if (ts < GREGORIAN_CUTOVER) {
```
alex-plekhanov: This is more correct fix for https://issues.apache.org/jira/browse/IGNITE-23772 ticket. Maybe we should check tests from PR for this ticket and if all these tests pass, close IGNITE-23772 after your fix?

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
tkalkirill: Constant folding uses `RexExecutorImpl(DataContexts.EMPTY)`, where `TIME_ZONE` is null. Without this fallback, temporal conversions throw an NPE and Calcite skips constant folding. I verified this with deterministic UDFs accepting and returning `java.util.Date`. The fallback uses the same JVM default as `BaseDataContext`.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/util/TypeUtils.java
tkalkirill: This fix addresses the incorrect conversion of historical dates, and the tests from PR #11822 pass. However, literals such as DATE '1582-10-05' still fail during parsing. After this fix is merged, we should address those remaining cases under IGNITE-23772.

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/exp/ReflectiveCallNotNullImplementor.java
```diff
@@ -66,6 +68,21 @@ public ReflectiveCallNotNullImplementor(Method method) {
...
+
+            if (converted != callExpr)
+                callExpr = converted;
+            else
+                callExpr = Expressions.convert_(
+                    Expressions.call(TypeUtils.class, "toInternal", translator.getRoot(), callExpr,
+                        Expressions.constant(method.getReturnType())),
+                    targetType
+                );
+        }
```
alex-plekhanov: Maybe it's worth to move this logic to new method `ConverterUtils.toInternal(root, ...)` for symmetry with `fromInternal`

modules/calcite/src/main/java/org/apache/ignite/internal/processors/query/calcite/exec/TableFunctionScan.java
```diff
@@ -58,14 +75,50 @@ private Row convertToRow(Object rowContainer) {
             throw new IgniteSQLException("Unable to process table function data: row type is neither Collection or Object[].");
 
         Object[] rowArr = rowContainer.getClass() == Object[].class
-            ? (Object[])rowContainer
+            ? ((Object[])rowContainer)
```
alex-plekhanov: Redundant parentheses

modules/calcite/src/test/java/org/apache/ignite/internal/processors/query/calcite/integration/UserDefinedFunctionsIntegrationTest.java
```diff
@@ -492,6 +504,383 @@ public void testBigDecimalFunctionArgument() {
...
+                    assertEquals("Unexpected value type at index " + i, exp[i].getClass(), actual.getClass());
+                    assertEquals("Unexpected value at index " + i, exp[i], actual);
+                }
+            })
+            .check();
+    }
+
+    /** */
+    @Test
+    public void testSerializableTableFunctionResult() {
```
alex-plekhanov: What is so special in serialiazable values? Why check with Object type is not enough?

alex-plekhanov (approved):

PR #13583, opened 2026-09-13
IGNITE-28907 Support merging QueryEntity metadata in CacheConfiguration

oleg-vlsk:

### Notes


oleg-vlsk:
**[Ignite PR Checker](https://ignite-pr-checker.is-a.dev/?pr=13583)** verdict · RunAll build [9344108](https://ci2.ignite.apache.org/build/9344108) · 147 suites ran, 0 reused

- 2 suite(s) have no reliable result (compilation error, timeout, crash)
- 1 suite(s) ran far fewer tests than the same suites on master
- a newer run is still going — its unfinished suites can still fail

⚠️ **2 broken suite(s)** (no reliable run):
- Platform C++ CMake (Linux Clang): IgniteOdbcTest terminated abnormally - the test binary died, its remaining tests never ran · Number of tests 595 is 43% less than 1047 in build #43233 — ran 595 of master's 1047 tests
- Platform C++ CMake (Win x64 | Release): non-zero exit code · Number of tests 465 is 56% less than 1047 in build #19133 — ran 465 of master's 1047 tests

- Platform C++ CMake (Linux): 595 tests vs 1047 on master (**−43%**)

❌ **8 blocker(s) in 5 suite(s):**
- ZooKeeper (Discovery) 1: `org.apache.ignite.spi.discovery.zk.ZookeeperDiscoverySpiTestSuite1: org.apache.ignite.spi.discovery.zk.internal.ZookeeperDiscoverySplitBrainTest.testSimpleSplitBrain`
- ZooKeeper (Discovery) 1: `org.apache.ignite.spi.discovery.zk.ZookeeperDiscoverySpiTestSuite1: org.apache.ignite.spi.discovery.zk.internal.ZookeeperDiscoverySplitBrainTest.testAlmostSplitBrain`
- ZooKeeper (Discovery) 1: `org.apache.ignite.spi.discovery.zk.ZookeeperDiscoverySpiTestSuite1: org.apache.ignite.spi.discovery.zk.internal.ZookeeperDiscoverySplitBrainTest.testNotActualSplitBrain`
- Platform .NET (Core Linux): `Apache.Ignite.Core.Tests.DotNetCore: Apache.Ignite.Core.Tests.Client.Services.ServicesAwarenessTests.TestClusterTopologyChanges("PlatformTestService")`
- Index Query API: `org.apache.ignite.cache.query.IndexQueryTestSuite: org.apache.ignite.cache.query.IndexQueryRangeTest.testRangeQueries[qryPar=1 atomicity=TRANSACTIONAL mode=REPLICATED node=CRD backups=0 idxName=PERSON_ID_IDX duplicates=1]`
- Queries 6: `org.apache.ignite.testsuites.IgniteBinaryCacheQueryTestSuite6: org.apache.ignite.internal.processors.cache.index.DuplicateKeyValueClassesSelfTest.testDuplicateValueClass`
- PDS (Indexing): `org.apache.ignite.testsuites.IgnitePdsWithIndexingCoreTestSuite: org.apache.ignite.internal.processors.cache.persistence.db.wal.IgniteWalRebalanceTest.testWithLocalWalChange`
- PDS (Indexing): `org.apache.ignite.testsuites.IgnitePdsWithIndexingCoreTestSuite: org.apache.ignite.internal.processors.cache.persistence.IgnitePdsCorruptedIndexTest.testCorruption`

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/DuplicateKeyValueClassesSelfTest.java
```diff
@@ -42,32 +48,51 @@ public class DuplicateKeyValueClassesSelfTest extends AbstractIndexingCommonTest
...
     @Test
-    public void testDuplicateValueClass() throws Exception {
+    public void testConflictingKeyTypesForSameValueClass() {
         CacheConfiguration ccfg = new CacheConfiguration()
-            .setName(CACHE_NAME)
-            .setIndexedTypes(UUID.class, Clazz1.class, String.class, Clazz1.class);
+            .setName(CACHE_NAME);
 
-        grid(0).createCache(ccfg);
+        String msg = String.format("Failed to merge query entities due to conflicting metadata " +
```
zstan, line 86: plz move this message into public string constant (QueryEntityMerger) and use it for comparison here.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/persistence/IgnitePdsIndexingDefragmentationTest.java
```diff
@@ -195,7 +216,7 @@ private static void validateIndexes(IgniteEx node) throws Exception {
      */
     @Test
     public void testIndexingWithIntegerKey() throws Exception {
-        test(Function.identity());
+        test(Integer.class, Function.identity());
```
zstan, line 219:
```suggestion
        test(indexedKeyType, Function.identity());
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1202 @@
...
+    /** {@inheritDoc} */
+    @Override protected void afterTest() throws Exception {
+        stopAllGrids();
+
+        super.afterTest();
+    }
+
+    /** Query entities with different value types must not be merged. */
+    @Test
+    public void testDifferentValueTypesAreNotMerged() throws Exception {
```
zstan, line 76: naming confusing, i see test for public API - this test - how it need\plan to work. I mean - one cache can contain numerous of different QueryEntity`s, isn`t it ? This test need to work perfectly well without your changes i suppose, if i\`m right - you need to rename it.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1202 @@
...
+
+        assertEquals(2, entities.size());
+
+        for (Class<?> cls : List.of(Person.class, AnnotatedPerson.class))
+            assertTrue(entities.stream().anyMatch(e -> cls.getName().equals(e.getValueType())));
+    }
+
+    /** Query entities with the same value type but different key type are a conflict. */
+    @Test
+    public void testConflictingKeyTypesFail() {
```
zstan, line 106: may be you can remove this test at all in such a case ? : testConflictingKeyTypesForSameValueClass

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1202 @@
...
+import org.apache.ignite.spi.systemview.view.sql.SqlTableView;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.junit.Test;
+
+import static org.apache.ignite.internal.processors.query.schema.management.SchemaManager.SQL_TBLS_VIEW;
+import static org.apache.ignite.internal.processors.query.schema.management.SchemaManager.SQL_TBL_COLS_VIEW;
+import static org.apache.ignite.testframework.GridTestUtils.assertThrows;
+
+/** Tests for merging QueryEntity metadata in CacheConfiguration. */
+public class CacheConfigurationQueryEntityMergeTest extends GridCommonAbstractTest {
```
zstan, line 48:
```suggestion
@SuppressWarnings("ThrowableNotThrown")
public class CacheConfigurationQueryEntityMergeTest extends GridCommonAbstractTest {
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1202 @@
...
+     * treated as having conflicting key types.
+     */
+    @Test
+    public void testConflictingImplicitAndExplicitKeyTypes() {
+        CacheConfiguration<Integer, Person> ccfg = new CacheConfiguration<>(CACHE_NAME);
+
+        QueryEntity first = new QueryEntity()
+            .setValueType(Person.class.getName())
+            .setFields(fields("id", Integer.class))
+            .setKeyFieldName("id"); // keyType is null
```
zstan, line 150: I suppose that in such case key type is implicitly derived from (.setFields(fields("id", Integer.class)).setKeyFieldName("id")) sequence. Thus i expect that near construction will fail but it didn\`t:

```
        QueryEntity first = new QueryEntity()
            .setValueType(Person.class.getName())
            .setFields(fields("id", Integer.class))
            .setKeyType(Long.class.getTypeName()) // <--
            .setKeyFieldName("id"); // keyType is null
```
and looks like shit :

```
        QueryEntity first = new QueryEntity()
            .setValueType(Person.class.getName())
            .setFields(fields("id", Integer.class))
            .setKeyType(Long.class.getTypeName()) // <--
            .setKeyType(Integer.class.getTypeName()) // <-- so what type ? 
            .setKeyFieldName("id"); // keyType is null
```

I understand that all such questions are also touch backward compatibility, but you also change public contract (bad contract of course) and if so - probably we also make clean all over here ?What do you think ?

Also why it\`s a collision here, as for me there is 2 entry, first - with implicitly defined type and second with explicitly, i\`m wrong ?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1202 @@
...
+        QueryEntity first = new QueryEntity()
+            .setFields(fields("val", Person.class))
+            .setValueFieldName("val"); // valueType is null
+
+        QueryEntity second = new QueryEntity().setValueType(Person.class.getName());
+
+        ccfg.setQueryEntities(Collections.singleton(first));
+        ccfg.setQueryEntities(Collections.singleton(second));
+
+        node.createCache(ccfg);
```
zstan, line 196: I also think about : if before your fix first and second are not been merged = for access them we need 2 different sql queries and after PR all become work through one query - it\`s incompatible behavior with broken back compatibility

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryEntityMerger.java
```diff
@@ -0,0 +1,261 @@
...
+        if (!Objects.equals(ex.findValueType(), in.findValueType())) {
+            throw new CacheException(
+                "Failed to merge query entities because value types differ " +
+                    "[cacheName=" + cacheName +
+                    ", existingValueType=" + ex.findValueType() +
+                    ", incomingValueType=" + in.findValueType() + ']'
+            );
+        }
+
+        QueryEntity res = new QueryEntity(ex);
```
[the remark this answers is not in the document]
oleg-vlsk: I reproduced the `QueryEntityEx` metadata loss and investigated the schema-add and cache-configuration merge paths.

The reported problem is real: with the current PR implementation, `QueryEntity res = new QueryEntity(ex)` can turn a `QueryEntityEx` into a plain `QueryEntity`. Which leads to losing `sql`, `preserveKeysOrder`, `fillAbsentPKsWithDefaults`, and the other extended metadata.

However, after tracing the actual lifecycle, it looks like the schema-add case does not require defining merge rules for two different `QueryEntityEx` instances. The downgrade was caused by a different issue in `patchCacheConfiguration()`.

The reproducer is `testSchemaAddPreservesQueryEntityExMetadata()`. It uses a statically configured cache with no query entities and then executes a query that creates a table. This DDL operation creates a `QueryEntityEx` with:
- sql = true
- preserveKeysOrder = true
- fillAbsentPKsWithDefaults = true

The schema-add lifecycle patches the cache configuration twice:
```text
SchemaAddQueryEntityOperation
>>>
GridCacheContext.onSchemaAddQueryEntity() -> patchCacheConfiguration()
>>>
SchemaFinishDiscovery -> DynamicCacheDescriptor.schemaChangeFinish() → patchCacheConfiguration()
```

Initially I expected this to be a legitimate merge of two extended entities, but it is not. The first and the second patch operate on different `CacheConfiguration` objects, but `CacheConfiguration` copy construction keeps the same mutable `qryEntities` collection. Therefore the first patch modifies the collection that is still visible from the descriptor configuration.

So the "merge" in this scenario is actually an accidentally repeated usage of the same schema-add entity. It is not merging two independently defined `QueryEntityEx` configurations.

I added an isolated regression test for the underlying issue:
```text
GridCacheUtilsTest.testPatchCacheConfigurationDoesNotModifyOriginalQueryEntities
```

The test prepares an `oldCfg` with a materialized empty query-entity collection, calls `GridCacheUtils.patchCacheConfiguration()`, and verifies that adding the new entity to the patched configuration does not modify `oldCfg`.

On `master` / the previous implementation it fails on:
```java
assertTrue(oldCfg.getQueryEntities().isEmpty());
```
because both configurations share the same mutable `qryEntities` collection.

After making `patchCacheConfiguration()` use an independent query-entity collection, this test passes.

More importantly, after this change `testSchemaAddPreservesQueryEntityExMetadata` also passes. The schema-finish patch now sees an empty descriptor configuration and simply installs the incoming `QueryEntityEx`. It no longer tries to merge two identical entities.

I also checked whether we have a supported runtime path that would require resolving a real conflict between two different `QueryEntityEx` objects. The relevant cases I checked were:
1. The normal `QueryEntity` conflicts targeted by this PR are still real and are covered by `CacheConfigurationQueryEntityMergeTest`. 
2. Adding SQL metadata dynamically to an already indexed cache is rejected, for example:
```text
javax.cache.CacheException: Cache is already indexed: sql-cache
```
3. Joining an active cluster with a static cache configuration that would require schema/configuration merging is rejected:
```text
org.apache.ignite.spi.IgniteSpiException: Failed to join node to the active cluster (the config of the cache 'TEST_CACHE' has to be merged which is impossible on active grid). Deactivate grid and retry node join or clean the joining node.
```
4. I also tested a more specific SQL-vs-API case: create the cache/table through `CREATE TABLE` on the existing node and configure the same cache statically on the joining node. This is rejected with:
```text
org.apache.ignite.IgniteCheckedException: Cache configuration mismatch (local cache was created via Ignite API, while remote cache was created via CREATE TABLE): ...
```
Again, `QueryEntityMerger` is not reached.
5. It is possible to force `QueryEntityEx + QueryEntity` manually by obtaining the internal `CacheConfiguration` through `node.context().cache().cacheConfiguration(...)` and calling `setQueryEntities()` directly. In that artificial scenario the current merger can indeed downgrade the entity. 
However, this mutates Ignite's internal runtime configuration directly. And as I understand it is not a supported public cache-configuration update path.

I also reviewed the internal `setQueryEntities()` paths. Normalization replaces the entity collection after clearing it; creation of a new SQL cache installs entities into an initially empty configuration; schema-add is expected to add entities to an empty schema; and the join cases above are validated before conflicting configurations can reach the merger.

Therefore I did not add rules for merging conflicting `QueryEntityEx`-specific fields. This would introduce new semantics, for example, for two different extended entities we would have to decide what these combinations mean:
- sql: false vs true
- implicitPk: false vs true
- preserveKeysOrder: false vs true
- fillAbsentPKsWithDefaults: false vs true
- primaryKeyInlineSize: 10 vs 20
- affinityKeyInlineSize: 10 vs 20

It is not clear whether they should be conflicts, whether one side should overtake or whether some boolean fields should be combined. I could not find a supported runtime lifecycle that requires making those decisions.

So the fix I propose is:
- keep QueryEntityMerger responsible for actual QueryEntity metadata merging
- make patchCacheConfiguration() independent from the original configuration's mutable query-entity collection
- keep testSchemaAddPreservesQueryEntityExMetadata as the integration regression test
- add GridCacheUtilsTest.testPatchCacheConfigurationDoesNotModifyOriginalQueryEntities
  as the focused regression test for the root cause.
With this change, the `QueryEntityEx` created by the schema-add operation is processed correctly without introducing merge rules for its extended metadata.

If there is a supported path where two independently created `QueryEntityEx` instances are expected to be merged, I can add handling for that path as well, but I could not reproduce or identify such a case in the current cache/schema lifecycle.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/DuplicateKeyValueClassesSelfTest.java
oleg-vlsk: Done.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/persistence/IgnitePdsIndexingDefragmentationTest.java
oleg-vlsk: `test()` expects `Class<T>`, so passing `Class<?>` will cause type mismatch. I changed `test()` so that `indexedKeyType` is used as the key type source directly

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
oleg-vlsk: Correct, we can have entities with the same key type, but with different value type. You’re right, this was working correctly before my changes. This is actually a regression test: I think it’s important to make sure that QueryEntityMerger doesn’t try to merge such entities and doesn’t view them as a conflict. I changed the name and javadoc to make it more obvious.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
oleg-vlsk: This test covers a different scenario and should remain. `testQueryEntitiesWithSameKeyTypeAndDifferentValueTypesAreNotMerged()` checks that we can have entities with the same key type but different value types. This test checks that 2 entities with the same value type are a conflict, even though they have different key types. Before my changes one of such entities was silently discarded.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
oleg-vlsk: Done.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
oleg-vlsk: Here 2 entities are merged because their key type is the same (null), so we end up having only one entity [null → Person]. Before my fix there no merging, but because the value type is the same (Person) the second entity was silently discarded. So the result is the same — only 1 entity [null → Person] . The only difference is that if you run this test on master, this assertion

```assertEquals(Person.class.getName(), entity.getValueType());```

will fail, because after my fix all properties of both entities are merged, while now in master the second entity is just ignored.

File: modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java (hunk 6 above)
Comment (oleg-vlsk) [reply modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java:174]: This test basically repeats the logic of `testConflictingKeyTypesForSameValueTypeFail()`: if we have 2 entities with the same value type, they are a conflict (before the cahnges the second entity with the same value type was silently discarded). 

But in this case we check that `QueryEntityMerger#merge` even though we do not explicitely set key type for the first entity. So when we start merging the key type of the first entity is null. Thus, we check that the `mergeKeyType()` helper in `QueryEntityMerger` works correctly by using `findKeyType()` and not `getKeyType()`, which would return null and do incorrect merging null + String → String.

So the collision here happens not because key type was set implicitly for the first entity, but because two entities have the same valye type (`Person.class`).

File: modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java [hunk 9]
```diff
@@ -0,0 +1,1260 @@
...
+                "ID2 INT, " +
+                "VAL VARCHAR NOT NULL, " +
+                "PRIMARY KEY (ID1, ID2)" +
+                ") WITH \"CACHE_NAME=" + CACHE_NAME + "\""
+        )).getAll();
+
+        Collection<QueryEntity> entities = node.context().cache()
+            .cacheDescriptor(CACHE_NAME)
+            .cacheConfiguration()
+            .getQueryEntities();
```
zstan, line 1074:
```suggestion
Collection<QueryEntity> entities = entities(node);
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1260 @@
...
+    public void testSchemaAddPreservesQueryEntityExMetadata() throws Exception {
+        staticCfg = true;
+
+        IgniteEx node = startGrid(0);
+
+        node.cluster().state(ACTIVE);
+
+        DynamicCacheDescriptor desc = node.context().cache().cacheDescriptor(CACHE_NAME);
+
+        assertTrue(desc.cacheConfiguration().getQueryEntities().isEmpty());
```
zstan, line 1060:
```suggestion
        assertTrue(entities(node).isEmpty());
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1260 @@
...
+    /** */
+    private static SqlTableView cacheTable(IgniteEx node) {
+        SystemView<SqlTableView> tables = node.context().systemView().view(SQL_TBLS_VIEW);
+        assertNotNull(tables);
+
+        SqlTableView res = null;
+
+        for (SqlTableView tbl : tables) {
+            if (CACHE_NAME.equals(tbl.cacheName()))
+                res = tbl;
```
zstan, line 1177:
```suggestion
            if (CACHE_NAME.equals(tbl.cacheName())) {
                res = tbl;
                break;
              }
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1260 @@
...
+
+        ccfg.setQueryEntities(Collections.singletonList(first));
+        ccfg.setQueryEntities(Collections.singletonList(second));
+
+        node.createCache(ccfg);
+
+        SqlTableView tbl = cacheTable(node);
+        assertNotNull(tbl);
+
+        assertEquals(tblName, tbl.tableName());
```
zstan, line 252: let\`s add also ?

assertEquals(1, entities(node).size());

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1260 @@
...
+    }
+
+    /** Different configured table names are a conflict. */
```
zstan, line 281:
```suggestion
    /** Different configured table names are lead to conflict. */
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1260 @@
...
+
+        String keyFieldName = "key";
+
+        QueryEntity first = personEntity(fields(keyFieldName, Integer.class, NAME_FIELD, String.class));
+        QueryEntity second = personEntity(fields(keyFieldName, Integer.class, NAME_FIELD, String.class));
+
+        second.setKeyFieldName(keyFieldName);
+
+        ccfg.setQueryEntities(Collections.singletonList(first));
+        ccfg.setQueryEntities(Collections.singletonList(second));
```
zstan, line 322:
```suggestion
        ccfg.setQueryEntities(List.of(first));
        ccfg.setQueryEntities(List.of(second));
```
more shorter but it up to you

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1260 @@
...
+
+        QueryEntity entity = singleQueryEntity(node);
+        assertNotNull(entity);
+
+        assertEquals(keyFieldName, entity.getKeyFieldName());
+    }
+
+    /** Same key field names must be merged. */
+    @Test
+    public void testSameKeyFieldNameIsMerged() throws Exception {
```
zstan, line 334: all merge exceptance need to be checked through :
```
Collection<QueryEntity> entities = entities(node);
assertEquals(EXCEPTED, entities.size());
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1260 @@
...
+
+        node.createCache(ccfg);
+
+        QueryEntity entity = singleQueryEntity(node);
+        assertNotNull(entity);
+
+        assertEquals(keyFieldName, entity.getKeyFieldName());
+    }
+
+    /** Different key field names for the same query entity are a conflict. */
```
zstan, line 358:
```suggestion
    /** Different key field names for the same query entity are lead to conflict. */
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1260 @@
...
+        assertNotNull(actualNameAlias);
+        assertTrue(nameAlias.equalsIgnoreCase(actualNameAlias));
+
+        String actualAgeAlias = entity.getAliases().get(AGE_FIELD);
+
+        assertNotNull(actualAgeAlias);
+        assertTrue(ageAlias.equalsIgnoreCase(actualAgeAlias));
+    }
+
+    /** Different aliases for the same field are a conflict. */
```
zstan, line 645:
```suggestion
    /** Different aliases for the same field are lead to conflict. */
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1260 @@
...
+        QueryEntity entity = singleQueryEntity(node);
+        assertNotNull(entity);
+
+        assertEquals(
+            Map.of(NAME_FIELD, precision),
+            entity.getFieldsPrecision()
+        );
+    }
+
+    /** Different precision definitions for the same field are a conflict. */
```
zstan, line 791:
```suggestion
    /** Different precision definitions for the same field are lead to conflict. */
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/CacheConfigurationQueryEntityMergeTest.java
```diff
@@ -0,0 +1,1260 @@
...
+            log,
+            () -> ccfg.setQueryEntities(Collections.singletonList(configured)),
+            CacheException.class,
+            msg
+        );
+    }
+
+    /** Configured indexedTypes must not prevent configured composite index from being created. */
+    @Test
+    public void testCompositeIndexIsCreatedWithIndexedTypes() throws Exception {
```
zstan, line 870: does this test relates to query merge logic ?

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryEntityMerger.java
```diff
@@ -0,0 +1,261 @@
...
+        throw mergeConflict(propName, existingVal, incomingVal);
+    }
+
+    /** */
+    private LinkedHashMap<String, String> mergeFields(
+        Map<String, String> existingFields,
+        Map<String, String> incomingFields
+    ) {
+        if (existingFields == null && incomingFields == null)
+            return null;
```
zstan, line 132: I see that result of this operation is used in : org.apache.ignite.cache.QueryEntity#setFields but there is no `@Nullable` annotation, thus you need to return smth like: `return U.newLinkedHashMap(0);`

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryEntityMerger.java
```diff
@@ -0,0 +1,261 @@
...
+                throw mergeConflict("fieldType[" + field + ']', existingType, incomingType);
+        }
+
+        return res;
+    }
+
+    /** */
+    private <T> Map<String, T> mergeMap(String propName, Map<String, T> existingVals, Map<String, T> incomingVals) {
+        if (existingVals == null && incomingVals == null)
+            return null;
```
zstan, line 164:
```suggestion
            return Map.of();
```

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryEntityMerger.java
```diff
@@ -0,0 +1,261 @@
...
+                throw mergeConflict(propName + '[' + field + ']', existingVal, incomingVal);
+        }
+
+        return res;
+    }
+
+    /** */
+    private <T> Set<T> mergeSet(Set<T> existing, Set<T> incoming) {
+        if (F.isEmpty(existing) && F.isEmpty(incoming))
+            return null;
```
zstan, line 196:
```suggestion
            return Set.of();
```

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryEntityMerger.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    /** */
+    private Collection<QueryIndex> mergeIndexes(
+        QueryEntity entity,
+        Collection<QueryIndex> existingIndexes,
+        Collection<QueryIndex> incomingIndexes
+    ) {
+        if (F.isEmpty(existingIndexes) && F.isEmpty(incomingIndexes))
+            return null;
```
zstan, line 216:
```suggestion
            return List.of();
```

modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
```diff
@@ -2484,6 +2478,29 @@ public CacheConfiguration<K, V> setIndexPath(String idxPath) {
         return S.toString(CacheConfiguration.class, this);
     }
 
+    /** */
+    private QueryEntity findQueryEntity(String valType) {
+        if (qryEntities == null)
+            return null;
```
zstan, line 2484: I plan to use JSpecify Nullaway frameqork in future, thus we need to be accurate with return types if function is not annotated with `@Nullable` plz fix it.

modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
```diff
@@ -2484,6 +2478,29 @@ public CacheConfiguration<K, V> setIndexPath(String idxPath) {
...
+
+        for (QueryEntity entity : qryEntities) {
+            if (Objects.equals(entity.findValueType(), valType))
+                return entity;
+        }
+
+        return null;
```
zstan, line 2491: and it

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryEntityMerger.java
```diff
@@ -0,0 +1,261 @@
...
+        if (!Objects.equals(ex.findValueType(), in.findValueType())) {
+            throw new CacheException(
+                "Failed to merge query entities because value types differ " +
+                    "[cacheName=" + cacheName +
+                    ", existingValueType=" + ex.findValueType() +
+                    ", incomingValueType=" + in.findValueType() + ']'
+            );
+        }
+
+        QueryEntity res = new QueryEntity(ex);
```
zstan, line 72: probably it\`s better to rewrite like : 
`QueryEntity res = ex instanceof QueryEntityEx ?  new QueryEntityEx(ex) : new QueryEntity(ex);`
and roll back code in 
`org.apache.ignite.internal.processors.cache.GridCacheUtils#patchCacheConfiguration`
I think it become more clear, wdyt ?

